PR #1, opened 2017-04-05
Feature: add static analysis

oubiwann:
This change adds support for various Clojure static analysis tools.

For subprojects:
 * a new profile has been added with new lein plugins for running the tools against the codebase
 * new aliases have been added that conveniently wrap the longer lein commands for the new profile

For the top-level project:
 * support has been added for running the most useful new static analysis commands across all projects
 * note that the CI/CD script has not yet been updated to use these (depends on future code cleanup work)

Future work:
 * there are pending branches which address the kibit issues
 * once those are addressed, we can update the CI/CD script to run `lein kibit` at the top-level
 * once we address the eastwood errors, we can add the combo command `lein lint` to the CI/CD script (which does both kibit and eastwood in one command)

### DIFF
--- a/access-control-app/project.clj
+++ b/access-control-app/project.clj
@@ -25,25 +25,39 @@
   :repl-options {:init-ns user}
   :jvm-opts ^:replace ["-server"
                        "-Dclojure.compiler.direct-linking=true"]
-  :profiles
-  {:dev {:dependencies ~(into '[[org.clojure/tools.namespace "0.2.11"]
-                                [pjstadig/humane-test-output "0.8.1"]
-                                [proto-repl "0.3.1"]]
-                          dev-cmr-deps)
-         :jvm-opts ^:replace ["-server"]
-         :source-paths ["src" "dev" "test" "int_test"]
-         :test-paths ["test" "int_test"]
-         :injections [(require 'pjstadig.humane-test-output)
-                      (pjstadig.humane-test-output/activate!)]}
+  :profiles {
+    :dev {:dependencies ~(into '[[org.clojure/tools.namespace "0.2.11"]
+                                 [pjstadig/humane-test-output "0.8.1"]
+                                 [proto-repl "0.3.1"]]
+                           dev-cmr-deps)
+          :jvm-opts ^:replace ["-server"]
+          :source-paths ["src" "dev" "test" "int_test"]
+          :test-paths ["test" "int_test"]
+          :injections [(require 'pjstadig.humane-test-output)
+                       (pjstadig.humane-test-output/activate!)]}
 
-   ;; This profile specifically here for generating documentation. It's faster than using the regular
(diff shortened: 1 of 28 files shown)


access-control-app/project.clj
```diff
@@ -59,4 +73,12 @@
                                         :let [project-dir (.replace (name group-artifact) "cmr-" "")]]
                                     ["shell" "ln" "-s" (str "../../" project-dir) "checkouts/,"]))
             ;; Alias to test2junit for consistency with lein-test-out.
-            "test-out" ["test2junit"]})
+            "test-out" ["test2junit"]
+            ;; Linting aliases
+            "kibit" ["do" ["with-profile" "lint" "shell" "echo" "== Kibit =="]
+                          ["with-profile" "lint" "kibit"]]
+            "eastwood" ["with-profile" "lint" "eastwood" "{:namespaces [:source-paths]}"]
+            "bikeshed" ["with-profile" "lint" "bikeshed"]
```
chris-durbin: Doesn't just adding the plugin automatically cause lein bikeshed to run with the bikeshed profile? I think it is fine to leave it here explicitly, just wanted to see if there's something I'm missing.

oubiwann: To keep things clean, the plugin was added to a `lint` profile ... the only way that bikeshed will get run is by executing `lein with-profile lint bikeshed`, and thus the alias. If bikeshed had been added to the top-level of each project, and not in a profile, then exactly what you said would apply.

chris-durbin: Missed lint in there. :)

access-control-app/project.clj
```diff
@@ -25,25 +25,39 @@
...
+    ;; using the default profile which causes the wait of 60 seconds before
+    ;; allowing the JVM to shutdown since no call to shutdown-agents is made.
+    ;; Generate docs with: lein with-profile docs generate-docs
+    :docs {}
 
-   :uberjar {:main cmr.access-control.runner
-             :aot :all}}
+    :uberjar {:main cmr.access-control.runner
+              :aot :all}
+    ;; This profile is used for linting and static analisys. To run for this
```
ygliuvt: Fix "analysis" typo in multiple places.

oracle-lib/src/cmr/oracle/connection.clj
```diff
@@ -14,6 +14,8 @@
    ;; This is required as an attempt to avoid NoClassDefFoundError occasionally during startup.
    ;; See CMR-3156
    (oracle.security.o5logon O5Logon)
+   ;;
+   ;(oracle.dms.console DMSConsole)
```
ygliuvt: Any reason we leave this commented line here?

access-control-app/project.clj
oubiwann: Whoops.

oubiwann: Done.

PR #2, opened 2017-04-05
Task: fix kibit errors

oubiwann:
This change addresses the kibit errors in subprojects that had only a few.

Note: This PR depends upon #1 and should not be merged before that one!

This includes the following:
* cmr-common-lib
* cmr-acl-lib (just one error)
* cmr-message-queue-lib (just three errors)
* cmr-mock-echo-app (just three errors)
* cmr-oracle-lib
* cmr-es-spatial-plugin (1 minor error)
* cmr-virtual-product-app (2 errors)
* cmr-ingest-app (3 errors)
* cmr-index-set-app (2 minor errors)
* cmr-elastic-utils-lib (1 error)

In addition to kibit fixes, some formatting of long lines was done as well.

### DIFF
--- a/common-lib/src/cmr/common/api/errors.clj
+++ b/common-lib/src/cmr/common/api/errors.clj
@@ -27,8 +27,8 @@
    :body {:errors ["An Internal Error has occurred."]}})
 
 (defn- keyword-path->string-path
-  "Converts a set of keyword field paths into the string equivalent field paths to return to the
-  user."
+  "Converts a set of keyword field paths into the string equivalent field paths
+  to return to the user."
   [field-path]
   (map (fn [path-item]
          (if (number? path-item)
@@ -37,15 +37,16 @@
        field-path))
 
 (defmulti errors->body-string
-  "Converts a set of errors into a string to return in the response body formatted according
-  to the requested response format."
+  "Converts a set of errors into a string to return in the response body
+  formatted according to the requested response format."
+
   (fn [response-format errors]
     response-format))
 
 (defmulti error->json-element
-  "Converts an individual error element to a clojure data structure representing the JSON element."
-  (fn [error]
-    (type error)))
+  "Converts an individual error element to a clojure data structure
+  representing the JSON element."
+  type)
 
 (defmethod error->json-element String
   [error]
@@ -61,8 +62,7 @@
 
 (defmulti error->xml-element
   "Converts an individual error element to the equivalent XML structure."
-  (fn [error]
-    (type error)))
+  type)
 
 (defmethod error->xml-element String
   [error]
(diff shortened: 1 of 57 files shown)


common-lib/src/cmr/common/api/errors.clj
```diff
@@ -27,8 +27,8 @@
    :body {:errors ["An Internal Error has occurred."]}})
 
 (defn- keyword-path->string-path
-  "Converts a set of keyword field paths into the string equivalent field paths to return to the
-  user."
+  "Converts a set of keyword field paths into the string equivalent field paths
```
chris-durbin: Nit: Not sure if this is coming from kibit or just an IDE preference, but we agreed on the CMR to a line limit of 100 characters. It looks like this might be 80. We're fine with anyone wrapping at something less than 100, but if it's a kibit setting we should probably set that value to 100.

ygliuvt: +1

oubiwann: Ah, I didn't realize that was a CMR thing. Bikeshed will complain about anything over 80, but not kibit. I was making the changes based on community standards/best practices ...

Is it okay to leave what has been done?

I'll make sure not to change long lines in the future. 

(Though my stuff will all be wrapped around 80; works better for me in terminals and with my split-panel editors -- I hope that's okay?)

oracle-lib/src/cmr/oracle/connection.clj
```diff
@@ -14,6 +14,8 @@
    ;; This is required as an attempt to avoid NoClassDefFoundError occasionally during startup.
    ;; See CMR-3156
    (oracle.security.o5logon O5Logon)
+   ;;
```
chris-durbin: Add a comment if there's a reason to keep this commented out line otherwise remove.

common-lib/src/cmr/common/api/errors.clj
chris-durbin: Not a problem, you can leave these changes and continue to use 80. Is there a way we could customize Bikeshed to complain at 100 instead since that's what all the code was using up to this point? I also like that bikeshed really lived up to its name - you get the most comments for line length. :)

oubiwann: *laughs* Oh, man ... too good :-)

Yeah, it looks like bikeshed lets you configure line length checks with `--max-line-length`.

oubiwann: Oh, btw, I would expect that kibit and eastwood is what we'd use in CI/CD scripts; I find that bikeshed and yagni are great for catching the odd bits that might get missed elsewhere, but are too noisy and often go above/beyond community norms for every-day linting/static analysis ... great as a developer tool, maybe a bit much as part of automation?

oracle-lib/src/cmr/oracle/connection.clj
oubiwann: Nope; I think that was from a bad edit/addition from the editor. Removing ...

PR #5, opened 2017-04-06
CMR-3915 Create ISO SMAP <=> UMM Location Keywords mapping

daniel-zamora:
Be able to read the attached ISO SMAP example (location keywords start on line 500) and convert it to the attached UMM example. Look only at the UMM LocationKeywords elements.

Also addresses ISO 19115_2 mapping bug in CMR-3903


daniel-zamora:
Looks like we don't have everyone on the CMR team so I can't add more reviewers.

PR #6, opened 2017-04-06
CMR-3944: Limit the ingest umm version to the configured ingest-accep…

sxu123:
…t-umm-version.

### DIFF
--- a/ingest-app/src/cmr/ingest/config.clj
+++ b/ingest-app/src/cmr/ingest/config.clj
@@ -6,8 +6,8 @@
             [cmr.message-queue.config :as rmq-conf]))
 
 (defconfig ingest-accept-umm-version
-  "Defines the latest umm version accepted by ingest - one version below the development version 
-   This environment variable needs to be manually set when newer UMM version becomes available"
+  "Defines the latest umm version accepted by ingest - it's the latest official version.  
+   This environment variable needs to be manually set when newer UMM version becomes official"
   {:default "1.9"}) 
 
 (defconfig ingest-username
--- a/ingest-app/src/cmr/ingest/config.clj
+++ b/ingest-app/src/cmr/ingest/config.clj
@@ -5,6 +5,11 @@
             [cmr.oracle.connection :as conn]
             [cmr.message-queue.config :as rmq-conf]))
 
+(defconfig ingest-accept-umm-version
+  "Defines the latest umm version accepted by ingest - one version below the development version 
+   This environment variable needs to be manually set when newer UMM version becomes available"
+  {:default "1.9"}) 
+
 (defconfig ingest-username
   "Ingest database username"
   {:default "CMR_INGEST"})
--- a/ingest-app/src/cmr/ingest/services/ingest_service.clj
+++ b/ingest-app/src/cmr/ingest/services/ingest_service.clj
@@ -59,11 +59,11 @@
 (defn- fix-ingest-concept-format 
    "Fixes formats"
   [fmt]
-  (if (mt/umm-json? fmt)
-    (if (mt/version-of fmt)
-      fmt 
-      (str fmt ";version=" (config/ingest-accept-umm-version)))
-    fmt))
+  (if (or 
+        (not (mt/umm-json? fmt)) 
+        (mt/version-of fmt))
+    fmt 
+    (str fmt ";version=" (config/ingest-accept-umm-version))))
 
 (defn-timed validate-and-prepare-collection
   "Validates the collection and adds extra fields needed for metadata db. Throws a service error
--- a/ingest-app/src/cmr/ingest/services/ingest_service.clj
+++ b/ingest-app/src/cmr/ingest/services/ingest_service.clj
@@ -3,6 +3,7 @@
     [cmr.common.cache :as cache]
     [cmr.common.config :as cfg :refer [defconfig]]
     [cmr.common.log :refer (debug info warn error)]
+    [cmr.common.mime-types :as mt]
     [cmr.common.services.messages :as cmsg]
     [cmr.common.services.errors :as errors]
     [cmr.common.util :as util :refer [defn-timed]]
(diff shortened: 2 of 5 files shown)


ingest-app/src/cmr/ingest/config.clj
```diff
@@ -5,6 +5,11 @@
             [cmr.oracle.connection :as conn]
             [cmr.message-queue.config :as rmq-conf]))
 
+(defconfig ingest-accept-umm-version
```
mschmele: You could do this dynamically in `cmr.umm-spec.versioning`

```
(def ingest-version
  "Defines the latest umm version accepted by ingest - one version below the development version 
   This environment variable needs to be manually set when newer UMM version becomes available"
  (last (butlast versions)))
```

ingest-app/src/cmr/ingest/services/ingest_service.clj
```diff
@@ -55,11 +56,20 @@
     {:collection sanitized-collection
      :warnings warnings}))
 
+(defn- fix-ingest-concept-format 
+   "Fixes formats"
+  [fmt]
+  (if (mt/umm-json? fmt)
```
mschmele: This `if` nest might be more readable as:
```
(if (or 
     (not (mt/umm-json? fmt)) 
     (mt/version-of fmt))
 fmt 
 (str fmt ";version=" (config/ingest-accept-umm-version)))
```
Double-check that I'm not changing your logic, but I'm fairly certain this is the same

ingest-app/src/cmr/ingest/config.clj
sxu123: I can not do that. Sorry for the misleading doc string. I was thinking of changing that but not sure exactly how to describe it.  Basically it's not really one version below, it's the highest official version, which could include the current-version.  
Also, we want to make this configurable because we want to allow the development/testers to be able to still ingest up to the current-version.

system-int-test/test/cmr/system_int_test/ingest/collection_ingest_test.clj
```diff
@@ -604,15 +606,29 @@
...
+        _ (side/eval-form `(config/set-ingest-accept-umm-version! ~accepted-version))]
+    (is (= 400 (:status response)))
+    (is (= [(str "umm version is higher than ingest-accept-umm-version\"" "1.8" "\"")]
+           (:errors response)))))
+
 (deftest ingest-invalid-umm-version
   (let [coll-map {:provider-id  "PROV1"
                   :native-id    "umm_json_coll_V1"
                   :concept-type :collection
-                  :format       "application/vnd.nasa.cmr.umm+json;version=9000.1"
```
laurenfrederick: Where to we validate that the version is actually a valid UMM version? I think ideally that would be done before the accepted version validation.

ingest-app/src/cmr/ingest/validation/validation.clj
```diff
@@ -88,9 +88,13 @@
...
-                     (umm-spec/validate-metadata (:concept-type concept)
-                                                 (:format concept)
-                                                 (:metadata concept))
+                     (let [umm-version (mt/version-of (:format concept))
+                           accept-version (config/ingest-accept-umm-version)]
+                       (if (>= 0 (compare umm-version accept-version))
+                         (umm-spec/validate-metadata (:concept-type concept)
+                                                     (:format concept)
+                                                     (:metadata concept))
+                         [(str "umm version is higher than ingest-accept-umm-version" (pr-str accept-version))]))
```
laurenfrederick: This needs a more user friendly message. Something like "UMM version X is in development. Can only ingest version 1.8 or lower." replace 1.8 with accept-version. This message would only work though if we validated that the version they're ingesting is legitimate first (see comment below). Otherwise it would have to be something different.

ingest-app/src/cmr/ingest/config.clj
laurenfrederick: We want a config for this so we can change it for testing purposes. Generally the idea is to block end-users from ingesting in-development versions of UMM so we do not invalidate their collections as we make schema changes. To test our changes, however, we'd have to allow ingest of the current in-development version.

Also, the accept version is not necessarily current - 1. Right now they are the same. If we were to make schema changes this PI, current would go to 1.10 and ingest would stay 1.9.

mschmele: Ah okay that makes sense. A docstring update is good then 😃

sxu123: Just want to point out that the default value and the actual value are two different things.
I think we need to set the default value to be always the current-version so that we can run/build our local tests without having the environment value set. 
In the fields, we will set the environment values differently, depending on if it's development environment or operation environment. I need to change the doc string.

ingest-app/src/cmr/ingest/services/ingest_service.clj
sxu123: Thanks. Done.

system-int-test/test/cmr/system_int_test/ingest/collection_ingest_test.clj
sxu123: The part that validates that the version is actually a valid UMM version is later in the process. it actually goes into the the file system to see if that schema file exists.  My thought is:
If the version is higher than the configured accepted version, it should be rejected right away without wasting the work to go in to find out that version doesn't exist.

ingest-app/src/cmr/ingest/validation/validation.clj
sxu123: 1. We don't know which version is in development, it might not exist, current-version might be official too. We only know which version is ingest accepted. 
This message has to be shared by both development and the operation environment, so we can't even tell through the comparison of the current-version and the ingest-accept-version.

2. If it's higher than the accepted version, we don't need to validate that their version is valid, just reject and tell them they can only ingest versions equal or below the accept-version.

3. If it's lower than the accepted version, we will allow them to ingest and then further down the code, it will catch it if it's an invalid umm version.

laurenfrederick: I thought that we could give them a better error message if we could tell if their version is valid. If that's not possible here, then we still need to give them a better error message.

Maybe something like "UMM JSON version X or lower can be ingested. Any version above that is considered in-development and cannot be ingested at this time." I want to convey why they can't ingest what they may think would be a valid version.

sxu123: Sounds good. Done.

ingest-app/src/cmr/ingest/config.clj
sxu123: Done.

sxu123:
Finally it built successfully.

PR #7, opened 2017-04-06
Task: enable dynamic switching between CMR run modes

oubiwann:
This change allows for the switching between `:in-memory` and `:external` CMR run modes. It supports both component-wide (all at once) run modes as well as setting run modes on individual components (one- or a-few-at-a-time). Previous behaviour required the commenting out and `(reset)`ing in the REPL. While calling a `(reset)` is still required, one may now simply set the appropriate CMR run modes in order to switch, without having to change any code.

Ways in which one may change the run modes:

* in the newly-extended function calls that now accept parameters (and set the run modes implicitly):
   * `(start :db external :messaging :internal)`  
   * `(start :all :external)`
   * `(reset :all :external)`
  
* explicitly setting run modes with the new convenience function `set-modes!`: 
   * set just one run mode: `(set-modes! :db :external)`
   * a few run modes at once: `(set-modes! :echo :external :elastic :in-memory)`
   * all at once: `(set-modes! :all :external)`

* from the command line, using the JVM options support in new `project.clj` profiles:
   * `$ lein with-profile +run-external repl`
   * `$ lein with-profile +run-in-memory repl`

Note that the old usage is still supported:
 * `(start)`
 * `(reset)`

Both of those will use whatever has been set in the settings data structure `run-modes`.

Notes:
 * all of the above methods eventually use the same code path to set the run modes (in other words, there are not three internal implementations)
* if no mode is supplied, the default of `:in-memory` will be used
* additional logic has been introduced for handling (and setting) `legacy` and `aws` boolean variables; if these are set to true, the applicable code paths will execute (again, without having to comment/uncomment anything)

### DIFF
--- a/dev-system/dev/refresh_persistent_settings.clj
+++ b/dev-system/dev/refresh_persistent_settings.clj
@@ -9,3 +9,33 @@
 (def logging-level
   "The logging level to use locally"
   (atom :error))
+
+(def legacy?
+  "Whether to run the dev-system with legacy services enabled and configured"
+  (atom false))
+
+(def aws?
+  "Whether to use AWS for messaging when running the dev-system"
+  (atom false))
+
+(defonce default-run-mode
+  ;; Note that `defonce` doesn't yet have the new(ish) support of docstrings
+  ;; that `def` does. See https://dev.clojure.org/jira/browse/CLJ-1148.
+  ^{:doc "Allow the default run mode to be set via the ENV or JVM options. In
+         practice, this is done in the following manner:
+
+         $ lein with-profile +run-external repl
+
+         or
+
+         $ lein with-profile +run-in-memory repl"}
+  (or (keyword (System/getProperty "cmr.runmode"))
+      :in-memory))
+
+(def run-modes
+  "The data structure that maintains the run modes for each individual
+  component"
+  (atom {:elastic default-run-mode
+         :echo default-run-mode
+         :db default-run-mode
+         :messaging default-run-mode}))
--- a/dev-system/dev/user.clj
+++ b/dev-system/dev/user.clj
@@ -27,7 +27,12 @@
 (selmer/cache-off!)
 
 (defonce system nil)
-(defonce mode (or (keyword (System/getProperty "cmr.runmode")) :in-memory))
+(defonce default-run-mode
+  (or (keyword (System/getProperty "cmr.runmode")) :in-memory))
+(defonce run-modes {:elastic default-run-mode
+                    :echo default-run-mode
+                    :db default-run-mode
+                    :messaging default-run-mode})
 (defonce legacy? false)
 (defonce aws? false)
 
@@ -38,21 +43,40 @@
           (fn [app-map]
             (u/map-values #(assoc-in % [:log :level] level) app-map))))
 
-(defn set-mode!
-  "Set the mode to use when starting/resetting. Valid options are:
+(defn set-modes!
+  "Set the mode to use when starting/resetting. One or more modes may
+  be set at a time by passing one or more of the following keyword
+  arguments, where `VALUE` is any alowed run mode:
 
+  * `:elastic VALUE`
+  * `:echo VALUE`
+  * `:db VALUE`
+  * `:messaging VALUE`
+
+  Allowed run mode values are currently:
   * :in-memory
-  * :external"
-  [new-mode]
-  (alter-var-root #'mode (constantly new-mode)))
+  * :external
 
-(defn use-legacy?
+  Examples:
+  ```
+  => (set-mode! :db :external)
+  => (set-modes! :elastic :in-memory :echo :external)
+  ```"
+  [& {:keys [elastic echo db messaging] :as new-modes}]
+  (->> new-modes
+       (remove #(nil? (val %)))
+       (into {})
+       (merge run-modes)
+       (constantly)
+       (alter-var-root #'run-modes)))
+
+(defn set-legacy
   "Passing `true` to this function will cause legacy configuration to be used
   during starts/resets."
   [bool]
   (alter-var-root #'legacy? (constantly bool)))
 
-(defn use-aws?
+(defn set-aws
   "Passing `true` to this function will cause AWS queues to be used during
   starts/resets."
   [bool]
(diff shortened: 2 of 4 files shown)


dev-system/dev/user.clj
```diff
@@ -52,55 +75,64 @@
...
+    (config/reset-config-values)
+
+    (jobs/set-default-job-start-delay! (* 3 3600))
+
+    (system/set-gorilla-repl-port! 8090)
+
+    (case mode
+      :in-memory (info "Running CMR in-memory ...")
+      :external (info "Running CMR externally ..."))
+    (system/set-dev-system-elastic-type! mode)
```
chris-durbin: We generally will only test one component in "external" mode at a time. I like these changes, we'll just need to make it a little more complicated to be able to choose the mode for each individual component.

dev-system/dev/user.clj
```diff
@@ -29,6 +32,26 @@
           (fn [app-map]
             (u/map-values #(assoc-in % [:log :level] level) app-map))))
 
+(defn set-mode!
+  "Set the mode to use when starting/resetting. Valid options are:
+
+  * :in-memory
+  * :external"
+  [new-mode]
```
indiejames: We can't use one var here for mode. We want to support any combination of in-memory/external for both the database and Elasticsearch, so you need to different vars.

dev-system/dev/user.clj
```diff
@@ -29,6 +32,26 @@
...
+  (alter-var-root #'mode (constantly new-mode)))
+
+(defn use-legacy?
```
indiejames: I think question marks should be reserved for predicates, but I'm willing to bow to the crowd view here.

oubiwann:
Ah, interesting ... good to be thinking about that ...

dev-system/project.clj
```diff
@@ -105,7 +105,19 @@
...
+    ;; The following run-* profiles are used in conjunction with other lein
+    ;; profiles to set the default CMR run mode and may be used in the
+    ;; following manner:
+    ;;
+    ;;   $ lein with-profile +run-external repl
+    ;;
+    ;; which will use dev and the other default profiles in addition to
+    ;; run-external (or whichever run mode profile is given).
+    :run-in-memory {
+      :jvm-opts ["-Dcmr.runmode=in-memory"]}
```
indiejames: Will :jvm-opts get merged with :jvm-opts from other profiles or overwrite them?

dev-system/dev/user.clj
```diff
@@ -29,6 +32,26 @@
           (fn [app-map]
             (u/map-values #(assoc-in % [:log :level] level) app-map))))
 
+(defn set-mode!
```
indiejames: So if I call `set-mode!` then I can just refresh and the system will restart int that mode, correct?

dev-system/dev/user.clj
```diff
@@ -29,6 +32,26 @@
...
+  "Set the mode to use when starting/resetting. Valid options are:
+
```
indiejames: You should probably add a validation that one of the valid modes is passed in.

dev-system/project.clj
indiejames: Someone using proto-repl should verify that they can start the REPL in a given mode this way - not sure how env vars are set for proto-repl.

oubiwann:
Okay, I think I have a solution which addresses your need. Not too much worse, really. Hurray data structures ;-)

I'll push the latest after the tests pass, and then update the description to account for this change.

dev-system/dev/user.clj
oubiwann: This behaviour will change in the next push.

dev-system/dev/user.clj
oubiwann: How are you guys on defensive programming?

I tend to stay away from that sort of thing myself, especially if writing code that devs will use in a REPL ...

dev-system/dev/user.clj
oubiwann: Right, that's what @chris-durbin mentioned -- I have an update for this change that addresses that.

dev-system/dev/user.clj
oubiwann: Fair enough. Just made that change locally; next push will have `set-legacy` and `set-aws` functions.

dev-system/project.clj
oubiwann: +1 for proto-repl testing.

As for `:jvm-opts`, yeah -- lein merges profile values unless you give it the `^replace` metadata. I did check, just to be sure, though ;-)

dev-system/dev/user.clj
oubiwann: Hey Chris, wanted to live a response to this comment, in case you'd missed it: I've pushed updates which add support for setting individual run modes.

chris-durbin: Yes, thanks. Let's get someone who has all the VMs to pair with you and make sure everything works as expected from ProtoREPL before approving.

oubiwann:
@indiejames Leo and I got the external db vm set up for me, and I've done a restart with `(restart :db :external)` ... what's the best way to check that it's running correctly?

oubiwann:
Okay, I've run all tests in the REPL. Steps:

1. start REPL
2. look at default run-modes: `settings/run-modes`
3, `(restart :db :external)`
4. confirming that the `:db` run-mode changed: `(:db @settings/run-modes)`
5. `(run-all-tests)` (note that this took about 25 minutes to run)
6. confirm that the `:db` run-mode  was still `:external`

That's just for the external db -- haven't done this with the echo, messaging, or elastic VMs yet.

oubiwann:
I've rebased from master and resolved a conflict with some of the recent changes to `user.clj`.

oubiwann:
Heads-up for anyone thinking about reviewing the latest: please wait!

Sadly, the two most recent commits from about 10 days ago have been lost and I'll need to re-write those before the branch is back in a good state :-/

oubiwann:
Update on my last comment: branch is back in a reviewable state.

dev-system/dev/refresh_persistent_settings.clj
```diff
@@ -9,3 +9,33 @@
...
+  (or (keyword (System/getProperty "cmr.runmode"))
+      :in-memory))
+
+(def run-modes
+  "The data structure that maintains the run modes for each individual
+  component"
+  (atom {:elastic default-run-mode
+         :echo default-run-mode
+         :db default-run-mode
+         :messaging default-run-mode}))
```
mschmele: Nit: add a newline here

dev-system/dev/user.clj
```diff
@@ -123,15 +191,30 @@
                   (fn [s]
                     (when s (system/stop s)))))
 
-(defn reset []
+(defn reset
+  "Resets the development environment, taking optional keyword arguments for
```
mschmele: This docstring update is great

dev-system/dev/refresh_persistent_settings.clj
oubiwann: Fixed -- thanks!

dev-system/dev/user.clj
```diff
@@ -35,6 +35,66 @@
           (fn [app-map]
             (u/map-values #(assoc-in % [:log :level] level) app-map))))
 
+(defn- maybe-set-all-modes
+  "A utility function for (conditionally) setting all the values of a map to
+  the same value."
+  [all new-modes]
```
indiejames: This docstring and the arguments are a bit confusing to me. It makes it sound like the second argument will get the 'all' value - there is no mention of the side-effect of looking up run-modes.

dev-system/dev/user.clj
```diff
@@ -35,6 +35,66 @@
...
+  * :in-memory
+  * :external
+
+  Examples:
+  ```
+  => (set-mode! :db :external)
+  => (set-modes! :elastic :in-memory :echo :external)
+  ```"
+  ;; Note that the keys are listed below as a means of self-documentation; they
+  ;; are not actually used individuall, but rather as a whole with the
```
indiejames: individually typo

dev-system/dev/user.clj
```diff
@@ -35,6 +35,66 @@
...
+  ;; `new-modes` hash map.
+  [& {:keys [all elastic echo db messaging] :as new-modes}]
```
indiejames: Nit: I think it would be more idiomatic to names these _all, _elastic, etc.

dev-system/dev/user.clj
```diff
@@ -58,53 +118,61 @@
...
+  "Starts the current development system, taking optional keyword arguments for
+  various run modes (see the docstring for `set-modes!` for more details).
 
-  ;; Comment/uncomment these lines to switch between external and internal settings.
+  Note that when one or more explicit modes are passed as arguments, the global
+  `run-modes` data structure is updated.
 
-  (system/set-dev-system-elastic-type! :in-memory)
-  ; (system/set-dev-system-elastic-type! :external)
+  If a run mode for a particular component is not passed, its value taken from
```
indiejames: "its value is taken"

dev-system/dev/user.clj
```diff
@@ -58,53 +118,61 @@
...
+  what is already in the `run-modes` global data structure. If no mode has been
+  set, the default from initialization will be used."
+  ;; Note that even through the named args are not used, they are provided as
+  ;; a means of self-documentation.
+  [& {:keys [elastic echo db messaging] :as new-modes}]
```
indiejames: Same comment about naming these _elastic, _echo, etc.

dev-system/dev/user.clj
```diff
@@ -123,15 +191,30 @@
...
+
+  Note that when one or more explicit modes are passed as arguments, the global
+  `run-modes` data structure is updated.
+
+  If a run mode for a particular component is not passed, its value taken from
+  what is already in the `run-modes` global data structure. If no mode has been
+  set, the default from initialization will be used."
+  ;; Note that even through the named args are not used, they are provided as
+  ;; a means of self-documentation.
+  [& {:keys [elastic echo db messaging] :as new-modes}]
```
indiejames: Again - _elastic, _echo, etc.

dev-system/dev/user.clj
oubiwann: Good point.

dev-system/dev/user.clj
oubiwann: Hrm, yeah. Definitely a bit vague. Also good point about the side effects ... the function probably doesn't merit a `!`, since sometimes it will not change anything. So, yeah, I should make up for this by mentioning the side effect nature ...

dev-system/dev/user.clj
oubiwann: Fixed.

dev-system/dev/user.clj
oubiwann: Fixed.

dev-system/dev/user.clj
oubiwann: Fixed.

dev-system/dev/user.clj
oubiwann: Done.

dev-system/dev/user.clj
oubiwann: Fixed.

dev-system/dev/user.clj
oubiwann: Fixed.

oubiwann:
Weirdness on bamboo build ... re-running. Will merge when green.

oubiwann:
Note that I copied this branch to one that actually had the right ticket associated with it (the second one we created for this work: CMR-4070). That has been merged with master via the command line, after checking the Bamboo build.

Manually closing this PR.

oubiwann:
Here's a link to the branch that was merged:
 * https://github.com/oubiwann/Common-Metadata-Repository/tree/task/CMR-4070-parameterize-dev-modes

PR #14, opened 2017-04-11
CMR 4045 - Put UMM Spec Validation Errors back in and split out extra ones into warnings

laurenfrederick:


system-int-test/test/cmr/system_int_test/ingest/collection_misc_validation_test.clj
```diff
@@ -1,26 +1,26 @@
...
+    [cmr.system-int-test.data2.core :as d]
+    [cmr.system-int-test.data2.umm-spec-collection :as data-umm-c]
+    [cmr.system-int-test.utils.ingest-util :as ingest]))
 
 (defn assert-valid
   ([coll-attributes]
    (assert-valid coll-attributes nil))
   ([coll-attributes options]
-   (let [collection (assoc (dc/collection coll-attributes) :native-id (:native-id coll-attributes))
+   (let [collection (assoc (data-umm-c/collection coll-attributes) :native-id (:native-id coll-attributes))
```
chris-durbin: How did these tests pass before the change to treat the ingests as failures? Were they somehow taking a different path through ingest that was causing them to fail?

laurenfrederick: Business rule validation was always returning errors, that had never been turned off. One of the other ones had validation turned on for the test.

chris-durbin: Oh good - business rule validations probably cover a good chunk of the acceptance criteria. What's left to verify beyond those?

laurenfrederick:
I've attached my spreadsheet of the side by side comparison of umm-lib validation to umm-spec validation to the ticket, accounting for every validation on either side. I've updated the spreadsheet this morning with the namespace of where the test is for each validation.

Many of the validations are in umm-spec-lib, so changes in ingest would not affect them. I did include in the spreadsheet the integration tests which test a subset of these validations. The tests initially broke, of course, and I made them work by setting turning the validate umm-spec config on, either in individual tests or via a namespace-wide fixture.

system-int-test/test/cmr/system_int_test/ingest/collection_misc_validation_test.clj
```diff
@@ -41,43 +40,56 @@
...
+                        :CollectionProgress :complete})
           ingest-response (d/ingest "PROV1" collection {:format :dif10})
-          validation-response (ingest/validate-concept (dc/collection-concept collection :dif10))]
+          validation-response (ingest/validate-concept (data-umm-c/collection-concept collection :dif10))]
       (is (some? (re-find #"/Platforms/0/ShortName string.*is too long \(length: 81, maximum allowed: 80\)" (:warnings ingest-response))))
       (is (some? (re-find #"/Platforms/0/ShortName string.*is too long \(length: 81, maximum allowed: 80\)" (:warnings validation-response))))
       (is (some? (re-find #"/Purpose string.*is too long \(length: 12000, maximum allowed: 10000\)" (:warnings ingest-response))))
       (is (some? (re-find #"/Purpose string.*is too long \(length: 12000, maximum allowed: 10000\)" (:warnings validation-response)))))))
 
+(deftest multiple-warnings
```
ygliuvt: It's great that you added this test. Do we have a similar test for checking on multiple validation errors? If not, we should add one for that too.

laurenfrederick: There are some in cmr.system-int-test.ingest.collection-umm-spec-validation-test

As for keyword validation errors - what is your question there? Keyword validation errors require the header to be passed in and there is a separate file testing all of them. I know because i break it all the time. cmr.system-int-test.ingest.collection-keyword-validation-test

ygliuvt: My question is if we have tests cover those two cases and you answered yes. I approve.

PR #15, opened 2017-04-11
Replace instances of REMOVED with something better

mschmele:
The wiki link that got removed was public, so I never did anything with it 👻. Rather than have a sentence with REMOVED in the middle of it, I opted to just remove it altogether.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,7 +54,6 @@ Items to include in each issue
   * Deployment Impacts:
     * Requires update of indexes and reindexing of collections. Note that collection indexing will be broken until update-indexes has been called.
     * Create a new alias for the collection index used by search.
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -54,7 +54,7 @@ Items to include in each issue
   * Deployment Impacts:
     * Requires update of indexes and reindexing of collections. Note that collection indexing will be broken until update-indexes has been called.
     * Create a new alias for the collection index used by search.
-      * See ***REMOVED*** for instructions.
+      * See https://wiki.earthdata.nasa.gov/display/CMR/CMR+Version+1.52 for instructions.
```
chris-durbin: That wiki page is not meant to be public, we should remove this bullet.

PR #17, opened 2017-04-12
Task: split out site and api routes

oubiwann:
This is part of the CMR-3988 feature work (adding landing pages to the search-app). 

In particular, it: 
* separates `api` routes from `site` routes
* adds a new `web` namespace for combining these two
* moves the handlers to that new namespace (which is also the new namespace used when setting up the web server)

In the next change set, the number of routes in the new `site` namespace will increase, making the need for the separation even more clear.

### DIFF
--- a/search-app/README.md
+++ b/search-app/README.md
@@ -4,8 +4,27 @@ Provides a public search API for concepts in the CMR.
 
 ## API Documentation
 
-API Docs are located in api_docs.md. They can be generated into the static site by running `lein generate-docs` This takes a while so we will commit the generated file for now so building the uberjar does not take a very long time. You can also generate the documents with a pre-started REPL by running `(load-file "./support/generate_docs.clj")` If you're running the REPL from dev-system you need to correct the path to generate_docs.clj.
+API docs for the search-app project are located in `api_docs.md`. They can be
+generated into the static site by running the following:
+
+```
+$ lein generate-docs
+```
+
+Due to the fact that this generally takes a while on most systems, we have
+generated it and committed the resulting file in order to bring the total
+build time (e.g, creating an uberjar) down.
+
+Additionally, you can also generate the documents with a pre-started REPL by
+running:
+
+```clj
+(load-file "./support/generate_docs.clj")
+```
+
+Note that if you're running the REPL from `dev-system` you need to correct the
+path to `generate_docs.clj`.
 
 ## License
 
-Copyright © 2014 NASA
+Copyright © 2014, 2017 NASA
--- a/search-app/README.md
+++ b/search-app/README.md
@@ -8,22 +8,8 @@ API docs for the search-app project are located in `api_docs.md`. They can be
 generated into the static site by running the following:
 
 ```
-$ lein generate-docs
+$ lein with-profile docs generate-docs
 ```
-
-Due to the fact that this generally takes a while on most systems, we have
-generated it and committed the resulting file in order to bring the total
-build time (e.g, creating an uberjar) down.
-
-Additionally, you can also generate the documents with a pre-started REPL by
-running:
-
-```clj
-(load-file "./support/generate_docs.clj")
-```
-
-Note that if you're running the REPL from `dev-system` you need to correct the
-path to `generate_docs.clj`.
 
 ## License
 
--- a/search-app/src/cmr/search/api/routes.clj
+++ b/search-app/src/cmr/search/api/routes.clj
@@ -333,7 +333,7 @@
         [(msg/mixed-arity-parameter-msg mixed-param)]))
     (f request)))
 
-(defn default-error-format-fn
+(defn default-error-format
   "Determine the format that errors should be returned in based on the request URI."
   [{:keys [uri]} _e]
   (if (or (re-find #"/caches" uri)
--- a/search-app/src/cmr/search/api/routes.clj
+++ b/search-app/src/cmr/search/api/routes.clj
@@ -44,7 +44,6 @@
    [cmr.search.services.acls.granule-acls]
 
    [cmr.search.services.health-service :as hs]
-   [cmr.search.services.messages.common-messages :as msg]
    [cmr.search.services.parameters.legacy-parameters :as lp]
    [cmr.search.services.query-service :as query-svc]
    [cmr.search.services.result-format-helper :as rfh]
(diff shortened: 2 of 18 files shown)


search-app/src/cmr/search/api/routes.clj
```diff
@@ -315,20 +305,46 @@
...
+  the query string. e.g. foo=1&foo[bar]=2 is mixed arity, so is foo[]=1&foo[bar]=2. foo=1&foo[]=2 is
+  not. Parameter with mixed arity will be flagged as invalid later."
+  [query-str]
+  (when query-str
+    (let [query-str (-> query-str
+                        (str/replace #"%5B" "[")
+                        (str/replace #"%5D" "]")
+                        (str/replace #"\[\]" ""))]
+      (last (some #(re-find % query-str)
+                  [#"(^|&)(.*?)=.*?\2\["
```
indiejames: Nice use of backreference to the capture groups, but I think this might find false positives if you have two parameter names where one contains the other, e.g., limit=1&error_limit=5.

search-app/README.md
```diff
@@ -4,8 +4,27 @@ Provides a public search API for concepts in the CMR.
...
 
-API Docs are located in api_docs.md. They can be generated into the static site by running `lein generate-docs` This takes a while so we will commit the generated file for now so building the uberjar does not take a very long time. You can also generate the documents with a pre-started REPL by running `(load-file "./support/generate_docs.clj")` If you're running the REPL from dev-system you need to correct the path to generate_docs.clj.
+API docs for the search-app project are located in `api_docs.md`. They can be
+generated into the static site by running the following:
+
+```
+$ lein generate-docs
+```
+
+Due to the fact that this generally takes a while on most systems, we have
```
chris-durbin: I realize this is existing documentation, but I think it is outdated. We run lein with-profile docs generate-docs and the documentation is generated in seconds. I believe we do this as part of the build and we aren't committing the generated file anymore.

search-app/src/cmr/search/api/routes.clj
```diff
@@ -315,20 +305,46 @@
...
+  from Ring."
+  [f]
+  (fn [request]
+    (when-let [mixed-param (find-query-str-mixed-arity-param (:query-string request))]
+      (svc-errors/throw-service-errors
+        :bad-request
+        [(msg/mixed-arity-parameter-msg mixed-param)]))
+    (f request)))
+
+(defn default-error-format-fn
```
indiejames: Nitpick: I would drop the -fn in the name here. With the -fn I would expect this to return a function, which it does not.

search-app/src/cmr/search/site/routes.clj
```diff
@@ -0,0 +1,28 @@
+(ns cmr.search.site.routes
+  "This namespace is the one responsible for the routes intended for human
```
indiejames: nitpick: I think I would rewrite this as "routes for static resources" (I'm assuming these are not generated upon request), since it includes CSS, etc., that aren't generally consumed by humans.

search-app/src/cmr/search/web/routes.clj
```diff
@@ -0,0 +1,68 @@
...
+   [ring.middleware.params :as params]
+
+   ;; CMR libs
+   [cmr.acl.core :as acl]
+   [cmr.common.api.errors :as errors]
+   [cmr.common.api.context :as context]
+   [cmr.common-app.api.routes :as common-routes]
+   [cmr.search.api.request-context-user-augmenter :as context-augmenter]
+   [cmr.search.api.routes :as api-routes]
+   [cmr.search.site.routes :as site-routes]))
```
indiejames: I kind of like the way you have separated the requires into CMR/non-CMR, but the standard we use is to sort alphabetically (the cmr ones end up together that way anyhow). I'm not opposed to switching to this, but we should be consistent.

search-app/src/cmr/search/api/routes.clj
oubiwann: Can't take credit for that code -- it was pre-existing -- I just moved it around.

search-app/src/cmr/search/api/routes.clj
oubiwann: That's also pre-existing ...

search-app/src/cmr/search/site/routes.clj
oubiwann: The template pages, which are coming, are generated dynamically.

search-app/src/cmr/search/web/routes.clj
oubiwann: That was carried over too -- I actually use alpha-sorted, like you guys. If you don't want to preserve the old stuff, I'm +1 for doing sorted.

oubiwann:
@chris-durbin ah, thanks -- those were done later in a different branch; I'll cherry pick those over right now ...

oubiwann:
Hrm, actually ... won't be able to do that; the were pulled into the tests for the new template pages (that's in the next branch). New commits on their way ...

search-app/src/cmr/search/api/routes.clj
indiejames: Sorry - you moved it last so it's yours now :)

search-app/README.md
oubiwann: Fixed.

search-app/src/cmr/search/api/routes.clj
oubiwann: Hahaha ;-)

oubiwann: This one's for Marc Huffnagle ...

oubiwann:
@chris-durbin, @indiejames: okay, pushed changes and checked tests on Bamboo -- all green!

search-app/src/cmr/search/api/routes.clj
indiejames: Seriously, though, the rule we go by is that you can't (knowingly) merge broken code, even if you didn't write it originally. It's a drag, but it keeps our code cleaner. (Leo has had to fix my mistakes countless times).

I think you can just put an ampersand before the back references like this
`#"(^|&)(.*?)=.*?&\2\["`
and this
`#"(^|&)(.*?)\[.*?&\2="`
and it will fix the issue.

oubiwann: So, this is probably outside the scope of this branch ...

I don't know the code, but if you think there is a bug here, shouldn't a bug report be filed? Sounds like this needs new tests, etc., which would definitely warrant some dedicated context.

search-app/src/cmr/search/api/routes.clj
oubiwann: Quick note on this one: did a `git grep` and this same function appears in ingest-app. If we change here, we should probably change there too ...

search-app/src/cmr/search/api/routes.clj
indiejames: Yeah, if you want to file a ticket for it instead of fixing it here that's probably fine - Chris can confirm. It probably should have had a unit test originally. For that matter, it should probably be moved into common-app-lib and added to the other projects, which is definitely out of scope.

search-app/src/cmr/search/api/routes.clj
oubiwann: Did some more poking around to see how else CMR handles error formatting, and these two examples are the only ones I could see that don't use the default.

Still, that being said, since it does occur in two places, it might make sense to briefly ponder the best home for this function ... maybe some place in common?

search-app/src/cmr/search/api/routes.clj
oubiwann: Okay, cool -- since you're thinking common for this, and maybe common is a good place for the one mentioned below, maybe we can do that work in the same ticket ...

I'll go ahead and open the ticket, and paste the applicable notes from here; if @chris-durbin gives the 👎 we can just close that ticket as invalid/wontdo ...

search-app/src/cmr/search/api/routes.clj
ygliuvt: Search app and ingest app have different definitions for the function. I think we should just rename the functions in both places and be done with it.

oubiwann: Doh! You are most correct.

Changing them now ...

oubiwann:
Thanks, guys!

PR #20, opened 2017-04-12
Cmr 3905

ygliuvt:
Fixed CMR to not use the GEM_HOME and GEM_PATH setting to load gems.


ygliuvt:
Need to put this on hold until we fully integrate with MMT.

PR #24, opened 2017-04-13
CMR-4304: Configurable URS password for mock-echo

mschmele:


mock-echo-app/src/cmr/mock_echo/api/urs.clj
```diff
@@ -1,15 +1,18 @@
...
+   [clojure.data.xml :as x]
+   [clojure.string :as str]
+   [cmr.common.config :refer [defconfig]]
+   [cmr.common.mime-types :as mt]
+   [cmr.common.services.errors :as errors]
+   [cmr.common.xml :as cx]
+   [cmr.mock-echo.data.urs-db :as urs-db]
+   [cmr.transmit.config :as transmit-config]
+   [compojure.core :refer :all]
+   [ring.util.response :as rsp]))
```
oubiwann: _super_ tiny nit: lisp data structures and funcalls tend to do two spaces of indent ... if this were me, I'd add one more space in front of those required elements ...

oubiwann:
This looks like low-impact; I'd +1 it. If a project newb's vote on this change is sufficient .... ;-)

mock-echo-app/src/cmr/mock_echo/api/urs.clj
mschmele: We're styling our `ns`es based on [this stuart sierra blogpost](https://stuartsierra.com/2016/clojure-how-to-ns.html), which I think says to do the above

oubiwann: Ah, good to know. 

Hrm, I wonder why he's treating `:require` like an element of a data structure instead of a (special) form?

/me looks at source ...

Ah, probably because the `ns` macro treats the next group of forms after the docstring as a data structure (assigned to 'attr-map?', in particular). That answers that question!

Thanks for the link!

PR #26, opened 2017-04-17
CMR-3908: Remapped the Project field.

sxu123:


umm-spec-lib/src/cmr/umm_spec/iso19115_2_util.clj
```diff
@@ -13,6 +13,14 @@
 (def short-name-xpath
   "gmi:identifier/gmd:MD_Identifier/gmd:code/gco:CharacterString")
 
+;; relative to gmi:childOperation
```
laurenfrederick: This is really helpful.

umm-spec-lib/src/cmr/umm_spec/iso19115_2_util.clj
```diff
@@ -13,6 +13,14 @@
...
+(def campaign-xpath
+  "gmi:MI_Operation/gmi:identifier/gmd:MD_Identifier/gmd:code/gco:CharacterString") 
+
+(def start-end-date-xpath
```
laurenfrederick: This name here is confusing because it's in the iso utils and called start-end-date, but description has nothing to do necessarily with start-end-date and "gmi:description/gco:CharacterString" is widely used throughout iso.

umm-spec-lib/src/cmr/umm_spec/xml_to_umm_mappings/iso19115_2.clj
```diff
@@ -102,12 +102,26 @@
...
-       :LongName (su/truncate long-name su/PROJECT_LONGNAME_MAX sanitize?)})))
+          long-name (value-of proj iso-util/long-name-xpath) 
+          start-end-date (when-let [date (value-of proj iso-util/start-end-date-xpath)]
+                           (str/split (str/trim date) #"\s+"))
+          ;; date is built as: StartDate: 2001:01:01T01:00:00Z EndDate: 2002:02:02T01:00:00Z
+          ;; One date can exist without the other.
+          start-date (when start-end-date 
+                       (if (= "StartDate:" (get start-end-date 0))
+                         (get start-end-date 1)
+                         (get start-end-date 3)))
```
laurenfrederick: Wouldn't this line be returning the end date? Or are you assuming the string is EndDate: 2001:01:01T01:00:00Z StartDate: 2002:02:02T01:00:00Z here?

laurenfrederick: What if it's just `EndDate: 2002:02:02T01:00:00Z`? Then `(get start-end-date 3)` would be out of bounds.

sxu123: It will return nil, which is fine because I will remove the nil elements in the end.

sxu123: I'm not assuming any order or the StartDate/EndDate. Either one could be missing.

sxu123: (get start-end-date 3) will return StartDate if "StartDate" is NOT the first element, or it will return nil if "StartDate" is completely missing.

umm-spec-lib/src/cmr/umm_spec/iso19115_2_util.clj
sxu123: "gmi:description/gco:CharacterString" could be many things. but start-end-date-xpath could only be one. It's the same with these other paths too. like short-name-xpath, long-name-xpath and other xpaths. None of them imply the meaning of the xpath.   If you really want to change it, what do you recommend that we use?

daniel-zamora: I would recommend you remove it. The xpath isn't long enough to make the case for being more easily read and it is only used once.  I think it would be better to just not use a def for this.

PR #31, opened 2017-04-19
CMR-4058: Indexed platform's characteristic's value, instrument/child…

sxu123:
… instrument's characteristic's name, description and value


indexer-app/src/cmr/indexer/data/concepts/collection/keyword.clj
```diff
@@ -82,9 +82,14 @@
...
-        char-names (keep :name characteristics)
-        char-descs (keep :description characteristics)
+        plat-characteristics (mapcat :characteristics platforms)
+        plat-char-names (keep :name plat-characteristics)
+        plat-char-descs (keep :description plat-characteristics)
+        plat-char-values (keep :value plat-characteristics)
+        instr-characteristics (mapcat :characteristics instruments)
+        instr-char-names (keep :name instr-characteristics)
+        instr-char-descs (keep :description instr-characteristics)
+        instr-char-values (keep :value instr-characteristics)
```
sxu123: The ticket description is a bit confusing to me. It seems to assume name and description are already indexed, only value needs to be added to the index. Since I didn't see them being indexed, I just indexed them all.
Here is the description:
"Index Platform and Instrument Characteristic value so that it can be searched through the keyword search parameter. Currently Characteristics names and descriptions are indexed as are Additional Attributes names, data types, values, and descriptions but Characteristics values are not indexed. The airborne group from two different DAACs would like to search on the Tail Number of an aircraft platform. But since the value is not indexed, the tail number can't be searched. The name of the Characteristic doesn't help because the name is Tail Number and all aircraft tail numbers are going to have this name."

system-int-test/test/cmr/system_int_test/search/collection_keyword_search_test.clj
```diff
@@ -44,14 +44,23 @@
         p2 (dc/platform
             {:short-name "platform_SnA spoonA"
              :long-name "platform_LnA"
-             :characteristics [(dc/characteristic {:name "char1" :description "char1desc"})
-                               (dc/characteristic {:name "char2" :description "char2desc"})]
+             :characteristics
```
laurenfrederick: Since you are editing this test file, it should be converted.

indexer-app/src/cmr/indexer/data/concepts/collection/keyword.clj
laurenfrederick: Yeah I guess the answer is to index them all then. You could check with Erich on this via email. He is at NSIDC currently so he could ask them.

system-int-test/test/cmr/system_int_test/search/collection_keyword_search_test.clj
sxu123: I was thinking since I'm not adding additional dc/collection, therefore I didn't add additional work for the future convert. The main difficulty is I had tried to convert this test and failed. it's a tricky one and takes a lot of work.

indexer-app/src/cmr/indexer/data/concepts/collection/keyword.clj
sxu123: I sent him an email and he replied. The reason he thinks they are indexed is because our document says so and I think he might just test the platform search.  My test result shows it doesn't work for instrument/child instrument search. so I'm adding them.

PR #33, opened 2017-04-20
CMR-4064

ygliuvt:
This pull request reactivated invocation of the preview gem to render collection preview page now that the preview gem is ready for integration, and also pulls the UMM schema version that is supported by the preview gem off the config file in the gem and convert the collection being rendered into that particular UMM JSON format that is supported by the gem. This fixed the rendering problems for fields that have been modified in between the UMM schema versions that are supported by CMR and preview gem.


collection-renderer-lib/src/cmr/collection_renderer/services/collection_renderer.clj
```diff
@@ -24,11 +26,13 @@
...
   (io/resource "collection_preview/collection_preview.erb"))
 
-(def preview-gem-umm-schema-version
-  "Temporay def for the UMM schema version that is supported by the CMR preview gem.
-   This will be removed once CMR preview gem is fixed to include the UMM schema version
-   it supports in the gem. (MMT-921)"
-  "1.6")
+(def preview-gem-umm-version-config-file
+  "Defines the path to the UMM schema version config file within the preview gem."
+  "gems/cmr_metadata_preview-0.0.1/.umm-version")
```
chris-durbin: We're going to use version 0.0.1 in production?

😄

collection-renderer-lib/src/cmr/collection_renderer/services/collection_renderer.clj
```diff
@@ -32,6 +42,19 @@
...
+  []
+  (try
+    (-> preview-gem-umm-version-config-file
+        io/resource
+        slurp
+        str/trim)
+    (catch Exception e
+      (error "Failed to parse preview gem UMM version, will use the default version:"
+             default-preview-gem-umm-version
+             (.printStackTrace e)))))
```
indiejames: I'm not sure this is the right thing to do. The config file SHOULD be correct. If it is not then something is wrong and you should not just swallow this exception. This will allow things to start up with possibly the wrong gem version.

collection-renderer-lib/src/cmr/collection_renderer/services/collection_renderer.clj
```diff
@@ -50,11 +73,14 @@
   l/Lifecycle
 
   (start
-    [this _system]
-    (assoc this :jruby-runtime (create-jruby-runtime)))
+   [this _system]
+   (-> this
+       (assoc :jruby-runtime (create-jruby-runtime))
+       (assoc :preview-gem-umm-version
+              (or (get-preview-gem-umm-version) default-preview-gem-umm-version))))
```
indiejames: See other comments regarding ignoring problems reading the config file.

collection-renderer-lib/src/cmr/collection_renderer/services/collection_renderer.clj
indiejames: No - 0.0.1-SNAPSHOT :)

ygliuvt: Currently we have no supported way in MMT to provide version control. I have been in dialog with MMT folks to get that fixed. They will be making changes to start versioning their gem and this will be different once they start doing that.

collection-renderer-lib/src/cmr/collection_renderer/api/routes.clj
```diff
@@ -1,9 +1,18 @@
...
-           [cmr.common.services.errors :as errors]
-           [clojure.string :as str]
-           [clojure.java.io :as io]))
+  (:require
+   [clojure.java.io :as io]
+   [clojure.string :as str]
+   [cmr.common.services.errors :as errors]
+   [compojure.core :refer :all]))
+
+(def cmr-metadata-preview-gem
```
ygliuvt: As I just commented above. I am working with MMT to provide versioned gem so that we can configure different gem versions for different deployment environments. You are right. This will be in puppet config once we do that.

collection-renderer-lib/src/cmr/collection_renderer/services/collection_renderer.clj
```diff
@@ -32,6 +42,19 @@
     (.eval jruby (io/reader bootstrap-rb))
     jruby))
 
+(defn- get-preview-gem-umm-version
+  "Get the UMM schema version that is defined in preview gem."
```
ygliuvt: The path is within the gem that was installed.

collection-renderer-lib/src/cmr/collection_renderer/services/collection_renderer.clj
```diff
@@ -24,6 +26,14 @@
...
+(def preview-gem-umm-version-config-file
+  "Defines the path to the UMM schema version config file within the preview gem."
+  "gems/cmr_metadata_preview-0.0.1/.umm-version")
+
+(def default-preview-gem-umm-version
+  "Default preview gem UMM schema version in case preview gem is mis-configured."
+  "1.8")
+
 (defn- create-jruby-runtime
   "Creates and initializes a JRuby runtime."
```
ygliuvt: :)

collection-renderer-lib/src/cmr/collection_renderer/services/collection_renderer.clj
ygliuvt: I put that in while working with MMT to settle down on the interface and it was good to not cause CMR to go belly up because the gem doesn't do the config correctly as we are now working off the current repo of MMT project. I agree with your comment, especially after move to versioned gems. I will take this try-catch block out.

collection-renderer-lib/src/cmr/collection_renderer/services/collection_renderer.clj
abbottry: We've not committed to any changes... the requirements and expectations for this gem have gone far beyond what was originally discussed. Before anymore tickets or requests are made we need to get on the same page. MMT uses commit hashes for 'versioning' but we just use the most recent so it was never a concern. I'm not even convinced this should be a ruby gem anymore given the way its being used by CMR...

ygliuvt: Talked to @abbottry, filed MMT-927 for providing versioning in preview gem. I will put this on hold until we resolve the versioning issue.

ygliuvt:
After discussion with MMT team, we have decided to do versioning support of the preview gem through the commit id of the cmr_metadata_preview project which requires MMT to only send CMR the last commit id whenever they do a new release (sort of) of the gem. 

So @indiejames, we will still be using the cmr_metadata_preview-0.0.1 as the gem name. :) 

The new updates include dropping the handling of errors when reading the UMM version off the config and versioning support (sort of) of the gem itself.

collection-renderer-lib/project.clj
```diff
@@ -7,6 +7,15 @@
   "Defines the repo url of cmr_metadata_preview project"
   "https://git.earthdata.nasa.gov/scm/cmr/cmr_metadata_preview.git")
 
+(def metadata-preview-commit
+  "Defines the commit id of cmr_metadata_preview project that we should build the gem off.
+   This is our contract with the MMT team in maintaining versioning of the gem without involving
+   RubyGems. The environment variable CMR_METADATA_PREVIEW_COMMIT can be used to override
+   the hardcoded commit id during dev integration with cmr_metadata_preview project.
+   The hardcoded commit id should be updated when MMT releases a new version of the gem."
+  (or (System/getenv "CMR_METADATA_PREVIEW_COMMIT")
```
chris-durbin: We should just make this a defconfig with default of "68cc0a3bfec"

ygliuvt: I don't think we have defconfig defined here.

chris-durbin: Oh ok works for me then.

PR #34, opened 2017-04-20
Fixed regex for find-query-str-mixed-arity-param and added test.

indiejames:
There was an error in the regex that could lead to false positives. Added two test cases that triggered the error then fixed the regexes to prevent the issue.

### DIFF
--- a/search-app/src/cmr/search/web/routes.clj
+++ b/search-app/src/cmr/search/web/routes.clj
@@ -33,8 +33,8 @@
                         (str/replace #"%5D" "]")
                         (str/replace #"\[\]" ""))]
       (last (some #(re-find % query-str)
-                  [#"(^|&)(.*?)=.*?\2\["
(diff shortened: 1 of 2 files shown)


search-app/src/cmr/search/web/routes.clj
```diff
@@ -33,8 +33,8 @@
                         (str/replace #"%5D" "]")
                         (str/replace #"\[\]" ""))]
       (last (some #(re-find % query-str)
-                  [#"(^|&)(.*?)=.*?\2\["
-                   #"(^|&)(.*?)\[.*?\2="])))))
+                  [#"(^|&)(.*?)=.*?&\2\["
+                   #"(^|&)(.*?)\[.*?&\2="])))))
```
oubiwann: Nice catch on these, man. Sharp eyes :-)

PR #36, opened 2017-04-21
CMR-3966 Fix bulk update ACLs and change endpoint return structure

laurenfrederick:


daniel-zamora:
Looks good

ingest-app/src/cmr/ingest/api/ingest.clj
```diff
@@ -456,13 +455,11 @@
   (generate-provider-task-status-response
    headers
    {:status 200
-    :task-status "Partial Fail"
-    :status-message "The following collections had errors: C-1, C-2"
+    :task-status "COMPLETE"
```
ygliuvt: I see you are pushing the partial failure status down the status-message. I guess it is OK.

PR #57, opened 2017-05-08
CMR 3965 - Bulk update

laurenfrederick:
Handle invalid concept ids in bulk update, save collection level errors, and update overall task status. Remove hardcoded concept ids in tests.

This is only a small part of the ticket, but getting more for MMT to test against in SIT.


ingest-app/src/cmr/ingest/data/bulk_update.clj
```diff
@@ -38,9 +46,10 @@
   (get-bulk-update-task-status
     [db task-id]
     ;; Returns a status for the particular task
-    (su/find-one db (su/select [:status :status-message]
-                     (su/from "bulk_update_task_status")
-                     (su/where `(= :task-id ~task-id)))))
+    (let [status (su/find-one db (su/select [:status :status-message]
+                                   (su/from "bulk_update_task_status")
```
oubiwann: The lines `(su/from ...` and `(su/where ...` should be indented to align with the first arg, `[:status ...`.

ingest-app/src/cmr/ingest/data/bulk_update.clj
```diff
@@ -38,9 +46,10 @@
...
+                                   (su/where `(= :task-id ~task-id))))]
+      (util/map-keys->kebab-case status)))
```
oubiwann: I tend to avoid `let`s with a single-use variable. The threading macros are usually clear enough to be just as self-documenting (and the function name + docstring can make up for any lack there):

```clj
(-> db
    (su/find-one (su/select [:status ...]
                            (su/from ...)
                            (su/where ...)))
    (util/map-keys->kebab-case))
```

Using `->>` here probably wouldn't be a good idea, since that would mean breaking up your select statement in an artificial manner ...

oubiwann:
Looks good to me, but I don't know enough about this to assess impact and offer a meaningful review :-/ (beyond the minor comments I made) ...

ingest-app/src/cmr/ingest/data/bulk_update.clj
laurenfrederick: Fixed

ingest-app/src/cmr/ingest/data/bulk_update.clj
laurenfrederick: I think the let helps with debugging, but since I am probably done debugging that function, I changed it.

oubiwann: Yeah, +1 on the debugging -- it's pretty awkward inserting a pass-through debug with the threading macros, e.g., ``((fn [x] (log/debug "stuff:" x) x))``

ingest-app/src/cmr/ingest/data/bulk_update.clj
```diff
@@ -9,6 +9,14 @@
    [cmr.oracle.connection]
    [cmr.oracle.sql-utils :as su]))
 
+(defn generate-task-status-message
```
daniel-zamora: good candidate for a unit test, and possibly move it to another namespace if its not going to be only message/utility functions.  Maybe the latter can be left for future updates.

ingest-app/src/cmr/ingest/data/bulk_update.clj
```diff
@@ -103,13 +112,16 @@
                             "SET status = ?, status_message = ?"
                             "WHERE task_id = ? AND concept_id = ?")]
          (j/db-do-prepared db statement [status status-message task-id concept-id])
-         (let [pending-collections (su/query db
-                                     (su/build (su/select [:concept-id]
+         (let [task-collections (su/query db
+                                     (su/build (su/select [:concept-id :status]
```
daniel-zamora: fix the indentation

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
```diff
@@ -56,6 +57,17 @@
...
 
 (defn handle-collection-bulk-update-event
-  "Perform update for the given concept id"
+  "Perform update for the given concept id. Log an error status of the concept
+  cannot be found."
   [context task-id concept-id bulk-update-params]
-  (data-bulk-update/update-bulk-update-task-collection-status context task-id concept-id "COMPLETE" nil))
+  (try
+    (if-let [concept (mdb2/get-latest-concept context concept-id)]
+      (data-bulk-update/update-bulk-update-task-collection-status context task-id concept-id "COMPLETE" nil)
```
mschmele: Nit: The two calls to this function having inconsistent formatting makes this a little tough to read. 
Putting the arguments all on their own line isn't pretty, but I think it makes it more obvious why you're calling this function twice.
```
(data-bulk-update/update-bulk-update-task-collection-status 
  context 
  task-id 
  concept-id 
  "COMPLETE" 
  nil)
(data-bulk-update/update-bulk-update-task-collection-status 
  context task-id 
  concept-id 
  "FAILED" 
  (format "Concept-id [%s] is not valid." concept-id))
```

ingest-app/src/cmr/ingest/data/bulk_update.clj
laurenfrederick: Done

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
laurenfrederick: Made them the same formatting.

PR #60, opened 2017-05-09
CMR-4091 Add request JSON body to task status endpoint

laurenfrederick:


mschmele:
This looks like it should only return the original request when performing bulk-updates. Is that true? That makes the most sense to me.

PR #64, opened 2017-05-11
CMR-3529 humanizer report as scheduled job and cached data

oubiwann:
This change updates report generation with some new features which address a performance bug recently found: when trying to download humanizer reports in our new production environment, the time taken has increased (on average) beyond the request timeout value. 

The new features include:
 * a scheduled job for generating the report every 24 hours
 * a report cache that utilizes elastic search as the data store, but offers a consistent memory store on top of that, and then a single-threaded cache for queuing reads/writes (to avoid race conditions via HTTP requests and the job scheduler)

When the scheduler starts up, it kicks off the report generation. If, for any reason, there is a delay and a user requests the report first, the report is generated and then cached. And, of course, returned. Otherwise, when the generator finishes, the scheduler will cache the report data and all subsequent requests will hit the cache.


search-app/src/cmr/search/services/humanizers/humanizer_report_service.clj
```diff
@@ -95,3 +105,88 @@
...
+  data."
+  [context]
+  (.set-value (get-cache context)
+              csv-report-cache-key
+              (safe-generate-humanizers-report-csv context)))
+
+(defconfig report-cache-consistent-timeout-seconds
+  "The number of seconds between when the report cache should check with cubby
+  for data consistency."
+  {:default (* 60 60 6) ; every 6 hours
```
oubiwann: The reason this was set to different than the time for the generated report is for when they "miss" each other; I just picked an arbitrary fraction of 24 hours ... there's no reason we couldn't have it go hourly, in the event that the 24 window was missed.

system-int-test/test/cmr/system_int_test/admin/cache_api_test.clj
```diff
@@ -71,18 +71,49 @@
...
-                                       "usage-metrics-cache"]
-        (url/index-set-read-caches-url) ["token-imp" "health"]
-        (url/mdb-read-caches-url) ["token-imp" "health"]
-        (url/ingest-read-caches-url) ["token-imp" "providers" "acls" "token-user-ids" "kms"
-                                      "health" "humanizer-alias-cache" "write-enabled"]
-        (url/access-control-read-caches-url) ["acls" "health" "group-ids-guids" "write-enabled"]
-        (url/search-read-caches-url) ["acls" "collections-for-gran-acls" "has-granules-map"
-                                      "index-names" "token-imp" "token-sid" "token-user-id" "health"
-                                      "xsl-transformer-templates" "metadata-cache" "kms" "write-enabled"])
+        (url/indexer-read-caches-url) [
```
indiejames: Don't bother changing this, but it occurs to me that it would be better if we used the respective vars here instead of string values. That way we can change cache names if needed without updating tests.

search-app/src/cmr/search/services/humanizers/humanizer_report_service.clj
indiejames: I think we may still run into problems here if I understand this correctly. If it has been more than six hours since someone has generated the report then a request will cause the report to be generated, correct?. This will still suffer from the timeout issue since generation isn't any faster. The workaround is to wait for five or six minutes and make the request again, by which time it should be cached, I think. Which is better than what we have now.

indiejames:
Nice job on getting this done so quickly.

search-app/src/cmr/search/services/humanizers/humanizer_report_service.clj
```diff
@@ -95,3 +105,88 @@
                        "concat humanized rows:" t3))))]
       (info "Create report " t4)
       (str string-writer))))
+
+(defn safe-generate-humanizers-report-csv
```
ygliuvt: Why do we do this? Wouldn't it be better to just throw an internal server error in this case?

search-app/src/cmr/search/services/humanizers/humanizer_report_service.clj
```diff
@@ -95,3 +105,88 @@
...
+  "This convenience function wraps the report generator in a try/catch."
+  [context]
+  (try
+    (generate-humanizers-report-csv context)
+    (catch Exception e
+      (warn (.getMessage e) "Returning empty report."))))
+
+(def report-cache-key
+  "The key used to store the humanizer report cache in the system cache map."
+  :report-cache)
```
oubiwann: I was thinking that this might be used for more than just humanizer reports in the future ... and that we might want to move this code to someplace more general at that time ...

But I think your point is valid: it's here now, let's name it as it is, and we can update the name when we move it. I'll change it.

search-app/src/cmr/search/services/humanizers/humanizer_report_service.clj
oubiwann: >  If it has been more than six hours since someone has generated the report
>  then a request will cause the report to be generated, correct?

I don't think so ... as I understand the consistency cache this is just the time that a regular check will be made to the fallback cubby cache: every six hours it will check if there is a difference between the in-memory consistency cache and the cubby cache.

We may need to have a chat to walk through this to make sure my logic is correct and matches reality, but I believe the way things are coded in this change mean that the only time the report is generated is:

1. when a job runs, or 
1. when an HTTP request is made and there is a cache miss

system-int-test/test/cmr/system_int_test/admin/cache_api_test.clj
oubiwann: Hrm, that's a good point. I'll add a `XXX` todo  here ...

search-app/src/cmr/search/services/humanizers/humanizer_report_service.clj
```diff
@@ -95,3 +105,88 @@
...
+
+(defn get-cache
+  "A utility function for returning the system report cache instance."
+  [context]
+  (get-in context [:system :caches report-cache-key]))
+
+(defn read-report
```
ygliuvt: Why not use the existing `get-value` function in `cmr.common.cache`? You can also specify the cache miss function so that you don't need the new wrapper function `humanizers-report-csv` for writing report. Just rename `generate-humanizers-report-csv` to `humanizers-report-csv`.

indiejames: Note for the retrospective - we don't do a good enough job of documenting our internal APIs. I'm not very familiar with the caching stuff either, so I might have done the same.

search-app/src/cmr/search/services/humanizers/humanizer_report_service.clj
oubiwann: It's perfectly valid to have an empty report before the system has had a chance to generate one. Why would we want to throw an error?

Additionally, by throwing an error here it would cause the developer REPL to be littered with stacktraces from the exception ...

search-app/src/cmr/search/services/humanizers/humanizer_report_service.clj
oubiwann: Nice! Will change ...

search-app/src/cmr/search/services/humanizers/humanizer_report_service.clj
indiejames: I'm confused now. Isn't this catching errors during report generation? Which is not the same as "the report is empty." I'm inclined to agree with Leo in that this should result in a 5xx status. There should not be a reason why the report could not be generated.

indiejames: Also, we don't run this from the REPL.

ygliuvt:
I think we should write a separate issue to generate a humanizers report on a given list of collections. This will give users capability to get instant feedback on the collections they are updating to fix any humanizer related issues; otherwise they will have to wait a day to see the updated report. The current endpoint won't work for them.

search-app/src/cmr/search/services/humanizers/humanizer_report_service.clj
```diff
@@ -95,3 +105,88 @@
...
+  data."
+  [context]
+  (.get-value (get-cache context)
+              csv-report-cache-key))
+
+(defn write-report
+  "A convenience function wrapping the cache object to retreive report
+  data."
+  [context]
+  (.set-value (get-cache context)
```
oubiwann: Already done :-) I had the same thought -- I just discovered that function yesterday and remembered it again now.

indiejames: For what it's worth, if I have my way, we will kill off cubby and replace it with the AWS Cache API.

search-app/src/cmr/search/services/humanizers/humanizer_report_service.clj
indiejames: Wait, you mean if the job runs and fails. I don't remember much about the jobs, but we need a way to disable some of the ones that are irrelevant during development. It's really irritating to see all the output from various jobs.

oubiwann: We don't run this from the REPL, but the job scheduler runs in the dev-system, so this code actually does now get run in the dev-system and the REPL will see errors here:
```
user=> 2017-05-12 09:43:27.384 [HumanizerReportGeneratorJob] ERROR [cmr.search.services.humanizers.humanizer-report-service] - HumanizerReportGeneratorJob caught exception.
                                                      org.quartz.simpl.SimpleThreadPool$WorkerThread.run         SimpleThreadPool.java:  573
                                                                         org.quartz.core.JobRunShell.run              JobRunShell.java:  202
             cmr.search.services.humanizers.humanizer-report-service.HumanizerReportGeneratorJob/execute  humanizer_report_service.clj:  153
                  cmr.search.services.humanizers.humanizer-report-service.HumanizerReportGeneratorJob/fn  humanizer_report_service.clj:  156
cmr.search.services.humanizers.humanizer-report-service/safe-generate-humanizers-report-csv/invokeStatic  humanizer_report_service.clj:  110
     cmr.search.services.humanizers.humanizer-report-service/generate-humanizers-report-csv/invokeStatic  humanizer_report_service.clj:   88
                            cmr.search.services.humanizers.humanizer-service/get-humanizers/invokeStatic         humanizer_service.clj:   84
                   cmr.search.services.humanizers.humanizer-service/fetch-humanizer-concept/invokeStatic         humanizer_service.clj:   56
                                             cmr.common.services.errors/throw-service-error/invokeStatic                    errors.clj:   27
                                                                       clojure.core/ex-info/invokeStatic                      core.clj: 4617
clojure.lang.ExceptionInfo: Humanizer does not exist
```

This will go away if you run the humanizer tests, though.

How about a compromise? I'll add a check here that will throw if it's not the error above ...

ygliuvt:
Update the api doc to state the limitation of humanizers report in terms of caching delay.

search-app/src/cmr/search/services/humanizers/humanizer_report_service.clj
oubiwann: Done.

system-int-test/test/cmr/system_int_test/admin/cache_api_test.clj
oubiwann: Added todo note.

search-app/src/cmr/search/services/humanizers/humanizer_report_service.clj
indiejames: Would it be possible to just not run the job when running from dev system?

oubiwann: This is place locally now and working nicely; will push changes soon.

ygliuvt:
I think the current fix is sufficient for reporting purpose. We can deal with the live reporting of a given list of collections when users actually request the feature.

search-app/src/cmr/search/services/humanizers/humanizer_report_service.clj
oubiwann: Done.

search-app/src/cmr/search/services/humanizers/humanizer_report_service.clj
indiejames: But I guess that would break the test...

oubiwann: @indiejames Check out the updated code; it now raises for everything except a lack of humanizers.

ygliuvt: This works.

PR #65, opened 2017-05-11
CMR-4115: Added concept-type to logging.

ygliuvt:


metadata-db-app/src/cmr/metadata_db/data/oracle/search.clj
```diff
@@ -152,7 +153,11 @@
                    [conn db]
                    (let [conditions [`(>= :id ~start-index)
                                      `(< :id ~(+ start-index batch-size))]
-                         _ (info "Finding batch for provider" provider-id "from id >=" start-index " and id <" (+ start-index batch-size))
+                         _ (info (format "Finding batch for provider [%s] concept type [%s] from id >= %s and id < %s"
```
indiejames: Doug has a dashboard that gets populated from the previous version of this logging statement. Please check with him to verify that this format still matches his regex.

ygliuvt: @littleidiot40, Can you review this logging change? I believe it gives you better info on the batch processing. You may need to  update your script if it doesn't work with the new logging.

littleidiot40: Here are the queries my dashboard does:
sourcetype="prod_ngap_app_cmr-bootstrap-$cmr-env$" Finding batch | rex "Finding batch for provider (?&lt;provider&gt;[^ ]*) from" | chart count by provider

sourcetype="prod_ngap_app_cmr-bootstrap-$cmr-env$" Finding batch | chart count by host

The first one will break. But it's an easy fix.

ygliuvt: This will make it apparent in the log which concept type CMR is batch processing, i.e. is it processing collections or granules? You could update your chart to be more accurate on where the bulk indexing is at.

PR #67, opened 2017-05-12
Add mailmap to fix contributor chart

mschmele:
I found a way to fix the contributor history, and this one wouldn't require rewriting history!


oubiwann:
We should update the acceptance criteria for this ... I know in the past, Github has not utilized `.mailmap` for its contributors list (and thus nor its graphs) -- this may have changed, but we'll need to look closely.  We'll want to check the git log via command line and the github interface, before and after this change ...

mschmele:
@oubiwann I can't find anything definitive after 2014 on whether or not GitHub utilizes the `.mailmap` file, though I've found countless repositories that do use it all the same, which would lead me to believe that it actually _is_ used by GitHub

PR #68, opened 2017-05-15
Cmr 3965 umm update

laurenfrederick:
This is another part of CMR-3965. Put in the framework to bulk update collections through umm conversion, save to metadata db, and report errors. This is not going to handle every bulk update case, but right now focusing on science keywords as that is the main use case and goal of this PI. I will have another PR coming for this ticket with more fields to bulk update.

### DIFF
--- a/ingest-app/src/cmr/ingest/services/bulk_update_service.clj
+++ b/ingest-app/src/cmr/ingest/services/bulk_update_service.clj
@@ -1,12 +1,17 @@
 (ns cmr.ingest.services.bulk-update-service
   (:require
+   [camel-snake-kebab.core :as csk]
    [cheshire.core :as json]
    [clojure.java.io :as io]
+   [clojure.string :as str]
    [cmr.common.services.errors :as errors]
    [cmr.common.validations.json-schema :as js]
    [cmr.ingest.data.bulk-update :as data-bulk-update]
    [cmr.ingest.data.ingest-events :as ingest-events]
-   [cmr.transmit.metadata-db2 :as mdb2]))
+   [cmr.ingest.services.ingest-service :as ingest-service]
+   [cmr.transmit.metadata-db :as mdb]
+   [cmr.transmit.metadata-db2 :as mdb2]
+   [cmr.umm-spec.field-update :as field-update]))
 
 (def bulk-update-schema
   (js/json-string->json-schema (slurp (io/resource "bulk_update_schema.json"))))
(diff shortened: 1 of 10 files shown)


ingest-app/src/cmr/ingest/services/bulk_update_service.clj
```diff
@@ -56,16 +61,56 @@
...
+
 (defn handle-collection-bulk-update-event
   "Perform update for the given concept id. Log an error status of the concept
   cannot be found."
   [context task-id concept-id bulk-update-params]
   (try
     (if-let [concept (mdb2/get-latest-concept context concept-id)]
-      (data-bulk-update/update-bulk-update-task-collection-status context task-id
-        concept-id "COMPLETE" nil)
+      (let [success-status-message (update-and-save-collection context concept bulk-update-params)]
```
oubiwann: Is there a guarantee that `update-and-save-collection` will always result in success?

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
```diff
@@ -56,16 +61,56 @@
...
+        (data-bulk-update/update-bulk-update-task-collection-status context task-id
+          concept-id "COMPLETE" success-status-message))
```
oubiwann: I think I'd use a threading macro here:

```clj
(->> bulk-update-params
     (update-and-save-collection context concept)
     (data-bulk-update/update-bulk-update-task-collection-status 
      context task-id concept-id "COMPLETE"))
```

oubiwann:
Overall, looks good; will defer to others with more knowledge here for a formal review. I believe I remember saying that you were testing changes like this with a local Oracle VM after running `(reset :db :external)` in the REPL, but I wanted to double-check. The reason I mention it is that Leo and I had noticed some failing tests when running master against an Oracle VM (tests that were still passing in Bamboo, but not locally).

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
```diff
@@ -1,12 +1,17 @@
 (ns cmr.ingest.services.bulk-update-service
   (:require
+   [camel-snake-kebab.core :as csk]
    [cheshire.core :as json]
    [clojure.java.io :as io]
+   [clojure.string :as str]
```
ygliuvt: I think we are changing our naming convention to use `string` as the alias for `clojure.string` as suggested by @oubiwann.

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
laurenfrederick: Yes, any errors would be thrown as errors and caught in the error handler and then logged. Validation only comes back with warnings, not errors.

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
laurenfrederick: To me, this would just be confusing. That way, it's not really clear what is coming back from `update-and-save-collection` and it would be extra work for someone to go figure that out or debug.

oubiwann: Caveat: the following response is meant as a conversation starter, not to change your mind in this PR ;-)

I think such an objection to a clear Clojure idiom indicates that there are other problems ... not sure what the underlying issues may be that ultimately make our API calls unclear as to return values, but possibly:

* use better names
* establish conventions for functions with side effects but which also return values
* update docstrings
* include return type metadata in function definitions

It may be that the objection really boils down to the lack of comfort with a language that is not statically typed? Dunno ...

Something else to keep in mind: the use of threading macros are often used for clarity in debugging, as they so clearly show how data passes through the given code.

laurenfrederick: I went back and did a bit of a re-factor, because I think warnings being returned from `update-and-save` and that not having anything to do with validations didn't make sense. I split up the updating of the concept and validate and save. I used better naming to help clear things up.

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
laurenfrederick: Ok I changed it, but I think now if we're in any file that uses it, we should be consciously updating it.

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
```diff
@@ -0,0 +1,73 @@
...
+                          :update-value {:Category "EARTH SCIENCE"
+                                         :Topic "HUMAN DIMENSIONS"
+                                         :Term "ENVIRONMENTAL IMPACTS"
+                                         :VariableLevel1 "HEAVY METALS CONCENTRATION"}}]
+       ;; Kick off bulk update
+       (ingest/bulk-update-collections "PROV1" bulk-update-body)
+       ;; Wait for queueing/indexing to catch up
+       (index/wait-until-indexed)
+       (let [collection-response (ingest/bulk-update-task-status "PROV1" 1)]
+         (def collection-response collection-response)
```
daniel-zamora: debug code

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
```diff
@@ -0,0 +1,73 @@
...
+
+       ;; Check that each concept was updated
+       (doseq [concept-id concept-ids
+               :let [concept (-> (search/find-concepts-umm-json
+                                   :collection {:concept-id concept-id})
+                                 :results
+                                 :items
+                                 first)]]
+        (is (= 2
+              (:revision-id (:meta concept))))
```
daniel-zamora: fix indentation.

PR #69, opened 2017-05-15
CMR-3529 Added logging for when the report generator runs as a job.

oubiwann:
This is important for troubleshooting in production.

I'm not yet convinced that this job is running as expected in AWS deployments. When I hit a deployment URL for the report this morning, it look a long time to get the download data; this should not have happened. The report data should have already been made available in the cache and thus downloaded immediately. It seems that the initial request I made today was what generated the report, indicating that the report did not generate upon post-deployment startup of CMR.

Fortunately, it did cache and my next request was immediate. However, I do have concerns about the job execution and would like to check ElasticSearch after deployments to ensure that it is running as a properly scheduled job.


oubiwann:
Thanks!

PR #71, opened 2017-05-16
CMR-4132 - Attempt to fix ingest application public root URL again

oubiwann:
It seems that the last attempt to fix this bug was only a partial success: the ingest application is now using the same public conf set as the other apps and a little bit more configuration data has made it into the base url for the ingest app. Sadly, `cmr.ingest.system/public-conf` did not set the host and port (though it did set the protocol/scheme). Instead of the old value of `null://null:null` we are now getting `http://null:null`. 

In order to address that, there are two steps necessary:

1. the new ingest config data in this PR
1. an update to the ingest config overrides in the configuration service for CMR

With both of those changes, we *should* see this further updated to the full `https://cmr.sit.earthdata.nasa.gov:443` when this gets rolled out to SIT.

Updates:
* Brought access-control, ingest, and search systems into closer parity
* Also, updated dev-system to use the `public-conf`s defined in the respective projects.

In particular, since I am new to this part of the CMR, please examine the following for proper usage and accuracy:
* All three systems are now setting public values for scheme, host, and port
* All of these have new docstrings ~~which attempt to better describe what's going on and what ENV vars will map to them~~
* The `cmr.transmit.config` ns docstring has been updated with more details

### DIFF
--- a/access-control-app/src/cmr/access_control/system.clj
+++ b/access-control-app/src/cmr/access_control/system.clj
@@ -40,11 +40,8 @@
 
   Note: this configuration value is used as-is in local, dev environments
   and is overridden with ENV variables in remote deployments. In both cases,
-  this configuration information is utilized and required.
-
-  The name given after `defconfig` is important and is used when checking for
-  an ENV variable. The ENV variable for this configuration will be:
-  * CMR_ACCESS_CONTROL_PUBLIC_PROTOCOL"
+  this configuration information is utilized and required. See `defconfig` for
+  more details."
   {:default "http"})
 
 (defconfig access-control-public-host
@@ -52,11 +49,8 @@
 
   Note: this configuration value is used as-is in local, dev environments
   and is overridden with ENV variables in remote deployments. In both cases,
-  this configuration information is utilized and required.
-
-  The name given after `defconfig` is important and is used when checking for
-  an ENV variable. The ENV variable for this configuration will be:
-  * CMR_ACCESS_CONTROL_PUBLIC_HOST"
+  this configuration information is utilized and required. See `defconfig` for
+  more details."
   {:default "localhost"})
 
 (defconfig access-control-public-port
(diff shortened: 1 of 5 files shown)


oubiwann:
Just had a great screenhero session with @chris-durbin, tons of info, going to replace this set of changes with completely different changes :-)

access-control-app/src/cmr/access_control/system.clj
```diff
@@ -31,26 +31,52 @@
   {:default nil
    :parser cfg/maybe-long})
 
+(defconfig log-level
+  "App logging level"
+  {:default "info"})
+
 (defconfig access-control-public-protocol
-  "The protocol to use in documentation examples for the access-control application."
+  "The protocol to use in documentation examples for the access-control application.
```
oubiwann: I think I'm going to tweak these again ... to remove the documentation-specific reference, since they are now used for other things, too.

access-control-app/src/cmr/access_control/system.clj
```diff
@@ -31,26 +31,52 @@
...
+
+  Note: this configuration value is used as-is in local, dev environments
+  and is overridden with ENV variables in remote deployments. In both cases,
+  this configuration information is utilized and required.
+
+  The name given after `defconfig` is important and is used when checking for
```
chris-durbin: The documentation on the variable name is probably redundant with the defconfig macro itself. Rather than listing what the environment variable name will be in each place, we should probably make sure the defconfig macro documentation is thorough and includes the environment variable naming.

access-control-app/src/cmr/access_control/system.clj
oubiwann: Done.

access-control-app/src/cmr/access_control/system.clj
```diff
@@ -31,26 +31,52 @@
...
    :type Long})
 
-(defconfig log-level
-  "App logging level"
-  {:default "info"})
-
-(defn public-conf
-  "Public access-control configuration used for generating example requests in documentation"
-  []
+(def public-conf
```
chris-durbin: I think this has to be a function in order for the configuration to be overridable at runtime.

Otherwise you are going to get the value used at the time of building the jars (which does not have the runtime configuration variables set since we use the same jars for every environment and just set the configuration for runtime).

search-app/src/cmr/search/system.clj
```diff
@@ -55,7 +79,11 @@
   "App logging level"
   {:default "info"})
 
-(def search-public-conf
+(def public-conf
```
chris-durbin: Oh I thought this needed to be a function in order to override with configuration from environment variables at runtime. Now I'm confused 😕

oubiwann: Let's go with the function, just to be safe -- I'll update them all, since this one might not have been getting overridden ...

access-control-app/src/cmr/access_control/system.clj
oubiwann: Removed.

access-control-app/src/cmr/access_control/system.clj
oubiwann: Changed.

search-app/src/cmr/search/system.clj
oubiwann: Oh, wait ... this is search, not ingest. I'm confused too :-) Regardless, I've updated all of them to be functions :-)

oubiwann:
Thanks, guys!

PR #76, opened 2017-05-18
CMR-4134 - Added variable validation.

oubiwann:
This is part 2 of a three-parter for UMM-Var ingest:

1. Add schema and models for variables (with some placeholders)
2. **Add validation code for variables (with some placeholders)**
3. Add ingest-services and routes for variables

The validators in this change were modeled off of the collection validators and written to use the generated schema. As such, when the UMM-Var schema matures, we should have to change very little (if anything) here.

### DIFF
--- a/ingest-app/src/cmr/ingest/validation/business_rule_validation.clj
+++ b/ingest-app/src/cmr/ingest/validation/business_rule_validation.clj
@@ -95,4 +95,5 @@
                 concept-id-validation
                 version-is-not-nil-validation
                 collection-update-validation]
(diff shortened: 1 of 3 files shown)


oubiwann:
@ygliuvt: Addressing only your point about nothing using this: the next PR takes advantage of this code ...

oubiwann:
Going to close this one without merge, per conversation with @ygliuvt .

This (tiny) bits of this that we'll end up keeping will show up in the next PR in this series ...

PR #78, opened 2017-05-19
CMR-4145: create variable association in metadata-db

ygliuvt:
This is part 1 of a four-parter for creating variable associations with collections:

1. Add creating variable association concept in metadata-db
2. Add search of variables in metadata-db
3. Add search of variable associations in metadata-db
4. Add variable associations endpoint in search-app

### DIFF
--- a/metadata-db-app/int_test/cmr/metadata_db/int_test/utility.clj
+++ b/metadata-db-app/int_test/cmr/metadata_db/int_test/utility.clj
@@ -176,6 +176,13 @@
       "VariableType" "",
       "ScienceKeywords" []}))
 
+(def variable-association-edn
+  "Valid EDN for variable association metadata"
+  (pr-str {:variable-name "totCldH2OStdErr"
+           :associated-concept-id "C120000000-PROV1"
+           :revision-id 1
+           :value "Some Value"}))
+
 (def concept-dummy-metadata
   "Index events are now created by MDB when concepts are saved. So the Indexer will attempt
   to look up the metadata for the concepts and parse it. So we need to provide valid
@@ -189,7 +196,8 @@
    :access-group group-edn
    :acl acl-edn
    :humanizer humanizer-json
-   :variable variable-json})
+   :variable variable-json
+   :variable-association variable-association-edn})
 
 (defn- concept
   "Create a concept map for any concept type. "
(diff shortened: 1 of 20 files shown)


metadata-db-app/src/cmr/metadata_db/data/oracle/search.clj
```diff
@@ -25,12 +25,16 @@
...
+                          [:associated_concept_id :associated_revision_id :tag_key :user_id])
    :access-group (into common-columns [:provider_id :user_id])
    :service (into common-columns [:provider_id :entry_title :entry_id :delete_time :user_id])
    :acl (into common-columns [:provider_id :user_id :acl_identity])
    :humanizer (into common-columns [:user_id])
-   :variable (into common-columns [:variable_name :measurement :user_id])})
+   :variable (into common-columns [:variable_name :measurement :user_id])
+   :variable-association (into common-columns
+                               [:associated_concept_id :associated_revision_id
+                                :variable-name :user_id])})
```
chris-durbin: I think this column should be `:variable_name` with an underscore

metadata-db-app/src/migrations/050_setup_variable_associations_table.clj
```diff
@@ -0,0 +1,60 @@
+(ns migrations.050-setup-variable-associations-table
+  (:require
+   [config.mdb-migrate-helper :as h]))
+
+(def ^:private variable-assocs-column-sql
+  "id NUMBER,
+  concept_id VARCHAR(255) NOT NULL,
+  native_id VARCHAR(1500) NOT NULL,
+  metadata BLOB NOT NULL,
```
chris-durbin: no tombstones which have NULL metadata?

metadata-db-app/src/cmr/metadata_db/data/oracle/search.clj
ygliuvt: Good catch. Fixed.

metadata-db-app/src/migrations/050_setup_variable_associations_table.clj
ygliuvt: This is a good observation. Our metadata field value is the gziped bytes of the original metadata and it is not NULL for empty string; in other words, we will never have NULL or empty string as value for our metadata field. This is true for all concept types in metadata-db, not just variable associations. 

I am guessing @jasongilman did it for a reason. If you want, we can file a separate issue to look into adding empty string handling in our utility lib.

indiejames: @ygliuvt that is incorrect. Tombstones have NULL for the metadata field for all concept types. Tag associations may be different from the rest, but I doubt it.

ygliuvt: @indiejames, We don't have NULL value as metadata for tombstones in production, it is always a 20 bytes BLOB which is the encrypted empty string. The production schema does allow NULL, but that is not from the db migration code. It must have been done manually as part of the migration to AWS. I agree that we should probably handle tombstone metadata more efficiently, but that is outside of the scope of this issue. I will leave it to @chris-durbin if he wants to file a separate issue or not.

metadata-db-app/int_test/cmr/metadata_db/int_test/utility.clj
```diff
@@ -1,22 +1,18 @@
...
+   [clj-http.conn-mgr :as conn-mgr]
+   [clj-time.format :as f]
+   [clojure.string :as string]
+   [clojure.test :refer :all]
+   [cmr.common-app.test.side-api :as side]
+   [cmr.common.util :as util]
+   [cmr.metadata-db.config :as mdb-config]
+   [cmr.metadata-db.services.concept-service :as concept-service]
+   [cmr.transmit.config :as transmit-config]
+   [inflections.core :as inf]))
```
oubiwann: Ah, I love seeing ns cleanup like this :-) Removing unused `require`s is good for one's (code) health ...

PR #88, opened 2017-05-26
CMR-4023 passes metadata db search POST params as form parameters

tgoff:


indiejames:
I guess one could ask how all the tests passed before, and is there a test to be added for this bug?

PR #89, opened 2017-05-26
CMR-4151 Make ingest granule/collection validation case-insensitive

daniel-zamora:
Fixes the granule validation to be lowercase as well.

### DIFF
--- a/indexer-app/src/cmr/indexer/data/concepts/attribute.clj
+++ b/indexer-app/src/cmr/indexer/data/concepts/attribute.clj
@@ -58,7 +58,7 @@
                               #(coll-psa/parse-value type %))
                         (:values psa-ref))}
        {:name (:name psa-ref)
-        (str field-name ".lowercase") (map (comp util/safe-lowercase
+        (str field-name ".lowercase") (map (comp str/lower-case
                                                  #(value->elastic-value type %)
                                                  #(coll-psa/parse-value type %))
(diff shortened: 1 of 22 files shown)


umm-lib/src/cmr/umm/validation/parent_weaver.clj
```diff
@@ -1,7 +1,8 @@
 (ns cmr.umm.validation.parent-weaver
   "Provides functions to thread together a granule and collection parent objects for validation.
   It weaves together the objects so matching items within the granule and collection are combined"
-  (:require [cmr.umm.umm-collection :as c]
```
laurenfrederick: You don't need changes to this file.

indexer-app/src/cmr/indexer/data/concepts/attribute.clj
```diff
@@ -70,9 +70,9 @@
   "Converts the psa-refs into a list of elastic documents"
   [collection granule]
   (let [parent-type-map (into {} (for [aa (:AdditionalAttributes collection)]
-                                   [(:Name aa) (csk/->kebab-case-keyword (:DataType aa))]))]
+                                   [(str/lower-case (:Name aa)) (csk/->kebab-case-keyword (:DataType aa))]))]
```
laurenfrederick: Wondering if we should use safe-lowercase everywhere.

PR #99, opened 2017-05-31
CMR-4169 - Removed SERF

oubiwann:
This is a prerequisite for adding the new service schema.

The old service schema also needs to be removed, but that will happen in a subsequent PR. @chris-durbin is removing the old services from metadata-db; there might be a tiny bit of overlap between his forthcoming PR on that and this one: this PR does update one line of code in the metadata-db (removing the SERF content type reference).


oubiwann:
Thanks, guys!

PR #101, opened 2017-05-31
CMR-4169: Remove old service schema

oubiwann:
This is follow-up to the SERF-removal PR (that one should be done first, and merged, so that this PR's diff only shows the service removal code).

The next PR in this series will be adding the new service schema.


oubiwann:
> Note - I think we're going to want to add back in the same changes to json_schema.clj 
> and we'll want the same generative testing for UMM-S eventually, but it is easy enough 
> to just add back.

Yup -- doing that right now in the new service branch. Also, it seems that the var schema didn't have this bit added, so I'm including that too (in the tests, that is).

> Also we have multiple copies of the umm-s-json-schema.json (in every revision of UMM 
> in the repo). We should remove all of those.

Ah, I didn't realize we were removing those too -- will do so, and then push.

oubiwann:
@chris-durbin: before I remove all those in the previous versions ... are we going to also update the previous versions to use the new service schema? Or are we only going to have the new service schema be in v1.9?

chris-durbin:
Ideally we would version UMM-S independently. But, if we continue the same versioning approach we should use the new UMM-S schema for all existing versions.

oubiwann:
Okay, in that case I'll over-write the previous versions' service schemas with the new one in the branch where I'm adding the schema.

oubiwann:
Thanks, guys!

PR #102, opened 2017-05-31
CMR-4169: Add new service schema

oubiwann:
This change adds the new service schema, updating the tests accordingly.

It also:
 * adds a missing test for the umm-var schema
 * updates all the old versions of the service schema to use the new one
 
This PR follows from PRs #99 and #101.


oubiwann:
Sweetness! Thanks again :-)

PR #105, opened 2017-06-01
CMR-4152: Request logs are still logging the incorrect time in NGAP.

chris-durbin:
Explicitly set the time zone via configuration.

### DIFF
--- a/common-lib/src/cmr/common/api/web_server.clj
+++ b/common-lib/src/cmr/common/api/web_server.clj
@@ -87,7 +87,15 @@
 
 (defconfig time-zone
   "The time zone where the web server is running. Needed because getting the time zone using
-   java.util.TimeZone is not returning the correct information on EC2 instances running in NGAP."
+   java.util.TimeZone is not returning the correct information on EC2 instances running in NGAP.
+
+   The default value of \"America/New_York\" will work for all of our NGAP EC2 instances (they are
+   all in US_EAST_1) and for developers working on the east coast.
+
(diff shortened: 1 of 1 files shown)


common-lib/src/cmr/common/api/web_server.clj
```diff
@@ -86,6 +85,19 @@
...
+   java.util.TimeZone is not returning the correct information on EC2 instances running in NGAP.
+
+   The default value of \"America/New_York\" will work for all of our NGAP EC2 instances (they are
+   all in US_EAST_1) and for developers working on the east coast.
+
+   If you enable access logging locally and need to set a different time zone you can figure out
+   the string to set by running:
+
+   (.getID (java.util.TimeZone/getDefault)))"
+  {:default "America/New_York"})
```
oubiwann: We should probably create a ticket to update this to work with an ENV var; we're going to forget this in a moment of panic when we need to deploy CMR to, e.g., US West 2 ....

chris-durbin: Yup, that's exactly what defconfig does. The environment variable that controls this is CMR_TIME_ZONE. We won't need to set it in any of our existing environment configurations because the default works, but if we deploy to US West 2 we would just set CMR_TIME_ZONE="America/Los_Angeles" in our configuration.

oubiwann: Oh, sweet! This is a great example for me to see the practical use of `defconfig` -- thanks!

PR #112, opened 2017-06-05
CMR 4172

ygliuvt:
Added support to save service concept in metadata-db.

### DIFF
--- a/metadata-db-app/README.md
+++ b/metadata-db-app/README.md
@@ -46,17 +46,14 @@
 
   {
       "concept-type": "service",
-      "native-id": "provider service id",
-      "concept-id": "S1-PROV1",
-      "provider-id": "PROV1",
+      "native-id": "mod10cm",
       "user-id": "user101",
-      "metadata": "xml here",
-      "format": "application/serf+xml",
+      "metadata": "service edn here",
+      "format": "application/edn",
       "revision-id": 1, //(optional field)
       "revision-date": "2012-01-01T00:00:00", //(optional field)
       "extra-fields": {
-        "entry-id": "short_V01"
-        "entry-title": "Dataset V01"
+        "service-name": "MOD10CM"
       }
     }
 
--- a/metadata-db-app/int_test/cmr/metadata_db/int_test/concepts/service_save_test.clj
+++ b/metadata-db-app/int_test/cmr/metadata_db/int_test/concepts/service_save_test.clj
@@ -3,7 +3,7 @@
   checking for proper error handling."
   (:require
    [clojure.test :refer :all]
-   [cmr.common.util :refer (are2)]
+   [cmr.common.util :refer (are3)]
    [cmr.metadata-db.int-test.concepts.concept-save-spec :as c-spec]
    [cmr.metadata-db.int-test.utility :as util]))
 
@@ -21,10 +21,9 @@
 (deftest save-service-test
   (c-spec/general-save-concept-test :service ["CMR"]))
 
-
 (deftest save-service-specific-test
   (testing "saving new services"
-    (are2 [service exp-status exp-errors]
+    (are3 [service exp-status exp-errors]
           (let [{:keys [status errors]} (util/save-concept service)]
             (is (= exp-status status))
             (is (= (set exp-errors) (set errors))))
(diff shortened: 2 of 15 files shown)


metadata-db-app/int_test/cmr/metadata_db/int_test/concepts/service_save_test.clj
```diff
@@ -0,0 +1,35 @@
...
+
+;; tests
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+(deftest save-service-test
+  (c-spec/general-save-concept-test :service ["CMR"]))
+
+
+(deftest save-service-specific-test
+  (testing "saving new services"
+    (are2 [service exp-status exp-errors]
```
daniel-zamora: Did you mean to use are3?

metadata-db-app/src/migrations/055_setup_services_table.clj
```diff
@@ -0,0 +1,58 @@
...
+  concept_id VARCHAR(255) NOT NULL,
+  native_id VARCHAR(1030) NOT NULL,
+  metadata BLOB NOT NULL,
+  format VARCHAR(255) NOT NULL,
+  revision_id INTEGER DEFAULT 1 NOT NULL,
+  revision_date TIMESTAMP WITH TIME ZONE DEFAULT SYSTIMESTAMP NOT NULL,
+  created_at TIMESTAMP WITH TIME ZONE DEFAULT SYSTIMESTAMP NOT NULL,
+  deleted INTEGER DEFAULT 0 NOT NULL,
+  user_id VARCHAR(30),
+  service_name VARCHAR(20),
```
chris-durbin: Seems too small - I know Simon mentioned changing variable name to 80 - I assume this will also change to 80.

metadata-db-app/src/migrations/055_setup_services_table.clj
```diff
@@ -0,0 +1,58 @@
...
+(defn- create-services-table
+  []
+  (h/sql
+   (format "CREATE TABLE METADATA_DB.cmr_services (%s, %s)"
+           services-column-sql services-constraint-sql)))
+
+
+(defn- create-services-indices
+  []
+  (h/sql "CREATE INDEX services_crdi ON METADATA_DB.cmr_services (concept_id, revision_id, deleted)")
```
chris-durbin: I've been thinking it might be a best practice to start documenting why we are creating specific indexes to make it easier for reviewers as well as to clarify for the author.

For example an index on just service_name would only fully satisfy the query "select service_name from cmr_services".

If we were actually issuing a query like "select concept_id from cmr_services where service_name = 'foo'" we'd want to include the concept_id as part of the index so that the index would fully cover the query.

I'm assuming the services table is small so it doesn't really matter, but for things like granules we should pay close attention.

ygliuvt: I think the indexes apply to joins and where clause, not he selects. In other words, it is not necessary to create index on concept_id and service_name to speed up "select concept_id from cmr_services where service_name='foo'". Am I wrong? If we select a lot of fields out of the table, the query engine might eventually choose a table scan rather than using the index, but that is a different issue.

metadata-db-app/int_test/cmr/metadata_db/int_test/concepts/service_save_test.clj
ygliuvt: Done.

metadata-db-app/README.md
```diff
@@ -46,17 +46,14 @@
...
       "concept-type": "service",
-      "native-id": "provider service id",
-      "concept-id": "S1-PROV1",
-      "provider-id": "PROV1",
+      "native-id": "mod10cm",
       "user-id": "user101",
-      "metadata": "xml here",
-      "format": "application/serf+xml",
+      "metadata": "service edn here",
+      "format": "application/edn",
```
ygliuvt: I have to say "I told you so". :)

metadata-db-app/src/migrations/055_setup_services_table.clj
ygliuvt: Agreed. I will change it to 80 now and save us a db migration down the road.

metadata-db-app/README.md
oubiwann: Question about that: do variable and service data need to support any non-JSON data types like dates or sets?

ygliuvt: My guess is it probably will eventually.

PR #114, opened 2017-06-06
CMR-4043 ACL search by user should not return ACLs for which the user has no permissions

daniel-zamora:
This pull request fixes ACL search by permitted user.  Originally it was returning ACLs with empty permission vectors.  This behavior is undesired.


oubiwann:
Question for you: was this issue only addressing how the tests were checking existing ACLs? The ticket seems to indicate there is more to the issue ... that updating the ACL API might be in order (i.e., ACL references that are returned).

PR #115, opened 2017-06-06
CMR-3975 Update ingest API docs to document bulk update functionality

laurenfrederick:
Added all of the bulk update documentation. It's kind of feeling like a wall of text right now, so any suggestions on how to break it up are appreciated!


ingest-app/docs/api.md
```diff
@@ -481,3 +481,145 @@ Example output:
...
+
+When a value is replaced by bulk update, only the field(s) specified in the update value will be replaced with the rest of the original value retained. For example, with a platform update value of {"ShortName": "A340-600"}, only the short name will be updated during a find and replace, while the long name, instruments, and other fields retain their values.
+
+If multiple bulk updates are run at the same time with the same concept-ids, there is no guarantee of the order that the updates will be performed on a collection. For example, if a clear all and replace is initiated, then an add to existing on the same collection, the clear all and replace could happen after the add to existing. Because of this, it is best to not run bulk update operations in parallel on overlapping collections.
+
+### Initiate Bulk Update
+
+Bulk update can be initiated by sending an HTTP POST request to `%CMR-ENDPOINT%/providers/<provider-id>/bulk-update/collections`
```
oubiwann: Quick question: you generated this locally and confirmed that `%CMR-ENDPOINT%` was what you expected (for a local run)? The reason I ask is that I want to make sure that the changes I made previously to the docs generation still support this ...

laurenfrederick: Yep! It did.

oubiwann: Phew!

ingest-app/docs/api.md
```diff
@@ -481,3 +481,145 @@ Example output:
     </gmd:identificationInfo>
 </gmi:MI_Metadata>
 ```
+## <a name="bulk-update"></a> Collection Bulk Update
+
+The bulk update API is used perform the same collection update to multiple concepts in one call. Bulk update is currently for collections only, granules are not supported.
+
+Bulk update is initiated through an ingest POST endpoint with the concept ids to update, the update type, the update field, and update information. The metadata is converted to the latest version of UMM, if not the native format, updated according to the parameters, and saved as the latest version of UMM, **making the native format of the collection now UMM**. Previous revisions of the collection are retained in the original native format.
```
indiejames: Does UMM here mean UMM-JSON?

laurenfrederick: Yeah I'll update those lines.

laurenfrederick: Fixed.

PR #116, opened 2017-06-06
CMR-4152: correct the request log time format to use 24 hour clock.

chris-durbin:
This change is going to work as expected in Splunk. It has to...


common-lib/src/cmr/common/api/web_server.clj
```diff
@@ -96,7 +95,7 @@
     (.setRequestLog
       (doto (NCSARequestLog.)
         (.setLogLatency true)
-        (.setLogDateFormat "yyyy-MM-dd hh:mm:ss.SSS")))))
```
oubiwann: Doh, sorry man -- I should have caught that last time.

PR #117, opened 2017-06-06
CMR-4001 expose search

mschmele:
Adds the search functionality requested in CMR-4001, as well as searching for granules by created-at


search-app/src/cmr/search/api/routes.clj
```diff
@@ -231,14 +238,44 @@
...
+                          (assoc :echo-collection-id collections-with-new-granules)
+                          (dissoc :has-granules-created-at)
+                          lp/process-legacy-psa)]
+    (find-concepts-by-parameters ctx path-w-extension search-params headers body)))
+
+(defn- multi-part-query?
+  "Return true if parameters match any from the list of `multi-part-query-params`
+   defined above. Supports CMR Harvesting."
+  [search-parameters]
+  (some true? (map #(contains? search-parameters %) multi-part-query-params)))
```
indiejames: This code is problematic because `some` will return `nil` if there is no match instead of `false`. It is idiomatic for functions ending in '?' to return a true boolean, not a truthy value. Also you are iterating through the list twice using `map` and `some`. I think a better approach might be

``` clojure
(boolean (some #(search-parameters %) multi-part-query-params))
```
Or if you make multi-part-query-params a set instead of a vector:

```clojure
(boolean (some multi-part-query-params (keys search-parameters))
```

search-app/src/cmr/search/api/routes.clj
```diff
@@ -231,14 +238,44 @@
...
+        search-params (-> params
+                          (assoc :echo-collection-id collections-with-new-granules)
+                          (dissoc :has-granules-created-at)
+                          lp/process-legacy-psa)]
+    (find-concepts-by-parameters ctx path-w-extension search-params headers body)))
+
+(defn- multi-part-query?
+  "Return true if parameters match any from the list of `multi-part-query-params`
+   defined above. Supports CMR Harvesting."
+  [search-parameters]
```
indiejames: Nit: `search-parameters` is a lot to type - how about `search-params`?

search-app/src/cmr/search/api/routes.clj
```diff
@@ -231,14 +238,44 @@
...
+
 (defn- find-concepts
-  "Invokes query service to find results and returns the response"
+  "Invokes query service to find results and returns the response.
+
+  This function supports several approaches/cases/whatevs for obtaining concept data:
+  * By JSON query
+  * By parameter string and URL parameters
+  * Collections from Granules - due to the fact that ES doesn't suport joins in the way
+    that we need, we have to make two queires here to support CMR Harvesting. This can
```
indiejames: queries typo

search-app/src/cmr/search/api/routes.clj
```diff
@@ -231,14 +238,44 @@
...
+    later be generalized easily, should the need arise."
   [ctx path-w-extension params headers body]
   (let [content-type-header (get headers (string/lower-case common-routes/CONTENT_TYPE_HEADER))]
     (cond
       (= mt/json content-type-header)
       (find-concepts-by-json-query ctx path-w-extension params headers body)
 
+      (multi-part-query? params)
```
indiejames: I appreciate you wanting to leave room to generalize this later, but right now I find it a bit confusing that you have a predicate checking if it's a multipart query jumping straight to `find-granule-parent-collections`. It would make more sense to me if you stopped trying to generalize it and just called the predicate `granule-parent-collection-query?` or if you adde a level of indirection here with an intermediate function like `find-concepts-with-multi-query` that conditionally calls the right function. For now the condition would always call `find-granule-parent-collections`.

search-app/src/cmr/search/api/routes.clj
```diff
@@ -231,14 +238,44 @@
...
+
+  This function supports several approaches/cases/whatevs for obtaining concept data:
+  * By JSON query
+  * By parameter string and URL parameters
+  * Collections from Granules - due to the fact that ES doesn't suport joins in the way
+    that we need, we have to make two queires here to support CMR Harvesting. This can
+    later be generalized easily, should the need arise."
   [ctx path-w-extension params headers body]
   (let [content-type-header (get headers (string/lower-case common-routes/CONTENT_TYPE_HEADER))]
     (cond
```
indiejames: When I see a `cond` starting to fill up with predicates like this I start to think maybe it's time to create a multimethod.

search-app/src/cmr/search/services/parameters/parameter_validation.clj
```diff
@@ -694,11 +697,17 @@
     :revision_date
     :result-format})
 
+(defn- filter-unrecognized-parameters
```
indiejames: This function is very similar to `unrecognized-params-validation` in `comm-app.services.search.parameter-validation`. I recommend putting this function in that namespace and calling it from there and here.

search-app/docs/api.md
```diff
@@ -43,6 +43,7 @@ Join the [CMR Client Developer Forum](https://wiki.earthdata.nasa.gov/display/CM
     * [Updated since](#c-updated-since)
     * [Revision date](#c-revision-date)
     * [Created at](#c-created-at)
+    * [Collections with new Granules](c-with-new-granules)
```
mschmele: Good catch - fixed

search-app/src/cmr/search/api/routes.clj
mschmele: I tend to default to longer parameter names because autocomplete makes it trivial to type them, but I see your point. Fixed

search-app/src/cmr/search/api/routes.clj
mschmele: Ah, old JS habits die hard - fixed

search-app/src/cmr/search/api/routes.clj
mschmele: Fixed

search-app/src/cmr/search/api/routes.clj
mschmele: Good call - I'll go the less general route for now

search-app/docs/api.md
```diff
@@ -1186,6 +1188,12 @@ Find collections that match all of the 'project' param values
 
    curl "%CMR-ENDPOINT%/collections?created_at\[\]=2000-01-01T10:00:00Z,2010-03-10T12:00:00Z&created_at\[\]=2015-01-01T10:00:00Z,"
 
+#### <a name="c-with-new-granules"></a> Find collections with new granules
+
+  Find collections containing granules added after a given date. The datetime has to be in yyyy-MM-ddTHH:mm:ssZ format.
+
+  curl "%CMR_ENDPOINT%/collections?has-granules-added-after=\[\]2015-01-01T10:00:00Z"
```
chris-durbin: All of our query parameters on CMR APIs should use underscores rather than dashes.

chris-durbin: How difficult would it be to support begin and end ranges just like the granules created_at API (and call the parameter `granules_created_at`)?

search-app/src/cmr/search/api/routes.clj
```diff
@@ -231,14 +238,44 @@
...
+  "Return true if parameters match any from the list of `multi-part-query-params`
+   defined above. Supports CMR Harvesting."
+  [search-parameters]
+  (some true? (map #(contains? search-parameters %) multi-part-query-params)))
+
 (defn- find-concepts
-  "Invokes query service to find results and returns the response"
+  "Invokes query service to find results and returns the response.
+
+  This function supports several approaches/cases/whatevs for obtaining concept data:
```
chris-durbin: Note to self: update our style guide to ban the use of whatevs.

search-app/src/cmr/search/services/parameters/parameter_validation.clj
```diff
@@ -110,9 +110,10 @@
    :temporal (conj exclude-plus-and-or-option :limit-to-granules)
    :revision-date cpv/and-option
    :created-at cpv/and-option
+   :granule-created-at cpv/and-option
```
chris-durbin: Is this working as expected? I thought the parameter was `has-granules-created-at`.

system-int-test/test/cmr/system_int_test/search/collection_with_new_granule_search_test.clj
```diff
@@ -0,0 +1,104 @@
...
+                         :native-id (:EntryTitle deleted-collection)}
+        _ (ingest/delete-concept deleted-concept)
+
+        _ (dev-system-util/freeze-time! "2017-01-01T10:00:00Z")
+        youngling-collection (d/ingest-umm-spec-collection
+                               "PROV1"
+                               (data-umm-c/collection
+                                 {:EntryTitle "youngling"
+                                  :Version "v1"
+                                  :ShortName "Oldie 2"}))
```
chris-durbin: This should probably be "Youngster". Actually "Whippersnapper" is better.

system-int-test/test/cmr/system_int_test/search/collection_with_new_granule_search_test.clj
```diff
@@ -0,0 +1,104 @@
...
+      (let [references (search/find-concepts-with-param-string
+                         "granule"
+                         "created-at=2014-01-01T10:00:00Z,")]
+        (d/refs-match? [regular-granule young-granule] references)))
+    (testing "Using unsupported or incorrect parameters in conjunction with multi-part-query-params"
+      (are [params]
+        (let [{:keys [status errors]} (search/find-concepts-with-param-string
+                                        "collection" params)]
+          (= [400 [(format "Parameter [%s] was not recognized."
+                           (first (clojure.string/split params #"=")))]]
```
chris-durbin: Our convention is to refer `clojure.string :as string`.

mschmele: Fixed

system-int-test/test/cmr/system_int_test/search/collection_with_new_granule_search_test.clj
mschmele: Fixed :)

search-app/src/cmr/search/services/parameters/parameter_validation.clj
mschmele: Oh right - this request isn't flowing through the usual way. The request is working as expected, but it isn't using this validation

search-app/docs/api.md
mschmele: Fixed

search-app/src/cmr/search/api/routes.clj
mschmele: Fixed

search-app/docs/api.md
mschmele: Not very hard, it turns out. Added.

search-app/src/cmr/search/services/query_service.clj
```diff
@@ -263,6 +265,21 @@
         results (qe/execute-query context query)]
     (common-search/search-results->response context query results)))
 
+(defn new-granules->collection-ids
```
indiejames: You should rename this. A `->` in a function name implies a transformation from one thing to another. So `new-granules->collection-ids` implies the primary argument should be a collection of granules, which it not. We tend to overuse `->` names in the CMR IMHO, but when we do use them one of the arguments is always of the type to the left of the `->` in the name.

search-app/docs/api.md
```diff
@@ -1186,6 +1188,14 @@ Find collections that match all of the 'project' param values
 
    curl "%CMR-ENDPOINT%/collections?created_at\[\]=2000-01-01T10:00:00Z,2010-03-10T12:00:00Z&created_at\[\]=2015-01-01T10:00:00Z,"
 
+#### <a name="c-with-new-granules"></a> Find collections with new granules
+
+  This supports option `and`.
+
+  Find collections containing granules added within the range of datetimes. The datetime has to be in yyyy-MM-ddTHH:mm:ssZ format. The default is inclusive on the range boundaries.
+
+  curl "%CMR_ENDPOINT%/collections?has_granules_created_at=\[\]2015-01-01T10:00:00Z"
```
chris-durbin: Looks like this is now a range - should there be a trailing `,` to indicate no end date?

search-app/src/cmr/search/api/routes.clj
```diff
@@ -231,14 +238,44 @@
...
+                          (assoc :echo-collection-id collections-with-new-granules)
+                          (dissoc :has-granules-created-at)
+                          lp/process-legacy-psa)]
+    (find-concepts-by-parameters ctx path-w-extension search-params headers body)))
+
+(defn- granule-parent-collection-query?
+  "Return true if parameters match any from the list of `multi-part-query-params`
+   defined above. Supports CMR Harvesting."
+  [search-params]
+  (boolean (some granule-parent-collection-params (keys search-params))))
```
chris-durbin: Nit: isn't the idiomatic way to do this `some?` rather than `(boolean (some ...)`?

search-app/src/cmr/search/services/query_service.clj
```diff
@@ -263,6 +265,21 @@
...
+  (when-let [[start-date end-date] (mapv time-format/parse
+                                         (string/split (:has-granules-created-at params) #","))]
+    (let [query (qm/query {:concept-type :granule
+                           :condition (qm/date-range-condition
+                                       :created-at start-date end-date)
+                           :page-size :unlimited
+                           :result-format :query-specified
+                           :result-fields [:collection-concept-id]})
+          new-granules (:items (qe/execute-query context query))]
+      (map :collection-concept-id new-granules))))
```
indiejames: I'm not sure what you are doing with the return value for this function later, but you might want to remove duplicates here before returning this.

system-int-test/test/cmr/system_int_test/search/collection_with_new_granule_search_test.clj
```diff
@@ -0,0 +1,108 @@
...
+                         :native-id (:EntryTitle deleted-collection)}
+        _ (ingest/delete-concept deleted-concept)
+
+        _ (dev-system-util/freeze-time! "2017-01-01T10:00:00Z")
+        whippersnapper-collection (d/ingest-umm-spec-collection
+                                   "PROV1"
+                                   (data-umm-c/collection
+                                     {:EntryTitle "youngling"
+                                      :Version "v1"
+                                      :ShortName "Oldie 2"}))
```
chris-durbin: Oh I meant the ShortName "Oldie 2" didn't make sense for this collection.

system-int-test/test/cmr/system_int_test/search/collection_with_new_granule_search_test.clj
```diff
@@ -0,0 +1,108 @@
...
+(deftest search-for-collections-with-new-granules
+  (let [_ (dev-system-util/freeze-time! "2010-01-01T10:00:00Z")
+        oldest-collection (d/ingest-umm-spec-collection
+                            "PROV1"
+                            (data-umm-c/collection
+                              {:EntryTitle "oldie"
+                               :Version "v1"
+                               :ShortName "Oldie"}))
+
+        _ (dev-system-util/freeze-time! "2010-01-01T10:00:00Z")
```
indiejames: You don't need to call `freeze-time` before you save each one. Just call it before each group of concepts that you want to have the same timestamp.

search-app/docs/api.md
mschmele: Yes, and I've added that, but it will work without a trailing `,`

search-app/src/cmr/search/api/routes.clj
indiejames: No, that would do something different. `some` checks the items in a collection. `some?` checks a single item to see if it's not `nil`.

search-app/src/cmr/search/services/query_service.clj
mschmele: Good point - fixed

system-int-test/test/cmr/system_int_test/search/collection_with_new_granule_search_test.clj
mschmele: I found this to make writing the tests much easier, and I've seen it in multiple other integration tests. I'm happy to make the change, but IMO it's easier to read the tests when it's obvious what the creation dates will be

search-app/src/cmr/search/validators/all_granule_validation.clj
```diff
@@ -58,14 +58,14 @@
 (defn no-all-granules-with-scroll
   "Validates that the query is not an all granules query if it is a scroll query."
   [query]
-  (when (and (:scroll query) 
-             (all-granules-query? query) 
-             
+  (when (and (:scroll query)
+             (all-granules-query? query)
+             (not= :created-at (get-in query [:condition :field]))
```
chris-durbin: We need to talk about this a little bit more. I don't think we want to allow an all granules query for scrolling at all. The user would first submit a query to see which collections have updated granules, then provide the granule query with a list of concept-ids and scroll through those results. Let's talk about it tomorrow.

system-int-test/test/cmr/system_int_test/search/harvesting_test.clj
```diff
@@ -55,32 +67,50 @@
...
+                    :created-at "2010-01-01T10:00:00Z,2014-02-01T10:00:00Z"
+                    :scroll true
+                    :page-size 2}
+            options {:accept nil
+                     :url-extension "native"}
+            response (search/find-metadata :granule :echo10 params options)
+            scroll-id (:scroll-id response)]
+        (is (not (nil? scroll-id)))
+        (data2-core/assert-metadata-results-match format-key [g1-echo g2-echo] response)))
+    (testing "Harvest collections by created-at"
```
chris-durbin: @indiejames - I thought we only support the scroll parameter for granules. Do we also support scrolling for collections?

PR #120, opened 2017-06-08
CMR-4167 Variable updates

oubiwann:
I'd already done most of this in a previous branch; this work essentially adds (and corrects) some system integration tests and applies minor fixes to variable ingest code.

### DIFF
--- a/ingest-app/src/cmr/ingest/services/ingest_service.clj
+++ b/ingest-app/src/cmr/ingest/services/ingest_service.clj
@@ -357,7 +357,7 @@
         variable (as-> variable-json-str data
                        (concept-json->concept data)
                        (assoc data :originator-id user-id)
-                       (assoc data :native-id (:name data)))]
+                       (assoc data :native-id (string/lower-case (:name data))))]
     ;; Check if the variable already exists
     (if-let [concept-id (mdb2/get-concept-id context
                                              :variable
--- a/ingest-app/src/cmr/ingest/services/ingest_service.clj
+++ b/ingest-app/src/cmr/ingest/services/ingest_service.clj
@@ -357,7 +357,7 @@
         variable (as-> variable-json-str data
                        (concept-json->concept data)
                        (assoc data :originator-id user-id)
-                       (assoc data :native-id (string/lower-case (:name data))))]
+                       (assoc data :native-id (:name data)))]
     ;; Check if the variable already exists
     (if-let [concept-id (mdb2/get-concept-id context
                                              :variable
--- a/system-int-test/src/cmr/system_int_test/utils/variable_util.clj
+++ b/system-int-test/src/cmr/system_int_test/utils/variable_util.clj
@@ -22,7 +22,7 @@
   (f))
 
 (def sample-variable
-  {:Name "A name"
+  {:Name "A-name"
    :LongName "A long UMM-Var name"
    :Units "m"
    :DataType "float32"
--- a/system-int-test/src/cmr/system_int_test/utils/variable_util.clj
+++ b/system-int-test/src/cmr/system_int_test/utils/variable_util.clj
@@ -16,10 +16,28 @@
    [cmr.transmit.variable :as transmit-variable]))
 
 (defn grant-all-variable-fixture
-  "A test fixture that grants all users the ability to create and modify variables."
+  "A test fixture that grants all users the ability to create and modify
+  variables."
   [f]
   (echo-util/grant-all-variable (s/context))
   (f))
+
+(defn setup-update-acl
+  "Set up the ACLs for UMM-Var update permissions and return the ids+token"
+  [context]
+  (let [user-name "umm-var-user42"
+        group-name "umm-var-guid42"
+        update-group-id (echo-util/get-or-create-group context group-name)
+        update-token (echo-util/login context user-name [update-group-id])
+        token-context (assoc context :token update-token)
+        update-grant-id (echo-util/grant-group-admin token-context
+                                                     update-group-id
+                                                     :update)]
+    {:user-name user-name
+     :group-name group-name
+     :group-id update-group-id
+     :grant-id update-grant-id
+     :token update-token}))
 
 (def sample-variable
   {:Name "A-name"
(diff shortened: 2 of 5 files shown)


ingest-app/src/cmr/ingest/services/ingest_service.clj
```diff
@@ -357,7 +357,7 @@
         variable (as-> variable-json-str data
                        (concept-json->concept data)
                        (assoc data :originator-id user-id)
-                       (assoc data :native-id (string/lower-case (:name data))))]
```
oubiwann: This change causes @ygliuvt's variable association tests to fail; without this, the variable ingest tests fail. Will ping @ygliuvt to see where casing needs to be lowered ...

oubiwann: I've fixed this now.

system-int-test/src/cmr/system_int_test/utils/variable_util.clj
```diff
@@ -158,7 +156,7 @@
   and a user-id."
   [variable concept-id revision-id user-id]
   {:concept-type :variable
-   :native-id (string/lower-case (or (:variable-name variable) ""))
+   :native-id (or (:variable-name variable) "")
```
ygliuvt: `virable-name` of a variable will never be nil and `native-id` should be the lowercase of `variable-name`, so this line should be `:native-id (string/lower-case (:variable-name variable))`

system-int-test/test/cmr/system_int_test/ingest/variable_ingest_test.clj
```diff
@@ -93,15 +112,37 @@
...
+      (testing "disallowed update responses:"
+        (are3 [token expected]
+          (let [update-response (variable-util/update-variable
+                                 token
+                                 (string/lower-case (:Name variable-data))
+                                 variable-data)]
+            (is (= expected (:status update-response))))
+          "no token provided"
+          nil 401
+          "guest user denied"
```
chris-durbin: I was hoping to see a few more tests - apologies if one or more of these is tested and I missed it. Some of these rely on deletes, so I just want to make sure for that PR we have at least covered all 4 of these:

1) Create on an already existing concept fails
2) Update on a concept that doesn't exist (does this fail?)
3) Update on a deleted concept - latest existing revision is a tombstone (I assume this is allowed)
4) Create on a deleted concept - latest existing revision is a tombstone (I assume this is allowed)

transmit-lib/src/cmr/transmit/variable.clj
```diff
@@ -8,14 +8,15 @@
...
-;; URL functions
+;;; URL functions
 
 (defn- variables-url
   [conn]
   (format "%s/variables" (conn/root-url conn)))
 
 (defn- variable-url
   [conn variable-name]
+  ;; XXX debug
```
chris-durbin: remove

oubiwann: Done.

system-int-test/src/cmr/system_int_test/utils/variable_util.clj
oubiwann: Done.

system-int-test/test/cmr/system_int_test/ingest/variable_ingest_test.clj
oubiwann: Will do.

oubiwann: @chris-durbin I've updated this branch with the first two -- I will update the delete branch with the latest two.

oubiwann: Last two are now done in the delete branch.

system-int-test/test/cmr/system_int_test/ingest/variable_ingest_test.clj
```diff
@@ -12,28 +14,25 @@
...
+        (is (e/permitted? token
+                          grant-id
+                          group-id
+                          (ingest-util/get-ingest-update-acls token)))
         (is (= 201 status))
         (is (= 1 revision-id))
         (is (mdb/concept-exists-in-mdb? concept-id revision-id))
-        (is (= (:name variable-data)
-               (:name (mdb/get-concept concept-id revision-id))))
         ;; XXX Once CMR-4172 (metdata-db services support) was added, we tried to
```
ygliuvt: I am not sure this is related to ACL caching. We should try to debug this and we can at least work around the ACL caching issue if it is really it.

oubiwann: Gonna put some debugging time into this tonight.

oubiwann: We'll, I've made some progress: I've reduced it from 31 failing tests to 7. It seems the primary problem was system-related: the search-app's system set up the caches while the ingest-app did not.

oubiwann: Now all tests are passing again. 

However, still getting a null exception when I try to uncomment this one:

```clj
;; XXX Once CMR-4172 (metdata-db services support) was added, we tried to
;;     enable the following test; this required the work that we've since
;;     put into ticket CMR-4193, which in turn is probably blocked by an
;;     as-yet unfiled ticket for addressing what seems to be an ACL caching
;;     issue. Once those two tickets are resolved, this test will be
;;     enabled and should then pass.
(variable-util/assert-variable-saved variable-data
                                     user-name
                                     concept-id
                                     revision-id)
```

Figuring that out might have to wait until tomorrow ...

oubiwann: More good news: user-id and originator-id are now getting set properly, with the ingest caching set up and the acl-injection handler in place for ingest.

oubiwann: Okay, null exception is gone now. This test is working and it fails in the same way that @chris-durbin objected to in the delete-variable branch. Once we sort out what's up with the `:created-at` and `:extra-fields` for delete, this will be passing.

oubiwann: 😴

ygliuvt: Thanks for doing this!

PR #123, opened 2017-06-08
CMR-4168 variable deletes

oubiwann:
This adds delete support to UMM-Vars (ingest API).

This is based upon the branch that added update support to UMM-Vars; once that branch is merged, I will rebase this branch so that it's easier to read the diff.

### DIFF
--- a/ingest-app/src/cmr/ingest/services/ingest_service.clj
+++ b/ingest-app/src/cmr/ingest/services/ingest_service.clj
@@ -397,6 +397,22 @@
           (dissoc :revision-date :transaction-id)
           (update-in [:revision-id] inc)))))
 
+(defn delete-variable
+  "Deletes a tag with the given concept id"
+  [context variable-key]
+  (let [existing-concept (fetch-variable-concept context variable-key)]
+    (mdb/save-concept
+      context
+      (-> existing-concept
+          ;; Remove fields not allowed when creating a tombstone.
+          (dissoc :metadata :format :provider-id :native-id :transaction-id)
+          (assoc :deleted true
+                 :user-id (context-util/context->user-id
+                           context
+                           msg/token-required-for-variable-modification))
+          (dissoc :revision-date)
+          (update-in [:revision-id] inc)))))
+
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Service Ingest Functions
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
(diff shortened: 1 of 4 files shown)


ingest-app/src/cmr/ingest/services/ingest_service.clj
```diff
@@ -397,6 +397,22 @@
...
+    (mdb/save-concept
+      context
+      (-> existing-concept
+          ;; Remove fields not allowed when creating a tombstone.
+          (dissoc :metadata :format :provider-id :native-id :transaction-id)
+          (assoc :deleted true
+                 :user-id (context-util/context->user-id
+                           context
+                           msg/token-required-for-variable-modification))
+          (dissoc :revision-date :created-at :extra-fields)
```
chris-durbin: Created-at should be the same as the previous revision. What is in extra-fields? I'm thinking those should stay the same as well.

ingest-app/src/cmr/ingest/services/ingest_service.clj
```diff
@@ -397,6 +397,22 @@
...
 
+(defn delete-variable
+  "Deletes a tag with the given concept id"
+  [context variable-key]
+  (let [existing-concept (fetch-variable-concept context variable-key)]
+    (mdb/save-concept
+      context
+      (-> existing-concept
+          ;; Remove fields not allowed when creating a tombstone.
+          (dissoc :metadata :format :provider-id :native-id :transaction-id)
```
oubiwann: This was carried over from tags, I believe ... /me looks it up ... yup; just confirmed. Does removing that data make sense for tags but not variables? Perhaps @ygliuvt knows?

ingest-app/src/cmr/ingest/services/ingest_service.clj
oubiwann: Delete failed when these were present, so I removed them. 

@ygliuvt: if these are supposed to be here after delete, why is delete failing with them present? Is there something that needs to get updated to allows these to be present?

oubiwann:
@chris-durbin The first two of your number bullet list were addressed in the "update" branch; the second two in this branch.

It seems that updates on a delete result in a 404 and the message "Variable with variable-name 'XXX' was deleted."

However, attempting to create a variable on a tombstoned variable *does* work (status 200).

ingest-app/src/cmr/ingest/services/ingest_service.clj
ygliuvt: @oubiwann I don't know off the top of my head. I will have to debug it to find out.

ingest-app/src/cmr/ingest/services/ingest_service.clj
ygliuvt: The comment is a little off. We may want to change "not allowed" to "that are unnecessary". The native-id and provider-id, etc. will be set based on the value of the previous concept that is deleted in metadata-db, so we don't have to pass them over from the ingest-app.

ingest-app/src/cmr/ingest/services/ingest_service.clj
ygliuvt: This is because metadata-db will use the existing concept that is being deleted to set these fields. So we don't need to set it here.

chris-durbin:
> It seems that updates on a delete result in a 404 and the message "Variable with variable-name 'XXX' was deleted."

> However, attempting to create a variable on a tombstoned variable does work (status 200).

As long as this is consistent with other concept types I'm ok with it.

PR #125, opened 2017-06-09
CMR-4193 Add ACL caching for Ingest app

oubiwann:
In the sys-int tests for the initial implementation of variables and services, the user-id and originator-id weren't getting set. It turned out there was a context-updating request handler that the tag functionality in the search app was using to set these, and we needed to do the same.

As such, the functionality in question was moved into common-app, search was updated to use this new home, and ingest was updated to use it.

Note that this branch is based off of two others; when those are merged, I will update this one to make the diff easier to read.


ingest-app/src/cmr/ingest/system.clj
```diff
@@ -97,6 +98,10 @@
                                 (af/refresh-acl-cache-job "ingest-acl-cache-refresh")
                                 jvm-info/log-jvm-statistics-job))
               :caches {acl/token-imp-cache-key (acl/create-token-imp-cache)
+                       ;; Caches a map of tokens to the security identifiers
+                       context-augmenter/token-sid-cache-name (context-augmenter/create-token-sid-cache)
```
chris-durbin: I don't think we need the SIDs in ingest. 

I agree caching the user-id makes sense we are saving it with variable and service concepts, but I think the SIDs caching in search is because we pass the groups a user belongs to as part of the Elasticsearch query. We don't do that in ingest.

system-int-test/test/cmr/system_int_test/ingest/service_ingest_test.clj
```diff
@@ -34,6 +34,10 @@
         (is (mdb/concept-exists-in-mdb? concept-id revision-id))
         (is (= (:name service-data)
                (:name (mdb/get-concept concept-id revision-id))))
+        ;; XXX DEBUG
```
chris-durbin: remove debug

ingest-app/src/cmr/ingest/system.clj
oubiwann: Hrm; the way this was written (this work was originally done in the search app), if the sid caching isn't present, the acl lookups fail (I tried running the tests with the sid caching removed from the ingest system -- a no-go as it is currently implemented).

I'll look into this now to see how search and ingest can both use this code (or parts of it) ...

chris-durbin: Ok if it is assumed the SID cache is needed for user-id caching I am fine with adding it to ingest too.

system-int-test/src/cmr/system_int_test/utils/variable_util.clj
```diff
@@ -173,23 +173,26 @@
...
-   :user-id user-id
-   :deleted false
-   :concept-id concept-id
-   :revision-id revision-id})
+  (let [native-id (string/lower-case (:Name variable))]
+    {:concept-type :variable
+     :native-id native-id
+     :provider-id "CMR"
+     :format mt/edn
+     :metadata (pr-str (assoc (util/kebab-case-data variable)
```
oubiwann: @ygliuvt I've updated the `:metadata` value here to be what's expected ... but this is a little different than what the tag tests do. In particular, all the ingest tests expect wherever `variable` is passed that the variable data has not yet been kebab-cased. What are your thoughts?

oubiwann:
Note that I've squashed the redundant commit from CMR-4169 (that somehow made it into this branch). You'll see it no longer after the next push up to my branch ...

system-int-test/src/cmr/system_int_test/utils/variable_util.clj
ygliuvt: I think we can skip the checking on metadata in edn format here as we are changing it to JSON format and it would be much easier to compare when the metadata doesn't change.

ygliuvt: It is an indication that you should make the JSON change next. :)

oubiwann:
@ygliuvt I'm getting some tag/metadata-db test failures with this change:

```
Testing cmr.metadata-db.int-test.concepts.tag-association-save-test

FAIL in (save-tag-association-test) (concept_save_spec.clj:60)
save with proper revision-id save concept with revision id
expected: (t/after? revision-date2 revision-date1)
  actual: (not
           (t/after?
            #=(cmr.common.joda-time/date-time 1420106400000 "UTC")
            #=(cmr.common.joda-time/date-time 1420106400000 "UTC")))

FAIL in (save-tag-association-test) (concept_save_spec.clj:60)
save with skipped revisions save concept with revision id
expected: (t/after? revision-date2 revision-date1)
  actual: (not
           (t/after?
            #=(cmr.common.joda-time/date-time 1420106400000 "UTC")
            #=(cmr.common.joda-time/date-time 1420106400000 "UTC")))
```

Looking through tag tests ... haven't found the issue yet.

oubiwann:
Just updated this test: `cmr.system-int-test.admin.cache-api-test`; I will wait to push that change until I get more review feedback ...

chris-durbin:
```
(clj-time.format/unparse  (clj-time.format/formatters :date-time) (clj-time.coerce/from-long 1420106400000))
--------------------
"2015-01-01T10:00:00.000Z"
```
That looks like a hand constructed time, so it looks like time-keeper may have frozen time on you. I think clear-current-time! might be the right thing to do before the test (though not sure why it is necessary).

metadata-db-app/int_test/cmr/metadata_db/int_test/concepts/concept_save_spec.clj
```diff
@@ -45,6 +46,7 @@
   "Save a concept once then save it again with the provided revision id, validating that the
   response matches the expected response, and possibly that the second save succeeded."
   [concept exp-status new-revision-id exp-errors]
+  (time-keeper/clear-current-time!)
```
ygliuvt: Ideally, we should find the test that freezes the time without clearing it and fix it there. But this is OK.

PR #127, opened 2017-06-09
CMR-4103: Added functionality to respect all temporal ranges for coll…

sxu123:
…ection search.


indexer-app/src/cmr/indexer/data/concepts/collection.clj
```diff
@@ -49,6 +50,9 @@
       :else
       (errors/internal-error! (str "Unknown spatial representation [" coord-sys "]")))))
 
+(defn- update-values [m f & args]
```
mschmele: A docstring and more descriptive variable names would make this function a lot easier to read

sxu123:
1. Confirmed with Erich that we currently don't have any collections that contain the periodic temporals.
   So, I only dealt with the RangeDateTime and SingleDateTime cases.
2. Added fake nested temporals for the granule searches because they share the same code as the collection searches in complex_to_simple_converters.
3. Added nested  limit-to-granules-temporals as a special case to respect all the collection temporal ranges when there are no granules for the collection and the limit_to_granules is set.

indexer-app/src/cmr/indexer/data/concepts/collection.clj
```diff
@@ -194,7 +207,15 @@
         insert-time (index-util/date->elastic insert-time)
         coordinate-system (get-in collection [:SpatialExtent :HorizontalSpatialDomain
                                                        :Geometry :CoordinateSystem])
-        permitted-group-ids (get-coll-permitted-group-ids context provider-id collection)]
+        permitted-group-ids (get-coll-permitted-group-ids context provider-id collection)
+        {:keys [granule-start-date granule-end-date]} (cgac/get-coll-gran-aggregates context concept-id)
+        last-3-days (t/interval (t/minus (tk/now) (t/days 3)) (tk/now))
+        granule-end-date (if (and granule-end-date (t/within? last-3-days granule-end-date))
```
mschmele: You could change this `if` to use `when-not`
https://clojuredocs.org/clojure.core/when-not

indexer-app/src/cmr/indexer/data/concepts/collection.clj
indiejames: A more descriptive function name would be helpful as well.

indexer-app/src/cmr/indexer/data/concepts/collection.clj
sxu123: I will take a look. Just cut and pasted from another section of the file.

indexer-app/src/cmr/indexer/data/concepts/collection.clj
sxu123: Will add doctoring to describe it more. but I think update-values is all I need. because I just want to update all the values in a map.

indexer-app/src/cmr/indexer/data/concepts/collection.clj
```diff
@@ -183,6 +187,15 @@
         archive-center-names (keep meaningful-short-name-fn archive-centers)
         data-centers (map #(data-center/data-center-short-name->elastic-doc kms-index %)
                           (map str/trim (data-center/extract-data-center-names collection)))
+        ;; returns a list of {:start-date xxx :end-date yyy} 
+        temporal-extents (->> temporal-extents
+                              ;; converts each temporal-extent into a list of 
+                              ;; {:BeginningDateTime xxx :EndingDateTime xxx}
+                              (map spec-time/temporal-ranges)
+                              (flatten)
```
indiejames: I think you can just call `mapcat` instead of `map` followed by `flatten`.

indexer-app/src/cmr/indexer/data/concepts/collection.clj
```diff
@@ -234,6 +255,14 @@
             :instruments instruments-nested
             :archive-centers archive-centers
             :data-centers data-centers
+            :temporals temporal-extents
+            ;; added so that we can respect all collection temporal ranges in search 
+            ;; when limit_to_granules is set and there are no granules for the collection.
+            :limit-to-granules-temporals 
+              (if granule-start-date
+                [{:start-date (index-util/date->elastic granule-start-date)
+                  :end-date (index-util/date->elastic granule-end-date)}]
```
indiejames: Does a `granule-end-date` of `nil` mean no end date? More importantly, does `date->elastic` handle nil correctly?

umm-spec-lib/src/cmr/umm_spec/time.clj
```diff
@@ -4,6 +4,21 @@
...
+(defn temporal-ranges
+  "Returns all the :RangeDateTimes and :SingeDateTimes contained in the given TemporalExtent record,
+   as a list of {:BeginningDateTime xxx :EndingDateTime xxx}"
+  [temporal]
+  (let [ranges (:RangeDateTimes temporal)
+        ;; when :EndsAtPresentFlag is set to true, set :EndingDateTime for each range to nil.
+        ;; otherwise, set :EndingDateTime to nil if it doesn't exist.
+        normalized-ranges (if (:EndsAtPresentFlag temporal) 
+                            (map #(assoc % :EndingDateTime nil) ranges)
+                            (map #(if (= nil (:EndingDateTime %)) (assoc % :EndingDateTime nil) %) ranges))
```
indiejames: This line seems a little strange. Why is it necessary to set the value for the key `:EndingDateTime` to `nil` if it is already `nil` or not in the map? Also, you should use `nil?` instead of `(= nil` here.

indexer-app/src/cmr/indexer/data/concepts/collection.clj
sxu123: I think so. I just moved this part of functionality from another section of the code which is working.

umm-spec-lib/src/cmr/umm_spec/time.clj
sxu123: Since the result of this will need to map to the nested mapping which contains both the start-date and end-date, I just want to make sure both BeginningDateTime and EndingDateTime are present, even with nil value. I don't know what would happen to the nested mapping if an element is missing.

indexer-app/src/cmr/indexer/data/concepts/collection.clj
sxu123: yeah. thanks.

indexer-app/src/cmr/indexer/data/concepts/collection.clj
```diff
@@ -49,6 +50,9 @@
       :else
       (errors/internal-error! (str "Unknown spatial representation [" coord-sys "]")))))
 
+(defn- update-values [m f & args]
+  (reduce (fn [r [k v]] (assoc r k (apply f v args))) {} m))
```
indiejames: Nit: using `reduce-kv` instead of `reduce` here will allow you to avoid destructuring the key/value pair in the anonymous function.

sxu123: Thanks. will look into that.

sxu123: done.

indexer-app/src/cmr/indexer/data/concepts/collection.clj
sxu123: done.

umm-spec-lib/src/cmr/umm_spec/time.clj
sxu123: done.

indexer-app/src/cmr/indexer/data/concepts/collection.clj
sxu123: done.

sxu123:
This branch is messed up, replacing it with a new branch.

PR #130, opened 2017-06-12
CMR-4203 - Adding Characteristics name and value to the platform shortname uniqueness validation.

eereiter:
Platform instances such as aircraft, boats, balloons, etc. are not uniquely defined by the GCMD keyword management system though the short name field.  Therefore we needed to add Characteristics' name and value to compare not only the short name but also the instance identifier such as an aircrafts tail number.


umm-spec-lib/src/cmr/umm_spec/validation/platform.clj
```diff
@@ -19,7 +19,41 @@
...
+(def ^:private unique-fields
+  "Defines the fields to use that combined create unique platform names.
+  GCMD's keyword management system (KMS) defines aircraft, boats, ballons, etc.
+  with the model number. So the short name is not specific enough to identify
+  specific platform instances. Therefore we are adding Characteristic names and values
+  so that specific aircraft tail numbers, or boat IDS, etc. can be specified with the
+  short name."
+  #{:ShortName :Characteristics :Name :Value})
+
+(defn- get-platform-identifiers
```
mschmele: For organization sake, I would move this function, as well as the `unique-platform-validator` into the `vu` namespace that the original `unique-platform-validator` was being called from. Maybe call it something like `non-satellite-platform-validator`

chris-durbin: I actually asked Erich to leave it here. The other namespace was for generic validators that anyone could use. This validation is specific to platforms, so I think this is the right place.

mschmele: Ah fair. I was thinking in terms of keeping the validation in the same spot much like we do with the search app. Chances are if somebody is looking to fix a platform validation issue they'll end up looking here anyway, I suppose.

eereiter:
I have created and updated the tests.

umm-spec-lib/test/cmr/umm_spec/test/validation/platform.clj
```diff
@@ -31,7 +40,17 @@
...
+      (testing "duplicate platform short names with characteristics"
+        (let [coll (coll/map->UMM-C
+                     {:Platforms [(c/map->PlatformType {:ShortName "DHC-6"
+                                                        :Characteristics [c3]})
+                                  (c/map->PlatformType {:ShortName "DHC-6"
+                                                        :Characteristics [c3]})]})]
+          (h/assert-invalid
+            coll
+            [:Platforms]
+            ["Platforms must be unique. The combination of platform ShortName with Characteristics Name and Value contain duplicates named [DHC-6AircraftIDC-FSJB]."])))
```
chris-durbin: I'm fine with merging this in since there is urgency for NSIDC. We should write a separate clean up ticket to display the ShortName, Characteristics Name and Value as provided in the metadata rather than a single concatenated string of all 3 (DHC-6AircraftIDC-FSJB).

eereiter: I have created ticket CMR-4210

eereiter:
Merging in the Changes for CMR-4203

PR #132, opened 2017-06-14
CMR-4212: Updated db migration to change variable_name to VARCHAR(80) NOT NULL in cmr_variables table.

ygliuvt:


metadata-db-app/src/migrations/056_update_variable_name_length.clj
```diff
@@ -0,0 +1,17 @@
...
+  (println "migrations.056-update-variable-name-length up...")
+  (h/sql "alter table METADATA_DB.cmr_variables modify variable_name VARCHAR(80) NOT NULL"))
+
+
+(defn down
+  "Migrates the database down from version 56."
+  []
+  (println "migrations.056-update-variable-name-length down...")
+  ;; We don't want to roll back the length change
+  (h/sql "alter table METADATA_DB.cmr_variables modify variable_name VARCHAR(80) NULL"))
```
chris-durbin: I think reversing would be putting the length back to 20.

ygliuvt: I didn't want to do that because it would be likely (actually certain) that there will be variable_names longer than 20 chars when we try to roll back and it would fail. It also doesn't make much sense to truncate the value to 20 chars, so I decided to keep the length at 80.

PR #133, opened 2017-06-14
Fixing CMR-4047 - The DataCenter ShortName will no longer allow the >…

eereiter:
… character to be valid. This fixes the generated tests where the shortname is >.  > also is used in ISO as a separator between the short and long names.

### DIFF
--- a/common-lib/src/cmr/common/test/test_check_ext.clj
+++ b/common-lib/src/cmr/common/test/test_check_ext.clj
@@ -53,6 +53,9 @@
           (chuck/pass? reports#))))))
 
 (defmacro qc-and-report-exception-with-seed
+  "This macro re-runs a failed test so that it can capture the information to produce a failed test report.
+   The only difference between this function and the one above is that this one uses a seed value so that
+   the generated values of a UMM record are repeatable."
   [name seed final-reports tests bindings & body]
   `(chuck/report-exception-or-shrunk
     (test-check/quick-check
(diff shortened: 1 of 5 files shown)


umm-spec-lib/test/cmr/umm_spec/test/generate_and_parse.clj
```diff
@@ -170,6 +170,24 @@
       (is (= expected actual)
           (str "Unable to roundtrip with format " metadata-format)))))
 
+(deftest roundtrip-generated-collection-records-with-seed
+  (checking-with-seed "collection round tripping seed" 100 1496683985472
```
sxu123: Could you write a brief doc to describe how this test is intended to be used when the intermittent failure happens in the old roundtrip-generated-collection-records, which you want to preserve to reveal potential issues in the future?

common-lib/src/cmr/common/test/test_check_ext.clj
```diff
@@ -51,6 +52,23 @@
 
           (chuck/pass? reports#))))))
 
+(defmacro qc-and-report-exception-with-seed
+  [name seed final-reports tests bindings & body]
```
oubiwann: This really needs a docstring ...

eereiter: I copied the above function into the one that I modified that does the same thing.  I think when a test fails, it re-runs it and captures the information from the failed test so that it can generate a report on it.  I just figured out how to add the seed value so that the generated values for the UMM record can be repeated.  I will put this into the doc string.

eereiter: I have added the documentation

umm-spec-lib/test/cmr/umm_spec/test/generate_and_parse.clj
eereiter: I have added the documentation to this function and to the function before it.

PR #134, opened 2017-06-14
IP Feature - Provide support for an additional test runner

oubiwann:
This change adds support for another test runner, without affecting the current/default test runner's behaviour. A convenience function has been provided in dev-system's `user.clj` file, allowing a developer to simple call `(run-suites)` from the REPL in order to use the new test runner (note that unit tests and integration tests are put into separate suites).

These changes also will make it easier to add more (well-behaved`*`)  test runners in the future.

This PR also made the following changes:
 * Moved code from the existing namespace `cmr.common.test.test-runner` to the new, more general `cmr.common.test.runners.default`. To support backwards compatibility, `cmr.common.test.test-runner` was recreated, but with aliases for all the vars it used to have.
 * Moved common test runner functions into `cmr.common.test.runners.util` (notably the functions that locate all the unit and integration tests)
 * Added new functions to `cmr.common.test.runners.util` that will help us audit CMR tests in the future

----
`*` "Well-behaved test runners shouldn't emulate the `clojure.test` API for `run-tests`, but should rather follow the example that the `eftest` and `ltest` test runners provide. In fact, the `ltest` example should be followed, as it sets the default reporter to `clojure.test` (the `ltest` API functions then override the reporter to use its own). The need for this rather awkward and cumbersome approach is due to the manner in which `clojure.test` was originally architected.

### DIFF
--- a/common-lib/src/cmr/common/test/runners/ltest.clj
+++ b/common-lib/src/cmr/common/test/runners/ltest.clj
@@ -1,0 +1,38 @@
+(ns cmr.common.test.runners.ltest
+  "This test runner is a CMR adaptor for the ltest Clojure test runner."
+  (:require
+   [clojure.stacktrace :as stack]
+   [clojure.string :as string]
+   [clojure.test :as test]
+   [cmr.common.test.runners.util :as util]
+   [ltest.core :as ltest]
+   [potemkin :refer [import-vars]]))
+
+(import-vars
+  [ltest.core
+   run-test
+   run-tests])
+
+(def unit-test-suite
+  {:name "Unit Tests"
+   :nss (util/unit-test-namespaces)})
+
+(def integration-test-suite
+  {:name "Integration Tests"
+   :nss (util/integration-test-namespaces)})
+
+(def test-suites
+  [unit-test-suite
+   integration-test-suite])
+
+(defn run-unit-tests
+  []
+  (ltest/run-suite unit-test-suite))
+
+(defn run-integration-tests
+  []
+  (ltest/run-suite integration-test-suite))
+
+(defn run-suites
+  []
+  (ltest/run-suites test-suites))
--- a/common-lib/src/cmr/common/test/test_runner.clj
+++ b/common-lib/src/cmr/common/test/test_runner.clj
@@ -1,0 +1,22 @@
+(ns cmr.common.test.test-runner
+  "DEPRECATED
+
+  This code has moved to `cmr.common.test.runners.default`; please update your
+  text editor's aliases to use the new namespace.
+
+  This namespace is currently maintained for backwards compatibility."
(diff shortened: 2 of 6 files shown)


common-lib/src/cmr/common/test/runners/ltest.clj
```diff
@@ -0,0 +1,38 @@
...
+  "This test runner is a CMR adaptor for the ltest Clojure test runner."
+  (:require
+   [clojure.stacktrace :as stack]
+   [clojure.string :as string]
+   [clojure.test :as test]
+   [cmr.common.test.runners.util :as util]
+   [ltest.core :as ltest]
+   [potemkin :refer [import-vars]]))
+
+(import-vars
```
chris-durbin: I took a look at the potemkin documentation and wasn't entirely clear on what the import-vars macro does for us. Could you explain?

common-lib/src/cmr/common/test/test_runner.clj
```diff
@@ -1,149 +1,25 @@
...
-    (reset! last-test-results test-results)
-    (when reset-fn (reset-fn))
-    (print-results (assoc (analyze-results test-results)
-                          :took took)
-                   options)))
-
-
-
-
+  "DEPLRECATED
```
chris-durbin: It's like a REPL

oubiwann: Oh, man. I totally lol'ed.

Still laughing ...

common-lib/src/cmr/common/test/runners/ltest.clj
oubiwann: Yeah, in general, potemkin lets one maintain a user-facing API without being tied to a particular namespace structure and layout (by importing vars).

In our case:

* we get a new home with the real code that gets updates/maintenance/etc., while we can forget the deprecated ns (the `test-runner` namespace)
* for this namespace, we get to pull in the var itself (including docstring, so this would show up in codox) which means that users can call `ltest/run-test` and `ltest/run-tests` without us having to implement a wrapper in the namespace (which, really, ties in with the API maintenance point I made above)

common-lib/src/cmr/common/test/test_runner.clj
oubiwann: Fixed.

oubiwann:
Just re-ran tests to be sure. All good.

common-lib/src/cmr/common/test/runners/ltest.clj
indiejames: It also has the one really annoying side-effect of breaking jump-to-definition. Not a big deal for a utility namespace like this, but really annoying when you are trying to navigate the code for an API to see how things are implemented.

PR #140, opened 2017-06-16
CMR-4135: Disabled Gorilla Repl.

sxu123:


dev-system/src/cmr/dev_system/system.clj
```diff
@@ -318,9 +318,7 @@
      :pre-components (u/remove-nil-keys
```
mschmele: Would we want to get rid of the `require` and the `defconfig` that sets the port on line 285?

sxu123:
@mschmele, @chris-durbin, Thanks for reviewing. I really wanted to preserve the code just in case we still need it, which was why I didn't cleanup the gorilla related code. I like Chris' clean approach. Thanks.

PR #149, opened 2017-06-22
CMR-4292 CMR-4293 CMR-4163

daniel-zamora:
This PR addresses:
CMR-4292 fixes pump interop, CMR-4293 fixes permitted group ids in collection index, CMR-4163 syncs entry-titles and concept-ids on acl create and update

### DIFF
--- a/access-control-app/int_test/cmr/access_control/int_test/acl_crud_test.clj
+++ b/access-control-app/int_test/cmr/access_control/int_test/acl_crud_test.clj
@@ -1018,3 +1018,117 @@
                                    (u/conn-context) acl-concept-id {:token token :raw? true})]
         (is (= 401 status))
         (is (= ["Permission to delete ACL is denied"] (:errors body)))))))
+
+(deftest entry-titles-concept-ids-sync
+  (let [token (e/login-guest (u/conn-context))
+        make-catalog-item (fn [name coll-id]
+                            (-> catalog-item-acl
+                                (assoc-in [:catalog_item_identity :name] name)
+                                (assoc-in [:catalog_item_identity :collection_identifier] coll-id)))
+        coll1 (u/save-collection {:entry-title "coll1 entry title"
+                                  :short-name "coll1"
+                                  :native-id "coll1"
+                                  :provider-id "PROV1"})
+
+        coll2 (u/save-collection {:entry-title "coll2 entry title"
+                                  :short-name "coll2"
+                                  :native-id "coll2"
+                                  :provider-id "PROV1"})
+
+        coll3 (u/save-collection {:entry-title "coll3 entry title"
+                                  :short-name "coll3"
+                                  :native-id "coll3"
+                                  :provider-id "PROV1"})
+
+        coll4 (u/save-collection {:entry-title "coll4 entry title"
+                                  :short-name "coll4"
+                                  :native-id "coll4"
+                                  :provider-id "PROV1"})
+
+        acl1 (ac/create-acl (u/conn-context)
+                            (make-catalog-item "acl1" {:entry_titles ["coll1 entry title"
+                                                                      "coll2 entry title"
+                                                                      "coll3 entry title"]})
+                            {:token token})
+
+        acl2 (ac/create-acl (u/conn-context)
+                            (make-catalog-item "acl2" {:concept_ids [coll1 coll2 coll3]})
+                            {:token token})
+
+        acl3 (ac/create-acl (u/conn-context)
+                            (make-catalog-item "acl3" {:concept_ids [coll1 coll2 coll3]
+                                                       :entry_titles ["coll1 entry title"
+                                                                      "coll2 entry title"
+                                                                      "coll3 entry title"]})
+                            {:token token})
+
+        acl4 (ac/create-acl (u/conn-context)
+                            (make-catalog-item "acl4" {:concept_ids [coll1 coll2]
+                                                       :entry_titles ["coll2 entry title"
+                                                                      "coll3 entry title"]})
+                            {:token token})
+        expected-collection-identifier {:concept_ids [coll3 coll2 coll1]
+                                        :entry_titles ["coll3 entry title"
+                                                       "coll2 entry title"
+                                                       "coll1 entry title"]}
+        expected-collection-identifier2 {:concept_ids [coll4 coll3 coll2 coll1]
+                                         :entry_titles ["coll4 entry title"
+                                                        "coll3 entry title"
+                                                        "coll2 entry title"
+                                                        "coll1 entry title"]}]
+    (testing "create acls"
+      (are3 [concept-id]
+        (is (= expected-collection-identifier
+               (get-in (ac/get-acl (u/conn-context) concept-id {:token token})
+                       [:catalog_item_identity :collection_identifier])))
+
+        "Only entry-titles"
+        (:concept_id acl1)
+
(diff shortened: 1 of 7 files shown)


metadata-db-app/src/migrations/052_add_coll_ids_catalog_acls.clj
```diff
@@ -48,20 +48,4 @@
 (defn down
   "Migrates the database down from version 52."
   []
-  (println "migrations.052-add-coll-ids-catalog-acls down...")
```
laurenfrederick: Why are the down migrations removed?

access-control-app/src/cmr/access_control/services/acl_util.clj
```diff
@@ -14,6 +14,7 @@
     [cmr.common-app.services.search.query-model :as qm]
     [cmr.transmit.echo.tokens :as tokens]
     [cmr.transmit.metadata-db2 :as mdb]
+    [cmr.transmit.metadata-db :as mdb1]
```
chris-durbin: Are we planning on keeping these two namespaces around indefinitely? @ygliuvt - do you know the plan?

access-control-app/src/cmr/access_control/services/acl_util.clj
```diff
@@ -102,3 +103,30 @@
...
+  "If the given ACL is a catalog item acl with a collection identifier that includes concept-ids or
+   entry-titles, return ACL such that both are a unioned with each other for :create action. If action is
+   :update and there are only entry-titles, then sync against entry-titles, otherwise sync against concept-ids.
+   If there are no concept-ids or entry-titles, acl remains unchanged."
+  [context action acl]
+  (if-let [collection-identifier (get-in acl [:catalog-item-identity :collection-identifier])]
+    (let [entry-titles (:entry-titles collection-identifier)
+          concept-ids (:concept-ids collection-identifier)
+          provider-id (get-in acl [:catalog-item-identity :provider-id])
+          colls-from-entry-titles (when (and (seq entry-titles)
```
chris-durbin: Why not `(when (seq entry-titles) ...`?

Basically this logic is saying to ignore entry titles when there is an update that included both concept-ids and entry titles.

As an API user this would be really confusing to me that you silently ignored my entry titles. If we don't want both entry-titles and concept-ids to come in an update request we should add a validation and reject the request instead.

metadata-db-app/src/migrations/052_add_coll_ids_catalog_acls.clj
chris-durbin: Daniel mentioned this could destroy data in the system making ACLs unrecoverable, so we agreed it should be removed.

access-control-app/int_test/cmr/access_control/int_test/acl_crud_test.clj
```diff
@@ -1018,3 +1018,117 @@
...
+                                                :entry_titles ["coll1 entry title"
+                                                               "coll2 entry title"
+                                                               "coll3 entry title"]})
+
+                     {:token token})
+      (is (= expected-collection-identifier2
+             (get-in (ac/get-acl (u/conn-context) (:concept_id acl4) {:token token})
+                     [:catalog_item_identity :collection_identifier]))))
+
+    (testing "update acls remove collection"
```
chris-durbin: Based on your code it seems like there should be two versions of this test - one the way you have it (entry titles and concept ids match) and one with 4 entry titles, but 3 concept ids.

access-control-app/int_test/cmr/access_control/int_test/acl_crud_test.clj
```diff
@@ -1018,3 +1018,117 @@
...
+                                  :short-name "coll3"
+                                  :native-id "coll3"
+                                  :provider-id "PROV1"})
+
+        coll4 (u/save-collection {:entry-title "coll4 entry title"
+                                  :short-name "coll4"
+                                  :native-id "coll4"
+                                  :provider-id "PROV1"})
+
+        acl1 (ac/create-acl (u/conn-context)
```
chris-durbin: Definitely not specific to this review (we tend to do this all the time), but I find the numbered collections and ACLs tough to review to understand what is being tested, what we want to happen, and whether that actually did happen. I prefer descriptive names like "acl-with-only-concept-ids", "acl-with-only-entry-titles", etc.

access-control-app/src/cmr/access_control/services/acl_util.clj
daniel-zamora: Well we want to keep them in sync, but it would get extremely complicated trying to parse the previous acl and the new acl to see what exactly changed in order to make sure both entry-titles and concept-ids are the same.  The idea is to remove entry-titles completely eventually I believe. So in the case where concept-ids are present in the update I use them as the last say on what should be in both entry-titles and concept-ids.  This simplifies things if our strategy is to try and make work what we are given.  We need entry-titles still because of legacy services.  But I also want to cover the case where the updated acl only has entry-titles, in that case I will use them because if I don't then that results in an empty list for both concept-ids and entry-titles.

We could force users not to use entry-titles at all through validation instead which work fine for MMT  but might break other things.

access-control-app/int_test/cmr/access_control/int_test/acl_crud_test.clj
daniel-zamora: This won't cause an update, I can add a test for this case but seeing your other comment I'm going to wait and see if I need to change how syncing is done.

access-control-app/int_test/cmr/access_control/int_test/acl_crud_test.clj
daniel-zamora: There are a lot of ACL tests that do the acl1 acl2 ... acln thing.  I agree with you but this has been baked in for a while in this app.

access-control-app/src/cmr/access_control/services/acl_util.clj
daniel-zamora: I could update the function to diff prev entry-titles and new entry-titles and do the same for concept-ids then union them, I avoided this because it seemed messy and onerous but it can be done.

metadata-db-app/src/migrations/052_add_coll_ids_catalog_acls.clj
daniel-zamora: yes, I explained this to her in slack.

access-control-app/int_test/cmr/access_control/int_test/acl_crud_test.clj
chris-durbin: The way the code is now, it appears you would remove the collection from the ACL (you ignore the entry-titles list when there are concept-ids).

It isn't clear to me which behavior is desirable. I assume we would't want to remove the collection (no update as you are saying), but to do that you would have to change the code to just `(when (seq entry-titles) ...` unless I'm missing something.

daniel-zamora: Because you are trying to sync entry-titles with concept-ids if you try and remove a collection from either via update it will always return when the collections are unioned.  This wouldn't affect the case where you are adding a collection to one or the other.

daniel-zamora: Perhaps this is ok.

PR #154, opened 2017-06-23
IP sprint cleanup

chris-durbin:
1. Remove Jetty debug logging by default
2. Improve REPL startup time by preventing JRuby runtime startup from blocking (one minute 20 seconds for me)
3. Fix warnings from lein modules tasks due to project.clj named files that are not Leiningen project files.


collection-renderer-lib/src/cmr/collection_renderer/services/collection_renderer.clj
```diff
@@ -30,12 +31,17 @@
   "gems/cmr_metadata_preview-0.0.1/.umm-version")
 
 (defn- create-jruby-runtime
```
mschmele: *thunderous applause*

PR #157, opened 2017-06-26
CMR-4222: Updated revision-date and MetadataDates/UPDATE date after b…

sxu123:
…ulk update.

### DIFF
--- a/ingest-app/docs/api.md
+++ b/ingest-app/docs/api.md
@@ -485,6 +485,8 @@ Example output:
 
 The bulk update API is used perform the same collection update to multiple concepts in one call. Bulk update is currently for collections only, granules are not supported.
 
+When we apply bulk update on a collection, either through API or MMT, regardless of if there is actual changes, a new collection record is created with a new revision-id, a revision-date with current date, and a updated Date field with current date, for "UPDATE" Type, in MetadataDates(add it if doesn't exist). 
+
 Bulk update is initiated through an ingest POST endpoint with the concept ids to update, the update type, the update field, and update information. The metadata is converted to the latest version of UMM, if not the native format, updated according to the parameters, and saved as the latest version of UMM-JSON, **making the native format of the collection now UMM-JSON**. Previous revisions of the collection are retained in the original native format.
 
 Updated collections are validated through ingest validation and the updates will not be saved if the updated collection is not valid UMM or fails business rule validation. The error will be recorded in the individual collection status, which can be queried via the status endpoint. Collection validation warnings will not prevent saving the updated collection and the warnings will be recorded in the individual collection status.
--- a/ingest-app/docs/api.md
+++ b/ingest-app/docs/api.md
@@ -485,9 +485,7 @@ Example output:
 
 The bulk update API is used perform the same collection update to multiple concepts in one call. Bulk update is currently for collections only, granules are not supported.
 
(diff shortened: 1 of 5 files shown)


umm-spec-lib/src/cmr/umm_spec/field_update.clj
```diff
@@ -72,5 +73,11 @@
         update-format (or update-format (:format concept))
         umm (spec-core/parse-metadata
              context concept-type format metadata {:sanitize? (= :umm-json (mt/format-key update-format))})
-        umm (apply-update update-type umm update-field update-value find-value)]
+        metadata-dates (:MetadataDates umm)
```
laurenfrederick: I'm not sure yet that we want to update the metadata and am talking to Kathy about it today. Please hold off on merging this change.

sxu123: I asked @eereiter about this before asking you. He got back to me later saying we need to update this field after bulk update.  He said the difference between the normal update through ingest and bulk update is that the normal update is through someone modifying the metadata file and ingest vs they modify the metadata file through our bulk update code. so we always have to update the MetadataDates/UPDATE - add it if it doesn't exist.   Now my only question is I'm not sure if we should keep multiple UPDATE field, or just having one latest one.

laurenfrederick: He asked Kathy about it and she asked to talk to me, so I am waiting to discuss with her. I would like to get an answer from Kathy, who is the creator of the ticket.

umm-spec-lib/src/cmr/umm_spec/field_update.clj
```diff
@@ -72,5 +73,11 @@
...
+        metadata-dates-without-update (remove #(= "UPDATE" (:Type %)) metadata-dates)
```
laurenfrederick: I think it would be cleaner and more re-usable if we had a function in `cmr.umm-spec.date-util` to update the update date and then we could call that. Also, we can remove this commented code.

umm-spec-lib/src/cmr/umm_spec/field_update.clj
sxu123: Had a meeting with everyone involved, the consent is to update/add the MetadataDates/UPDATE field. Done.

umm-spec-lib/src/cmr/umm_spec/field_update.clj
sxu123: Done. Also updated the ticket and ingest API docs.

laurenfrederick: Did you push the commit up? Still only seeing one commit.

ingest-app/docs/api.md
```diff
@@ -485,6 +485,8 @@ Example output:
 
 The bulk update API is used perform the same collection update to multiple concepts in one call. Bulk update is currently for collections only, granules are not supported.
 
+When we apply bulk update on a collection, either through API or MMT, regardless of if there is actual changes, a new collection record is created with a new revision-id, a revision-date with current date, and a updated Date field with current date, for "UPDATE" Type, in MetadataDates(add it if doesn't exist).
```
laurenfrederick: The wording is a bit confusing. 

I would remove "either through API or MMT". Change to "regardless if there are actual collections".

I would just say that a new revision is created. Move this all to after the next paragraph where we explain the conversion to UMM-C. Say in the UMM-JSON metadata, the Metadata Date of type "UPDATE" will be set to the current date.

PR #158, opened 2017-06-27
CMR-4284: Add back the temporal search using the nested index.

sxu123:


sxu123:
This just adds back the code that was removed temporarily just to wait for the granule index build.

PR #160, opened 2017-06-27
CMR-4285 Incorrect results on Find&Replace of Science Keywords when not all levels of hierarchy are specified in Find term

daniel-zamora:
This PR alters bulk update.

It moves the current implementation of find-and-replace to find-and-update
It changes find-and-replace to completely replace matched values with the update value instead of merging the original value and the updated value

### DIFF
--- a/ingest-app/docs/api.md
+++ b/ingest-app/docs/api.md
@@ -503,8 +503,8 @@ The following update types are supported:
 
   * Add to existing - the update value is added to the existing list. An update value is required.
   * Clear all and replace - clear the list and replace with the update value.
-  * Find and update - replace any instance in the list that matches the find value with the update value.
-  * Find and replace - merge update value into any instance in the list that matches the find value.
+  * Find and replace - replace any instance in the list that matches the find value with the update value.
+  * Find and update - merge update value into any instance in the list that matches the find value.
   * Find and remove - remove any instance from the list that matches the find value.
 
 Update types that include a find will match on the fields supplied. For example, for a science keyword update with a find value of {"Category": "EARTH SCIENCES"}, any science keyword with a category of "EARTH SCIENCES" will be considered a match regardless of the values of the science keyword topic, term, etc.
--- a/ingest-app/docs/api.md
+++ b/ingest-app/docs/api.md
@@ -503,7 +503,8 @@ The following update types are supported:
 
   * Add to existing - the update value is added to the existing list. An update value is required.
   * Clear all and replace - clear the list and replace with the update value.
-  * Find and replace - replace any instance in the list that matches the find value with the update value.
+  * Find and update - replace any instance in the list that matches the find value with the update value.
+  * Find and replace - merge update value into any instance in the list that matches the find value.
   * Find and remove - remove any instance from the list that matches the find value.
 
 Update types that include a find will match on the fields supplied. For example, for a science keyword update with a find value of {"Category": "EARTH SCIENCES"}, any science keyword with a category of "EARTH SCIENCES" will be considered a match regardless of the values of the science keyword topic, term, etc.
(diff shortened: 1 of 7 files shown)


system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
```diff
@@ -106,9 +106,9 @@
...
                       (:collection-statuses collection-response))))))
 
-    (testing "Data center find and replace"
+    (testing "Data center find and update"
       (let [bulk-update-body {:concept-ids concept-ids
-                              :update-type "FIND_AND_REPLACE"
+                              :update-type "FIND_AND_UPDATE"
                               :update-field "DATA_CENTERS"
                               :find-value {:ShortName "NSID"}
                               :update-value {:ShortName "NSIDC"
```
sxu123: Could you add a test for the new FIND_AND_REPLACE now that the old one becomes FIND_AND_UPDATE?

daniel-zamora: I added a test to the unit test.

sxu123: I noticed. but I think integration test is different.

daniel-zamora: The integration test doesn't cover everything.

ingest-app/docs/api.md
```diff
@@ -503,7 +503,8 @@ The following update types are supported:
 
   * Add to existing - the update value is added to the existing list. An update value is required.
   * Clear all and replace - clear the list and replace with the update value.
-  * Find and replace - replace any instance in the list that matches the find value with the update value.
+  * Find and update - replace any instance in the list that matches the find value with the update value.
```
laurenfrederick: I think these descriptions are swapped.

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
sxu123: Personally, adding a new type without integration test just doesn't sound right. unit test can help testing more details but can't replace integration test.

laurenfrederick: Probably a good idea to have 1 test for replace, just to make sure the update type gets converted properly and all.

daniel-zamora: The combination of tests already cover this, which is why I assume the integration test doesn't bother to cover all scenarios as written.  I didn't write the tests so I don't really know why it is so narrow.  I assume you did it this way for a reason?

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
```diff
@@ -134,3 +145,40 @@
                   :Roles ["PROCESSOR"]}]
                 (map #(select-keys % [:Roles :ShortName])
                      (:DataCenters (:umm concept))))))))))
+
+(deftest bulk-update-replace-test
+  (let [concept-ids (ingest-collection-in-each-format replace-keywords-umm)
```
sxu123: "replace-keywords-umm" should be "update-replace-keywords-umm"?

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
```diff
@@ -34,6 +34,17 @@
                   :ContactPersons [{:Roles ["Data Center Contact"]
                                     :LastName "Smith"}]}]})
 
+(def update-replace-keywords-umm
```
sxu123: Since we are separating the update and the replace, maybe it's better to name it more clearly?

PR #161, opened 2017-06-27
CMR-4238 Add provider-id index and constraint

oubiwann:
~It seems that the provider-id is already on the table (I got Oracle errors when trying to add the column).~ Actually, no -- that was due to some left overs in a messy testing session.

In addition to the migration changes and migration docs updates, I've included some other misc. docstring updates.

New to metadata-db, so I'm not sure if there's something I'm missing here ...

Note I intentionally did not touch the variable associations in this PR.

### DIFF
--- a/metadata-db-app/README.md
+++ b/metadata-db-app/README.md
@@ -165,7 +165,21 @@ _Note the absence of provider-id for tag, tag association, humanizer and variabl
 
 ### Setting up the database
 
-There are two ways database operations can be done. It can happen through leiningen commands for local development or using the built uberjar.
+There are two ways database operations can be done. It can happen through
+leiningen commands for local development or using the built uberjar.
+
+These commands run against the CMR database and as such, you will need
+Oracle installed or to be running an Oracle VM in order to perform migration
+tasks. Furthermore, the associated environment variables will need to be
+set. As mentioned in the top-level CMR `README` in the "Testing CMR" section,
+the required environemnt variables and their values can be provided to you by
+a CMR core dev.
+
+The CMR projects that currently support migrations are:
+
+* `ingest-app`
+* `bootstrap-app`
+* `metadata-db-app`
 
 #### leiningen commands
 
@@ -191,6 +205,16 @@ completely.
 lein drop-user
 ```
 
+#### CMR-wide
+
+The Oracle setup script in `dev-system/support` demonstrates how the above
+`lein` commands are used to create database users and run migrations for
+the ingest, boostrap, and metadata-db apps.
+
+Note that, depending upon how up-to-date your local `.jar` files are, you
+may first need to run `lein install-with-content!` in the top-level CMR
+directory first.
+
 #### java commands through uberjar
 
 1. Create the user
@@ -212,6 +236,7 @@ You can provider additional arguments to migrate the database to a given version
 ```
 CMR_DB_URL=thin:@localhost:1521:orcl CMR_METADATA_DB_PASSWORD=****** java -cp target/cmr-metadata-db-app-0.1.0-SNAPSHOT-standalone.jar cmr.db drop-user
 ```
+
 
 ## General Workflow
 
--- a/metadata-db-app/src/migrations/058_add_provider_id_to_variables_table.clj
+++ b/metadata-db-app/src/migrations/058_add_provider_id_to_variables_table.clj
@@ -2,6 +2,7 @@
   "Adds `provider_id` column to variables table."
   (:require
    [clojure.java.jdbc :as j]
+   [cmr.metadata-db.services.provider-validation :as provider-validation]
    [config.mdb-migrate-helper :as h]))
 
 (defn up
@@ -12,10 +13,10 @@
   ;; don't care about what's in the database right now, and the new field
   ;; can't be null, so:
   (h/sql "TRUNCATE TABLE METADATA_DB.cmr_variables")
-  (h/sql (str "ALTER TABLE METADATA_DB.cmr_variables "
-              "ADD provider_id VARCHAR(255) NOT NULL "
-              "ADD CONSTRAINT variables_p_uniq UNIQUE (provider_id)"))
-  (h/sql "CREATE INDEX variables_p_vn ON METADATA_DB.cmr_variables (provider_id, variable_name)"))
+  (h/sql "TRUNCATE TABLE METADATA_DB.cmr_variable_associations")
+  (h/sql (format (str "ALTER TABLE METADATA_DB.cmr_variables "
+                      "ADD provider_id VARCHAR(%s) NOT NULL")
+                 provider-validation/PROVIDER_ID_MAX_LENGTH)))
 
 (defn down
   "Migrates the database down from version 58 to 57."
--- a/metadata-db-app/src/migrations/058_add_provider_id_to_variables_table.clj
+++ b/metadata-db-app/src/migrations/058_add_provider_id_to_variables_table.clj
@@ -1,0 +1,24 @@
+(ns migrations.058-add-provider-id-to-variables-table
+  "Adds `provider_id` column to variables table."
+  (:require
+   [clojure.java.jdbc :as j]
+   [config.mdb-migrate-helper :as h]))
+
+(defn up
+  "Migrates the database up from version 57 to 58."
+  []
+  (println "migrations.058-add-provider-id-to-variables-table up...")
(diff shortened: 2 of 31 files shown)


metadata-db-app/src/migrations/058_add_provider_id_to_variables_table.clj
```diff
@@ -0,0 +1,24 @@
...
+   [config.mdb-migrate-helper :as h]))
+
+(defn up
+  "Migrates the database up from version 57 to 58."
+  []
+  (println "migrations.058-add-provider-id-to-variables-table up...")
+  ;; At this point in time, UMM-Vars support is still being developed, so we
+  ;; don't care about what's in the database right now, and the new field
+  ;; can't be null, so:
+  (h/sql "TRUNCATE TABLE METADATA_DB.cmr_variables")
```
chris-durbin: Let's go ahead and delete all of the variable associations as well since they are no longer valid.

metadata-db-app/src/migrations/058_add_provider_id_to_variables_table.clj
```diff
@@ -0,0 +1,24 @@
...
+  (h/sql (str "ALTER TABLE METADATA_DB.cmr_variables "
+              "ADD provider_id VARCHAR(255) NOT NULL "
```
chris-durbin: It looks like we are using 10 characters in some tables and 255 in others for the provider-id. 10 characters is the correct max allowed length:

From `cmr.metadata-db.services.provider-validation`
`(def ^:private ^:const PROVIDER_ID_MAX_LENGTH 10)`

metadata-db-app/README.md
```diff
@@ -165,7 +165,21 @@ _Note the absence of provider-id for tag, tag association, humanizer and variabl
 
 ### Setting up the database
 
-There are two ways database operations can be done. It can happen through leiningen commands for local development or using the built uberjar.
+There are two ways database operations can be done. It can happen through
```
mschmele: I really like the updates that you've made to the app-level readme files

metadata-db-app/src/migrations/058_add_provider_id_to_variables_table.clj
```diff
@@ -0,0 +1,24 @@
...
+              "ADD CONSTRAINT variables_p_uniq UNIQUE (provider_id)"))
```
chris-durbin: We should not have a unique constraint on provider_id. All of the variables for a given a provider will have the same value (potentially 10s of thousands).

metadata-db-app/src/migrations/058_add_provider_id_to_variables_table.clj
```diff
@@ -0,0 +1,24 @@
...
+  (h/sql "CREATE INDEX variables_p_vn ON METADATA_DB.cmr_variables (provider_id, variable_name)"))
```
chris-durbin: What query is this index supporting? It probably makes sense that we will be looking up variables by provider_id and variable_name, but just want to make sure we've thought it through.

oubiwann:
> I don't understand why "all" is added in this ticket, but adding the new project-id column to the table looks fine.

@sxu123 Those changes are unrelated but something I noticed recently; without the `all`, when checking for deps, only the deps themselves are looked at, not the plugins. This has meant that we've been overlooking looking them when they're out-dated :-/

metadata-db-app/src/migrations/058_add_provider_id_to_variables_table.clj
oubiwann: So you mean `TRUNCATE` in the associations table, right? (just to be super-clear).

metadata-db-app/src/migrations/058_add_provider_id_to_variables_table.clj
oubiwann: Oooh. Nice.

metadata-db-app/src/migrations/058_add_provider_id_to_variables_table.clj
oubiwann: Ah, right -- I'll remove that line.

metadata-db-app/src/migrations/058_add_provider_id_to_variables_table.clj
oubiwann: I was trying to address the very general comment in the ticket: 
 * "Add unique index on variable_name and provider_id pair in cmr_variables table"

I have *no* idea how this will be used ...

metadata-db-app/src/migrations/058_add_provider_id_to_variables_table.clj
oubiwann: Ugh, but why is that private?! We should be using it ... /me looks in the ns to see if there's another way to use that value ...

oubiwann: That's the only place I could find 'provider' and 'max' on the same line in the code base, so I made those two constants public and required the ns in the migration. Let me know if you think these constants should be moved someplace more general ...

metadata-db-app/src/migrations/058_add_provider_id_to_variables_table.clj
oubiwann: Done.

metadata-db-app/src/migrations/058_add_provider_id_to_variables_table.clj
oubiwann: Done (if that's what you meant).

metadata-db-app/src/migrations/058_add_provider_id_to_variables_table.clj
```diff
@@ -0,0 +1,26 @@
...
+  (println "migrations.058-add-provider-id-to-variables-table up...")
+  ;; At this point in time, UMM-Vars support is still being developed, so we
+  ;; don't care about what's in the database right now, and the new field
+  ;; can't be null, so:
+  (h/sql "TRUNCATE TABLE METADATA_DB.cmr_variables")
+  (h/sql "TRUNCATE TABLE METADATA_DB.cmr_variable_associations")
+  (h/sql (format (str "ALTER TABLE METADATA_DB.cmr_variables "
+                      "ADD provider_id VARCHAR(%s) NOT NULL")
+                 provider-validation/PROVIDER_ID_MAX_LENGTH))
+  (h/sql "CREATE INDEX variables_p_vn ON METADATA_DB.cmr_variables (provider_id, variable_name)"))
```
chris-durbin: It looks like my comment about our indexes never made it through. I just want to double-check what query this index is supporting. When we search by variable name are we going to also pass the provider-id?

I think this existing index should be dropped and we should add provider-id to it (I'm assuming searching for variables in metadata-db will include the provider-id):
`(h/sql "CREATE INDEX variables_crdi ON METADATA_DB.cmr_variables (concept_id, revision_id, deleted)")`

Having said that it probably makes sense to add a database migration changing the indexes in the same PR as the updates to metadata-db searching for variables now that we include provider-id. Are we planning to do that as part of CMR-4239 or CMR-4240?

oubiwann: Thanks for the detailed response!

As for changing the search indices, yeah: this is just the first metadata-db change. We have some more coming. Then we're going to change the routes to include support for provider-id getting passed in the REST-ish APIs (ingest and search). We're not at the point where a user can search variables by provider-id yet, but once we are, we'll be updating the necessary migrations.

Do you see there being a negative impact on the db relationships if we make the current change without updating the search internals?

metadata-db-app/src/migrations/058_add_provider_id_to_variables_table.clj
sxu123: So the fact you could truncate the cmr_variables table without truncating the associations table first means there's no foreign key constraint?  should we have it? I'm not familiar with these tables but trying to learn since I'm gonna work on another umm var ticket.

metadata-db-app/src/migrations/058_add_provider_id_to_variables_table.clj
oubiwann: Just pushed up the change to not do an index right now (will wait until we change the search API to include provider-id).

oubiwann:
Thanks, everyone!

PR #162, opened 2017-06-28
CMR-4309 modifies sync entry-titles concept-ids to batch find concepts for collections

daniel-zamora:


access-control-app/src/cmr/access_control/services/acl_util.clj
```diff
@@ -113,9 +113,13 @@
           concept-ids (:concept-ids collection-identifier)
           provider-id (get-in acl [:catalog-item-identity :provider-id])
           colls-from-entry-titles (when (seq entry-titles)
-                                    (mdb1/find-concepts context {:provider-id provider-id :entry-title entry-titles} :collection))
+                                    (for [batch (mdb1/find-in-batches context :collection 100 {:provider-id provider-id :entry-title entry-titles})
```
laurenfrederick: Might want to have 100 as a def

laurenfrederick: Also we discussed for testing, def the batch-size, change it to something small for the test, and have more than that # of collections so we know it hits the batching code.

oubiwann:
I don't know enough about this code to suggest how, but this sort of change looks like it needs some integration tests added ...

access-control-app/src/cmr/access_control/services/acl_util.clj
chris-durbin: I had the same thought for the test.

sxu123:
It might be just me but from the ticket I can't tell what exactly was the issue and how changing to batch find solves the issue.  I was hoping to see some tests here to show how the problem is resolved... but then I don't know enough of this part of the code to know if it's feasible...

PR #164, opened 2017-06-29
CMR-4240 metadata db variable save provider

oubiwann:
ON HOLD:

Not ready for review -- I discovered other areas that I need to address (failing tests in search, etc.) before this is ready.

Tasks for this PR:
* Updated the metadata-db concept service to remove variables from the system-level concept types
* Removed variables from the extended concept validations (mirroring the validations that collections and granules do)
* Updated the metadata-db testing utility to create testing variable concepts with a provider id
* Removed variable-specific test that was created for non-provider-id variables
* Updated remaining variables test and add a new one


oubiwann:
It turns out the metadata-db changes can't be done in isolation (while not breaking tests), so I'm going to close this PR and create a new one when both CMR-4240 and CMR-4239 are readying to go (they will be in the same PR).

PR #165, opened 2017-06-29
CMR-4312 adds chunking to query, originally only had batches of results

daniel-zamora:


access-control-app/src/cmr/access_control/services/acl_util.clj
```diff
@@ -105,6 +105,20 @@
                     identity-type-condition)]
     (get-acls-by-condition context condition)))
 
+(defn get-collections-chunked
+  "Searches for collections in chunks."
+  [context values keyword provider-id]
```
chris-durbin: We should avoid using clojure.core function names as bindings. Change keyword to field.

access-control-app/src/cmr/access_control/services/acl_util.clj
```diff
@@ -105,6 +105,20 @@
...
+  (let [chunk-size (config/sync-entry-titles-concept-ids-collection-batch-size)
+        chunked-values (partition chunk-size chunk-size nil values)]
+    (apply merge
```
chris-durbin: I believe you have a sequence of sequences of maps. I think `mapcat` is the idiomatic way to do this rather than `(apply merge (map ...`. 

You should definitely run the batching test you had added before, but had to comment out because access-control integration tests do not have the same ability to modify behavior in dev-system to make sure everything is working.

PR #166, opened 2017-06-30
CMR-4298: Updated the bulk update tables.

sxu123:
In the task status table, need:
* add created_at as timestamp - populated when row created
* change the task-id to varchar(255)

In the collection status table:
* add action_taken flag as varchar2(1)- default to "F" - false for now
* change the task-id to a varchar(255)
* change status_message to varchar(500)


ingest-app/src/migrations/006_update_bulk_update_tables.clj
```diff
@@ -0,0 +1,87 @@
...
+                                  REQUEST_JSON_BODY  BLOB NOT NULL,
+                                  STATUS VARCHAR(20),
+                                  STATUS_MESSAGE   VARCHAR(255),
+                                  CREATED_AT TIMESTAMP(9) WITH TIME ZONE DEFAULT sysdate NOT NULL)")
+    (j/db-do-commands trans-conn "CREATE TABLE CMR_INGEST.bulk_update_coll_status_new (
+                                  TASK_ID VARCHAR(255) NOT NULL,
+                                  CONCEPT_ID VARCHAR(255) NOT NULL,
+                                  STATUS VARCHAR(20),
+                                  STATUS_MESSAGE  VARCHAR(500),
+                                  ACTION_TAKEN VARCHAR2(1) DEFAULT 'F' NOT NULL)")
```
chris-durbin: What are the possible values for ACTION_TAKEN?

If we're really going to use just a single character CHAR would be better than VARCHAR. In general though I am not a fan of using a single letter to represent things. There's really not much overhead in using a short VARCHAR string for the the action taken especially without a lot of rows. A more descriptive term will just make things easier for future maintainers.

sxu123: Chris, I was researching on Boolean in oracle, and saw varchar2(1) was recommended to replace boolean since it doesn't exist.  it's only gonna be false and true, which will be "F" and "T".  I guess char will be fine.  Thanks. Done.

chris-durbin: I didn't realize how controversial booleans in Oracle actually are:
https://stackoverflow.com/questions/30062/boolean-field-in-oracle

We use 0 (false) and 1 (true) for booleans in the CMR (like the DELETED column in all of our concept tables). Unless there's a good reason to do something different, I would suggest in the CMR we should be consistent (and so use 0 and 1).

sxu123: Changed to INTEGER DEFAULT 0 NOT NULL. Thanks.

PR #167, opened 2017-06-30
CMR-4310: add test to prevent hasty granule-index updates

mschmele:


system-int-test/test/cmr/system_int_test/ingest/granule_ingest_test.clj
```diff
@@ -73,6 +75,19 @@
                       (:concept-id coll1) (:concept-id coll2))]
              errors)))))
 
+;; Verify that granule indexing is not being changed too quickly.
```
chris-durbin: The key point is that we cannot change the search application to start using the new fields until all of the granules have been indexed. That means the developer needs to make the change to start indexing the new field in one sprint, and make the search application changes in the following sprint.

Can you modify the comment to make it clearer?

mschmele: Done

sxu123: I was just thinking we don't necessarily want to see the two huge list difference. we just want to detect the change.
So add the following to let:
expected-string "okay with changing the search"
actual-string (if (= allowed-granule-index-fields actual-granule-index-fields)
"okay with changing the search"
"granule index changed, wait till next sprint to change search")]
then do (is (= expected-string actual-string)) check. sth like that. that way, people don't have to see the huge list of difference and then have to go to the test to read the doc string to find out what it is about.

sxu123: I see you also rely on the printed out new index list to update the existing one. so you will need to keep that (is (= allowed-index actual-index)) thing. then maybe you could add what I suggested and when granule index changed, actual-string="granule index changed wait till next sprint to change search using the index. Also update the allowed-index in <file> using the actual index printed out to make the test pass"  what do you think?

PR #176, opened 2017-07-07
CMR-4323

daniel-zamora:
This PR is to fix broken ACLs from the first migration test in SIT and is also a better way to do this migration for UAT and PROD

### DIFF
--- a/metadata-db-app/src/migrations/057_regenerate_collection_identifiers.clj
+++ b/metadata-db-app/src/migrations/057_regenerate_collection_identifiers.clj
@@ -17,7 +17,8 @@
           t (h/get-provider-collection-tablename provider)]
       (for [entry-title entry-titles
             :let [result (j/query (config/db) [(format "select distinct concept_id, entry_title from metadata_db.%s where entry_title = ?" t)
-                                               entry-title])]]
+                                               entry-title])]
+            :when (not (nil? entry-title))]        
         {:concept_id (:concept_id (first result))
          :entry_title (:entry_title (first result))}))))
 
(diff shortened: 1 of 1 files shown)


metadata-db-app/src/migrations/057_regenerate_collection_identifiers.clj
```diff
@@ -44,21 +46,17 @@
...
-                     (if (seq concept-ids)
-                       (assoc-in metadata [:catalog-item-identity :collection-identifier :concept-ids] concept-ids)
-                       metadata))
-          metadata (if deleted
-                     metadata
-                     (if (seq entry-titles)
-                       (assoc-in metadata [:catalog-item-identity :collection-identifier :entry-titles] entry-titles)
+                     (if (seq collections)
+                       (-> metadata
+                           (assoc-in [:catalog-item-identity :collection-identifier :concept-ids] (map :concept_id collections))
```
chris-durbin: It looks like this is to fix cases where you had some entry-titles and some concept-ids in an ACL, but not a full list of both?

daniel-zamora: The first migration put nils in for entry-titles, second migration swapped that to concept-ids and populated the entry-titles with entry-titles, so to fix SIT we need to run it this way.

laurenfrederick: We also did verify locally that it works for the original issue. We tested that, nil concept-ids, and nil entry-titles.

PR #178, opened 2017-07-09
CMR-4313 Adding CollectionDataType to the UMM <=> ISO MENDS mapping

eereiter:


umm-spec-lib/src/cmr/umm_spec/xml_to_umm_mappings/iso19115_2.clj
```diff
@@ -75,6 +75,10 @@
   (str "/gmi:MI_Metadata/gmd:metadataExtensionInfo/gmd:MD_MetadataExtensionInformation"
        "/gmd:extendedElementInformation/gmd:MD_ExtendedElementInformation"))
 
+(def collection-data-type-xpath
+  (str identifier-base-xpath "[gmd:codeSpace/gco:CharacterString='gov.nasa.esdis.umm.collectiondatatype']"
```
laurenfrederick: Do we have any data in our test files with collection data type in it?

eereiter: The generated tests produce collection data types as well as the example_data/iso19115/ISOExample-project.xml

umm-spec-lib/src/cmr/umm_spec/xml_to_umm_mappings/iso19115_2.clj
```diff
@@ -143,7 +147,7 @@
   DataIdentification element"
   [md-data-id-el sanitize?]
   (if-let [value (char-string-value md-data-id-el "gmd:abstract")]
-    (let [[abstract version-description](str/split
+    (let [[abstract version-description](string/split
```
daniel-zamora: fix spacing here

umm-spec-lib/src/cmr/umm_spec/xml_to_umm_mappings/iso19115_2.clj
```diff
@@ -234,6 +238,7 @@
       :TilingIdentificationSystems (tiling/parse-tiling-system md-data-id-el)
       :TemporalExtents (or (seq (parse-temporal-extents doc extent-info md-data-id-el))
                            (when sanitize? su/not-provided-temporal-extents))
+      :CollectionDataType (value-of (select doc collection-data-type-xpath) ".")
```
chris-durbin: I assume collection data type is an enumeration of science and near-real-time or something like that? So even though we're not validating here, we'll throw an error during UMM validation if the value isn't valid?

Sorry it's been awhile since I've reviewed a UMM change. 

I also assume the reason for no tests is that the generator tests and the change to iso19115_expected_conversion.clj are sufficient.

eereiter: Yes it is.  We keep the value that the user entered into the field and we return a warning message that includes all of the valid values. Yes, the generated tests produce collection data types as well as the example_data/iso19115/ISOExample-project.xml

umm-spec-lib/src/cmr/umm_spec/xml_to_umm_mappings/iso19115_2.clj
eereiter: done.

PR #181, opened 2017-07-11
CMR-4327 Fix or write new tickets for failing Oracle tests 

chris-durbin:
The main change was to fix the database migrations to work correctly in a system which did not have any small providers in the providers table (e.g. starting with a fresh database). Four other tickets were written to fix the individual failures that remained:

1. CMR-4334 - bulk updates
2. CMR-4335 - services
3. CMR-4336 - ACLs
4. CMR-4337 - bootstrap


common-app-lib/test/cmr/common_app/test/api/health.clj
```diff
@@ -30,21 +31,22 @@
...
+        context-with-health-cache (get-context-with-health-cache)]
     (is (= healthy-response
            (select-keys (get-app-health context-with-health-cache healthy-fn) [:status :body])))
     (testing "health response is cached so even though the unhealthy-fn is passed in it is not used"
       (is (= healthy-response
              (select-keys (get-app-health context-with-health-cache unhealthy-fn)
                           [:status :body])))
       ;; Wait 2 seconds and the cache has not expired, so it still returns healthy
-      (Thread/sleep 2000)
+      (Thread/sleep 2001)
```
laurenfrederick: what was the reason for the extra millisecond?

metadata-db-app/src/config/mdb_migrate_helper.clj
```diff
@@ -73,11 +79,18 @@
...
+  to the small_prov tables. In an environment which does not have any small providers we need to
+  have a dummy small provider to ensure the tables have their schema updated."
+  {:small true :cmr-only true :provider-id "DUMMY" :short-name "DUMMY"})
+
 (defn get-collection-tablenames
   "Gets a list of all the collection tablenames. Primarily for enabling migrations of existing
   provider tables."
   []
-  (distinct (map #(concept-tables/get-table-name % :collection) (p/get-providers (config/db)))))
+  (let [providers (conj (p/get-providers (config/db)) dummy-provider-for-small-prov)]
```
laurenfrederick: Nitpick: Seems like you could have a get-tablenames function and pass in :granules or :collections.

laurenfrederick: So we add the dummy-prov no matter what even if we have small providers?

common-app-lib/test/cmr/common_app/test/api/health.clj
chris-durbin: I was seeing some inconsistent results over a few runs, so I just made sure that the time from making the first call to the final call was guaranteed to be more than 5 seconds.

metadata-db-app/src/config/mdb_migrate_helper.clj
chris-durbin: There's a bunch of places using these functions so I didn't refactor down to a single function. 

We add the dummy-prov because we want this function to always return all of the collection tables. SMALL_PROV_COLLECTIONS will exist regardless of whether or not there are any small providers, so we'll want to update its schema whenever we update other collection tables schemas. Similarly for granules.

chris-durbin:
> Do we want to disable these tests in all modes or just for external db mode? Is it possible to limit when these tests are run that way?

That's a great question. I was thinking the same thing at one point, and then totally forgot to look into it. I'm going to check now.

chris-durbin:
I uncommented the tests that work with the in-memory database and used the macro to only test with in-memory mode.

PR #185, opened 2017-07-14
CMR-4045 Fix additional attribute validation for collections

laurenfrederick:
The additional attribute parsed values were not being added for collection validation (they were being added for granule parent validation) so the parsed values were always nil and certain validations were not running because of that.


system-int-test/test/cmr/system_int_test/ingest/collection_umm_spec_validation_test.clj
```diff
@@ -137,10 +142,31 @@
       ["AdditionalAttributes"]
       ["Additional Attributes must be unique. This contains duplicates named [bool]."]))
 
+  (testing "Additional Attribute parameter range validation"
```
laurenfrederick: Just added these two tests. For the rest of the changes, I turned on Parinfer and then re-formatted.

PR #186, opened 2017-07-14
CMR-4334, CMR-4338 Get bulk update tests working with oracle, fix int…

laurenfrederick:
…ernal server error when task-id does not exist

### DIFF
--- a/ingest-app/src/cmr/ingest/data/bulk_update.clj
+++ b/ingest-app/src/cmr/ingest/data/bulk_update.clj
@@ -47,12 +47,12 @@
   (get-bulk-update-task-status
     [db task-id]
     ;; Returns a status for the particular task
-    (-> db
-        (su/find-one (su/select [:status :status-message :request-json-body]
-                                (su/from "bulk_update_task_status")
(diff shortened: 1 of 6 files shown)


ingest-app/src/cmr/ingest/data/bulk_update.clj
```diff
@@ -47,12 +47,12 @@
   (get-bulk-update-task-status
     [db task-id]
     ;; Returns a status for the particular task
-    (-> db
-        (su/find-one (su/select [:status :status-message :request-json-body]
-                                (su/from "bulk_update_task_status")
-                                (su/where `(= :task-id ~task-id))))
-        util/map-keys->kebab-case
-        (update :request-json-body util/gzip-blob->string)))
+    (some-> db
```
mschmele: I'm always surprised at how many threading macros there are - this one is pretty cool! I assume you're okay with it short-circuiting and returning `nil` if any of the calls below do?

laurenfrederick: Yes, that is what fixed the internal server bug.

PR #188, opened 2017-07-18
Fixing CMR-3924 adding codespace to processing level UMM->ISO MENDS m…

eereiter:
…apping.

### DIFF
--- a/umm-spec-lib/src/cmr/umm_spec/umm_to_xml_mappings/iso19115_2.clj
+++ b/umm-spec-lib/src/cmr/umm_spec/umm_to_xml_mappings/iso19115_2.clj
@@ -361,6 +361,7 @@
          [:gmd:processingLevel
           [:gmd:MD_Identifier
            [:gmd:code (char-string (-> c :ProcessingLevel :Id))]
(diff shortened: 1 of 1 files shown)


umm-spec-lib/src/cmr/umm_spec/umm_to_xml_mappings/iso19115_2.clj
```diff
@@ -361,6 +361,7 @@
          [:gmd:processingLevel
           [:gmd:MD_Identifier
            [:gmd:code (char-string (-> c :ProcessingLevel :Id))]
+           [:gmd:codeSpace (char-string "gov.nasa.esdis.umm.proessinglevelid")]
```
laurenfrederick: processinglevelid is spelled wrong. You might want to put that in a def.

eereiter: Thanks, I copy and pasted from the spreadsheet and it was spelled wrong there too.  I have updated my spreadsheet and I put the string in a def.

PR #189, opened 2017-07-18
CMR-4285 removes dupes after find and replace and find and update

daniel-zamora:
This PR purpose is to fix find-and-replace and find-and-update to no longer leave duplicates in the umm it updates.

### DIFF
--- a/system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
+++ b/system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
@@ -39,6 +39,21 @@
                   :ContactPersons [{:Roles ["Data Center Contact"]
                                     :LastName "Smith"}]}]})
 
+(def find-update-keywords-umm
+  {:ScienceKeywords [{:Category "EARTH SCIENCE"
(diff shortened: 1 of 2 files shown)


system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
```diff
@@ -195,12 +210,46 @@
...
               (:format (:meta concept))))
-       (is (= (:ScienceKeywords (:umm concept))
-              [{:Category "EARTH SCIENCE"
+       (is (= [{:Category "EARTH SCIENCE"
                 :Topic "ATMOSPHERE"
                 :Term "AIR QUALITY"
+                :VariableLevel1 "EMISSIONS"}]
+              (:ScienceKeywords (:umm concept)))))))
+
+(deftest bulk-update-update-test
```
laurenfrederick: We should do a find and replace test too and show that only 1 science kw is left in the list.

daniel-zamora: I think there is a find and replace test there?

PR #190, opened 2017-07-18
CMR-4307: Fixed the code to refresh the collection cache if no collec…

sxu123:
Refresh the collection cache once if no collections are found
when generating the humanizer-report.

### DIFF
--- a/search-app/src/cmr/search/services/humanizers/humanizer_report_service.clj
+++ b/search-app/src/cmr/search/services/humanizers/humanizer_report_service.clj
@@ -10,9 +10,8 @@
    [cmr.common.cache.single-thread-lookup-cache :as stl-cache]
    [cmr.common.concepts :as concepts]
    [cmr.common.config :refer [defconfig]]
-   [cmr.common.jobs :refer [defjob]]
+   [cmr.common.jobs :refer [defjob default-job-start-delay]]
    [cmr.common.log :as log :refer [debug info warn error]]
-   [cmr.common.services.errors :as errors]
    [cmr.common.util :as util]
    [cmr.search.data.metadata-retrieval.metadata-cache :as metadata-cache]
    [cmr.search.data.metadata-retrieval.revision-format-map :as rfm]
@@ -41,6 +40,19 @@
   "The size of the batches to use to process collections for the humanizer report"
   {:default 500 :type Long})
 
+(defconfig humanizer-report-generator-job-delay
+  "Number of seconds humanizer-report-generator-job needs to wait after collection cache
+   refresh job starts..
+   We want to add the delay so that the collection cache can be populated first.
+   Splunk shows the average time taken for collection cache to be refreshed is around 
+   300 seconds"
+  {:default 400 :type Long})
+
+(defconfig humanizer-report-generator-job-wait
+  "Number of mili-seconds humanizer-generator-job waits for the collection cache 
+   to be populated in the event when the delay is not long enough."
+  {:default 60000 :type Long}) ;; one minute
+
 (defn- rfm->umm-collection
   "Takes a revision format map and parses it into a UMM-spec record."
   [context revision-format-map]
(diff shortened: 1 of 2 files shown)


sxu123:
Not sure how to test this in SIT:
1. Need to clear up all the cache (how?)
2. Need to make sure humanizer-report-generator-job is started before the refresh-collections-metadata-cache-job. but both of these jobs have a delay of 0 seconds. so not sure if the sequence is random.
3. Assuming 2 is satisfied, need to check the SIT LOG(where?) to make sure the first time the humanizer-report-generator-job is run, it doesn't generate empty report.

search-app/src/cmr/search/services/humanizers/humanizer_report_service.clj
```diff
@@ -56,13 +57,25 @@
   [context rfms]
   (map #(rfm->umm-collection context %) rfms))
 
+(defn- refresh-and-get-collections-from-cache
+   "Force the collection cache to be populated and then get all the collections from it."
+   [context]
+   (let [_ (metadata-cache/refresh-collections-metadata-cache-job)
```
oubiwann: We can just put `(metadata-cache/refresh-collections-metadata-cache-job)` above the `let` block, yes?

sxu123: Sure.  Done.

sxu123: Actually I made a mistake here. I should call (metadata-cache/refresh-cache) here. not refresh-collections-metadata-cache-job, which is used for scheduler...

search-app/src/cmr/search/services/humanizers/humanizer_report_service.clj
```diff
@@ -56,13 +57,25 @@
   [context rfms]
   (map #(rfm->umm-collection context %) rfms))
 
+(defn- refresh-and-get-collections-from-cache
+   "Force the collection cache to be populated and then get all the collections from it."
+   [context]
+   (metadata-cache/refresh-cache context)
+   (let [rfms (metadata-cache/all-cached-revision-format-maps context)]
+     (if (nil? (seq rfms))
```
daniel-zamora: I think `(if-not (seq rfms))` is idiomatic for this? not sure.

search-app/src/cmr/search/services/humanizers/humanizer_report_service.clj
```diff
@@ -56,13 +57,25 @@
...
+
 (defn- get-all-collections
   "Retrieves all collections from the Metadata cache, partitions them into batches of size
   humanizer-report-collection-batch-size, so the batches can be processed lazily to avoid out of memory errors."
   [context]
   ;; Currently not throwing an exception if the cache is empty. May want to
   ;; change in the future to throw an exception.
-  (let [rfms (metadata-cache/all-cached-revision-format-maps context)]
+  (let [rfms (metadata-cache/all-cached-revision-format-maps context)
+        rfms (if (nil? (seq rfms))
```
daniel-zamora: I think (if-not (seq rfms)) is idiomatic for this? not sure.

search-app/src/cmr/search/services/humanizers/humanizer_report_service.clj
sxu123: Thanks. I thought it didn't look quite right. I will change it to (if (seq rams) ....   Done.

search-app/src/cmr/search/services/humanizers/humanizer_report_service.clj
sxu123: Done.

sxu123:
@laurenfrederick, 
The humanizer report job is kicked off when I first start my proto-repl(0 second delay) and then every 24 hours.
When it first got kicked off, if there are collections in the system, but we can't get collections simply because they are not in the collection cache yet, then this fix will work because it will force the refresh of the cache. 

In the humanizer-report test you mentioned, it starts with no collections. now the job has already been kicked off, forced the cache refresh and still no collections.  Then it ingests a collection. If it doesn't refresh the collection cache, it will have to wait for 24 hours for the collection to be pulled into the cache by the humanizer-report-generator-job to generate report.

sxu123:
Okay. I didn't realize the endpoint get-humanizer-report shares the same code as the one fixed in this ticket and the humanizer-report integration tests make use of that. So yes I could remove the first refresh cache call there. Done.  Also, I could use this same endpoint to test in SIT without having to worry about which job gets kicked off first in my earlier comments. I just need to clear the cache, then get the report. (hopefully not right after the jobs are kicked off again)

sxu123:
@ygliuvt, the only reason I didn't go that route was I wasn't sure how to long to wait, now that I got the splunk statistics, I've added the delay of 400 seconds and an extra wait for 60 seconds, if the collections are still not in the cache.  Throw a warning if the collection cache is still empty after the wait. All the info are logged. Done.

search-app/src/cmr/search/services/humanizers/humanizer_report_service.clj
```diff
@@ -40,6 +40,19 @@
   "The size of the batches to use to process collections for the humanizer report"
   {:default 500 :type Long})
 
+(defconfig humanizer-report-generator-job-delay
+  "Number of seconds humanizer-report-generator-job needs to wait after collection cache
+   refresh job starts..
+   We want to add the delay so that the collection cache can be populated first.
+   Splunk shows the average time taken for collection cache to be refreshed is around 
+   300 seconds"
+  {:default 400 :type Long})
```
ygliuvt: Should change `400` to `400000`.

PR #191, opened 2017-07-18
CMR-3930 adds echo10 mapping for collection citations

daniel-zamora:
This PR adds CollectionCitation mappings for UMM <=> ECHO10


sxu123:
Do you have a sample echo10 xml file that contains the collection citations under umm-spec-lib/resources/example_data/echo10 ?
If so, the round trip testings will cover the testing of this ticket, otherwise, you will need to ask Erich to get you one and put it in there.

PR #196, opened 2017-07-19
Fixing CMR-3923 adding UMM<=>ISO SMAP processing level translation.

eereiter:


umm-spec-lib/src/cmr/umm_spec/test/iso_smap_expected_conversion.clj
```diff
@@ -298,7 +298,7 @@
         (assoc :SpatialKeywords nil)
         (assoc :TemporalKeywords nil)
         (assoc :AdditionalAttributes nil)
-        (assoc :ProcessingLevel (umm-c/map->ProcessingLevelType {:Id su/not-provided}))
+        (update :ProcessingLevel su/convert-empty-record-to-nil)
```
oubiwann: A thought for later: all these assoc calls can be collapsed into one, since assoc is n-arity (if ordering doesn't matter). This seems pretty inefficient to me...

PR #198, opened 2017-07-20
CMR-4307: Added retries

sxu123:


search-app/src/cmr/search/services/humanizers/humanizer_report_service.clj
```diff
@@ -70,18 +75,25 @@
...
-   from collection cache. If collections are still not present, throw warning."
+  "Wait configurable number of milli seconds before retrying configurable number of times
+   to get all the collections from collection cache. If collections are still not present, throw warning."
   [context]
   (info (format "Collection cache is not populated after %d seconds of delay" 
                 (humanizer-report-generator-job-delay)))
-  (Thread/sleep (humanizer-report-generator-job-wait)) 
+  ;; sleep, retry 19 times, then sleep and retry one last time.
+  (Thread/sleep (humanizer-report-generator-job-wait))
+  (let [a (atom (- (humanizer-report-generator-job-retry-number) 1))]
```
laurenfrederick: I would just name a something a little more descriptive like atom-retry-count. You could also call `(dec (humanizer-report-generator-job-retry-number))` here.

search-app/src/cmr/search/services/humanizers/humanizer_report_service.clj
```diff
@@ -70,18 +75,25 @@
...
+    (while (and (pos? @a) (not (seq (metadata-cache/all-cached-revision-format-maps context))))
+      (info (format "Humanizer report generator job is sleeping for %d milli-seconds before retry"
```
laurenfrederick: Might want to say something about sleeping before retrying populating the collection cache again. It could be a bit confusing for someone looking in the logs if they don't know we're waiting on the collection cache, not the humanizer report.

search-app/src/cmr/search/services/humanizers/humanizer_report_service.clj
```diff
@@ -70,18 +75,25 @@
...
+  "Wait configurable number of milli seconds before retrying configurable number of times
+   to get all the collections from collection cache. If collections are still not present, throw warning."
   [context]
   (info (format "Collection cache is not populated after %d seconds of delay" 
                 (humanizer-report-generator-job-delay)))
-  (Thread/sleep (humanizer-report-generator-job-wait)) 
+  ;; sleep, retry 19 times, then sleep and retry one last time.
+  (Thread/sleep (humanizer-report-generator-job-wait))
+  (let [a (atom (- (humanizer-report-generator-job-retry-number) 1))]
+    (while (and (pos? @a) (not (seq (metadata-cache/all-cached-revision-format-maps context))))
```
laurenfrederick: So here you're calling `metadata-cache/all-cached-revision-format-maps` and if it succeeds, you're going to call it again on line 91. Might want to try something like this: https://stackoverflow.com/questions/27097978/implementing-break-in-clojure to avoid making that call again unnecessarily.

search-app/src/cmr/search/services/humanizers/humanizer_report_service.clj
sxu123: Done.

search-app/src/cmr/search/services/humanizers/humanizer_report_service.clj
sxu123: Done. Changed to sleeping before retrying to see if the collection cache is populated.

search-app/src/cmr/search/services/humanizers/humanizer_report_service.clj
sxu123: Done.

search-app/src/cmr/search/services/humanizers/humanizer_report_service.clj
```diff
@@ -70,18 +75,18 @@
...
-    (if (seq rfms)
-      rfms
-      (warn (format "Collection cache is not populated after %d seconds of delay and %d milli-seconds of wait."
-                    (humanizer-report-generator-job-delay)
-                    (humanizer-report-generator-job-wait))))))
+  "Wait configurable number of milli seconds before retrying configurable number of times
+   to get all the collections from collection cache."
+  [count context]
+  (when (< count (retry-count))
+    (info (format (str "Humanizer report generator job is sleeping for %d milli-seconds"
```
ygliuvt: update the logging message to be in seconds so that it is more readable.

sxu123: Done.

search-app/src/cmr/search/services/humanizers/humanizer_report_service.clj
```diff
@@ -70,18 +75,18 @@
...
+                       "before retrying to see if the collection cache is populated.")
+                  (humanizer-report-generator-job-wait)))
+    (Thread/sleep (humanizer-report-generator-job-wait))
+    (let [rfms (metadata-cache/all-cached-revision-format-maps context)]
+      (if (seq rfms)
+        rfms
+        (recur (inc count) context)))))
 
 (defn- get-all-collections
   "Retrieves all collections from the Metadata cache, partitions them into batches of size
```
ygliuvt: I think you can refactor this function to something like:
```
(if-let [rfms (get-cached-revision-format-maps context)]
 (map
     #(rfms->umm-collections context %)
     (partition-all (humanizer-report-collection-batch-size) rfms))
(warn (format "Collection cache is not populated after %d seconds of delay and %d times of retry." ...)))
```
And in function `get-cached-revision-format-maps`, you have a loop that tries to call ` (metadata-cache/all-cached-revision-format-maps context)` returns it if it is not empty; otherwise retries until the retry count is exhausted, i.e. something like:
```
(loop [retries (retry-count)]
  (if-let [rfms (metadata-cache/all-cached-revision-format-maps context)] 
    rfms
   (when (> retries 0)
     ;; logging and sleeping, then 
     (recur (dec max-tries))))
```
Add a `(seq ...)` call in `all-cached-revision-format-maps` function, so the return value will be nil when empty.

sxu123: 1. I thought the normal flow would be we call to get the collections from the cache first, if it doesn't exist, we fire the retry mechanism. If you just want to start the retry mechanism right away, I guess it's okay...  but then, we can't say that collection cache is not populated after the 400 seconds delay because we never try it.  I will need to remove that msg. 
2. Your loop is basically the same as mine except you are decreasing the count and I'm increasing the count.

sxu123: Done.

search-app/src/cmr/search/data/metadata_retrieval/metadata_cache.clj
```diff
@@ -160,9 +160,11 @@
 (defn all-cached-revision-format-maps
   "Returns a sequence of all revision format maps in the cache sorted by concept id"
   [context]
-  (let [cache (deref (:cache-atom (c/context->cache context cache-key)))]
-    (for [concept-id (sort (keys cache))]
-      (get cache concept-id))))
+  (let [cache (deref (:cache-atom (c/context->cache context cache-key)))
+        maps (for [concept-id (sort (keys cache))]
+               (get cache concept-id))]
+     (when (seq maps)
```
laurenfrederick: Just return `(seq maps)` here instead of doing the when.

PR #200, opened 2017-07-21
CMR-4359: Updated UMM-Var Schema to the latest version at commit a6fa…

ygliuvt:
…6fe.


ygliuvt:
I think the spec version is driven by UMM-C. UMM-Var is in DRAFT state and is just tagging along.

PR #201, opened 2017-07-21
CMR-4247 - Rollback

oubiwann:
Reverted changes due to failures in master CI/CD not present in CI/CD branch.


oubiwann:
The team and I talked about this in chat, and folks have given their +1 for a rollback. I'll go ahead and do that shortly.

PR #203, opened 2017-07-25
CMR-4363: Increase max request size we can accept in Jetty 

chris-durbin:
from 5MB to 50MB so that large humanizer reports can be saved in cubby.

There is already a test which verifies we can successfully process a body of exactly 50MB and we reject a request with 50MB + 1 byte (cmr.common.test.api.web-server/test-max-post-size).

The only reason we have this limit is so that we can prevent invalid requests from causing OOM errors on our apps. Every app has at least a 1GB heap size so a 50MB request is not likely to cause any problems.

### DIFF
--- a/common-lib/src/cmr/common/api/web_server.clj
+++ b/common-lib/src/cmr/common/api/web_server.clj
@@ -41,10 +41,10 @@
   ONE_MB)
 
 (def MAX_REQUEST_BODY_SIZE
- "The maximum size of a request to body. This is set to 500KB prevent large requests coming in
-  that cause out of memory exceptions. A large ISO document like AST_L1A can be 120K characters which
-  would take up about 240KB in Java. This value is set to 5 MB."
- (* 5 ONE_MB))
+ "The maximum request body size which the application will accept. This is set to prevent large,
+  invalid requests coming in that cause out of memory exceptions. Requests to save the humanizer
+  report in Cubby can be in the 5 to 10 MB range."
+ (* 50 ONE_MB))
 
 (defn- routes-fn-verify-size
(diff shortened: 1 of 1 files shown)


common-lib/src/cmr/common/api/web_server.clj
```diff
@@ -41,10 +41,10 @@
   ONE_MB)
 
 (def MAX_REQUEST_BODY_SIZE
- "The maximum size of a request to body. This is set to 500KB prevent large requests coming in
-  that cause out of memory exceptions. A large ISO document like AST_L1A can be 120K characters which
-  would take up about 240KB in Java. This value is set to 5 MB."
- (* 5 ONE_MB))
+ "The maximum request body size which the application will accept. This is set to prevent large,
```
laurenfrederick: Do we know how what size would cause an out of memory exception?

chris-durbin: You can search Splunk to see the most memory any of our apps use:
sourcetype=*cmr*-prod logjvmstatisticsjob "Used memory" | rex "Used memory.*\[(?<used>\d+)\] Percent" | stats max(used) by sourcetype

Search, indexer, and metadata-db set the max heap size to 5GB. The other apps use the defaults. The default heap size is 1/4 of memory on the machine and our smallest instance type is m3.medium with 3.75 GB of RAM or roughly 940 MB. You can see that none of the apps are within 50MB of hitting the max heap size. Even if they were garbage collection would take place, and so an OOM from a 50MB request is really unlikely to ever happen (never is not in my vocabulary).

laurenfrederick: Need <used> in your rex. Ok just checking.

PR #205, opened 2017-07-26
Cmr 4230 - Sort by version

laurenfrederick:
We are already sorting by version for keyword search, but have run across an issue in a few cases where the version formats are mismatched (i.e. 5 vs 006) which would result in the versions not sorting correctly. I added a parsed-version-id that tries to parse to int to normalize the versions if possible.

### DIFF
--- a/indexer-app/src/cmr/indexer/data/concepts/collection.clj
+++ b/indexer-app/src/cmr/indexer/data/concepts/collection.clj
@@ -50,8 +50,8 @@
       :else
       (errors/internal-error! (str "Unknown spatial representation [" coord-sys "]")))))
 
-(defn- apply-function-to-all-values-in-map 
-  "Applies function f to all the values in map m." 
+(defn- apply-function-to-all-values-in-map
+  "Applies function f to all the values in map m."
   [m f]
   (reduce-kv (fn [m k v] (assoc m k (f v))) {} m))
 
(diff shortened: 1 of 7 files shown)


indexer-app/src/cmr/indexer/data/concepts/collection.clj
```diff
@@ -239,6 +249,7 @@
             :short-name.lowercase (util/safe-lowercase short-name)
             :version-id version-id
             :version-id.lowercase (util/safe-lowercase version-id)
+            :parsed-version-id.lowercase (util/safe-lowercase parsed-version-id)
```
oubiwann: Inserting dots in these keys seems an odd convention to me ... like JavaScript-isms making it into Clojure code. Why not `:doi-lowercase`, `:version-id-lowercase` and `:parsed-version-id-lowercase` instead?

Is there something else consuming these that we don't control, that expects this?

indexer-app/src/cmr/indexer/data/index_set.clj
```diff
@@ -267,6 +267,7 @@
           :short-name.lowercase  m/string-field-mapping
           :version-id            (m/stored m/string-field-mapping)
           :version-id.lowercase  m/string-field-mapping
+          :parsed-version-id.lowercase m/string-field-mapping
```
chris-durbin: Make sure to note we need to reindex collections as part of the deployment instructions.

indexer-app/test/cmr/indexer/test/data/concepts/collection.clj
```diff
@@ -0,0 +1,23 @@
...
+  (are3 [version-id parsed-version-id]
+    (is (= parsed-version-id
+           (#'collection-indexer/parse-version-id version-id)))
+
+    "Regular integer version"
+    "15" "15"
+
+    "Leading 0's version"
+    "006" "6"
+
```
chris-durbin: How about a test with leading and trailing 0s as well?

"010" "10"

indexer-app/src/cmr/indexer/data/concepts/collection.clj
laurenfrederick: Not sure of the history on this and why it was originally done this way. I haven't seen anything in the elasticsearch documentation that would need it this way, but I am wondering if it's something elasticsearch specific that I am missing.

indexer-app/test/cmr/indexer/test/data/concepts/collection.clj
laurenfrederick: I put that in.

PR #207, opened 2017-07-27
Cmr 4001 wl fixes

mschmele:
A few small fixes that need to be tested in other environments


mschmele:
Closing this until the fix is confirmed to be working

PR #208, opened 2017-07-27
CMR-4369 search var by concept

oubiwann:
This is a quick feature-add for MMT so that they can do their testing and dev. The same ticket (CMR-4369) will have more branches to come that build upon this.

### DIFF
--- a/search-app/docs/api.md
+++ b/search-app/docs/api.md
@@ -1961,7 +1961,9 @@ Examples of sorting by start_date in descending(Most recent data first) and asce
 
 ### <a name="retrieving-concepts-by-concept-id-and-revision-id"></a> Retrieve concept with a given concept-id or concept-id & revision-id
 
-This allows retrieving the metadata for a single concept. This is only supported for collections and granules. If no format is specified the native format of the metadata will be returned.
+This allows retrieving the metadata for a single concept. This is only supported for collections, granules, and variables. If no format is specified the native format of the metadata will be returned.
+
+WARNING: Currently for variables, the only `Accept` header value you can pass is `*/*`. As such the format extension is not yet supported for variables.
 
 By concept id
 
@@ -1971,6 +1973,8 @@ By concept id and revision id
 
     curl -i "%CMR-ENDPOINT%/concepts/:concept-id/:revision-id"
 
+Examples:
+
     curl -i "%CMR-ENDPOINT%/concepts/G100000-PROV1"
     curl -i "%CMR-ENDPOINT%/concepts/G100000-PROV1.iso"
     curl -i -H 'Accept: application/xml' "%CMR-ENDPOINT%/concepts/G100000-PROV1"
@@ -1978,11 +1982,13 @@ By concept id and revision id
     curl -i "%CMR-ENDPOINT%/concepts/G100000-PROV1.json"
     curl -i "%CMR-ENDPOINT%/concepts/C100000-PROV1/1"
     curl -i "%CMR-ENDPOINT%/concepts/G100000-PROV1/2.echo10"
+    curl -i "%CMR-ENDPOINT%/concepts/V100000-PROV1"
+    curl -i "%CMR-ENDPOINT%/concepts/V100000-PROV1/1"
+    curl -i "%CMR-ENDPOINT%/concepts/V100000-PROV1/2"
 
 Note that attempting to retrieve a revision that is a tombstone is an error and will return a 400 status code.
 
-The following extensions and MIME types are supported by the
-`/concepts/` resource:
+The following extensions and MIME types are supported by the `/concepts/` resource for collection and granule concept types:
 
   * `html`      "text/html" (Collections only)
   * `json`      "application/json"
--- a/system-int-test/src/cmr/system_int_test/utils/search_util.clj
+++ b/system-int-test/src/cmr/system_int_test/utils/search_util.clj
@@ -179,11 +179,12 @@
          url (url/retrieve-concept-url concept-type concept-id revision-id)
          url (if url-extension
                (str url "." url-extension)
-               url)]
-     (client/get url (merge {:accept (when-not url-extension format-mime-type)
-                             :throw-exceptions false
-                             :connection-manager (s/conn-mgr)}
-                            options)))))
+               url)
+         args (merge {:accept (when-not url-extension format-mime-type)
+                      :throw-exceptions false
+                      :connection-manager (s/conn-mgr)}
+                     options)]
+     (client/get url args))))
 
 (defn find-concepts-in-format
   "Returns the concepts in the format given."
--- a/system-int-test/test/cmr/system_int_test/search/variable/concept_retrieval_test.clj
+++ b/system-int-test/test/cmr/system_int_test/search/variable/concept_retrieval_test.clj
@@ -1,0 +1,103 @@
+(ns cmr.system-int-test.search.variable.concept-retrieval-test
+  "Integration test for variable retrieval via the following endpoints:
+
+  * /concepts/:concept-id
+  * /concepts/:concept-id/:revision-id"
+  (:require
+   [cheshire.core :as json]
+   [clojure.test :refer :all]
+   [cmr.common.mime-types :as mt]
+   [cmr.mock-echo.client.echo-util :as e]
+   [cmr.system-int-test.system :as s]
+   [cmr.system-int-test.utils.index-util :as index]
+   [cmr.system-int-test.utils.ingest-util :as ingest]
+   [cmr.system-int-test.utils.metadata-db-util :as mdb]
+   [cmr.system-int-test.utils.search-util :as search]
+   [cmr.system-int-test.utils.variable-util :as variable]
+   [cmr.transmit.config :as transmit-config]))
+
+(use-fixtures
+ :each
+ (join-fixtures
+  [(ingest/reset-fixture {"provguid1" "PROV1"
+                          "provguid2" "PROV2"
+                          "provguid3" "PROV3"})
+   variable/grant-all-variable-fixture]))
+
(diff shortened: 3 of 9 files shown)


system-int-test/test/cmr/system_int_test/search/variable/concept_retrieval_test.clj
```diff
@@ -0,0 +1,109 @@
...
+                                     (search/retrieve-concept
+                                      (:concept-id del-var)
+                                      nil
+                                      {:accept mt/any
+                                       :throw-exceptions true
+                                       :headers {transmit-config/token-header user1-token}}))]
+        (is (= 404 status))
+        (is (= ["Concept with concept-id [V1200000016-PROV1] could not be found."]
+               errors))))
+    (testing "retrieval by variable concept-id returns the latest revision."
```
chris-durbin: This doesn't look to be testing what it says. I only see one revision for this concept. We should modify the setup to add multiple revisions and verify it returns the latest.

system-int-test/test/cmr/system_int_test/search/variable/concept_retrieval_test.clj
```diff
@@ -0,0 +1,109 @@
+(ns cmr.system-int-test.search.variable.concept-retrieval-test
+  "Integration test for variable retrieval via the /concepts/:concept-id and
+  /concepts/:concept-id/:revision-id endpoints."
+  (:require
+   [cheshire.core :as json]
```
ygliuvt: remove unused namespaces from `:require`.

chris-durbin: There are a ton of unused requires here. Can you use something like slamhound on this file to get rid of all the unused requires?

I even see cmr.common.mime-types in there twice 😄 . Looks like a copy/paste from somewhere that might also need to be fixed.

system-int-test/test/cmr/system_int_test/search/variable/concept_retrieval_test.clj
```diff
@@ -0,0 +1,109 @@
...
+               errors))))
+    (testing "retrieval by variable concept-id returns the latest revision."
+      (let [response (search/retrieve-concept
+                       (:concept-id var1)
+                       nil
+                       {:accept mt/any
+                        :headers {transmit-config/token-header user1-token}})]
+        (is (= var1-name
+               (:Name (json/parse-string (:body response) true))))))
+    (testing "retrieval by variable concept-id and revision-id returns the specified variable"
```
ygliuvt: Add a new revision with a different variable_name, this way you can be sure that the test is actually doing what it claims to do.

system-int-test/test/cmr/system_int_test/search/variable/concept_retrieval_test.clj
oubiwann: yeah, it was a copy paste from collections concept retrieval test. We're going to be adding a bunch of tests here, though ... I guess I can remove them (we'll just be adding them back).

oubiwann: Fixed.

system-int-test/test/cmr/system_int_test/search/variable/concept_retrieval_test.clj
oubiwann: Fixed.

system-int-test/test/cmr/system_int_test/search/variable/concept_retrieval_test.clj
oubiwann: Fixed.

system-int-test/test/cmr/system_int_test/search/variable/concept_retrieval_test.clj
oubiwann: Done.

oubiwann:
Added doc updates. It's the same usage as what currently is done for concepts, so this really amounted to adding some example usage with the "V" prepended and adding some caveats about format support with variables.

search-app/docs/api.md
```diff
@@ -1961,7 +1961,12 @@ Examples of sorting by start_date in descending(Most recent data first) and asce
 
 ### <a name="retrieving-concepts-by-concept-id-and-revision-id"></a> Retrieve concept with a given concept-id or concept-id & revision-id
 
-This allows retrieving the metadata for a single concept. This is only supported for collections and granules. If no format is specified the native format of the metadata will be returned.
+This allows retrieving the metadata for a single concept. This is only supported
```
chris-durbin: When you generate the docs and look at the HTML do these lines still appear together as a single responsive paragraph or is there a hard break? I thought I saw a problem with artificial line breaks before when I tried to breakup long lines.

system-int-test/test/cmr/system_int_test/search/variable/concept_retrieval_test.clj
oubiwann: Oh dear, in a rebase, I seem to have lost the changes I made to the `requires` :-( ...

system-int-test/test/cmr/system_int_test/search/variable/concept_retrieval_test.clj
chris-durbin: I'm still seeing a bunch of unused requires. Things like spatial search, all the various collection specific formats, etc will never be used for variables.

system-int-test/test/cmr/system_int_test/search/variable/concept_retrieval_test.clj
oubiwann: Fixed again (fortunately only a few undo's away).

system-int-test/test/cmr/system_int_test/search/variable/concept_retrieval_test.clj
oubiwann: Yeah, this is due to a bad rebase; I think all the new changes are back in and pushed. Double-checking now ...

system-int-test/src/cmr/system_int_test/utils/search_util.clj
```diff
@@ -179,11 +179,12 @@
...
-               url)]
-     (client/get url (merge {:accept (when-not url-extension format-mime-type)
-                             :throw-exceptions false
-                             :connection-manager (s/conn-mgr)}
-                            options)))))
+               url)
+         args (merge {:accept (when-not url-extension format-mime-type)
+                      :throw-exceptions false
+                      :connection-manager (s/conn-mgr)}
+                     options)]
```
oubiwann: Note that this change was done so I could easily look at the args getting passed to the client (debugging different results from curl and CMR; first thing I wanted to check was that the args for the HTTP GET were the same). I left the change in place so that it would be easier for folks in the future, should they also want to debug the args.

system-int-test/test/cmr/system_int_test/search/variable/concept_retrieval_test.clj
```diff
@@ -0,0 +1,99 @@
...
+;; Retrieve by concept-id - general test
+;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(deftest retrieve-variable-by-concept-id
+  (let [var1-name "variable1"
+        var1-concept (variable/make-variable-concept {:Name var1-name})
+        var1 (ingest/ingest-concept var1-concept)
+        _ (index/wait-until-indexed)
+        var1-v2 (ingest/ingest-concept var1-concept)
```
chris-durbin: You should use a different :Name for this revision and then test below that when you retrieve the latest it is the V2 name, when you explicitly request revision 1 you get V1 name, and when you explicitly request revision 2 you get V2 name.

search-app/docs/api.md
oubiwann: I hope not ... I've been doing this in other docs, too -- I'm generating locally right now to double-check.

oubiwann: Yikes -- you are correct. It looks like CMR's markdown parser is converting single newlines in text to `<br>`s. This is atypical ... I think you have to go out of your way to cause that behaviour in markdown parsers ... which leads me to think that if I change the configuration for CMR's markdown parsing, it will have a host of other side-effects ...

system-int-test/test/cmr/system_int_test/search/variable/concept_retrieval_test.clj
oubiwann: @chris-durbin: Question about this ... I've tested for the same concept-id but different revisions ... and your comment above implies that's not enough? Which leads some some confusion on my part ... (because I don't know enough about CMR concepts and revisioning, no doubt). Can you give me an example of where the concept-id + revision increment check would give a false positive, where a check against the updated name would not? This would help me greatly ...

chris-durbin: You need some way to say "yes this is the right metadata for revision 2". The way it is written right now there is no differentiation in the metadata between revision 1 and revision 2.

See lines 68 and 69 below to test that the latest revision for a concept is coming back:

`(= var1-name` ...

That test will pass even if revision 1 were incorrectly returned since both revision 1 and revision 2 have the same variable name.

oubiwann: *Got* it -- it's about the metadata, not the concept. That makes total sense. Thanks!

oubiwann: Okay, I've updated the test to ensure that metadata updates are present and accounted for when a concept is updated.

PR #212, opened 2017-07-28
CMR-4254

ygliuvt:
Changed variable association to use variable concept-id as the key rather than native-id.


oubiwann:
@ygliuvt Looks good so far ... do the docs need to be updated, too?

ygliuvt:
I have held off the api doc change thinking it would have conflict with my other branch that added variable keyword search. I just checked again and there is actually no merge conflict. The conflict stuff is with the search by variable_name PR which has been merged yesterday. So I just pushed the doc change. Thanks.

PR #213, opened 2017-07-28
CMR-4343: Refactor dispatching logic in bootstrap-app

chris-durbin:
This is the first of a few pull requests for CMR-4343. There is no change in functionality as a result of this PR.

Currently we use core.async channels for queueing up work for the bootstrap application. Unfortunately this does not allow for scaling up because the queue is local to a single instance.

My goal for CMR-4343 is to at a minimum change our bulk indexing of providers to put messages on a message queue so that operations can spin up additional bootstrap instances whenever they need to reindex all the granules in the system.

So back to this PR - the purpose of this was to refactor the way we handle dispatching bootstrap work to make it easy to add a new implementation (the message queue) without worrying about breaking existing functionality.

Our integration tests always request synchronous requests for bulk indexing since there are timing issues if trying to use core.async.

I temporarily modified all of the bootstrap endpoint calls in system-int-test to use asynchronous requests and added sleeps after making the requests in order to test the core.async implementation after the refactor. All of the tests passed (but took a really long time).

### DIFF
--- a/bootstrap-app/src/cmr/bootstrap/api/routes.clj
+++ b/bootstrap-app/src/cmr/bootstrap/api/routes.clj
@@ -41,7 +41,8 @@
      :body {:message (str "Processing collection " collection-id "for provider " provider-id)}}))
 
 (defn- migrate-provider
-  "Copy a single provider's data from catalog-rest to metadata db (including collections and granules)"
+  "Copy a single provider's data from catalog-rest to metadata db (including collections and
+  granules)."
   [context provider-id-map params]
   (let [provider-id (get provider-id-map "provider_id")
         synchronous (synchronous? params)]
@@ -57,7 +58,8 @@
         result (bs/index-provider context provider-id synchronous start-index)
         msg (if synchronous
               result
-              (str "Processing provider " provider-id " for bulk indexing from start index " start-index))]
+              (str "Processing provider " provider-id " for bulk indexing from start index "
+                   start-index))]
     {:status 202
      :body {:message msg}}))
 
@@ -126,7 +128,8 @@
         concept-type (keyword (get request-details-map "concept_type"))
         concept-ids (get request-details-map "concept_ids")
         synchronous (synchronous? params)
-        result (bs/delete-concepts-from-index-by-id context synchronous provider-id concept-type concept-ids)
+        result (bs/delete-concepts-from-index-by-id context synchronous provider-id concept-type
+                                                    concept-ids)
         msg (if synchronous
               (str "Processed " result "conccepts for bulk deletion from indexes.")
               (str "Processing concepts for bulk deletion from indexes."))]
@@ -204,20 +207,18 @@
         (DELETE "/concepts" {:keys [request-context body params]}
           (bulk-delete-concepts-from-index-by-id request-context body params)))
 
-
       (context "/rebalancing_collections/:concept-id" [concept-id]
+        ;; Start rebalancing
+        (POST "/start" {:keys [request-context params]}
+          (start-rebalance-collection request-context concept-id params))
 
-       ;; Start rebalancing
-       (POST "/start" {:keys [request-context params]}
-         (start-rebalance-collection request-context concept-id params))
+        ;; Get counts of rebalancing data
+        (GET "/status" {:keys [request-context]}
+          (rebalance-status request-context concept-id))
 
-       ;; Get counts of rebalancing data
-       (GET "/status" {:keys [request-context]}
-         (rebalance-status request-context concept-id))
-
-       ;; Complete reindexing
-       (POST "/finalize" {:keys [request-context]}
-         (finalize-rebalance-collection request-context concept-id)))
+        ;; Complete reindexing
+        (POST "/finalize" {:keys [request-context]}
+          (finalize-rebalance-collection request-context concept-id)))
 
       (context "/virtual_products" []
         (POST "/" {:keys [request-context params]}
--- a/bootstrap-app/src/cmr/bootstrap/api/routes.clj
+++ b/bootstrap-app/src/cmr/bootstrap/api/routes.clj
@@ -30,13 +30,21 @@
   [params]
   (= "true" (when (:synchronous params) (str/lower-case (:synchronous params)))))
 
+(defn- get-dispatcher
+  "Returns the correct dispatcher to use based on the system configuration and the request."
+  [context params request-type]
+  (if (synchronous? params)
+    (get-in context [:system :synchronous-dispatcher])
+    (let [dispatcher-type (get bs/request-type->dispatcher request-type :core-async-dispatcher)]
+      (get-in context [:system dispatcher-type]))))
+
 (defn- migrate-collection
   "Copy collections data from catalog-rest to metadata db (including granules)"
   [context provider-id-collection-map params]
-  (let [provider-id (get provider-id-collection-map "provider_id")
-        synchronous (synchronous? params)
+  (let [dispatcher (get-dispatcher context params :migrate-collection)
+        provider-id (get provider-id-collection-map "provider_id")
         collection-id (get provider-id-collection-map "collection_id")]
-    (bs/migrate-collection context provider-id collection-id synchronous)
+    (bs/migrate-collection context dispatcher provider-id collection-id)
     {:status 202
      :body {:message (str "Processing collection " collection-id "for provider " provider-id)}}))
 
@@ -44,19 +52,19 @@
   "Copy a single provider's data from catalog-rest to metadata db (including collections and
   granules)."
   [context provider-id-map params]
-  (let [provider-id (get provider-id-map "provider_id")
-        synchronous (synchronous? params)]
-    (bs/migrate-provider context provider-id synchronous)
+  (let [dispatcher (get-dispatcher context params :migrate-provider)
+        provider-id (get provider-id-map "provider_id")]
+    (bs/migrate-provider context dispatcher provider-id)
     {:status 202 :body {:message (str "Processing provider " provider-id)}}))
 
 (defn- bulk-index-provider
   "Index all the collections and granules for a given provider."
   [context provider-id-map params]
-  (let [provider-id (get provider-id-map "provider_id")
-        synchronous (synchronous? params)
+  (let [dispatcher (get-dispatcher context params :index-provider)
+        provider-id (get provider-id-map "provider_id")
         start-index (Long/parseLong (get params :start_index "0"))
-        result (bs/index-provider context provider-id synchronous start-index)
-        msg (if synchronous
+        result (bs/index-provider context dispatcher provider-id start-index)
+        msg (if
               result
               (str "Processing provider " provider-id " for bulk indexing from start index "
                    start-index))]
@@ -66,11 +74,11 @@
 (defn- bulk-index-data-later-than-date-time
   "Index all the data with a revision-date later than a given date-time."
   [context params]
-  (let [synchronous (synchronous? params)
+  (let [dispatcher (get-dispatcher context params :index-data-later-than-date-time)
         date-time (:date_time params)]
     (if-let [date-time-value (date-time-parser/try-parse-datetime date-time)]
-      (let [result (bs/index-data-later-than-date-time context date-time-value synchronous)
-            msg (if synchronous
+      (let [result (bs/index-data-later-than-date-time context dispatcher date-time-value)
+            msg (if
                   (:message result)
                   (str "Processing data after " date-time " for bulk indexing"))]
         {:status 202
@@ -81,11 +89,11 @@
 (defn- bulk-index-collection
   "Index all the granules in a collection"
   [context provider-id-collection-map params]
-  (let [provider-id (get provider-id-collection-map "provider_id")
+  (let [dispatcher (get-dispatcher context params :index-collection)
+        provider-id (get provider-id-collection-map "provider_id")
         collection-id (get provider-id-collection-map "collection_id")
-        synchronous (synchronous? params)
-        result (bs/index-collection context provider-id collection-id synchronous)
-        msg (if synchronous
+        result (bs/index-collection context dispatcher provider-id collection-id)
+        msg (if
               result
               (str "Processing collection " collection-id " for bulk indexing."))]
     {:status 202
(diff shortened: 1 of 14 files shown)


bootstrap-app/src/cmr/bootstrap/services/bootstrap_service.clj
```diff
@@ -15,26 +12,29 @@
     [cmr.indexer.system :as indexer-system]
     [cmr.transmit.index-set :as index-set]))
 
+(def request-type->dispatcher
+  "A map of request types to which dispatcher to use for asynchronous requests."
+  {:migrate-provider :core-async-dispatcher
+   :migrate-collection :core-async-dispatcher
+   :index-provider :core-async-dispatcher
```
chris-durbin: For this ticket (CMR-4343) I plan to just change the index-provider type to use the message queue (in my next PR). I will write additional tickets for changing over anything else that could benefit from horizontal scaling using a message queue.

bootstrap-app/src/cmr/bootstrap/services/dispatch/core_async_dispatch.clj
```diff
@@ -0,0 +1,143 @@
...
+   ;; Channel for processing data newer than a given date-time.
+   data-index-channel
+   ;; Channel for processing bulk index requests for system concepts (tags, acls, access-groups)
+   system-concept-channel
+   ;; channel for processing bulk index requests by concept-id
+   concept-id-channel
+   ;; channel for bootstrapping virtual products
+   virtual-product-channel])
+
+(def dispatch-behavior
```
chris-durbin: Duncan showed me an example of how to implement a protocol using normal functions and clojure.core/extend. This seems really clean and allows for easier unit testing (though I didn't add unit tests in this example). I think we should use this approach whenever we define new protocols.

mschmele: Agreed - this is so much easier to understand

oubiwann: Another really nice thing about this approach is that it's easy to emulate multiple inheritance (only if really needed!)  by simply calling `extend` again with the new parent class, and it's easy to override behaviours in subclasses with a simple `merge` or `assoc`. Extremely powerful -- while maintaining clarity and simplicity -- for multiple implementation scenarios. Which sort of makes it a coding holy grail ... ❤️ Rich Hickey.

oubiwann:
Btw, for reviewers -- this might be a good branch to look at locally ... (instead of relying upon the diffs in Github).

bootstrap-app/src/cmr/bootstrap/api/routes.clj
```diff
@@ -30,45 +30,55 @@
   [params]
   (= "true" (when (:synchronous params) (str/lower-case (:synchronous params)))))
 
+(defn- get-dispatcher
```
oubiwann: Ultimately, it might make sense to move this into the same place you define your protocols/core API. It's not a route-specific function, but instead a general "get me the implementation appropriate for the system configuration" ...

bootstrap-app/src/cmr/bootstrap/services/dispatch/core_async_dispatch.clj
```diff
@@ -0,0 +1,143 @@
+(ns cmr.bootstrap.services.dispatch.core-async-dispatch
```
oubiwann: Some notes on this -- and, again, I think you may be thinking of these things for a future PR:

* the filename is a little redundant, given the directory name -- I'd rename it to `.../dispatch/async/clj`
* a common idiom in Clojure is for the protocol API to be defined in a `core` ns, e.g., `.../dispatch/core.clj` which is used by the rest of the application and then for implementations of this to go in `.../dispatch/impl`, 
* as such I would really rename this ns to `.../dispatch/impl/async.clj`

Note that the *only* place that would ever get `require`d would be `dispatch.core` (with the possible exception of unit tests). See my comments below on other things that need to be moved that will help this make more sense (e.g., the `extend` calls).

bootstrap-app/src/cmr/bootstrap/services/dispatch/core_async_dispatch.clj
```diff
@@ -0,0 +1,143 @@
...
+   :index-system-concepts index-system-concepts
+   :index-concepts-by-id index-concepts-by-id
+   :delete-concepts-from-index-by-id delete-concepts-from-index-by-id
+   :bootstrap-virtual-products bootstrap-virtual-products})
+
+(extend CoreAsyncDispatcher
+        dispatch-protocol/Dispatch
+        dispatch-behavior)
+
+(defn- create-default-channels
```
oubiwann: I'm curious about this (I know this has nothing to do with your PR and was there before) ... it seems odd that so many channels are created here. Usually, for this sort of thing, one would create one channel and then send different message types to accomplish the same goal. I guess I'd need to hear about the communications between components, original design reqs, etc.

bootstrap-app/src/cmr/bootstrap/services/dispatch/dispatch_protocol.clj
```diff
@@ -0,0 +1,43 @@
+(ns cmr.bootstrap.services.dispatch.dispatch-protocol
```
oubiwann: As noted above, I'd rename this to simply `dispatch.core` and move the `extend`s and implementation-agnostic functions here.

bootstrap-app/src/cmr/bootstrap/services/dispatch/core_async_dispatch.clj
```diff
@@ -0,0 +1,143 @@
+(ns cmr.bootstrap.services.dispatch.core-async-dispatch
+  "Provides methods to insert migration requets on the appropriate channels."
+  (:require
+    [clojure.core.async :as async :refer [>!]]
+    [cmr.bootstrap.services.dispatch.dispatch-protocol :as dispatch-protocol]
```
oubiwann: I don't believe you want to be doing anything protocol related here ... this becomes more clear once you re-org the code into implementation namespaces.

bootstrap-app/src/cmr/bootstrap/services/dispatch/core_async_dispatch.clj
```diff
@@ -0,0 +1,143 @@
...
+   :migrate-collection migrate-collection
+   :index-provider index-provider
+   :index-data-later-than-date-time index-data-later-than-date-time
+   :index-collection index-collection
+   :index-system-concepts index-system-concepts
+   :index-concepts-by-id index-concepts-by-id
+   :delete-concepts-from-index-by-id delete-concepts-from-index-by-id
+   :bootstrap-virtual-products bootstrap-virtual-products})
+
+(extend CoreAsyncDispatcher
```
oubiwann: You will want to do the `extend`ing in the `core` namespace where the API/protocol is defined. This will let you use that namespace and only that namespace for all implementations.

bootstrap-app/src/cmr/bootstrap/services/dispatch/synchronous_dispatch.clj
```diff
@@ -0,0 +1,73 @@
+(ns cmr.bootstrap.services.dispatch.synchronous-dispatch
+  "Functions implementing the dispatch protocol to support synchronous calls."
+  (:require
+   [cmr.bootstrap.data.bulk-index :as bulk-index]
+   [cmr.bootstrap.data.bulk-migration :as bulk-migration]
+   [cmr.bootstrap.data.virtual-products :as virtual-products]
+   [cmr.bootstrap.services.dispatch.dispatch-protocol :as dispatch-protocol]))
```
oubiwann: Same comment as above -- this import shouldn't be here.

bootstrap-app/src/cmr/bootstrap/services/dispatch/synchronous_dispatch.clj
```diff
@@ -0,0 +1,73 @@
...
+   :migrate-collection migrate-collection
+   :index-provider index-provider
+   :index-data-later-than-date-time index-data-later-than-date-time
+   :index-collection index-collection
+   :index-system-concepts index-system-concepts
+   :index-concepts-by-id index-concepts-by-id
+   :delete-concepts-from-index-by-id delete-concepts-from-index-by-id
+   :bootstrap-virtual-products bootstrap-virtual-products})
+
+(extend SynchronousDispatcher
```
oubiwann: Same comment as above -- this should be in `core`.

bootstrap-app/src/cmr/bootstrap/services/dispatch/synchronous_dispatch.clj
```diff
@@ -0,0 +1,73 @@
+(ns cmr.bootstrap.services.dispatch.synchronous-dispatch
```
oubiwann: Same comment as above -- I'd move to `.../dispatch/impl/sync.clj`.

bootstrap-app/src/cmr/bootstrap/services/dispatch/core_async_dispatch.clj
oubiwann: Some examples from the wild (done differently, but with similar code organization with regard to `impl`):
 * https://github.com/jonase/eastwood/tree/master/src/eastwood/copieddeps/dep10/clojure/tools/reader
 * https://github.com/clojure/core.async/tree/master/src/main/clojure/clojure/core/async
 * https://github.com/omcljs/om/tree/master/src/main/om/next
 * https://github.com/reagent-project/reagent/tree/master/src/reagent
 * https://github.com/funcool/catacumba/tree/master/src/clojure/catacumba
 * https://github.com/clojang/zhang/tree/release/0.1.0/src/zhang/process

bootstrap-app/src/cmr/bootstrap/system.clj
```diff
@@ -28,7 +27,9 @@
    [cmr.metadata-db.config :as mdb-config]
    [cmr.metadata-db.system :as mdb-system]
    [cmr.oracle.connection :as oracle]
-   [cmr.transmit.config :as transmit-config]))
+   [cmr.transmit.config :as transmit-config]
+   [cmr.bootstrap.services.dispatch.core-async-dispatch :as core-async-dispatch]
+   [cmr.bootstrap.services.dispatch.synchronous-dispatch :as synchronous-dispatch]))
```
oubiwann: With the changes mentioned above, you wouldn't need to import both of these; instead, you'd just do:
```
  [cmr.bootstrap.services.dispatch.core :as dispatch]
```
and then below you'd call `(dispatch/<some factory function> config-data)` and then create the appropriate dispatcher given the config values.

bootstrap-app/src/cmr/bootstrap/api/routes.clj
chris-durbin: Agreed I moved it once but didn't like it here or in the bootstrap service namespace and figured I should just get a PR submitted and continue thinking of where to move it. That sounds like a good home.

bootstrap-app/src/cmr/bootstrap/services/dispatch/core_async_dispatch.clj
chris-durbin: I did not work on the original implementation of bootstrap, but I think the reasoning was they wanted a different number of concurrent workers for different types of messages and so different channels made sense. I think eventually too many channels were added as more functionality was added where a few types of messages should have shared the same channels.

But... I'm trying to move towards a message queue implementation for the sake of scaling so I'm not planning on changing any of the channel configuration.

bootstrap-app/src/cmr/bootstrap/services/dispatch/core_async_dispatch.clj
chris-durbin: I'm sure you are right. A couple of reasons I went with the approach I did:

1) Naming stuff with redundant names. We all tend to use atom/sublime type editors where only the filename is displayed in the tab. At a retrospective awhile back people asked that we include descriptive, though redundant based on the path, information rather than have 5 tabs open named core.clj. Probably a bad decision.
2) I treated the protocol definition just like I would an interface class in Java. The interface would not know anything about its implementors. I like a couple of things about this approach. The implementor clearly shows what protocol it is implementing, and it feels like the right place for calling extend directly in the namespace.

I think I like the core namespace idea being on the only public namespace, so I'll look to make the change in my next PR.

oubiwann: Ah, so it seems that you and I have a similar philosophy about this :-) I gave pretty middle-ground Clojure example -- _but_ :-) how I usually do this, when no one's looking, is define my protocols in `proj.some.ns.protocols` then require it as well as use potemkin in `proj.some.ns.thing.core` and then put implementations in `proj.some.ns.thing.impl.*`. Which I think is kind of what you were getting at ...? (basically creating protocols like an interface in Java).

oubiwann: Which means I get to:
1. keep protocols organized together for large projects (makes a great reference, and it's nice to know where they all like)
1. still use protocols in Clojure like they are meant to (the engine of APIs that have multiple implementations), in particular:
1. I get to `(require '[proj.some.ns.thing.core :as thing-api])` and then call `(thing-api/func ...)` because I used potemkin to import the vars defined in the protocol itself. And since  this is where the `extend`s happen, all of the implementations are treated as one from the API user's perspective (since extends dispatches based on type, under the hood; never a need to go poking into `impl`, iow).

mschmele:
+1ing Lauren's goto idea

bootstrap-app/src/cmr/bootstrap/api/routes.clj
```diff
@@ -30,45 +30,55 @@
...
   [context provider-id-map params]
-  (let [provider-id (get provider-id-map "provider_id")
-        synchronous (synchronous? params)
+  (let [dispatcher (get-dispatcher context params :index-provider)
+        provider-id (get provider-id-map "provider_id")
         start-index (Long/parseLong (get params :start_index "0"))
-        result (bs/index-provider context provider-id synchronous start-index)
-        msg (if synchronous
+        result (bs/index-provider context dispatcher provider-id start-index)
+        msg (if
```
chris-durbin: Thanks, fixed all 3.

PR #215, opened 2017-08-01
Cmr 3689

laurenfrederick:
Add project error validation that start date if before end date. Add project warning validation that start and end dates are in the past. Add temporal extents warning validation that dates are in the past.

### DIFF
--- a/umm-spec-lib/src/cmr/umm_spec/date_util.clj
+++ b/umm-spec-lib/src/cmr/umm_spec/date_util.clj
@@ -87,18 +87,18 @@
         Exception e nil))))
 
 (defn is-in-past?
-  "Parse date and return true if date is in the past, false if not, nil if date
-  is nil"
+  "Parse date and return true if date is in the past, false if not or date is nil"
   [date]
-  (when date
-    (time-core/before? date (time-keeper/now))))
+  (if date
+    (time-core/before? date (time-keeper/now))
+    false))
 
 (defn is-in-future?
-  "Parse date and return true if date is in the future, false if not, nil if date
-  is nil"
+  "Parse date and return true if date is in the future, false if not or date is nil"
   [date]
-  (when date
-    (time-core/after? date (time-keeper/now))))
+  (if date
+    (time-core/after? date (time-keeper/now))
+    false))
 
 (defn update-metadata-dates
   "Update the Date to current date, for a given Type: date-type, in MetadataDates of a umm record"
--- a/umm-spec-lib/src/cmr/umm_spec/date_util.clj
+++ b/umm-spec-lib/src/cmr/umm_spec/date_util.clj
@@ -2,7 +2,7 @@
   "Useful UMM date values and functions."
   (:require
    [clj-time.format :as f]
-   [clj-time.core :as t-core]
+   [clj-time.core :as time-core]
    [clojure.string :as str]
    [cmr.common.date-time-parser :as p]
    [cmr.common.time-keeper :as time-keeper]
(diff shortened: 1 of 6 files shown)


umm-spec-lib/src/cmr/umm_spec/validation/umm_spec_validation_utils.clj
```diff
@@ -13,6 +14,18 @@
         {field-path [(format "%%s must be unique. This contains duplicates named [%s]."
                              (str/join ", " duplicate-names))]}))))
 
+(defn date-in-past-validator
+  "Validate that the date is in the past"
+  [field-path value]
+  (when (and value (not (date/is-past? value)))
+    {field-path ["Date should be in the past."]}))
+
+(defn date-in-future-validator
```
laurenfrederick: This isn't used yet, but there is a ticket next sprint that would use it. I figured it was easy to add while I was doing date in past.

umm-spec-lib/src/cmr/umm_spec/date_util.clj
```diff
@@ -81,11 +81,25 @@
...
         Exception e nil))))
 
+(defn is-past?
+  "Parse date and return true if date is in the past, false if not, nil if date
+  is nil"
+  [date]
+  (when date
+    (time-core/before? date (time-keeper/now))))
+
+(defn is-future?
```
mschmele: Nitpick: this name might be confusing, because `futures` are a thing in clojure

https://clojuredocs.org/clojure.core/future

umm-spec-lib/test/cmr/umm_spec/test/date_util_test.clj
```diff
@@ -3,6 +3,7 @@
   (:require
    [clj-time.format :as f]
    [clojure.test :refer :all]
+   [cmr.common.time-keeper :as time-keeper]
```
mschmele: I don't remember why, but @chris-durbin or @ygliuvt told me to use `[cmr.system-int-test.utils.dev-system-util :as dev-system-util]` here. Something to do with the dev-system time keeper properly resetting time after you freeze.

It works effectively the same: `(dev-system-util/freeze-time! "2010-01-01T10:00:00Z")`

umm-spec-lib/src/cmr/umm_spec/date_util.clj
laurenfrederick: I changed both functions to `is-in-past?` and `is-in-future?`

umm-spec-lib/test/cmr/umm_spec/test/date_util_test.clj
laurenfrederick: Thank you for that. I changed it.

laurenfrederick: And I put timekeeper back...

chris-durbin: Any time you are in the system integration test namespaces and you need to change time for some app running in dev-system you need to use dev-system-util (system integration test and dev-system are running in two different processes in CI).

In this case Lauren's test was not running in system-int-test so it was correct to use timekeeper directly.

umm-spec-lib/src/cmr/umm_spec/date_util.clj
```diff
@@ -81,11 +81,25 @@
   [date]
   (let [date (f/parse date)]
     (try
-      (t-core/equal? date
+      (time-core/equal? date
                      date)
       (catch
         Exception e nil))))
 
+(defn is-in-past?
```
chris-durbin: I think we should make these predicate functions (ie they return true or false). It seems like the right thing to do is return false instead of nil in both cases or even throw an exception when the argument is nil.

laurenfrederick: I changed it.

PR #217, opened 2017-08-02
CMR-4379: Allow crawlers to index our collection landing pages

chris-durbin:
under /search/concepts route.

We cannot verify this ticket in SIT or UAT because we only submit our production site for indexing in Google (obviously we don't want to direct science users looking for data to our SIT our UAT environments).

I tested with Google's robots testing tools to verify that removing the one line from robots.txt fixed the problem with loading the landing pages.


oubiwann:
@chris-durbin If you're still up, I think this qualifies for an immediate merge -- I don't think we need 2 +1s for this ...

PR #218, opened 2017-08-02
CMR-4381: As an admin, I should be able to force update elasticsearch…

ygliuvt:
… indexes.

### DIFF
--- a/indexer-app/src/cmr/indexer/api/routes.clj
+++ b/indexer-app/src/cmr/indexer/api/routes.clj
@@ -64,16 +64,16 @@
         {:status 204})
 
       ;; Sends an update to the index set to update mappings and index settings.
-      (POST "/update-indexes" {:keys [request-context params headers]}
+      (POST "/update-indexes" {:keys [request-context params]}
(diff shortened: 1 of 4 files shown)


indexer-app/src/cmr/indexer/api/routes.clj
```diff
@@ -64,16 +64,16 @@
...
+      (POST "/update-indexes" {:keys [request-context params]}
         (acl/verify-ingest-management-permission request-context :update)
-        (index-svc/update-indexes request-context)
+        (index-svc/update-indexes request-context params)
         {:status 200})
 
       ;; This is just an alias for /update-indexes to make it easy to update indexes
       ;; after a deployment using the same deployment code that other apps use for db-migrate.
-      (POST "/db-migrate" {:keys [request-context]}
+      (POST "/db-migrate" {:keys [request-context params]}
```
oubiwann: Nice.

ygliuvt:
I don't think any test is needed for the change. So none is included.

ygliuvt:
The reason I forgo automated testing here is the logic is too simple. It is not worth the testing setup.

PR #222, opened 2017-08-03
CMR-4386: Use a small buffer when verifying request body sizes

chris-durbin:
Every request submitted to any CMR application was causing a 50MB buffer to be allocated in order to validate the size of the request body was less than 50MB.

(We recently increased the max request body size from 5MB to 50MB in SIT and UAT, but that change has not made it to production). So in production we are allocating a 5MB buffer on every request.

We started seeing Java heap space errors for some apps in SIT in UAT, which can easily happen when many requests are actively validating the size of the request body in parallel.

This change is to use buffers of 512 bytes to validate the request body. If a request is larger than 512 bytes we will continue allocating additional buffers until we've read the entire body or hit 50MB and we reject the request.

I've done some performance testing which indicates this code change if faster for all request bodies 50KB or smaller. Almost all of our requests have request bodies smaller than 50KB.

Larger requests take longer to validate with the smaller buffers, but even a 50MB request would complete in under 5 seconds vs. 2 seconds previously.

I want to rerun some of the performance numbers, because I realized my test setup was artificially increasing the times. I will post them in the ticket when complete.

### DIFF
--- a/common-lib/src/cmr/common/api/web_server.clj
+++ b/common-lib/src/cmr/common/api/web_server.clj
@@ -6,7 +6,8 @@
    [cmr.common.lifecycle :as lifecycle]
    [cmr.common.log :refer [debug info warn error]]
    [cmr.common.mime-types :as mt]
-   [ring.adapter.jetty :as jetty])
+   [ring.adapter.jetty :as jetty]
+   [clojure.core.reducers :as reducers])
   (:import
    (java.io ByteArrayInputStream InputStream)
    (org.eclipse.jetty.server Server NCSARequestLog Connector HttpConnectionFactory)
@@ -46,6 +47,12 @@
   report in Cubby can be in the 5 to 10 MB range."
  (* 50 ONE_MB))
 
+(def buffer-size
+  "Number of bytes to allocate for the buffer used for verifying request body size is not too
+  large. We want to use a smaller buffer size than the max request body size so for small requests
+  we use a small amount of memory to perform the verification."
+  512)
+
 (defn- routes-fn-verify-size
   "Takes the passed in routes function and wraps it with another function that will verify request
    sizes do not exceed the maximum size.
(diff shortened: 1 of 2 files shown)


common-lib/src/cmr/common/api/web_server.clj
```diff
@@ -55,35 +94,37 @@
...
-                 :body "Request body exceeds maximum size"}
-                ;; Put the request body into a new input stream since the current has been read from
-                (routes-fn (assoc request :body (ByteArrayInputStream.
-                                                 input-byte-array
-                                                 0 ;; offset
-                                                 (+ total-bytes-read bytes-read)))))
-              (recur (+ total-bytes-read bytes-read)))))))))
+      (loop [total-bytes-read 0
+             bytes-read-for-current-buffer 0
+             input-byte-arrays (transient [])
```
chris-durbin: Using transients because we want calls to add additional buffers to be as fast as possible.

common-lib/src/cmr/common/api/web_server.clj
```diff
@@ -46,6 +47,44 @@
...
+  {:status 413
+   :content-type :text
+   :body "Request body exceeds maximum size"})
+
+(defn- byte-stream-from-buffers
+  "Reconstructs request body into a new input stream from all of the buffers used to validate the
+  request body size. This is required so that the request body can be read successfully from
+  within application routes code."
+  [buffers total-bytes]
+  (let [request-body-bytes (->> buffers
```
chris-durbin: This was the fastest code I could come up with to take multiple byte-arrays and consolidate them into a single byte array. I should probably update the docstring to indicate this is performance sensitive.

If anyone has ideas to make this faster please let me know.

oubiwann: +1 for updating the docstring ... I'd have to experiment to explore options for speed increases.

common-lib/src/cmr/common/api/web_server.clj
```diff
@@ -46,6 +47,44 @@
...
+                                persistent!
+                                (reducers/mapcat vec)
+                                reducers/foldcat
+                                (into [])
```
oubiwann: You should be able to use `vec` here instead of `(into [])` -- let me know if that doesn't work in this case (that would really surprise me).

common-lib/src/cmr/common/api/web_server.clj
```diff
@@ -46,6 +47,44 @@
...
+                                byte-array)]
+    (ByteArrayInputStream. request-body-bytes 0 total-bytes)))
+
 (defn- routes-fn-verify-size
```
oubiwann: I know you didn't touch this part ... but!

In Ring/Compojure parlance, this is middleware which is usually named with a `wrap-` and gets passed `handler`. I'd recommend renaming this function to `wrap-size-verification`.  I'd also change `[routes-fn]` to `[handler]`. They return a `fn` that takes the request as its arg, so the rest is good.

common-lib/src/cmr/common/api/web_server.clj
```diff
@@ -55,35 +94,37 @@
...
-                {:status 413
-                 :content-type :text
-                 :body "Request body exceeds maximum size"}
-                ;; Put the request body into a new input stream since the current has been read from
-                (routes-fn (assoc request :body (ByteArrayInputStream.
-                                                 input-byte-array
-                                                 0 ;; offset
-                                                 (+ total-bytes-read bytes-read)))))
-              (recur (+ total-bytes-read bytes-read)))))))))
+      (loop [total-bytes-read 0
```
oubiwann: +1 for `loop` here; is often faster than `for` (I try to avoid `for` in general, though ...)

common-lib/src/cmr/common/api/web_server.clj
```diff
@@ -55,35 +94,37 @@
...
+             bytes-read-for-current-buffer 0
+             input-byte-arrays (transient [])
+             current-buffer (byte-array buffer-size)]
+        (let [{:keys [bytes-read at-end? buffer-full?]}
+              (read-into-buffer body-input-stream current-buffer bytes-read-for-current-buffer
+                                (- buffer-size bytes-read-for-current-buffer))
+              input-byte-arrays (if (or at-end? buffer-full?)
```
oubiwann: either in-line comment what you're doing with this conditional in the `let` block, or add some explanation in the docstring.

common-lib/src/cmr/common/api/web_server.clj
```diff
@@ -55,35 +94,37 @@
...
+                                  (conj! input-byte-arrays current-buffer)
+                                  input-byte-arrays)
+              current-buffer (if (and buffer-full? (not at-end?))
```
oubiwann: same comment here about commenting/documenting the `if` ...

common-lib/src/cmr/common/api/web_server.clj
```diff
@@ -55,35 +94,37 @@
...
+                               (byte-array buffer-size)
+                               current-buffer)
+              bytes-read-for-current-buffer (if buffer-full?
```
oubiwann: same comment about `if` in `let` ... we should call this "`let`-abuse tax" 😉

common-lib/src/cmr/common/api/web_server.clj
```diff
@@ -1,6 +1,7 @@
 (ns cmr.common.api.web-server
   "Defines a web server component."
   (:require
+   [clojure.core.reducers :as reducers]
```
oubiwann: Woo-hoo! 🏆 and 🍺 for @chris-durbin!

(I feel shame for not having said this the first time around ... I'm going to my room now)

common-lib/src/cmr/common/api/web_server.clj
chris-durbin: The exception is:

Unable to convert: class clojure.core.reducers.Cat to Object[]

common-lib/src/cmr/common/api/web_server.clj
chris-durbin: Done

common-lib/src/cmr/common/api/web_server.clj
chris-durbin: Went for in-line comments rather than refactor.

common-lib/src/cmr/common/api/web_server.clj
chris-durbin: Done

common-lib/src/cmr/common/api/web_server.clj
chris-durbin: Done

common-lib/src/cmr/common/api/web_server.clj
oubiwann: oh, wow -- very interesting. Thanks for pasting the error (now I'm super-curious ... my "dig" instincts are twitching ...)

common-lib/src/cmr/common/api/web_server.clj
```diff
@@ -46,44 +47,92 @@
...
+  [buffers total-bytes]
+  (let [request-body-bytes (->> buffers
+                                persistent!
+                                (reducers/mapcat vec)
+                                reducers/foldcat
+                                (into [])
+                                byte-array)]
+    (ByteArrayInputStream. request-body-bytes 0 total-bytes)))
+
+(defn- wrap-request-body-size-validation
```
oubiwann: Looks great!

common-lib/src/cmr/common/api/web_server.clj
```diff
@@ -46,44 +47,92 @@
...
-                                                 (+ total-bytes-read bytes-read)))))
-              (recur (+ total-bytes-read bytes-read)))))))))
+      (loop [total-bytes-read 0
+             bytes-read-for-current-buffer 0
+             all-buffers (transient [])
+             current-buffer (byte-array buffer-size)]
+        (let [{:keys [bytes-read at-end? buffer-full?]}
+              (read-into-buffer body-input-stream current-buffer bytes-read-for-current-buffer
+                                (- buffer-size bytes-read-for-current-buffer))
+              ;; Put the current buffer into a vector of all buffers if we've finished reading
```
oubiwann: Thanks for all these!

common-lib/src/cmr/common/api/web_server.clj
```diff
@@ -46,44 +47,92 @@
...
+              ;; full. If creating a new buffer reset the bytes read back to 0.
+              bytes-read-for-current-buffer (if buffer-full?
+                                              0
+                                              (+ bytes-read-for-current-buffer bytes-read))
+              total-bytes-read (+ total-bytes-read bytes-read)]
+
+          ;; If the entire request body has been read or if the amount of bytes read is
+          ;; greater than MAX_REQUEST_BODY_SIZE, process based on num bytes read
+          ;; otherwise loop to continue reading the request body
+          (if (or at-end? (> total-bytes-read MAX_REQUEST_BODY_SIZE))
```
ygliuvt: Change this to 
```
(if (> total-bytes-read MAX_REQUEST_BODY_SIZE)
 request-body-too-large-response
 (if at-end?
    (let [request-body (byte-stream-from-buffers all-buffers total-bytes-read)]
          (handler (assoc request :body request-body)))
   (recur ...)))

File: common-lib/src/cmr/common/api/web_server.clj [hunk 13]
```diff
@@ -46,44 +47,92 @@
   report in Cubby can be in the 5 to 10 MB range."
  (* 50 ONE_MB))
 
-(defn- routes-fn-verify-size
+(def buffer-size
```
ygliuvt: I am assuming that you get to this number after considering our average request size and playing with different settings with benchmark testing.

ygliuvt: I am wondering how big a difference different buffer sizes will make? and is it worthy to considering using different buffer sizes (vs one buffer size here) to help with the performance?

PR #224, opened 2017-08-03
CMR-4350 fine tune var perms

oubiwann:
The primary purpose of the changes in this PR is to add a constraint that will improve fine-grained permissions on variables. In particular, we previously had no check in place for attempting to create a variable with the same variable name and provider id but different native-id from one that already exists in the db.

Additionally, this PR has:

* made updates for better organization of various messages into the proper namespace(s)
* added another variable integration test
* updated testing sections of a couple README files

### DIFF
--- a/metadata-db-app/int_test/cmr/metadata_db/int_test/concepts/collection_save_test.clj
+++ b/metadata-db-app/int_test/cmr/metadata_db/int_test/concepts/collection_save_test.clj
@@ -55,7 +55,7 @@
             _ (Thread/sleep 10)
             {second-revision-id :revision-id} (util/save-concept initial-collection)
             _ (Thread/sleep 10)
-            {tombstone-revision-id :revision-id} (util/delete-concept concept-id)
+            {tombstone-revision-id :revision-id} (util/save-concept {:deleted true :concept-id concept-id})
             _ (Thread/sleep 10)
             {final-revision-id :revision-id} (util/save-concept initial-collection)
             [initial-revision
--- a/metadata-db-app/int_test/cmr/metadata_db/int_test/concepts/collection_save_test.clj
+++ b/metadata-db-app/int_test/cmr/metadata_db/int_test/concepts/collection_save_test.clj
@@ -39,14 +39,6 @@
           coll2 (util/collection-concept "SMAL_PROV2" 2 {:native-id "foo"})]
       (c-spec/save-distinct-concepts-test coll1 coll2))))
 
-(defn- created-at-same?
-  "Returns true if the `created-at` for the given concept revisions are the same
-  and none of them are nil"
-  [& concepts]
-  (let [created-ats (map :created-at concepts)]
-    (and (apply = created-ats)
-         (not-any? nil? created-ats))))
-
 (deftest save-collection-created-at-test
   (testing "Save collection multiple times gets same created-at"
     (doseq [provider-id ["REG_PROV" "SMAL_PROV1"]]
@@ -63,7 +55,7 @@
             _ (Thread/sleep 10)
             {second-revision-id :revision-id} (util/save-concept initial-collection)
             _ (Thread/sleep 10)
-            {tombstone-revision-id :revision-id} (util/save-concept {:deleted true :concept-id concept-id})
+            {tombstone-revision-id :revision-id} (util/delete-concept concept-id)
             _ (Thread/sleep 10)
             {final-revision-id :revision-id} (util/save-concept initial-collection)
             [initial-revision
@@ -74,7 +66,7 @@
                                     second-revision-id
                                     tombstone-revision-id
                                     final-revision-id])]
-        (is (created-at-same? initial-revision second-revision tombstone final-revision))))))
+        (is (util/created-at-same? initial-revision second-revision tombstone final-revision))))))
 
 (deftest save-collection-post-commit-constraint-violations
   (testing "duplicate entry titles"
--- a/metadata-db-app/int_test/cmr/metadata_db/int_test/concepts/variable_save_test.clj
+++ b/metadata-db-app/int_test/cmr/metadata_db/int_test/concepts/variable_save_test.clj
@@ -7,30 +7,36 @@
    [cmr.metadata-db.int-test.concepts.concept-save-spec :as c-spec]
    [cmr.metadata-db.int-test.utility :as util]))
 
-(def provider-id "REG_PROV")
-
-;;; fixtures
+;;; fixtures & one-off utility functions
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
 (use-fixtures :each (util/reset-database-fixture
-                     {:provider-id provider-id :small false}))
+                     {:provider-id "PROV1" :small false}
+                     {:provider-id "PROV2" :small false}))
 
 (defmethod c-spec/gen-concept :variable
   [_ provider-id uniq-num attributes]
-  (util/variable-concept provider-id uniq-num attributes))
+  (util/variable-concept "PROV1" uniq-num attributes))
+
+(defn- get-revisions
+  [concept-id initial-revision-id second-revision-id tombstone-revision-id
+   final-revision-id]
+  (mapv #(:concept (util/get-concept-by-id-and-revision concept-id %))
+        [initial-revision-id second-revision-id tombstone-revision-id
+         final-revision-id]))
 
 ;; tests
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
-(deftest save-variable-test
-  (c-spec/general-save-concept-test :variable [provider-id]))
+(deftest save-variable
+  (c-spec/general-save-concept-test :variable ["PROV1"]))
 
-(deftest save-variable-with-missing-required-parameters-test
+(deftest save-variable-with-missing-required-parameters
   (c-spec/save-test-with-missing-required-parameters
-    :variable [provider-id] [:concept-type :provider-id :native-id :extra-fields]))
+    :variable ["PROV1"] [:concept-type :provider-id :native-id :extra-fields]))
 
-(deftest save-variable-created-at-test
+(deftest save-variable-created-at
   (testing "Save variable multiple times gets same created-at"
-    (let [initial-var (util/variable-concept provider-id 2)
+    (let [initial-var (util/variable-concept "PROV1" 2)
           ;; 1) Save a variable
           ;; 2) Then wait for a small period of time before saving it again
           ;; 3) Then wait again and save a tombstone.
@@ -50,15 +56,43 @@
                                                 concept-id)
           _ (Thread/sleep 100)
           {final-revision-id :revision-id} (util/save-concept initial-var)
-          [initial-revision
-           second-revision
-           tombstone
-           final-revision] (mapv #(:concept (util/get-concept-by-id-and-revision concept-id %))
-                                 [initial-revision-id
-                                  second-revision-id
-                                  tombstone-revision-id
-                                  final-revision-id])]
-      (is (util/created-at-same? initial-revision
-                                 second-revision
-                                 tombstone
-                                 final-revision)))))
+          revisions (get-revisions concept-id initial-revision-id
+                                   second-revision-id tombstone-revision-id
+                                   final-revision-id)]
+      (is (apply util/created-at-same? revisions)))))
+
+(deftest save-variable-with-conflicting-native-id
+  (testing "Save a variable"
+    (let [var1 (util/variable-concept "PROV1" 1)
+          {:keys [status revision-id concept-id]} (util/save-concept var1)]
+      (is (= status 201))
+      (is (= 1 revision-id))
+      (testing "and another with all the same data"
+        (let [var2 var1
+              {:keys [status revision-id]} (util/save-concept var1)]
+          (is (= 201 status))
+          (is (= 2 revision-id))))
+      (testing "and another with same data but different provider"
+        (let [var3 (util/variable-concept "PROV2" 1)
+              {:keys [status revision-id]} (util/save-concept var3)]
+          (is (= status 201))
+          (is (= 1 revision-id))))
+      (testing "and another the same data but with a different native-id"
+        (let [var4 (assoc var1 :native-id "var-native-different")
+              failed-concept (util/save-concept var4)
+              failed-revision-id 2
+              failed-concept-id "V1200000002-PROV1"]
+          (is (= 409 (:status failed-concept)))
+          (is (= [(format (str "Revision [%s] of concept [%s] has the same "
+                               "provider [%s] and variable name [%s] but "
+                               "different native id [%s] than the new concept "
+                               "[%s] with revision [%s] and native id [%s].")
+                          failed-revision-id
+                          concept-id
+                          (:provider-id var1)
+                          (get-in var1 [:extra-fields :variable-name])
+                          (:native-id var1)
+                          failed-concept-id
+                          revision-id
+                          (:native-id var4))]
+                  (:errors failed-concept))))))))
--- a/metadata-db-app/int_test/cmr/metadata_db/int_test/concepts/variable_save_test.clj
+++ b/metadata-db-app/int_test/cmr/metadata_db/int_test/concepts/variable_save_test.clj
@@ -19,6 +19,10 @@
   (util/variable-concept "PROV1" uniq-num attributes))
 
 (defn- get-revisions
+  "This is a utility function that returns revisions of interest (given the
+  respective revision ids).
+
+  The results of this function are intended to be used with `(apply ...)`."
   [concept-id initial-revision-id second-revision-id tombstone-revision-id
    final-revision-id]
   (mapv #(:concept (util/get-concept-by-id-and-revision concept-id %))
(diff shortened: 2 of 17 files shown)


metadata-db-app/int_test/cmr/metadata_db/int_test/concepts/collection_save_test.clj
```diff
@@ -63,7 +55,7 @@
             _ (Thread/sleep 10)
             {second-revision-id :revision-id} (util/save-concept initial-collection)
             _ (Thread/sleep 10)
-            {tombstone-revision-id :revision-id} (util/save-concept {:deleted true :concept-id concept-id})
+            {tombstone-revision-id :revision-id} (util/delete-concept concept-id)
```
ygliuvt: This test is testing metadata-db save collection endpoint. I don't think we should change this to delete concept, which makes the test not testing what it is supposed to.

ygliuvt: Wondering why you decided to change this.

oubiwann: Hrm, I thought that these two were identical (I'll look again) ... which led me to believe that `delete-concept` was newer and the test hadn't been updated yet.

oubiwann: Nope, you're totally correct -- changing it back ...

metadata-db-app/int_test/cmr/metadata_db/int_test/concepts/variable_save_test.clj
```diff
@@ -7,20 +7,92 @@
...
+(use-fixtures :each (util/reset-database-fixture
+                     {:provider-id "PROV1" :small false}
+                     {:provider-id "PROV2" :small false}))
 
 (defmethod c-spec/gen-concept :variable
   [_ provider-id uniq-num attributes]
-  (util/variable-concept provider-id uniq-num attributes))
+  (util/variable-concept "PROV1" uniq-num attributes))
+
+(defn- get-revisions
```
ygliuvt: Add a docstring.

oubiwann: Done.

metadata-db-app/src/cmr/metadata_db/services/concept_constraints.clj
```diff
@@ -14,15 +14,22 @@
...
+  that matches the passed in concept"
+  [db provider concept query]
+  (->> query
+       (merge {:concept-type (:concept-type concept)
+               :provider-id (:provider-id concept)})
+       (c/find-latest-concepts db provider)
+       ;; Remove tombstones from the list of concepts
+       (remove :deleted)))
+
 (defn- find-latest-matching-concepts
```
ygliuvt: Not sure why you want to refactor this function when you could have just used it as is. Am I missing anything?

oubiwann: Ah, yeah -- that was deep in a debugging session where I actually had different code for `find-latest` and `find-latest-matching` -- I'll reduce this back to one function now ...

oubiwann: Oh, wait -- no. I did refactor other parts of the code to use this. It was a redundancy elimination change. I'll update the docstring(s)/code comments.

oubiwann: Updated.

metadata-db-app/int_test/cmr/metadata_db/int_test/concepts/variable_save_test.clj
```diff
@@ -7,20 +7,96 @@
...
+          {:keys [status revision-id concept-id]} (util/save-concept var1)]
+      (is (= status 201))
+      (is (= 1 revision-id))
+      (testing "and another with all the same data"
+        (let [var2 var1
+              {:keys [status revision-id]} (util/save-concept var1)]
+          (is (= 201 status))
+          (is (= 2 revision-id))))
+      (testing "and another with same data but different provider"
+        (let [var3 (util/variable-concept "PROV2" 1)
```
ygliuvt: Change this to `(let [var3 (assoc var1 :provider-id "PROV2")` to make it clearer what is different.

metadata-db-app/int_test/cmr/metadata_db/int_test/concepts/variable_save_test.clj
```diff
@@ -7,20 +7,96 @@
...
+              {:keys [status revision-id]} (util/save-concept var3)]
+          (is (= status 201))
+          (is (= 1 revision-id))))
+      (testing "and another the same data but with a different native-id"
+        (let [var4 (assoc var1 :native-id "var-native-different")
+              failed-concept (util/save-concept var4)
```
ygliuvt: Change `failed-concept` to `response`.

metadata-db-app/src/cmr/metadata_db/services/messages.clj
```diff
@@ -175,3 +178,43 @@
...
+               "which is lower than revision [%d] with transaction-id [%d].")
+          revision-id
+          concept-id
+          transaction-id
+          this-revision-id
+          this-transaction-id))
+
+(defn pvn-equality-failure
+  [new-concept old-concept]
+   (format (str "Revision [%d] of concept [%s] has the same provider [%s] "
```
ygliuvt: Change the error message to something like "The provider id [%s] and variable name [%s] combined must be unique. The following concepts with the same provider id and variable name were found: [%s]." to be consistent with the one returned from collection validation.

metadata-db-app/src/migrations/062_add_variable_constraint.clj
```diff
@@ -0,0 +1,54 @@
...
+(def ^:private table (format "METADATA_DB.%s" table-name))
+(def ^:private new-constraint-name "variables_con_n_r_p")
+(def ^:private new-index-name "variables_idx_n_r_p")
+(def ^:private old-constraint-name "variables_con_rev")
+(def ^:private old-index-name "variables_ucr_i")
+
+(defn- add-variable-constraint
+  []
+  (h/sql
+   (format (str "ALTER TABLE %s "
```
ygliuvt: I actually like the string literal approach more readable. Since we already know the value of the names, we should just use a string literal here.

metadata-db-app/src/cmr/metadata_db/services/messages.clj
oubiwann: I used `concept-lower-transaction-id` and `concept-higher-transaction-id` as the basis ...

metadata-db-app/src/migrations/062_add_variable_constraint.clj
oubiwann: Okay, I'd buy that.

metadata-db-app/int_test/cmr/metadata_db/int_test/concepts/variable_save_test.clj
oubiwann: Done.

metadata-db-app/int_test/cmr/metadata_db/int_test/concepts/variable_save_test.clj
oubiwann: Done.

metadata-db-app/src/cmr/metadata_db/services/messages.clj
ygliuvt: This error message is for ingest providers. They wouldn't care about the new concept id and revision id that CMR tries to create but failed. I think the collection style error message is more user friendly.

oubiwann: Got it -- thanks. Changing ...

metadata-db-app/int_test/cmr/metadata_db/int_test/concepts/variable_save_test.clj
```diff
@@ -7,20 +7,96 @@
...
+                               "[%s] with revision [%s] and native id [%s].")
+                          failed-revision-id
+                          concept-id
+                          (:provider-id var1)
+                          (get-in var1 [:extra-fields :variable-name])
+                          (:native-id var1)
+                          failed-concept-id
+                          revision-id
+                          (:native-id var4))]
+                  (:errors failed-concept))))))))
```
ygliuvt: We should add a check to verify that the new revision-id that CMR tries to create has indeed been rolled back, i.e. does not exist in metadata-db any more.

metadata-db-app/src/cmr/metadata_db/services/concept_constraints.clj
```diff
@@ -94,41 +101,80 @@
...
+  This function ensures that for these three attributes of a concept, updates
+  will only succeed if all three are the same for the most current revision-id.
+  If `provider-id` and `variable-name` are the same, but `native-id` is
+  different, this constraint will not be met and the valdiations will fail.
+
+  Any failures result in a list of error messages returned. Successful passing
+  of the constraint will result in an empty list being returned."
+  [db provider {:keys [native-id extra-fields] :as concept}]
+  (let [variable-name (:variable-name extra-fields)
+        concepts (find-latest-concepts
```
ygliuvt: Can we use `find-latest-matching-concepts` function here instead?

metadata-db-app/src/cmr/metadata_db/services/messages.clj
oubiwann: Done.

metadata-db-app/src/cmr/metadata_db/services/concept_constraints.clj
oubiwann: `find-latest-matching-concepts` only handles one field; `find-latest-concepts` is the more general form needed here, since multiple fields are used ... what are you thinking?

ygliuvt: I think we only have one field here -- variable name, provider-id is added by default. I am thinking if we call `find-latest-matching-concepts` instead, we don't need to create the `find-latest-concepts` function at all.

metadata-db-app/src/migrations/062_add_variable_constraint.clj
oubiwann: Done.

metadata-db-app/int_test/cmr/metadata_db/int_test/concepts/variable_save_test.clj
oubiwann: Done. Also added a check that response-id for the `response` here should be nil.

metadata-db-app/src/cmr/metadata_db/services/concept_constraints.clj
oubiwann: Ah-ha! I do believe you are correct ...

oubiwann: Done.

metadata-db-app/src/migrations/062_add_variable_constraint.clj
oubiwann: Oh, I thought your comment was for something else -- the reason I did it here was that the same values are used elsewhere, and it made more sense just to have one place where a string value was defined.

ygliuvt: It would be so much easier to just read the sql statement as it is, rather than going through the variable lookup. Just my 2 cents. You can keep it the way it is if you feel strongly about it.

oubiwann: Nah, I don't feel that strongly about it. And it's not like this is getting reused by anything ... I'll get rid of the variables and just use strings.

oubiwann: Oh, hey -- looks like the docstrings here need to get updated, too (from a previous migration number collision with another PR that beat mine to the code base).

metadata-db-app/src/migrations/062_add_variable_constraint.clj
```diff
@@ -0,0 +1,54 @@
...
+                "ADD CONSTRAINT %s "
+                "UNIQUE (native_id, revision_id) "
+                "USING INDEX (CREATE UNIQUE INDEX %s ON %s "
+                "(native_id, revision_id))")
+           table old-constraint-name old-index-name table-name)))
+
+(defn up
+  "Migrates the database up to version 61."
```
ygliuvt: Change `61` to `62`

metadata-db-app/src/migrations/062_add_variable_constraint.clj
```diff
@@ -0,0 +1,54 @@
...
+  []
+  (println "migrations.062-add-variable-constraint up...")
+  (drop-constraint old-constraint-name)
+  (add-variable-constraint))
+
+(defn down
+  "Migrates the database down from version 61."
```
ygliuvt: Change `61` to `62`.

metadata-db-app/src/migrations/062_add_variable_constraint.clj
oubiwann: Yup -- already caught that above ;-)

metadata-db-app/src/migrations/062_add_variable_constraint.clj
oubiwann: Ditto ;-)

metadata-db-app/src/cmr/metadata_db/services/concept_constraints.clj
```diff
@@ -94,41 +92,79 @@
...
+
+  Note that old-style validation functions (see `cmr.common.util`) expect
+  validation functions to return an empty list upon success and a list of
+  one or more error messages upon failure."
+  [new-concept old-concept]
+  (if (and (= (:provider-id new-concept)
+              (:provider-id old-concept))
+           (= (get-in new-concept [:extra-fields :variable-name])
+              (get-in old-concept [:extra-fields :variable-name]))
+           (not (= (:native-id new-concept)
```
chris-durbin: Nit: `(not=`. There are only so many parens in the world.

metadata-db-app/src/cmr/metadata_db/services/concept_constraints.clj
```diff
@@ -94,41 +92,79 @@
...
+
+  Any failures result in a list of error messages returned. Successful passing
+  of the constraint will result in an empty list being returned."
+  [db provider {:keys [native-id extra-fields] :as concept}]
+  (let [variable-name (:variable-name extra-fields)
+        concepts (find-latest-matching-concepts
+                  db provider concept :variable-name variable-name)]
+    (->> concepts
+         (mapv (partial validate-pvn-equalities concept))
+         (flatten))))
```
chris-durbin: Anytime I see `map` and `flatten` I think `mapcat`. But I guess you'd have to do `mapcat` and then `vec` in this case.

metadata-db-app/src/migrations/062_add_variable_constraint.clj
```diff
@@ -0,0 +1,54 @@
...
+                "UNIQUE (native_id, revision_id) "
+                "USING INDEX (CREATE UNIQUE INDEX %s ON %s "
+                "(native_id, revision_id))")
+           table old-constraint-name old-index-name table-name)))
+
+(defn up
+  "Migrates the database up to version 61."
+  []
+  (println "migrations.062-add-variable-constraint up...")
+  (drop-constraint old-constraint-name)
```
chris-durbin: I assume we don't have to worry about any existing violations in the table? Otherwise you might have to truncate the table first (I'm assuming we're still fine with throwing out any variables).

ygliuvt: This is a good point. We should keep an eye out for failing migrations and fix the conflicts if there are any. I don't think it is worth the effort to handle this in the code given nobody is creating variables except ourselves and it is unlikely we have conflicts. We might want to do a manual check though before deployment.

metadata-db-app/src/cmr/metadata_db/services/concept_constraints.clj
oubiwann: Thanks :-) Fixed.

metadata-db-app/src/cmr/metadata_db/services/concept_constraints.clj
oubiwann: Yeah, it's a 6, 1/2*dozen situation ...

metadata-db-app/src/migrations/062_add_variable_constraint.clj
oubiwann: Might be worth the truncate anyway ... that's what I did in the other variable migration.

ygliuvt: The other variable migration was weeks earlier, but it appears the only variables in SIT right now are created by us. So it should be safe to truncate. I say we add a truncate and send emails to MMT and EDSC to let them know.

ygliuvt: Still don't like the truncation idea, as it would leave orphans in ealsticsearch. I am leaning towards no truncation now.

oubiwann: Done.

oubiwann: Oh, okay -- I'll back out that change.

oubiwann: Also: good last-minute save, @ygliuvt ;-)

oubiwann:
When I tested with `:db :external` about half of the new assertions failed. @ygliuvt hopped on a GTM with me and we debugged it. He found the issue (which wasn't captured by `:db :in-memory`, since the in-memory "implementation" doesn't have the same set of implementations for multimethods that the external does). The bit we were missing was in `cmr.metadata-db.data.oracle.concepts`, particularly there needed to be a non-`:default` implementation for `by-provider`.

With his fix in place (a copy-and-paste of `by-provider` for `:access-group`), all the tests passed (both in-memory and external).

Running in CI now ...

oubiwann:
Also, CI didn't catch this since it had stopped responding to my branch pushes. Got it to build again by creating a new branch name ...

oubiwann:
It's not easy getting green .... 🐸

PR #226, opened 2017-08-04
CMR-4343: Change provider bootstrapping to use message queue

chris-durbin:
This pull request moves the bootstrapping of providers to a message queue.

The message queue is configured to allow messages to take up to 12 hours before considering them failed and to not ever retry a message. This will effectively allow operations to distribute messages for indexing providers across several nodes. Even if a message takes several days to process (like indexing all the granules for LPDAAC_ECS), the node that is processing that message will not attempt to pick up any other messages (listener count is set to 1). By not retrying we ensure that after 12 hours we don't have another bootstrap node kick off indexing for LPDAAC_ECS. The only downsides are that long messages will eventually move to the dead letter queue and the number of in flight messages on the message queue will not be correct. But since almost no one can get to the GP-MCE AWS console anyway, no one will notice any problem. 😄 

I also refactored the code as requested by Duncan in the prior pull request for this ticket.

### DIFF
--- a/message-queue-lib/src/cmr/message_queue/queue/sqs.clj
+++ b/message-queue-lib/src/cmr/message_queue/queue/sqs.clj
@@ -100,8 +100,6 @@
   "Returns the Topic with the given display name."
   [sns-client exchange-name]
   (info "Calling SNS to get topic " exchange-name)
-  (debug "SNS Client" (pr-str sns-client))
-  (debug "Topics" (pr-str (.listTopics sns-client)))
   (let [exchange-name (normalize-queue-name exchange-name)
         topics (vec (.getTopics (.listTopics sns-client)))]
     (some (fn [topic]
@@ -190,7 +188,7 @@
      (.setQueueAttributes sqs-client set-queue-attrs-request)))
 
 (defn- create-exchange
-  "Creaete an SNS topic to be used as an exchange."
+  "Create an SNS topic to be used as an exchange."
   [sns-client exchange-name]
   (.createTopic sns-client (normalize-queue-name exchange-name)))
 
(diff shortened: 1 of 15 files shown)


message-queue-lib/src/cmr/message_queue/queue/sqs.clj
```diff
@@ -95,6 +100,8 @@
   "Returns the Topic with the given display name."
   [sns-client exchange-name]
   (info "Calling SNS to get topic " exchange-name)
+  (debug "SNS Client" (pr-str sns-client))
```
chris-durbin: Remove debug.

chris-durbin:
I deployed the code to workload and tested it out. I called the endpoint to bootstrap all 39 providers. There was a single bootstrap node running and it picked up 1 message and started bootstrapping ASF. I then scaled bootstrap up and each instance picked up one message and started processing it. In short it is working as expected.

message-queue-lib/src/cmr/message_queue/queue/sqs.clj
```diff
@@ -156,34 +163,36 @@
...
+        set-queue-attrs-request (doto (SetQueueAttributesRequest.)
+                                      (.setAttributes q-attrs)
+                                      (.setQueueUrl queue-url))]
+     (.setQueueAttributes sqs-client set-queue-attrs-request)))
 
 (defn- create-exchange
- "Creaete an SNS topic to be used as an exchange."
- [sns-client exchange-name]
- (.createTopic sns-client (normalize-queue-name exchange-name)))
+  "Creaete an SNS topic to be used as an exchange."
```
ygliuvt: Fix "Create" typo.

chris-durbin: Done.

PR #230, opened 2017-08-10
CMR-4228 sort by end date, bin keywords

laurenfrederick:
Add sorting by end date after sorting by community usage. Nil end date should be above other end dates (ends at present/ongoing). Turn on keyword binning (was an earlier PR, but always turned off).

Still going to test the keyword binning performance in WL before merge.


system-int-test/test/cmr/system_int_test/search/collection_relevancy/collection_temporal_relevancy.clj
```diff
@@ -82,4 +83,17 @@
...
 
       "Date range including collection with early ranges"
-      ["1955-01-01T10:00:00Z,1999-03-01T0:00:00Z"] [coll5 coll6 coll2])))
+      ["1955-01-01T10:00:00Z,1999-03-01T0:00:00Z"] [coll5 coll6 coll2])
+
+   (testing "Keyword search, sort by end date"
+     (is (d/refs-match-order? [coll3 coll4 coll1 coll2 coll5 coll6]
+                              (search/find-refs :collection {:keyword "coll"})))
+
+     (testing "Multiple ongoing collections"
```
mschmele: This `testing` block looks like it's inside the one above. Not sure if you meant to do that or not.

laurenfrederick: Yes, it's under keyword search sorting by end date

search-app/src/cmr/search/data/query_to_elastic.clj
```diff
@@ -47,13 +50,13 @@
   "When sorting by keyword score, set to true if we want to bin the keyword scores
   i.e. round to the nearest keyword-score-bin-size"
   {:type Boolean
-   :default false})
+   :default true})
 
 (defconfig keyword-score-bin-size
   "When sort-bin-keyword-scores is true, what should the keyword score should
   be rounded to the nearest keyword-score-bin-size"
```
sxu123: Could you rephrase this?  It doesn't sound right -  should you remove "what should"

search-app/src/cmr/search/data/query_to_elastic.clj
```diff
@@ -371,7 +374,12 @@
        ;; We only include this if one of the others is present
        (when (and (or use-temporal-sort? use-keyword-sort?)
                   (sort-use-relevancy-score))
-         [{:usage-relevancy-score {:order :desc :missing 0}}])))))
+         [{:usage-relevancy-score {:order :desc :missing 0}}])
+       ;; If end-date is nil, collection is ongoing so use today so ongoing
+       ;; collections will be at the top
+       (when use-keyword-sort?
+         [{:end-date {:order :desc
+                      :missing (time-coerce/to-long (time/today))}}])))))
```
chris-durbin: It looks like temporal for collections can have a time component. Shouldn't we use `(time/now)` instead of `(time/today)`?

system-int-test/test/cmr/system_int_test/search/collection_relevancy/collection_relevancy.clj
```diff
@@ -30,6 +30,7 @@
 (deftest relevancy-temporal-ranges
```
chris-durbin: I think this could impact other tests. We should probably put these 3 config change calls into a fixture or set the values back to the default at the end of the test.

system-int-test/test/cmr/system_int_test/search/collection_relevancy/collection_temporal_relevancy.clj
```diff
@@ -82,4 +83,17 @@
...
+       (let [coll7 (d/ingest "PROV1" (dc/collection {:entry-title "coll 7"
```
chris-durbin: Do we allow collections to have end dates in the future? I couldn't find a test for it in cmr.umm.test.validation.collection-validation-tests.

If we did allow future ending collections, it would put those collections ahead of collections that are ongoing.

search-app/src/cmr/search/data/query_to_elastic.clj
laurenfrederick: Yeah I don't think it matters, but I was trying something for some test fails and forgot to set it back

system-int-test/test/cmr/system_int_test/search/collection_relevancy/collection_temporal_relevancy.clj
laurenfrederick: Yeah I'll add another test. It's "discouraged" to have an end date in the future, it's an ARC thing. I think for now we just sort them and if someone has an end date in the future, it'll show up first. If we really want to combat it, the options are to set :missing to a date way in the future or write a script that sorts nil first, which i think is overkill.

search-app/src/cmr/search/data/query_to_elastic.clj
laurenfrederick: Done.

system-int-test/test/cmr/system_int_test/search/collection_relevancy/collection_relevancy.clj
laurenfrederick: It's set back to true on line 102

search-app/docs/api.md
```diff
@@ -1529,7 +1529,12 @@ When `has_granules` is set to "true" or "false", results will be restricted to c
 
 Collection results are sorted by ascending entry title by default when a search does not result in a score.
 
-If a keyword search is performed then the search results will be sorted by relevance (score descending), then by temporal overlap if one or more temporal ranges are provided with the tie breaker being the EMS community usage score (also descending). The usage score comes from EMS metrics which contain access counts of the collections by short name and version. The metrics are ingested into the CMR.
+If a keyword search is performed then the search results will be sorted by:
+
+  * Relevance Score (descending), binned to the nearest 0.2. For example a score of 0.75 and 0.85 will be considered equal for sorting purposes.
+  * Temporal Overlap (descending), if one or more temporal ranges are provided.
```
chris-durbin: Might be worth defining temporal overlap. I assume that means we take the temporal coverage being searched against and if a collection covers that entire time range, it is 100% coverage. For example if searching for Jan2010-Jan2016, a collection with coverage of Jan2010-2013 would be considered 50%.

laurenfrederick: You're correct... kind of. So if the collection has multiple temporal ranges, it will just take the start and the end now as the collection temporal range, so it's not 100% accurate. We have a ticket next sprint where we can update this now that we are indexing all temporal ranges and possibly bin these values.

PR #233, opened 2017-08-11
CMR-3682: Added date range validation for DataDates and MetadataDates.

sxu123:


umm-spec-lib/src/cmr/umm_spec/validation/data_date.clj
```diff
@@ -0,0 +1,45 @@
...
+        latest-update (date-util/latest-date-of-type value "UPDATE")
+        earliest-review (date-util/earliest-date-of-type value "REVIEW")
+        latest-review (date-util/latest-date-of-type value "REVIEW")
+        earliest-delete (date-util/earliest-date-of-type value "DELETE")]
+    ;;First check the absolute range.
+    (if (or (and latest-create (not (date-util/is-in-past? latest-create)))
+            (and latest-update (not (date-util/is-in-past? latest-update)))
+            (and earliest-review (not (date-util/is-in-future? earliest-review)))
+            (and earliest-delete (not (date-util/is-in-future? earliest-delete))))
+      {field-path [(format
```
laurenfrederick: I think these should be individual messages. None of these are required, so we'd be showing nil in the message if we don't have one or more of these dates.

umm-spec-lib/src/cmr/umm_spec/validation/data_date.clj
```diff
@@ -0,0 +1,45 @@
...
+                          " need to be in the future if not nil.")
+                     latest-create latest-update earliest-review earliest-delete)]}
+      ;;Then check the relative range
+      (when (or (and latest-create 
+                     earliest-update
+                     (time/after? latest-create earliest-update))
+                (and latest-review                                
+                     earliest-delete
+                     (time/after? latest-review earliest-delete)))
+        {field-path [(format
```
laurenfrederick: I think these should be separate warnings as well - create comes before latest and review comes before delete. You don't need to say if both are not nil, because if they get to this point they won't be.

sxu123: I know.  I added if not nil because I'm combining the msg, just to avoid going through all the permutations since none of them is required...

umm-spec-lib/src/cmr/umm_spec/validation/data_date.clj
sxu123: Done.

umm-spec-lib/src/cmr/umm_spec/validation/data_date.clj
```diff
@@ -0,0 +1,50 @@
...
+   field-path is :DataDates/:MetadataDates, 
+   value is the value of :DataDates/:MetadataDates."
+  [field-path value]
+  (let [latest-create (date-util/latest-date-of-type value "CREATE")
+        earliest-update (date-util/earliest-date-of-type value "UPDATE")
+        latest-update (date-util/latest-date-of-type value "UPDATE")
+        earliest-review (date-util/earliest-date-of-type value "REVIEW")
+        latest-review (date-util/latest-date-of-type value "REVIEW")
+        earliest-delete (date-util/earliest-date-of-type value "DELETE")]
+    (def warning-msgs (atom ""))
```
laurenfrederick: I don't think you need to use an atom here. I think what you want to do is return a list of errors, not the errors all as one string. 

So it should be something like `(let [errors (concat (when .... [error message]))` then do a `(when (seq errors))` at the end.

laurenfrederick: Also change "needs to be in the past" to "should be in the past".

sxu123: Done.

system-int-test/src/cmr/system_int_test/data2/umm_spec_collection.clj
```diff
@@ -150,6 +150,49 @@
                                        :Type "CREATE"})]
    :Abstract "A very abstract collection"})
 
+;; Used for testing invalid data date ranges.
```
laurenfrederick: Since this is only used by one test, i'd move it into the file with that test.

umm-spec-lib/src/cmr/umm_spec/validation/data_date.clj
```diff
@@ -0,0 +1,44 @@
+(ns cmr.umm-spec.validation.data-date
+  "Defines validations for UMM collection DataDates and MetadataDates"
+  (:require
+   [clj-time.core :as time]
+   [clojure.string :as string]
+   [cmr.umm-spec.date-util :as date-util]))
+
+(defn data-dates-warning-validation
```
laurenfrederick: We should have some tests with nils, with multiple review dates, etc. You can put them in the validation tests in umm-spec and use the helper functions like `assert-warnings-invalid`.

umm-spec-lib/src/cmr/umm_spec/validation/data_date.clj
```diff
@@ -0,0 +1,44 @@
...
+   field-path is :DataDates/:MetadataDates, 
+   value is the value of :DataDates/:MetadataDates."
+  [field-path value]
+  (let [latest-create (date-util/latest-date-of-type value "CREATE")
+        earliest-update (date-util/earliest-date-of-type value "UPDATE")
+        latest-update (date-util/latest-date-of-type value "UPDATE")
+        earliest-review (date-util/earliest-date-of-type value "REVIEW")
+        latest-review (date-util/latest-date-of-type value "REVIEW")
+        earliest-delete (date-util/earliest-date-of-type value "DELETE")
+        warning-msgs (str
```
laurenfrederick: This should be a vec of warning messages, not one string. I think the result will be the same in the int test, but a vec would keep it consistent with the other validation.

sxu123: It's a vec when putting it in the {field-path [warning-msgs]}

umm-spec-lib/src/cmr/umm_spec/validation/data_date.clj
sxu123: I have tried that. but multiple values will all turned into one value when converted to xml, using the latest one. Also, like different format only support partial of DataDates and MetadataDates.

umm-spec-lib/src/cmr/umm_spec/validation/data_date.clj
laurenfrederick: But it's one item when it should be several individual error messages

system-int-test/src/cmr/system_int_test/data2/umm_spec_collection.clj
sxu123: Done.

umm-spec-lib/src/cmr/umm_spec/validation/data_date.clj
sxu123: Done.

umm-spec-lib/src/cmr/umm_spec/validation/data_date.clj
laurenfrederick: I'm only seeing the one test, which does not cover everything.

laurenfrederick: I am saying you could put additional tests in the umm-spec validation unit tests and cover all the scenarios. Easier to put them in there.

umm-spec-lib/src/cmr/umm_spec/validation/data_date.clj
```diff
@@ -0,0 +1,43 @@
...
+   field-path is :DataDates/:MetadataDates, 
+   value is the value of :DataDates/:MetadataDates."
+  [field-path value]
+  (let [latest-create (date-util/latest-date-of-type value "CREATE")
+        earliest-update (date-util/earliest-date-of-type value "UPDATE")
+        latest-update (date-util/latest-date-of-type value "UPDATE")
+        earliest-review (date-util/earliest-date-of-type value "REVIEW")
+        latest-review (date-util/latest-date-of-type value "REVIEW")
+        earliest-delete (date-util/earliest-date-of-type value "DELETE")
+        warning-msgs (concat
```
mschmele: Nitpick: It'd be nice to pull this `concat` out into its own function above

sxu123: I'm hesitant to do that because concat part is just about all this function is doing...

oubiwann: At a quick glance, it might be good to split the function by the conceptually meaningful (to humans) checks that are being done ... maybe the functions `valdiate-date-order`, `validate-create-and-update`, and `validate-review-and-delete`? Then this function would define the parameters required by those and concat the calls to each of those functions ...

sxu123: I understand your point.  If I could completely separate the validation of CREATE, UPDATE, REVIEW and DELETE, I would do that. But since I still have to validate the relative relationships, the detailed function names could be a little confusing. Given the length of this function, which is very short, I really don't feel it's necessary to break it down further.  In general, I feel that you can almost always break down a function further based on some kind of meaningful conceptions, depending on how you look at it and how far you want to break it...

sxu123: I'm going to merge it for now. If you guys still feel strongly about breaking it up, we can talk about it and think of some good function names... What I don't like is to validate CREATE in both validate-create-and-update and validate0date-order.  To me, it's less clear than giving a good doc string and combine them together.

PR #235, opened 2017-08-11
CMR-4390: Improve performance of has_granules_created_at query

chris-durbin:
Refactored the code to move the has_granules_created_at parameter through the normal query flow in the API routes and query service. Implemented all the special case code as a new query feature. Added a bunch of tests to make sure that the parameter passing from the collection query to the granule query worked as expected. Also made a change to the ACL queries so that if any provider ids are sent in we do not include any collections for other providers in the ACL query (which results in searching against a smaller number of granule indexes).


search-app/src/cmr/search/services/query_walkers/update_field_resolver.clj
```diff
@@ -0,0 +1,95 @@
+(ns cmr.search.services.query-walkers.update-field-resolver
```
laurenfrederick: Would prob be good to have some unit tests for this.

search-app/src/cmr/search/services/parameters/conversion.clj
```diff
@@ -18,91 +18,92 @@
 
 (defmethod common-params/param-mappings :collection
   [_]
-  {:entry-title :string
```
mschmele: I like the idea of alphabetizing large maps like this

search-app/src/cmr/search/services/query_execution/has_granules_created_at_feature.clj
```diff
@@ -0,0 +1,76 @@
...
+                query
+                parameters/collection-only-params)
+        (has-granules-created-at-query (:condition query))))
+
+(defn- generate-collection-query-condition
+  "Takes the original query and the concept IDs from the granules query and constructs a new
+  collection query condition to execute. If there are not any concept-ids we should return a
+  match none condition."
+  [context query concept-ids]
+  (let [concept-ids-condition (when (seq concept-ids)
```
mschmele: Nitpick: pulling this into a function might be a little too granular, but I think generally we want to try to avoid conditionals in `let` blocks

search-app/src/cmr/search/services/query_walkers/update_field_resolver.clj
oubiwann: Yeah, I took a look at common-app-lib services unit tests and the search-app services unit tests; looks like adding these would be fairly straight-forward.

search-app/src/cmr/search/services/parameters/conversion.clj
```diff
@@ -313,10 +325,13 @@
                                     [:facets-v2])
                                 (when (= (:include-highlights params) "true")
                                   [:highlights])
+                                (when (:has-granules-created-at params)
```
ygliuvt: Thanks for refactoring this as a results feature.

search-app/src/cmr/search/services/query_execution/has_granules_created_at_feature.clj
```diff
@@ -0,0 +1,76 @@
...
+                                                                    concept-ids nil))
+        ;; Remove the has-granules-created-at query condition
+        query (update-field-resolver/remove-field query :has-granules-created-at)]
+    (if (seq concept-ids)
+      (group-query-conditions/and-conds [(:condition query) concept-ids-condition])
+      query-model/match-none)))
+
+(defmethod query-execution/pre-process-query-result-feature :has-granules-created-at
+  [context query feature]
+  (let [granule-query (generate-granule-query query)
```
ygliuvt: If i am reading the code correctly, this searching sequence is not very straightforward and could potentially be a lot more expensive with certain search params. I have thought a more straightforward search sequence would be to filter the collections down first based on the search params, then use the filtered down collection identifiers to do the granule facet search.

search-app/src/cmr/search/services/query_walkers/update_field_resolver.clj
chris-durbin: Added the unit tests and refactored the source namespace some.

search-app/src/cmr/search/services/query_execution/has_granules_created_at_feature.clj
chris-durbin: I thought about doing this initially, but forgot about when I saw that a collection query was being executed for my granule query - it turned out that was just for ACLs. Once I found that there were 4 parameters in the granule query that actually require a collection query to be performed first (short-name, entry-title, version, and provider) I had to make this change to make them all work (well provider already worked because of a different change).

Anyway made the change now, thanks.

PR #236, opened 2017-08-14
CMR-4192 fix issue with location keyword search

mschmele:
Some expected collections were not being returned when searching by location keywords in UAT


system-int-test/test/cmr/system_int_test/search/collection_spatial_keyword_search_test.clj
```diff
@@ -31,7 +31,7 @@
            "LA" [coll4]
```
laurenfrederick: Can we add a test in here to test cmr-4192-iso-collection.xml with spatial search? So pretty much exactly what you did in the above file, but instead of keyword search do spatial-keyword search with the scenarios Kathy used when she wrote the ticket. I think the spatial test would really test what is in the ticket.

mschmele: Done

indexer-app/src/cmr/indexer/data/index_set.clj
```diff
@@ -321,7 +321,7 @@
           :archive-center.lowercase       m/string-field-mapping
           :data-center                    (m/stored m/string-field-mapping)
           :data-center.lowercase          m/string-field-mapping
-          :spatial-keyword                m/string-field-mapping
+          :spatial-keyword                (m/doc-values m/string-field-mapping)
```
chris-durbin: You cannot change mappings for a field in an existing Elasticsearch index (maybe we should add a test for this similar to the granule index one).

If these are collection fields it is ok that they aren't doc-values (collections use a tiny amount of memory in Elasticsearch when compared to granules).

umm-spec-lib/src/cmr/umm_spec/location_keywords.clj
```diff
@@ -43,20 +43,15 @@
...
+  (as-> location-keyword keyword
+        (vals keyword)
+        (distinct keyword)
+        (remove nil? keyword)))
 
 (defn location-keywords->spatial-keywords
   "Converts a list of LocationKeyword maps to a list of SpatialKeywords"
   [location-keyword-list]
-  (map #(leaf-value %) location-keyword-list))
+  (flatten (map location-values location-keyword-list)))
```
chris-durbin: This could cause a major behavior change for groups outside the MMT. I'm trying to remember why we only wanted to match the last value in a location keyword.

Give me a little time to double-check what impact this will have.

umm-spec-lib/src/cmr/umm_spec/location_keywords.clj
```diff
@@ -43,20 +43,15 @@
 (defn- location-values
   "Returns the location keyword values in order so that we can get the last one"
   [location-keyword]
-  (for [k location-keyword-order
```
sxu123: Now that it no longer makes use of the location-keyword-order:
1. Should the doc string be modified? 
2. location-keyword-order def should be removed from the file since no where else is using it.

indexer-app/src/cmr/indexer/data/index_set.clj
sxu123: Are these doc-values changes necessary for this ticket?

umm-spec-lib/src/cmr/umm_spec/location_keywords.clj
chris-durbin: I don't see the three acceptance criteria tested:

1. wildcard searches return values regardless of case (add a case searching for aMEricA or something crazy)
2. first term in Location Keyword hierarchy is returned in a search (add a case searching for Continent that returns all the Continent collections)
3. all records which match search criteria are returned in result set (covered by the test above).

chris-durbin: I looked some more - I don't see a problem with indexing all of the fields from the location keyword hierarchy so that a spatial keyword search will return a match of any string in the hierarchy.

By the way in this case I think `mapcat` is the idiomatic way rather than `(flatten (map`.

system-int-test/test/cmr/system_int_test/search/collection_spatial_keyword_search_test.clj
```diff
@@ -31,7 +35,18 @@
...
+           "Tuolumne River Basin" [coll-iso-sample]
+           "North America" [coll-iso-sample]))
+    (testing "search by spatial keywords using wildcard * for cmr-4192 collection"
+      (is (d/refs-match? [coll-iso-sample]
+                         (search/find-refs :collection
+                                           {:spatial-keyword "*river*"
+                                            "options[spatial-keyword][pattern]" "true"})))
+      (is (d/refs-match? [coll-iso-sample]
+                         (search/find-refs :collection
+                                           {:spatial-keyword "*aMEricA*"
```
chris-durbin: I'd drop the wildcard from this search just to make sure that case insensitive aMEricA is matched without setting `options[spatial-keyword][pattern]`.

system-int-test/test/cmr/system_int_test/search/collection_spatial_keyword_search_test.clj
```diff
@@ -31,7 +35,18 @@
            "LA" [coll4]
            ["Detroit"] [coll5 coll7]
            ["LL" "Detroit"] [coll5 coll6 coll7]
-           "BLAH" []))
+           "Gaza Strip" [coll-gaza]
+           "Tuolumne River Basin" [coll-iso-sample]
+           "North America" [coll-iso-sample]))
```
chris-durbin: Add a search for "Continent" finds all collections which have Continent at the top level.

PR #239, opened 2017-08-15
CMR-4397 don't use :temporals for granule search

laurenfrederick:
No longer index to temporals for granules, update temporal search to have conditions for granules. This fixes granule performance issues.


ygliuvt:
I have not followed this topic. Am I mistaken in believing that this change will not consider granule temporal info at all during granule search? How is that acceptable?

ygliuvt:
"We did not change how granule search works. " That is the missing info I had. And based on all the granule temporal search tests are still passing, I must have been mistaken. :) Thanks for clarifying it.

PR #242, opened 2017-08-16
CMR-4365 integrate TopHat2 with CMR static web pages

zjam1:


pom.xml
```diff
@@ -0,0 +1,64 @@
+<?xml version="1.0" encoding="UTF-8"?><project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
```
oubiwann: I don't think we track the `pom.xml` file in git. @chris-durbin or @ygliuvt can confirm that ...

oubiwann:
I guess I should have started with: Thanks! It's super-cool to see your PR here :-D

zjam1:
Dana was looking to integrate TopHat2 with the CMR static web pages. Which were the sites below and the sites that these pages linked.

https://cmr.earthdata.nasa.gov/access-control
https://cmr.earthdata.nasa.gov/search
https://cmr.earthdata.nasa.gov/ingest

I inserted the javascript that integrates Tophat2 into the base html page for access-control, ingest, and search. The other pages extended the base html pages except for the home page which do not oddly enough. So the javascript was also inserted into the home page html.

oubiwann:
Ah, and some googleing just turned up this:
 * https://developer.earthdata.nasa.gov/earthdata-tools/tophat2/integration

Remember -- I'm super-new to CMR/NASA :-)

access-control-app/resources/templates/access-control-base.html
```diff
@@ -2,4 +2,11 @@
 
 {% block app-nav %}
   <li><a href="{{ base-url }}site/docs/access-control">Documentation</a></li>
+  <script src="https://cdn.uat.earthdata.nasa.gov/tophat2/tophat2.js"
```
oubiwann: Now that I better understand what Tophat2 is, I think there's a better way to do this -- you're close, we just need to generalize a bit so there's not so much duplication in the code.

Let's do this:
* Update `template/base.html` with a new, empty block `{% block tophat-js %}{% endblock %}`
* Create a new template file called `base-tophat.html` (in the `resources/templates` directory) which extends `templates/base.html` and put the script tophat setup there in the new block
* Remove the script from all the other templates
* Update the `*-home*` templates to extend `base-tophat.html` instead of `base.html`

This will mean there only being one declaration of the script setup.

If the template stuff is out of your comfort zone, you can add me as a collaborator on your branch, and I can make commits to it ...

zjam1:
Okay thanks for the feedback on the code. This is actually the second time I've ever worked with Javascript so I'm just a little bit of a noobie :),  but your suggestions look doable. I'm currently heading out for today but I can implement the changes tomorrow.

oubiwann:
Sweet -- if you run into any issues, we can pair on screenhero.

pom.xml
ygliuvt: Correct. We should put pom.xml in the .gitignore file.

zjam1:
So I basically followed Duncan's advise on this. Just that the empty {% block tophat-js %}{% endblock %} needed to be in the templates/structure.html instead of the templates/base.html because base.html extends structure.html

Also note that the access-control-home.html  extends landing.html instead of base.html so I made a landing-tophat.html and had the access-control-home.html extend that instead. The same thing happened with the other home htmls.

pom.xml
oubiwann: This still needs to be removed... And could you update `.gitignore` per Liu's suggestion?

oubiwann:
Oh, wait -- before you merge, be sure to see what @ygliuvt says (it looks like he's in the midst of a review).

oubiwann:
@zjam1 Okay, you're good to go now :-)

PR #248, opened 2017-08-18
CMR-4370 Change variable association provider ACL

oubiwann:
We're moving from system-based to provider-based, just like we did for variables themselves. We didn't do this originally, since variable associations aren't directly linked to providers, but the difference of permissions has been difficult for OPS and MMT.

~~Note that no system integration tests existed specifically for the `system-object` ACL for variables associations, so this change didn't cause any tests to fail. Liu and I have chatted about some tests we can add -- I'm working on that now and will push to this branch when something's in place.~~

Tests for variable association ACLs are now in place.

### DIFF
--- a/search-app/src/cmr/search/api/variables.clj
+++ b/search-app/src/cmr/search/api/variables.clj
@@ -4,6 +4,7 @@
    [cheshire.core :as json]
    [cmr.acl.core :as acl]
    [cmr.common-app.api.enabled :as common-enabled]
+   [cmr.common.concepts :as concepts]
    [cmr.common.log :refer (debug info warn error)]
    [cmr.common.mime-types :as mt]
    [cmr.common.services.errors :as errors]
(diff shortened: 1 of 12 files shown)


search-app/src/cmr/search/api/variables.clj
```diff
@@ -30,7 +30,7 @@
   "Verifies the current user has been granted permission to make variable associations"
   [context permission-type]
   (when-not (seq (acl/get-permitting-acls
-                  context :system-object "INGEST_MANAGEMENT_ACL" permission-type))
+                  context :provider-object "INGEST_MANAGEMENT_ACL" permission-type))
```
chris-durbin: Do we have to make this change right away?

I think this will make it so that anyone that has permission to any provider can associate variables from any provider to collections for any provider.

I would like to see an answer to Leo's email about whether we can limit associations to within a single provider. If so I think we should make the change at that point. Validate that the user has ingest update permissions for that single provider.

oubiwann:
PR has been updated with tests now.

oubiwann:
Tests are 🐸; merging ...

PR #251, opened 2017-08-21
Cmr 4158 - Author search

laurenfrederick:


indexer-app/src/cmr/indexer/data/concepts/collection/keyword.clj
```diff
@@ -89,6 +89,10 @@
         related-url-urls (map :URL related-urls)
         related-url-titles (map :Title related-urls)
         related-url-descriptions (map :Description related-urls)
+        ;; Pull author info from both creator and other citation details
+        authors (concat
```
mschmele: Did you want to also thread this to `(map str/trim)`? If so, it's probably worth breaking that into its own function somewhere

indexer-app/src/cmr/indexer/data/concepts/collection.clj
```diff
@@ -277,6 +282,8 @@
             :instrument-sn.lowercase  (map str/lower-case instrument-short-names)
             :sensor-sn sensor-short-names
             :sensor-sn.lowercase  (map str/lower-case sensor-short-names)
+            :authors authors
+            :authors.lowercase (map str/lower-case authors)
```
sxu123: I think we should use util/safe-lowercase in case authors is nil.

indexer-app/src/cmr/indexer/data/index_set.clj
```diff
@@ -327,6 +327,8 @@
           :two-d-coord-name.lowercase     m/string-field-mapping
           :attributes                     attributes-field-mapping
           :downloadable                   (m/stored m/bool-field-mapping)
+          :authors                        m/string-field-mapping
```
ygliuvt: I think we want every new field to be indexed as `doc-values`.

indexer-app/src/cmr/indexer/data/concepts/collection/keyword.clj
laurenfrederick: The other ones don't, so I don't think we need to.

indexer-app/src/cmr/indexer/data/concepts/collection.clj
laurenfrederick: authors can be an empty list, but not nil. The keep functions return an empty list if there is nothing there.

indexer-app/src/cmr/indexer/data/index_set.clj
laurenfrederick: Done.

PR #255, opened 2017-08-25
Cmr 4231

laurenfrederick:
Update temporal overlap relevancy script to calculate based on all temporal ranges in the collection, not just start and end.


indexer-app/src/cmr/indexer/data/index_set.clj
```diff
@@ -293,6 +293,8 @@
           :start-date                     (m/stored m/date-field-mapping)
           :end-date                       (m/stored m/date-field-mapping)
```
sxu123: I'm confused. If we have :temporal-ranges, do we still need :start-date :end-date here?

search-app/resources/temporal_overlap.groovy
```diff
@@ -1,26 +1,35 @@
 /* Elastic groovy script to calculate the temporal overlap of a collection over the input ranges.
-   Collection start and end dates are doc['start-date'] and doc['end-date'].
+
+   Collection temporal ranges are in doc['temporal-ranges'], which is a list of dates with no
+   overlaps. When sorted, they will be in the order: start-date, end-date, start-date, end-date...
+   If the list has an odd number of elements, the collection is ongoing and has no end date.
```
sxu123: I thought we always have a nil for the end-date, so there won't be odd number of elements?

indexer-app/src/cmr/indexer/data/concepts/collection.clj
```diff
@@ -192,13 +192,19 @@
...
-        temporal-extents (->> temporal-extents
+        temporal-extents (->> collection-temporal-extents
                               ;; converts temporal-extents into a list of many
                               ;; {:BeginningDateTime xxx :EndingDateTime xxx}
                               (mapcat spec-time/temporal-ranges)
                               (map #(set/rename-keys % {:BeginningDateTime :start-date
                                                         :EndingDateTime :end-date}))
                               (map #(apply-function-to-all-values-in-map % index-util/date->elastic)))
+        temporal-extents-ranges (->> collection-temporal-extents
+                                     (map #(dissoc % :SingleDateTimes))
```
sxu123: Why do you need to exclude :SingleDateTimes?

search-app/resources/temporal_overlap.groovy
laurenfrederick: Elastic will remove the nils, so when the list comes through the script, there will be no nils. You can play around with Marvel to see.

indexer-app/src/cmr/indexer/data/concepts/collection.clj
laurenfrederick: SingleDateTimes will always have an overlap of 0, so there is no reason to process them. In addition, when indexing a list elastic will remove dups so then it would be a single value, which would mess up the script.

search-app/resources/temporal_overlap.groovy
sxu123: Thanks.

indexer-app/src/cmr/indexer/data/concepts/collection.clj
sxu123: Got it. thanks.

system-int-test/test/cmr/system_int_test/search/collection_relevancy/collection_temporal_relevancy.clj
```diff
@@ -126,3 +126,46 @@
          (index/wait-until-indexed)
          (is (d/refs-match-order? [coll8 coll3 coll7 coll4 coll1 coll2 coll5 coll6]
                                   (search/find-refs :collection {:keyword "coll"}))))))))
+
+(deftest relevancy-multiple-temporal-ranges
+  (dev-sys-util/eval-in-dev-sys `(query-to-elastic/set-sort-use-temporal-relevancy! true))
```
chris-durbin: Is this not the default? It doesn't look like this config param was added with this PR. If it is not the default we should make sure to set it back to false at the end of the test.

laurenfrederick: It is the default. Should I just remove it from both sets of tests?

PR #257, opened 2017-08-29
CMR-4368

ygliuvt:
This added versioning support that are separate for UMM-C, UMM-Var and UMM-S.


dev-system/README.md
```diff
@@ -41,6 +41,10 @@ CMR developer for the appropriate values for each key in `profiles.clj`.
 that allows unrestricted access to shutdown the system, evaluate arbitrary
 code, remove all data, etc.
 
+## Update of umm-cmn-json-schema.json
+
+Since the common umm-cmn-json-schema.json is shared by UMM-C, UMM-Var and UMM-S, and is compiled into clojure model. We cannot really untie it from the different concept schemas. As a result, there can only be one version of umm-cmn-json-schema.json for the latest version of different concept schemas. We should update umm-cmn-json-schema.json for all concept schemas when it is updated.
```
laurenfrederick: Remove Since in the first sentence.

It's a bit confusing. Maybe say something like we need to keep the latest version of the umm-cmn-json-schema.json in sync for all concept types. When the umm-cmn-json-schema.json is updated for one concept type, the corresponding files should be updated for all concept types.

ygliuvt: Like your suggestion. Thanks.

PR #258, opened 2017-08-29
CMR-4372 UMM Var sort params

laurenfrederick:


system-int-test/test/cmr/system_int_test/search/variable/variable_search_test.clj
```diff
@@ -455,3 +455,54 @@
...
+                            {:sort-key sort-key}
+                            {}))))
+
+      "Default sort"
+      nil
+      [variable3 variable4 variable1 variable2]
+
+      "Sort by name"
+      "name"
+      [variable3 variable1 variable4 variable2]
```
ygliuvt: When variable names are the same, the sort order should fall back to the default. Shouldn't variable4 be before variable1? Curious as to how collection sorting behaves in similar situation.

system-int-test/test/cmr/system_int_test/search/variable/variable_search_test.clj
```diff
@@ -455,3 +455,54 @@
...
+      [variable2 variable3 variable4 variable1]
+
+      "Sort by revision-date"
+      "revision_date"
+      [variable1 variable2 variable3 variable4]
+
+      "Sort by long name"
+      "long-name"
+      [variable3 variable4 variable1 variable2]
+
```
ygliuvt: Add at least one test case for multiple specified sort keys.

ygliuvt:
Update the api doc for sorting in variable search.

system-int-test/test/cmr/system_int_test/search/variable/variable_search_test.clj
laurenfrederick: It's on the next line.... ["name" "long_name"]

system-int-test/test/cmr/system_int_test/search/variable/variable_search_test.clj
laurenfrederick: Yeah, you're right. I fixed it by adding variable sub-sort. Collections are a bit more complicated when it comes to sorting.

PR #262, opened 2017-08-30
CMR-3366: Changed Characteristics' DataType from string to enum

sxu123:


umm-spec-lib/src/cmr/umm_spec/migration/characteristics_data_type_migration.clj
```diff
@@ -0,0 +1,73 @@
+(ns cmr.umm-spec.migration.characteristics-data-type-migration
```
ygliuvt: General comment: I feel like `migration` is a wrong word to use here. It is more of a `normalization` than `migration`.

umm-spec-lib/resources/json-schemas/collection/umm/v1.10/umm-cmn-json-schema.json
```diff
@@ -863,10 +863,7 @@
         },
         "DataType": {
           "description": "The datatype of the Characteristic/attribute.",
-          "type": "string",
-          "minLength": 1,
-          "maxLength": 80,
-          "pattern": "[\\w]{1,79}"
+          "$ref": "#/definitions/DataTypeEnum"
```
ygliuvt: We need to keep the latest version of the umm-cmn-json-schema.json in sync for all concept types. When the umm-cmn-json-schema.json is updated for one concept type, the corresponding files should be updated for all concept types.

umm-spec-lib/src/cmr/umm_spec/xml_to_umm_mappings/echo10.clj
```diff
@@ -45,7 +46,9 @@
 (defn parse-characteristics
   "Returns a seq of UMM characteristic records from the element's child Characteristics."
   [el]
-  (map parse-characteristic (select el "Characteristics/Characteristic")))
+  (seq (remove nil?
+         (map char-data-type-migration/migrate-data-type
```
ygliuvt: Combine the two maps into a single one by `comp` the two functions.

umm-spec-lib/resources/json-schemas/collection/umm/v1.10/umm-cmn-json-schema.json
sxu123: That's bad. now we have same code kept in multiple places?

umm-spec-lib/src/cmr/umm_spec/migration/characteristics_data_type_migration.clj
```diff
@@ -0,0 +1,73 @@
...
+   "TIME" "TIME"
+   "DATETIME" "DATETIME"
+   "DATE_STRING" "DATE_STRING"
+   "TIME_STRING" "TIME_STRING"
+   "DATETIME_STRING" "DATETIME_STRING"})
+
+(defn migrate-data-type
+  "Migrate the Charateristics' data type from string to enum."
+  [characteristics]
+  (when-let [data-type (:DataType characteristics)]
```
ygliuvt: Since `DataType` is a required field, the `when` check is unnecessary.

umm-spec-lib/src/cmr/umm_spec/migration/characteristics_data_type_migration.clj
sxu123: It's written for the migration purpose, used by the version_migration.clj. and it just happened that it could be used by other part of the code as well. It is consistent with the collection-progress-migration.

umm-spec-lib/src/cmr/umm_spec/migration/characteristics_data_type_migration.clj
```diff
@@ -0,0 +1,73 @@
...
+    (let [upper-case-data-type (string/upper-case data-type)]
+      (when-not (= umm-spec-util/NOT-APPLICABLE upper-case-data-type)    
+       (assoc characteristics :DataType (get mapping-up upper-case-data-type umm-spec-util/STRING))))))
```
ygliuvt: So we drop any Characteristics that has a DataType of "NOT APPLICABLE"? Just want to make sure.

umm-spec-lib/src/cmr/umm_spec/migration/characteristics_data_type_migration.clj
sxu123: Good that you spotted that. I didn't think it's necessary. but I worked on part of the code, which I don't remember exactly where, it seems that it's possible that individual Characteristic could be nil?

umm-spec-lib/src/cmr/umm_spec/migration/characteristics_data_type_migration.clj
sxu123: Yes.

umm-spec-lib/src/cmr/umm_spec/migration/characteristics_data_type_migration.clj
ygliuvt: When `map` on nil, it does nothing. I don't think we should have a case of `[nil]` as the value of characteristics; If we do, we should fix the code that generated [nil], not here.

umm-spec-lib/src/cmr/umm_spec/migration/characteristics_data_type_migration.clj
ygliuvt: Wouldn't it make more sense to just use `STRING` rather than dropping the entire Characteristics?

umm-spec-lib/resources/json-schemas/collection/umm/v1.10/umm-cmn-json-schema.json
ygliuvt: Yes, the common schema is shared and we should keep it consistent, and I couldn't think of a better way of doing this.

umm-spec-lib/src/cmr/umm_spec/migration/characteristics_data_type_migration.clj
ygliuvt: I would like to think it as migration code happens to use the normalization code rather than the other way around.

umm-spec-lib/src/cmr/umm_spec/migration/characteristics_data_type_migration.clj
sxu123: No Idea. @eereiter confirmed that we remove the whole Characteristic if the DataType is "NOT APPLICABLE".

umm-spec-lib/resources/json-schemas/collection/umm/v1.10/umm-cmn-json-schema.json
sxu123: Done.

umm-spec-lib/src/cmr/umm_spec/migration/characteristics_data_type_migration.clj
sxu123: okay. Done.

umm-spec-lib/src/cmr/umm_spec/migration/characteristics_data_type_migration.clj
sxu123: Done.

umm-spec-lib/src/cmr/umm_spec/xml_to_umm_mappings/echo10.clj
sxu123: Thanks. Done.

PR #265, opened 2017-09-01
CMR 4271

ygliuvt:
This added XML reference response format to variables search and added support of all revisions search to all variable search response formats.

### DIFF
--- a/search-app/src/cmr/search/api/variables.clj
+++ b/search-app/src/cmr/search/api/variables.clj
@@ -56,10 +56,6 @@
   (variable-api-response
    (variable-service/dissociate-variable-to-collections context variable-concept-id body)))
 
-(defn search-for-variables
-  [context params]
-  (variable-api-response (variable-service/search-for-variables context params)))
-
 (def variable-api-routes
   (context "/variables" []
 
--- a/search-app/src/cmr/search/data/query_to_elastic.clj
+++ b/search-app/src/cmr/search/data/query_to_elastic.clj
@@ -354,10 +354,11 @@
                                :cloud-cover :cloud-cover-doc-values})
       default-mappings)))
 
-(def collection-all-revision-sub-sort-fields
-  "This defines the sub sort fields for an all revisions search."
-  [{(q2e/query-field->elastic-field :concept-seq-id :collection) {:order "asc"}}
-   {(q2e/query-field->elastic-field :revision-id :collection) {:order "desc"}}])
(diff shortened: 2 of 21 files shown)


search-app/src/cmr/search/api/variables.clj
```diff
@@ -56,10 +56,6 @@
   (variable-api-response
    (variable-service/dissociate-variable-to-collections context variable-concept-id body)))
 
-(defn search-for-variables
```
oubiwann: I'm guessing this wasn't used anywhere? (I don't see a removal of its reference ...)

search-app/src/cmr/search/data/query_to_elastic.clj
```diff
@@ -421,17 +427,22 @@
...
-(defmethod q2e/concept-type->sub-sort-fields :variable
-  [_]
-  [{(q2e/query-field->elastic-field :name :variable) {:order "asc"}}
-   {(q2e/query-field->elastic-field :provider :variable) {:order "asc"}}])
+(defmethod q2e/query->sort-params :variable
+  [query]
+  (let [{:keys [concept-type sort-keys]} query
+        specified-sort (q2e/sort-keys->elastic-sort concept-type sort-keys)
+        default-sort (q2e/sort-keys->elastic-sort concept-type (q/default-sort-keys concept-type))
+        sub-sort-fields (if (:all-revisions? query)
```
mschmele: Nitpick: I think we should avoid having conditions inside `let` blocks. To me, the block below is a little easier to read and refactor in the future if needed

```
(concat
  (or (q2e/sort-keys->elastic-sort concept-type sort-keys)
      (q2e/sort-keys->elastic-sort concept-type (q/default-sort-keys concept-type))
  (if (:all-revisions? query)
    (all-revision-sub-sort-fields :variable)
    variable-latest-sub-sort-fields)))

search-app/src/cmr/search/api/variables.clj
ygliuvt: Yes, this should have been dropped when I moved the search endpoint to routes.clj.

search-app/src/cmr/search/data/query_to_elastic.clj
ygliuvt: I am fine with `if` within let block. In this case, it is basically setting a intermediate variable and the variable name (`sub-sort-fields` makes the code more readable). The combined way is harder to read in my opinion.

PR #269, opened 2017-09-06
Cmr 4300

daniel-zamora:
This PR implements the search for deleted granules, to support CMR harvesting.

*NOTE: Still need to update api-docs

### DIFF
--- a/search-app/docs/api.md
+++ b/search-app/docs/api.md
@@ -131,6 +131,7 @@ Join the [CMR Client Developer Forum](https://wiki.earthdata.nasa.gov/display/CM
   * [Search for Tiles](#search-for-tiles)
   * [Retrieve Controlled Keywords](#retrieve-controlled-keywords)
   * [Find collections that have been deleted after a given date](#deleted-collections)
+  * [Find granules that have been deleted after a given date](#deleted-granules)
   * [Tagging](#tagging)
     * [Tag Access Control](#tag-access-control)
     * [Creating a Tag](#creating-a-tag)
@@ -2640,7 +2641,7 @@ To support metadata harvesting, a harvesting client can search CMR for collectio
 
 The following search will return the last revision of the collections that are deleted since 01/20/2017.
 
-    curl -i "%CMR-ENDPOINT%/deleted-collections?revision_date=2017-01-20T00:00:00Z&pretty=true"
+    curl -i "%CMR-ENDPOINT%/deleted-granules?revision_date=2017-01-20T00:00:00Z&pretty=true"
 
 __Example Response__
 
@@ -2674,6 +2675,42 @@ CMR-Hits: 3
         </reference>
     </references>
 </results>
+```
+
+### <a name="tagging"></a> Tagging
+
+Tagging allows arbitrary sets of collections to be grouped under a single namespaced value. The sets of collections can be recalled later when searching by tag fields.
+
+Tags have the following fields:
+
+* tag_key (REQUIRED): free text specifying the key of the tag. Tag key cannot contain `/` character. Tag key is case-insensitive, it is always saved in lower case. When it is specified as mixed case, CMR will convert it into lower case. It normally consists of the name of the organization or the project who created the tag followed by a dot and the name of the tag. For example, org.ceos.wgiss.cwic.quality. The maximum length for tag key is 1030 characters.
+* description (OPTIONAL): a free text description of what this tag is and / or how it is used. The maximum length for description is 4000 characters.
+* originator_id (REQUIRED): the Earthdata Login ID of the person who created the tag.
+
+### <a name="deleted-granules"></a> Find granules that have been deleted after a given date
+
+To support metadata harvesting, a harvesting client can search CMR for granules that are deleted after a given date. The only search parameter supported is `revision_date` and its format is slightly different from the `revision_date` parameter in regular granule search in that only one revision date can be provided and it can only be a starting date, not a date range. The only supported result format is json. The revision_date is limited to 1 year in the past.
+
+Additionally, deleted granules search can be filtered by the provider parameter, which takes the provider id, and parent_collection_id which takes the granule's parent collection concept id.
+
+The following search will return the concept-id, parent-collection-id, granule-ur, revision-date, and provider-id of the granules that are deleted since 01/20/2017.
+
+    curl -i "%CMR-ENDPOINT%/deleted-granules?revision_date=2017-01-20T00:00:00Z&pretty=true"
+
+__Example Response__
+
+```
+HTTP/1.1 200 OK
+Date: Thu, 07 Sep 2017 18:52:04 GMT
+Content-Type: ; charset=utf-8
+Access-Control-Expose-Headers: CMR-Hits, CMR-Request-Id
+Access-Control-Allow-Origin: *
+CMR-Hits: 4
+CMR-Request-Id: 03da8f3d-57b3-4ce8-bbc0-29970a4a8b30
+Content-Length: 653
+Server: Jetty(9.2.10.v20150310)
+
+[{"granule-ur":["ur2"],"revision-date":["2017-09-07T18:51:39+0000"],"parent-collection-id":["C1200000009-PROV1"],"concept-id":["G2-PROV1"],"provider-id":["PROV1"]},{"granule-ur":["ur1"],"revision-date":["2017-09-07T18:51:39+0000"],"parent-collection-id":["C1200000009-PROV1"],"concept-id":["G1-PROV1"],"provider-id":["PROV1"]},{"granule-ur":["ur3"],"revision-date":["2017-09-07T18:51:39+0000"],"parent-collection-id":["C1200000010-PROV1"],"concept-id":["G3-PROV1"],"provider-id":["PROV1"]},{"granule-ur":["ur4"],"revision-date":["2017-09-07T18:51:39+0000"],"parent-collection-id":["C1200000011-PROV2"],"concept-id":["G4-PROV2"],"provider-id":["PROV2"]}]
 ```
 
 ### <a name="tagging"></a> Tagging
--- a/search-app/docs/api.md
+++ b/search-app/docs/api.md
@@ -2641,7 +2641,7 @@ To support metadata harvesting, a harvesting client can search CMR for collectio
 
 The following search will return the last revision of the collections that are deleted since 01/20/2017.
 
-    curl -i "%CMR-ENDPOINT%/deleted-granules?revision_date=2017-01-20T00:00:00Z&pretty=true"
+    curl -i "%CMR-ENDPOINT%/deleted-collections?revision_date=2017-01-20T00:00:00Z&pretty=true"
 
 __Example Response__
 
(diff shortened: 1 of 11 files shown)


search-app/src/cmr/search/services/parameters/parameter_validation.clj
```diff
@@ -789,3 +798,62 @@
                         deleted-colls-revision-date-validation])]
     (when (seq errors)
       (errors/throw-service-errors :bad-request errors))))
+
+(defn- deleted-grans-result-format-validation
+  "Validates that the only result format support by deleted granules search is :xml"
```
laurenfrederick: :xml should be :json

search-app/src/cmr/search/services/parameters/parameter_validation.clj
```diff
@@ -789,3 +798,62 @@
...
+        (validate-deleted-grans-revision-date-str (first revision-date)))
+      (validate-deleted-grans-revision-date-str revision-date))))
+
+(defn- deleted-grans-revision-date-range-validation
+  "Validates that deleted time can only have a start date for deleted collections search"
+  [params]
+  (when-let [revision-date (or (:revision-date params)
+                               (:revision_date params))]
+    (when (t/before? (dt-parser/parse-datetime revision-date) (t/minus- (t/now) (t/days 365)))
+      [(format "Revision date must be withing one year of today.")])))
```
laurenfrederick: Change withing to within

search-app/src/cmr/search/services/query_service.clj
```diff
@@ -427,6 +430,45 @@
...
      :result-format result-format}))
 
+(defn get-deleted-granules
+  "Executes elasticsearch searches to find collections that are deleted from a given revision-date.
+   This only finds granules that are truly deleted and not searchable.
+   granules that are deleted, then later ingested again are not included in the result."
+
+  [context params]
+  (pv/validate-deleted-granules-params params)
+  (let [start-time (System/currentTimeMillis)
```
laurenfrederick: Could move some of this into a separate function, like creating the filters.

search-app/src/cmr/search/services/parameters/parameter_validation.clj
```diff
@@ -748,6 +748,15 @@
     :revision_date
     :result-format})
 
+(def ^:private valid-deleted-granules-search-params
+  "Valid parameters for deleted granules search"
+  #{:revision-date
+    :provider
+    :parent-collection-id
+    :revision_date
```
chris-durbin: These should all be kebab-case (no underscores). Processing params should convert all underscores in query-params to dashes.

So removing :parent_collection_id and :revision_date should work without problem.

search-app/src/cmr/search/services/parameters/parameter_validation.clj
```diff
@@ -748,6 +748,15 @@
...
+    :parent_collection_id
+    :result-format})
```
chris-durbin: We only support JSON right now and for other endpoints we support the result-format through a URL extension or header, not a query parameter. I'm wondering why this is in the list?

search-app/src/cmr/search/services/parameters/parameter_validation.clj
```diff
@@ -789,3 +798,62 @@
...
+(defn- validate-deleted-grans-revision-date-str
+  [revision-date]
+  (when (.contains revision-date ",")
+    [(format (str "Invalid format for revision date, only a starting date is allowed "
+                  "for deleted granules search, but was [%s]") revision-date)]))
+
+(defn- deleted-grans-revision-date-validation
+  "Validates that deleted time can only have a start date for deleted collections search"
+  [params]
+  (when-let [revision-date (or (:revision-date params)
```
chris-durbin: Similar comment: params should be processed so that we only need to look for the dash case.

search-app/src/cmr/search/services/parameters/parameter_validation.clj
```diff
@@ -789,3 +798,62 @@
...
+(defn- deleted-grans-revision-required-validation
+  "Validates that deleted time can only have a start date for deleted collections search"
+  [params]
+  (let [revision-date (or (:revision-date params)
+                          (:revision_date params))]
+    (when (not (seq revision-date))
+      [(format "One revision date is required for deleted granules search.")])))
+
+(defn validate-deleted-granules-params
+  "Validates the query parameters passed in with deleted collections search.
```
chris-durbin: granules rather than collections

search-app/src/cmr/search/services/parameters/parameter_validation.clj
```diff
@@ -789,3 +798,62 @@
...
+   Throws exceptions to send to the user if a validation fails."
+  [params]
+  (let [errors (mapcat #(% params)
+                       [unrecognized-deleted-grans-params-validation
```
chris-durbin: Were there any validations from deleted-collections that could be reused? I was just thinking any of the date parsing, single value validations should have been identical to the collection ones.

search-app/src/cmr/search/services/query_service.clj
```diff
@@ -16,6 +16,8 @@
    [clj-time.format :as time-format]
    [clojure.set :as set]
    [clojure.string :as string]
+   [clojurewerkz.elastisch.query :as esq]
```
chris-durbin: With our existing design we do not perform any direct Elasticsearch queries in this namespace.

search-app/src/cmr/search/services/query_service.clj
```diff
@@ -427,6 +430,45 @@
...
+(defn get-deleted-granules
+  "Executes elasticsearch searches to find collections that are deleted from a given revision-date.
+   This only finds granules that are truly deleted and not searchable.
+   granules that are deleted, then later ingested again are not included in the result."
+
+  [context params]
+  (pv/validate-deleted-granules-params params)
+  (let [start-time (System/currentTimeMillis)
+        {:keys [parent-collection-id provider revision-date result-format]} (common-params/sanitize-params params)
+        filters [{:range {:revision-date {:gte revision-date}}}]
```
chris-durbin: You should be able to use existing code to help build up a query model and then call execute query so it follows through most of the same paths in the code. I'm wondering if the deleted-collections endpoint might be a good example of how we perform elasticsearch queries.

system-int-test/test/cmr/system_int_test/ingest/deleted_granules_index_test.clj
```diff
@@ -53,3 +71,110 @@
...
+    (ingest/delete-concept granule4-prov2)
+    (index/wait-until-indexed)
+
+    (testing "Search for all"
+      (are3 [params granules]
+        (is (= (set (map :concept-id granules))
+               (set (find-deleted-granules params))))
+
+        "after tomorrow"
+        {:revision-date (t/plus- (t/now) (t/days 1))}
```
chris-durbin: I'm not sure what others think about this, but I always want to make sure we verify our query parameters on the API support underscores. I think we should be using `{"revision_date" (t/plus- (t/now) (t/days 1))}` or maybe just add one additional test for every query parameter that makes sure underscores are supported.

system-int-test/test/cmr/system_int_test/ingest/deleted_granules_index_test.clj
```diff
@@ -53,3 +71,110 @@
...
+
+(deftest deleted-granule-parameter-validation
+  (are3 [errors params]
+    (let [response (search/get-search-failure-data
+                    (search/find-deleted-granules params))]
+      (is (= (:status response) 400))
+      (is (= errors (:errors response))))
+
+    "Revision date range validation"
+    '("Revision date must be withing one year of today.")
```
chris-durbin: I don't think you need to use literal lists. `["Revision date must be withing one year of today."]` should work or just have a single error string passed into each test and change line 159 to `(is (= [error] (:errors response))))`

search-app/src/cmr/search/services/parameters/parameter_validation.clj
daniel-zamora: It seems to be added automatically to the params at some point, deleted-collections has this also even though it only accepts xml

search-app/src/cmr/search/services/parameters/parameter_validation.clj
daniel-zamora: This works the same as deleted collections does, I can swap the order of sanitize and validation to get it to work your way though.

search-app/src/cmr/search/services/parameters/parameter_validation.clj
daniel-zamora: see other comment, this works the same as deleted collections does.  I was trying to be consistent.

search-app/src/cmr/search/services/query_service.clj
daniel-zamora: need this for esq/match-all in the filter.

search-app/src/cmr/search/services/query_service.clj
daniel-zamora: The existing query building framework requires a concept-type. There is no concept type for this query.

search-app/docs/api.md
```diff
@@ -2640,7 +2641,7 @@ To support metadata harvesting, a harvesting client can search CMR for collectio
 
 The following search will return the last revision of the collections that are deleted since 01/20/2017.
 
-    curl -i "%CMR-ENDPOINT%/deleted-collections?revision_date=2017-01-20T00:00:00Z&pretty=true"
+    curl -i "%CMR-ENDPOINT%/deleted-granules?revision_date=2017-01-20T00:00:00Z&pretty=true"
```
chris-durbin: Undo this change.

search-app/docs/api.md
```diff
@@ -2686,6 +2687,42 @@ Tags have the following fields:
...
+Access-Control-Allow-Origin: *
+CMR-Hits: 4
+CMR-Request-Id: 03da8f3d-57b3-4ce8-bbc0-29970a4a8b30
+Content-Length: 653
+Server: Jetty(9.2.10.v20150310)
+
+[{"granule-ur":["ur2"],"revision-date":["2017-09-07T18:51:39+0000"],"parent-collection-id":["C1200000009-PROV1"],"concept-id":["G2-PROV1"],"provider-id":["PROV1"]},{"granule-ur":["ur1"],"revision-date":["2017-09-07T18:51:39+0000"],"parent-collection-id":["C1200000009-PROV1"],"concept-id":["G1-PROV1"],"provider-id":["PROV1"]},{"granule-ur":["ur3"],"revision-date":["2017-09-07T18:51:39+0000"],"parent-collection-id":["C1200000010-PROV1"],"concept-id":["G3-PROV1"],"provider-id":["PROV1"]},{"granule-ur":["ur4"],"revision-date":["2017-09-07T18:51:39+0000"],"parent-collection-id":["C1200000011-PROV2"],"concept-id":["G4-PROV2"],"provider-id":["PROV2"]}]
+```
+
+### <a name="tagging"></a> Tagging
```
chris-durbin: Do you know why this tagging section is showing up as new with your PR?

PR #274, opened 2017-09-11
CMR-3365 update var and service schemas, add case insensitivity

mschmele:
This adds the TilingIdentificationSystem and VerticalSpatialDomain Enums to the UMM Var and Service common schema, as well as validate incoming data insensitive to case

### DIFF
--- a/common-lib/src/cmr/common/util.clj
+++ b/common-lib/src/cmr/common/util.clj
@@ -120,6 +120,16 @@
   [v]
   (when v (str/upper-case v)))
 
+(defn match-enum-case
+  "Given a string and a collection of valid enum values, return the proper-cased
+   value from the enum. The values will not differ, but this ensures that the
+   one returned is the proper case, even if that is a crazy mix"
+  [value enum-values]
+  (->> enum-values
+       (filter #(re-matches (re-pattern (str "(?i)" value)) %))
+       seq
+       first))
+
 (defn sequence->fn
   [vals]
   "Creates a stateful function that returns individual values from the
--- a/umm-spec-lib/src/cmr/umm_spec/spatial_conversion.clj
+++ b/umm-spec-lib/src/cmr/umm_spec/spatial_conversion.clj
@@ -36,19 +36,23 @@
    in the list of valid ones"
   [tile-identification-system-name]
   (when (tile-id-system-name-is-valid? tile-identification-system-name)
-    tile-identification-system-name))
+    (util/match-enum-case tile-identification-system-name valid-tile-identification-system-names)))
 
 (defn expected-tiling-id-systems-name
   "Translate TilingIdentificationSystemNames in accordance with UMM Spec v1.10"
   [tiling-identification-systems]
-  (let [tiling-id-systems (mapv
-                           (fn [tiling-id-system]
-                             (update tiling-id-system
-                                     :TilingIdentificationSystemName
-                                     translate-tile-id-system-name))
-                           tiling-identification-systems)]
-    (when-not (empty? tiling-id-systems)
-      tiling-id-systems)))
+  (->> tiling-identification-systems
+       (mapv
+        #(update % :TilingIdentificationSystemName translate-tile-id-system-name))
+       seq))
+
+(defn filter-and-translate-tiling-id-systems
+  "Drop invalid TilingIdentificationSystems, and ensure that valid ones are cased
+   properly."
+  [tiling-identification-systems]
+  (->> tiling-identification-systems
+       (filter #(tile-id-system-name-is-valid? (:TilingIdentificationSystemName %)))
+       expected-tiling-id-systems-name))
 
 (def valid-vertical-spatial-domain-types
   "Valid values for VerticalSpatialDomainType according to UMM spec v1.10.0"
@@ -74,9 +78,12 @@
    is not considered valid according to UMM spec v1.10.0 and should be dropped.
    This behavior will eventually generate errors"
   [vertical-spatial-domains]
-  (seq (filter
-         #(vertical-spatial-domain-type-is-valid? (:Type %))
-         vertical-spatial-domains)))
+  (->> vertical-spatial-domains
+       (filter #(vertical-spatial-domain-type-is-valid? (:Type %)))
+       (map (fn [domain]
+              (update domain :Type #(util/match-enum-case
+                                     % valid-vertical-spatial-domain-types))))
+       seq))
 
 (defn convert-vertical-spatial-domains-from-xml
   "Given a name to key into an xml file, extract Vertical Spatial Domains
--- a/umm-spec-lib/src/cmr/umm_spec/spatial_conversion.clj
+++ b/umm-spec-lib/src/cmr/umm_spec/spatial_conversion.clj
@@ -2,6 +2,7 @@
   "Defines functions that convert umm spec spatial types to spatial lib spatial shapes."
   (:require
    [clojure.string :as string]
+   [cmr.common.util :as util]
    [cmr.common.xml.parse :as xml-parse]
    [cmr.spatial.line-string :as ls]
    [cmr.spatial.mbr :as mbr]
@@ -20,11 +21,15 @@
    "WRS-1"
    "WRS-2"])
 
+(def upcase-valid-tile-identification-system-names
+  "Upcase values to allow for case-insenitive validation"
+  (map util/safe-uppercase valid-tile-identification-system-names))
+
 (defn tile-id-system-name-is-valid?
   "Return whether or not the given TileIdentificationSystemName is one of the
    valid-tile-identification-system-names"
   [tile-system-id-name]
-  (some #(= tile-system-id-name %) valid-tile-identification-system-names))
+  (some #(= (util/safe-uppercase tile-system-id-name) %) upcase-valid-tile-identification-system-names))
 
 (defn translate-tile-id-system-name
   "Return nil or equivalent value if the given name does not match any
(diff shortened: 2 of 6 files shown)


umm-spec-lib/src/cmr/umm_spec/spatial_conversion.clj
```diff
@@ -20,11 +21,15 @@
    "WRS-1"
    "WRS-2"])
 
+(def upcase-valid-tile-identification-system-names
+  "Upcase values to allow for case-insenitive validation"
+  (map util/safe-uppercase valid-tile-identification-system-names))
+
```
sxu123: My understanding about the case insensitive part is the following:
1. the valid-title-identification-system-names stay the same - It should still contain the lower/upper case mixed.
   This is the what the enum list allows. 
2. when converting from other format to UMM v1.10, and converting from UMM v1.9 to v1.10, 
     take the value "MODIS Tile EASE", as example,  you consider other "modis tILE ease" as valid values, which is 
     correct, but you still need to translate "modis tiLE ease" to "MODIS Tile EASE", which you didn't do. So, you need to 
     change the translate-tile-id-system-name function. Same with the vertical spatial domain.
3. Change the add unit tests to reflect these changes.

umm-spec-lib/test/cmr/umm_spec/test/migration/version_migration.clj
```diff
@@ -1563,7 +1563,7 @@
 
 (deftest migrate-1-9-tiling-identification-systems-to-1-10
   (let [tiling-id-systems {:TilingIdentificationSystems
-                           [{:TilingIdentificationSystemName "MISR"
+                           [{:TilingIdentificationSystemName "misr"
```
sxu123: After the translate, it should be a valid "MISR". "misr" is not allowed in the enum list.

sxu123: Add more tests for the tiling-id-systems, and Vertical spatial domain.

common-lib/src/cmr/common/util.clj
```diff
@@ -120,6 +120,22 @@
...
+(defn capitalize-words
+  "Capitalize every word in a string"
+  [s]
+  (->> (str/split (str s) #"\b")
+       (map str/capitalize)
+       str/join))
+
+(defn match-enum-case
+  "Given a string and a collection of valid enum values, return the proper-cased
+   value from the enum."
```
sxu123: I like this function. Please update the doc string to indicate that it returns the case insensitive mapping of the value in the enum list.

umm-spec-lib/src/cmr/umm_spec/spatial_conversion.clj
```diff
@@ -20,30 +21,37 @@
...
    valid-tile-identification-system-names"
   [tile-system-id-name]
-  (some #(= tile-system-id-name %) valid-tile-identification-system-names))
+  (some #(= (util/safe-uppercase tile-system-id-name) %) upcase-valid-tile-identification-system-names))
 
 (defn translate-tile-id-system-name
   "Return nil or equivalent value if the given name does not match any
    in the list of valid ones"
   [tile-identification-system-name]
   (when (tile-id-system-name-is-valid? tile-identification-system-name)
```
sxu123: After you add the match-enum-case, there is no need to check if tile-id-system-name-is-valid? anymore. If the value doesn't have a case insensitive match in the enum list, it will return nil.

common-lib/src/cmr/common/util.clj
```diff
@@ -120,6 +120,22 @@
   [v]
   (when v (str/upper-case v)))
 
+(defn capitalize-words
+  "Capitalize every word in a string"
+  [s]
+  (->> (str/split (str s) #"\b")
+       (map str/capitalize)
+       str/join))
```
sxu123: I don't like adding this function. It's too specific and can only handle the case when the enum list contains the values with the first letter capitalized. You should be able to use the generic match-enum-case to handle it.

umm-spec-lib/src/cmr/umm_spec/spatial_conversion.clj
```diff
@@ -20,30 +21,37 @@
...
+(def upcase-valid-tile-identification-system-names
+  "Upcase values to allow for case-insenitive validation"
+  (map util/safe-uppercase valid-tile-identification-system-names))
+
 (defn tile-id-system-name-is-valid?
   "Return whether or not the given TileIdentificationSystemName is one of the
    valid-tile-identification-system-names"
   [tile-system-id-name]
-  (some #(= tile-system-id-name %) valid-tile-identification-system-names))
+  (some #(= (util/safe-uppercase tile-system-id-name) %) upcase-valid-tile-identification-system-names))
```
sxu123: This function is no longer needed after the generic match-enum-case.

umm-spec-lib/src/cmr/umm_spec/spatial_conversion.clj
```diff
@@ -20,30 +21,37 @@
...
+       (mapv #(update % :TilingIdentificationSystemName translate-tile-id-system-name))
+       seq))
+
+(defn filter-and-translate-tiling-id-systems
+  "Drop invalid TilingIdentificationSystems, and ensure that valid ones are cased
+   properly."
+  [tiling-identification-systems]
+  (->> tiling-identification-systems
+       (filter #(tile-id-system-name-is-valid? (:TilingIdentificationSystemName %)))
+       expected-tiling-id-systems-name))
```
sxu123: You could simplify the code by combining the filter-and-translate-tiling-id-systems, expected-tiling-id-system-name and translate-title-id-system-name because now that you have the generic match-enum-case function, you don't need to check if the tiling-id is valid anymore.

umm-spec-lib/src/cmr/umm_spec/spatial_conversion.clj
```diff
@@ -53,20 +61,26 @@
...
+  "Upcase values to allow for case-insenitive validation"
+  (map util/safe-uppercase valid-vertical-spatial-domain-types))
+
 (defn vertical-spatial-domain-type-is-valid?
   "Return true or false based on whether or not the given value matches
    one of the valid values in valid-vertical-spatial-domain-types."
   [vs-domain-type]
-  (some #(= vs-domain-type %) valid-vertical-spatial-domain-types))
+  (some #(= (util/safe-uppercase vs-domain-type) %)
+        upcase-valid-vertical-spatial-domain-types))
```
sxu123: This function is no longer needed.

umm-spec-lib/src/cmr/umm_spec/spatial_conversion.clj
```diff
@@ -53,20 +61,26 @@
...
    is not considered valid according to UMM spec v1.10.0 and should be dropped.
    This behavior will eventually generate errors"
   [vertical-spatial-domains]
-  (seq (filter
-         #(vertical-spatial-domain-type-is-valid? (:Type %))
-         vertical-spatial-domains)))
+  (->> vertical-spatial-domains
+       (filter #(vertical-spatial-domain-type-is-valid? (:Type %)))
+       (map #(update % :Type util/capitalize-words))
+       seq))
```
sxu123: Please combining the functions the same way as the tiling id case I mentioned above.  Use the more generic match-enum-case, rather than capitalize-words.

common-lib/src/cmr/common/util.clj
mschmele: This function takes a value, and if it's valid, returns the corresponding one in the given enum to guarantee that its casing is correct. I believe the name is correct, but the docstring could be expanded on. Done

common-lib/src/cmr/common/util.clj
mschmele: I disagree about specificity being a bad thing here, but I do agree that `match-enum-case` should be used in case the enum this is being used for is updated to include values that don't include this case

umm-spec-lib/src/cmr/umm_spec/spatial_conversion.clj
mschmele: Fixed

umm-spec-lib/src/cmr/umm_spec/spatial_conversion.clj
mschmele: These functions are used in different places, so deleting them entirely would be a more significant effort. I'd prefer to avoid this if I can, but if you're adamant I'll refactor as needed

umm-spec-lib/src/cmr/umm_spec/spatial_conversion.clj
mschmele: Done

umm-spec-lib/src/cmr/umm_spec/spatial_conversion.clj
mschmele: This function is used in other places, so I'd rather not delete it

umm-spec-lib/test/cmr/umm_spec/test/migration/version_migration.clj
mschmele: I added several mixed case tests

umm-spec-lib/src/cmr/umm_spec/spatial_conversion.clj
sxu123: I'm just asking you to drop the tile-id-system-name-is-valid check.  You already remove the function based on your response to my previous comment.

umm-spec-lib/src/cmr/umm_spec/spatial_conversion.clj
sxu123: It's really not a lot of changes. I checked the baseline code. There are no other places that reference these functions other than one or two places in this PR.  After you drop tile-id-system-name-is-valid? function, you will see there is no need for both filter-and-translate-tiling-id-systems and expected-tiling-id-systems-name - only one of them is needed because they are the same. Regarding translate-title-id-system-name, after tile-id-system-name-is-valid? is removed, there is really nothing more than just use the match-enum-case directly.  so if you want to keep translate-title-id-system-name it's fine. Same with the Vertical Spatial Domain case.

PR #275, opened 2017-09-11
CMR-4430

ygliuvt:
This changed how variable associations are indexed as a result of variable deletion. There is an indexing delay of 5 minutes that only occurs in SIT that this fix fixed. This has been tested in SIT.

### DIFF
--- a/metadata-db-app/src/cmr/metadata_db/services/concept_service.clj
+++ b/metadata-db-app/src/cmr/metadata_db/services/concept_service.clj
@@ -426,8 +426,8 @@
   (fn [context concept]
     (boolean (:deleted concept))))
 
-(defn- delete-associations
-  "Delete the associations that matches the given search params,
+(defn- tombstone-associations
+  "Tombstone the associations that matches the given search params,
   skip-publication? flag controls if association deletion event should be generated,
   skip-publication? true means no association deletion event should be generated."
   [context assoc-type search-params skip-publication?]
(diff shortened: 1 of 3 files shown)


metadata-db-app/src/cmr/metadata_db/services/concept_service.clj
```diff
@@ -427,6 +427,19 @@
     (boolean (:deleted concept))))
 
 (defn- delete-associations
+  "Delete the associations that matches the given search params,
```
chris-durbin: Since the function is creating tombstones - maybe rename to tombstone-associations?

ygliuvt: > I'm assuming that there were existing tests to verify that associations were being cleaned up appropriately. It's not clear to me why this addressed a 5 minute delay, can you add a sentence on what fixed that?

This is explained in the JIRA issue description. I am copying it here:

Delete variable should cause associated variables within collection search result to be reflected right away. Right now, the variable association delete event is sent to elasticsearch and that causes unnecessary indexing operations be performed and in turn causes a metadata db retrieval error of the variable and the indexing will be retried in 5 mins and succeed. As a result, we have to wait 5 minutes for the search result to be correct.

The fix is to actually take the event generation outside of the variable tombstone transaction. This way we can avoid the metadata db retrieval error and the indexing will succeed right away.

ygliuvt:
> I'm assuming that there were existing tests to verify that associations were being cleaned up appropriately. It's not clear to me why this addressed a 5 minute delay, can you add a sentence on what fixed that?

This is explained in the JIRA issue description. I am copying it here:

Delete variable should cause associated variables within collection search result to be reflected right away. Right now, the variable association delete event is sent to elasticsearch and that causes unnecessary indexing operations be performed and in turn causes a metadata db retrieval error of the variable and the indexing will be retried in 5 mins and succeed. As a result, we have to wait 5 minutes for the search result to be correct.

The fix is to actually take the event generation outside of the variable tombstone transaction. This way we can avoid the metadata db retrieval error and the indexing will succeed right away.

PR #276, opened 2017-09-14
CMR-4449

ygliuvt:
Force delete variable revision should cause variables all revision index be updated.

### DIFF
--- a/indexer-app/src/cmr/indexer/config.clj
+++ b/indexer-app/src/cmr/indexer/config.clj
@@ -53,10 +53,10 @@
   "The ingest exchange to which provider change messages are published."
   {:default "cmr_ingest_provider.exchange"})
 
-(defconfig deleted-collection-revision-exchange-name
-  "An exchange that will have messages passed to it whenever a collection revision is removed
+(defconfig deleted-concept-revision-exchange-name
+  "An exchange that will have messages passed to it whenever a concept revision is removed
   from metadata db."
-  {:default "cmr_deleted_collection_revision.exchange"})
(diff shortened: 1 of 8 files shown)


indexer-app/src/cmr/indexer/config.clj
```diff
@@ -53,10 +53,10 @@
   "The ingest exchange to which provider change messages are published."
   {:default "cmr_ingest_provider.exchange"})
 
-(defconfig deleted-collection-revision-exchange-name
-  "An exchange that will have messages passed to it whenever a collection revision is removed
+(defconfig deleted-concept-revision-exchange-name
+  "An exchange that will have messages passed to it whenever a concept revision is removed
   from metadata db."
-  {:default "cmr_deleted_collection_revision.exchange"})
+  {:default "cmr_deleted_concept_revision.exchange"})
```
ygliuvt: @chris-durbin, do we need to update any cloud configurations due to this naming change?

chris-durbin: By doing this we will be affecting all environments. We could miss messages as a result of changing the name of the exchange (multiple apps and multiple nodes will cause a mismatch in what queues are bound to which exchanges and to which exchange messages get published).

I know the name is no longer correct, but I think it would be better to just leave the name the same and put a comment in the docstring.

ygliuvt: Yeah, that's what I was afraid of. I am just going to leave this name alone then. Thanks.

PR #279, opened 2017-09-15
CMR-4502: Add relevancy tests from EDSC relevancy metrics

chris-durbin:
The EDSC logs relevancy metrics anytime a user downloads data for a collection. The log includes the selected_index (how far from the top of the results was the collection), the list of collections that were above the selected index, and the query that was performed. From that information we can generate anomaly tests.

From the past 7 days of operational logs I was able to generate 796 test cases. Most of this PR is the collection metadata which I did separately in the first commit so that it can be reviewed easier by looking at the other commits.

There were a few changes to support downloading many collections and parsing the EDSC logs to generate the tests. I put the EDSC relevancy tests in a separate file from the ones that the providers have specifically called out.


ygliuvt:
This is too much testing data we are adding to the code base. I would prefer to have these data separated from our code base because:

1/ these data are fairly irrelevant to the project other than relevancy testing and we are making the repo unnecessarily big.
2/ not all collections are public. Are we putting us at risk of exposing non-public collections by doing this?

chris-durbin:
The reason the test files are in the repo is because they need to be controlled so that when we run the tests we are testing against the same metadata files. If we downloaded them on demand different people would get different test results.

We want to be sure that changes to relevancy are based entirely on relevancy algorithm changes, and are not being impacted by changes to the metadata.

The test files are 11MB so I'm not concerned about affecting the size of the repo. Also these are test files, anyone doing science research is going to be using the collection metadata from the CMR system, not these metadata files.

We may eventually move out the search-relevancy-tests to a separate repo, it was not clear at the time we started whether we could create is as a new repo on Github.

oubiwann:
What about creating a data git repo? This is a fairly common practice in projects where a subset of users require extensive data sources. We could even add a git submodule for the data repo so that it would only be downloaded on-demand (with an explicit git command).

chris-durbin:
The plan is to move the entire search-relevancy-test directory to its own git repo.

oubiwann:
Nice. +1.

oubiwann:
Okay, since the plan is to eventually move this, for expeidiency's sake, I'm okay with a merge of this PR. Once the move happens, we'll be removing this from the git repo history. Essentially that means the data files in this PR will have zero user impact, long-term.

search-relevancy-test/src/search_relevancy_test/anomaly_fetcher.clj
```diff
@@ -72,22 +74,24 @@
...
+             (spit full-path collection-metadata))))))))
 
 (comment
- (download-and-save-all-collections)
- (find-collection-ids-to-download)
+ (download-and-save-all-collections core/edsc-anomaly-filename)
+ (find-collection-ids-to-download core/edsc-anomaly-filename)
  (formats-for-collections ["C1200196931-SCIOPS" "C1000000803-DEV08"]) ;; SIT
  (formats-for-collections ["C1397496671-NSIDC_ECS"]) ;; Prod
  (download-concept-metadata "C1237114193-GES_DISC"))
```
eereiter: Is this section here to run these tests manually?  Or is this just testing code that needs to be removed? Are we going to set up a periodic run?

search-relevancy-test/src/search_relevancy_test/anomaly_fetcher.clj
```diff
@@ -35,8 +35,10 @@
   "Returns a map of concept-id to the metadata format for that collection."
   [concept-ids]
   (let [url (str (source-url) "/collections.umm-json")
-        response (client/get url {:query-params {:concept-id concept-ids
+        response (client/post url {:form-params {:concept-id concept-ids
                                                  :page-size 2000}})
+                                  ;; Uncomment when collections are only visible for a certain token
+                                  ;  :headers {:echo-token (System/getenv "CMR_SYSTEM_TOKEN")}})
```
eereiter: Can we just always pass the token, that way you wouldn't have to remember to uncomment it out?

search-relevancy-test/src/search_relevancy_test/anomaly_fetcher.clj
```diff
@@ -48,6 +50,8 @@
   [concept-id fmt]
   (let [url (format "%s/concepts/%s" (source-url) concept-id)
         params {:query-params {:concept-id concept-id}}
+                ;; Uncomment when collections are only visible for a certain token
+                ; :headers {:echo-token (System/getenv "CMR_SYSTEM_TOKEN")}}
```
eereiter: Can we just always pass the token, that way you wouldn't have to remember to uncomment it out?

search-relevancy-test/src/search_relevancy_test/edsc_log_parser.clj
```diff
@@ -0,0 +1,123 @@
...
+
+(defn get-relevancy-json-from-line
+  "Strips out everything except for the relevancy JSON from an EDSC metrics log line. Return nil
+  if there are any parsing problems"
+  [line]
+  (try
+    (-> (re-find #"\{.*\}" line)
+        (json/parse-string true)
+        :data)
+    (catch Exception e
```
eereiter: I know this is easy to return nil. Don't you at least want to log exceptions so that we can look to find out what the parsing problem was - maybe fix the EDSC metrics log?  Just a suggestion because if you have 1 error you are probably going to have many.

search-relevancy-test/src/search_relevancy_test/edsc_log_parser.clj
```diff
@@ -0,0 +1,123 @@
...
+  "Parses the provided EDSC relevancy log file and returns search relevancy tests."
+  [log-file]
+  (let [all-tests (remove-duplicate-tests (get-all-tests log-file))]
+    (write-edsc-anomaly-csv all-tests (str "resources/" anomaly-filename))
+    all-tests))
+
+(comment
+ (count (get-all-tests (io/resource "EDSC_Relevancy_metrics.log")))
+ (count (remove-duplicate-tests (get-all-tests (io/resource "EDSC_Relevancy_metrics.log"))))
+ (parse-edsc-logs (io/resource "one-ignored-request.log"))
```
eereiter: Is this section debug code or is it to run the tests manually?  If it is debug code, shouldn't we get rid of it?  You could create test functions to do the manual runs.

search-relevancy-test/src/search_relevancy_test/relevancy_test.clj
```diff
@@ -90,9 +100,21 @@
...
+  search via search-params"
+  ([args]
+   (edsc-relevancy-test args nil))
+  ([args search-params]
+   (test-setup :edsc)
+   (run-anomaly-tests edsc-log-parser/anomaly-filename args search-params)
    nil)) ; Return nil so the return value is not printed to the REPL
 
+
 (comment
```
eereiter: Is this debug code?  Should it be removed?

search-relevancy-test/src/search_relevancy_test/anomaly_fetcher.clj
oubiwann: If we do keep this commented out, it should be two semi-colons (per the style guide; end-of-line comments are the only ones in the Lisp world that get one semi-colon).

search-relevancy-test/src/search_relevancy_test/anomaly_fetcher.clj
chris-durbin: Fixed to use a token if configuration is set and not pass in the header otherwise.

search-relevancy-test/src/search_relevancy_test/anomaly_fetcher.clj
chris-durbin: This is in a comment block for manual use.

search-relevancy-test/src/search_relevancy_test/anomaly_fetcher.clj
chris-durbin: Fixed to use a token if configuration is set and not pass in the header otherwise.

search-relevancy-test/src/search_relevancy_test/edsc_log_parser.clj
chris-durbin: Removed debug.

search-relevancy-test/src/search_relevancy_test/edsc_log_parser.clj
chris-durbin: I logged the exception at warning level.

PR #281, opened 2017-09-18
Fixing CMR-4515 intermittent test failure and fixing the DIF 9

eereiter:
SpatialCoverageType change where we should determine the SpatailCoverageType based on Horizontal, Vertical, or both types of mappings.


eereiter:
DIF 9 doesn't have a SpatialCoverageType element - your assumption is correct, there was nothing that needed to change.

PR #285, opened 2017-09-20
Cmr 4466 - Adding in runtime change logging for the CMR.

eereiter:
I have the code working and I tested it using curl scripts outside of the CMR system.  I wanted to get developers to look it it.  I am in the process of adding system integration tests, but I don't have them done yet. It is taking me a while to understand it.


common-app-lib/src/cmr/common_app/api/log.clj
```diff
@@ -0,0 +1,133 @@
...
+          (log/merge-logging-configuration edn-map-config))))))
+
+(def log-api-routes
+  "Routes for changing the logging configuration endpoints"
+  (context "/log" []
+
+    ;; update the logging configuration
+    (PUT "/"
+         {:keys [request-context headers body]}
+         (println "request-context:" request-context)
```
oubiwann: I'd either get rid of these or change them to `(debug ...)` log calls.

common-app-lib/src/cmr/common_app/api/log.clj
```diff
@@ -0,0 +1,133 @@
+(ns cmr.common-app.api.log
```
oubiwann: Very nicely done ns/file. 👍

common-lib/src/cmr/common/log.clj
```diff
@@ -32,6 +34,30 @@
...
             "")))
 
+;; The next two functions are needed to allow run time logging. The base
+;; knowledge for these functions came from here:
+;; https://github.com/ptaoussanis/timbre/issues/208
+;; https://github.com/yonatane/timbre-ns-pattern-level
+;; I modified the above solutions to be able to set a log level by namespace
+;; during run time.
+
+(defn ns-filter [fltr] (-> fltr enc/compile-ns-filter enc/memoize_))
```
oubiwann: Needs a docstring.

common-lib/src/cmr/common/log.clj
```diff
@@ -32,6 +34,30 @@
...
+
+(defn log-by-ns-pattern
+  [& [{:keys [?ns-str config level] :as opts}]]
```
oubiwann: Docstring.

common-lib/src/cmr/common/log.clj
```diff
@@ -32,6 +34,30 @@
...
+  (let [ns-pattern-map (get config :ns-pattern-map)
+        namesp   (or (some->> ns-pattern-map
```
oubiwann: That's a bit much, logic-wise, for a `let` block -- can you split this out into a private function (with docstring 😉)

common-lib/src/cmr/common/log.clj
```diff
@@ -49,7 +75,30 @@
...
+  []
+  t/*config*)
+
+(defn merge-logging-configuration
+  "This function merges the passed in configuration with the current logging configuration.
+  The passed in configuration map includes a map of name spaces with logging levels and at the end
+  use :all and its level to signify what all the other non identified name spaces' logging level
+  should be.
+  An example of a passed in configuration is as follows
+  { :level :info
```
oubiwann: looks like there's a space here between the `{` and the `:level` ...?

common-app-lib/src/cmr/common_app/api/log.clj
```diff
@@ -0,0 +1,133 @@
...
+   [cheshire.core :as json]
+   [clojure.edn :as edn]
+   [clojure.pprint :as pp]
+   [clojure.string :as string]
+   [cmr.common.log :as log]
+   [cmr.common-app.api.routes :as cr]
+   [cmr.common.mime-types :as mt]
+   [compojure.core :refer :all]))
+
+(defn get-log-configuration-response
```
oubiwann: This and other non-route functions should go in `cmr.common-app.services.logging`. Just the route definition should go in `cmr.common-app.api.log` (I'd opt to name it with a noun, e.g., `cmr.common-app.api.logging`; namespaces are nouns, functions get the verbs.) Then the API ns would make calls to the logging service.

common-app-lib/src/cmr/common_app/api/log.clj
```diff
@@ -0,0 +1,133 @@
...
+      (let [edn-map-config (edn/read-string body-string)]
+        (log/warn (format "Merging a new CMR logging configuration map %s. The old configuration is %s."
+                          edn-map-config
+                          (with-out-str (pp/pprint (log/get-logging-configuration)))))
+        (get-log-configuration-response
+          (.contains (mt/get-header headers "accept") "text/html")
+          200
+          (log/merge-logging-configuration edn-map-config))))))
+
+(def log-api-routes
```
oubiwann: I'd probably just name this `logging-routes`, since it's not doing anything (it's just a definition).

common-app-lib/src/cmr/common_app/api/log.clj
```diff
@@ -0,0 +1,133 @@
...
+  "Routes for changing the logging configuration endpoints"
+  (context "/log" []
```
oubiwann: FYI, since this is doing something (from the API consumer's point of view, that is), I'm good with `log` here 😄

common-app-lib/src/cmr/common_app/api/log.clj
```diff
@@ -0,0 +1,133 @@
...
+   with the request or the user doesn't have permission, the acl will through an exception. The
+   exception is caught in base code and an appropriate error message is passed to the end user."
+  [request-context headers]
+  (acl/verify-ingest-management-permission request-context :update)
+  (get-log-configuration-response
+    (.contains (mt/get-header headers "accept") "text/html")
+    200
+    (log/get-logging-configuration)))
+
+(defn merge-log-configuration
```
oubiwann: This is a side-effect-y operation ... it's setting data and not returning anything, so you should change the name with an appended `!`: `merge-logging-configuration!` . (I think `logging` is a better name here, and it matches your function above, `get-logging-configuration`.)

littleidiot40:
Before I'm going to allow this in an operations environment for any application the following needs to happen,
1. The API to change the log level must be ACL controlled with a new ACL
2. You must never be able to reduce the log level to a point where our metrics logging are not output
3. The application that is being manipulated must utilize log levels in a meaningful way (or this is a waste of time)

oubiwann:
For @littleidiot40's first point:

* Update the code to ensure that the log routes are protected (See `cmr.search.site.routes/build-routes` which points to `cmr.search.services.content-service/retrieve-page`, where the actual ACL check occurs)
* Add a sys-int test that performs multiple permissions tests and only allows admin tokens in (See `cmr.system-int-test.utils.site-util/assert-search-directory-regen-permissions` as used in `cmr.system-int-test.search.collection-directory-pages-test/regeneration-permissions` or `cmr.system-int-test.search.sitemaps-test/regeneration-permissions` for a concise example on how to do this)

oubiwann:
For @littleidiot40's second point, you'll want to add a filtering function right before the final config merge that ensures that, for a given set of namespaces (defined explicitly, e.g., `protected-logging-nss`), logging cannot be set higher than the default (e.g., define an explicit sets `#{:fatal :error :warn}` that includes the default and higher, and if it `contains?` the requested one, drop that from the config update). Making it nosier should be fine, of course :-)

oubiwann:
@littleidiot40's final point highlights something that we are focusing more on now: improving the logging (proper categorization and appropriate messages ... both meaningful content as well as general usefulness of the particular log statement). That work's on-going and doesn't affect your particular PR; we all simply need to do a better job of improving CMR's logging.

common-app-lib/src/cmr/common_app/api/log.clj
```diff
@@ -0,0 +1,133 @@
...
+   CMR logging configuration and 2) Change the logging configuration. To get the logging
+   configuration from a service such as search locally one would use the following command:
+   curl -H \"Echo-Token: XXXX\" http://localhost:3003/log where the port goes to the correct service.
+   From the internet the command is:
+   curl -H \"Echo-Token: XXXX\" http://cmr.earthdata.nasa.gov/search/log
+   The above commands return the current CMR logging configuration to the caller. It will format the
+   message to either a basic html if one of the accepted headers is text/html, or as formatted raw text
+   if text/html is not included in the accept HTML header.
+   To change the configuration one needs to provide an EDN map of the configuration you want to replace.
+   For example if the configuraiton looks like the following:
```
chris-durbin: configuration typo

common-app-lib/src/cmr/common_app/api/log.clj
```diff
@@ -0,0 +1,133 @@
...
+   {:level :warn,
```
chris-durbin: There's too much in this map. We should limit to only a subset of fields that we could change. For example we aren't going to be passing in functions. What fields should we actually allow modifying - from your example it looks like just `:level` and `:ns-pattern-map`?

common-app-lib/src/cmr/common_app/api/log.clj
```diff
@@ -0,0 +1,133 @@
...
+    use the following commmand:
+    curl -i -XPUT -H \"Echo-Token: XXXX\" -H \"Content-Type: application/edn\" <URL> -d
+        '{:level :debug :ns-pattern-map {\"cmr.common-app.api.log\" :debug :all :warn}}'
+    Where <URL> is either for example http://localhost:3003/log or
+     http://cmr.earthdata.nasa.gov/search/log
+    This command changes the overall logging level to debug - then sets the cmr.common-app.api.log
+    namespace to debug and all other names spaces are set to warn.
+    ******* Important! - The main logging level :level - must be set at the lowest setting that you
+    want logged at the namespace level. Otherwise Timbre will filter these logs out before the logs
+    make it to the middleware.  The middleware :all will set the logging level for all namespaces
```
chris-durbin: If someone leaves out :all from the map, could we default to adding :all and the current logging level?

common-app-lib/src/cmr/common_app/api/log.clj
```diff
@@ -0,0 +1,133 @@
...
+   :headers {cr/CONTENT_TYPE_HEADER "text/plain"}})
+
+(defn get-logging-configuration
+  "This function gets the logging configuration map to the caller. If a token is not passed in
+   with the request or the user doesn't have permission, the acl will through an exception. The
+   exception is caught in base code and an appropriate error message is passed to the end user."
+  [request-context headers]
+  (acl/verify-ingest-management-permission request-context :update)
+  (get-log-configuration-response
+    (.contains (mt/get-header headers "accept") "text/html")
```
chris-durbin: The idiomatic way to do contains is: `(some #{"text/html"} [(mt/get-header headers "accept")])`. But, in this case I think you just want `(= mt/html mt/accept-mime-type)` using the already defined ways for looking at mime-types in that namespace.

common-app-lib/src/cmr/common_app/api/log.clj
```diff
@@ -0,0 +1,133 @@
...
+(defn get-log-configuration-response
+  "Creates a successful log configuration response which is an output of the logging configuration map.
+   This function takes in three parameters, the first is a boolean - if true the output is in html format
+   otherwise the output will be a string representation of the configuration map. The second parameter is
+   the status code and the third is the actual configuration map."
+  [html? status-code data]
+  (let [body-string (if html?
+                      (str "<html><pre><code>" (with-out-str (pp/pprint data)) "</code></pre></html>")
+                      (with-out-str (pp/pprint data)))
+        content (if html?
```
chris-durbin: I'd call this content-type-header since it isn't the content.

common-app-lib/src/cmr/common_app/api/log.clj
eereiter: Great!

common-app-lib/src/cmr/common_app/api/log.clj
eereiter: I got rid of them - it was an oversight on my part.

common-app-lib/src/cmr/common_app/api/log.clj
eereiter: done.

common-app-lib/src/cmr/common_app/api/log.clj
eereiter: I have moved the code into cmr.common-app.services.logging.  I moved the routing function into routes.clj in the same directory and deleted the log.clj file.

common-app-lib/src/cmr/common_app/api/log.clj
eereiter: Thanks. The comments have been moved to cmr.common-app.services.logging

common-app-lib/src/cmr/common_app/api/log.clj
eereiter: In this function I am setting the configuration and then I am returning the new configuration.  Since I am doing that should I still append the function with !?

common-lib/src/cmr/common/log.clj
eereiter: yes, I have removed it.

common-app-lib/src/cmr/common_app/api/log.clj
eereiter: fixed.

common-app-lib/src/cmr/common_app/api/log.clj
eereiter: done

common-app-lib/src/cmr/common_app/api/log.clj
eereiter: mt/accept-mime-type gets you only the first accept header that is recognized by the mime-type namespace and not a list of them.  If text/html is not the first header that is recognized by mime-type then I will miss it.  For example if the accept header includes "*/*, text/csv, text/html" then accept-mime-type returns only text/csv.  I wanted to get all accept headers and see if one of them is text/html. I can't use the idiomatic way because the result of (mt/get-header headers "accept") is a string and not a collection.  I would have to parse out the string into its individual components first by space and then try the some function.  Is that still the better approach?

common-lib/src/cmr/common/log.clj
eereiter: done

common-lib/src/cmr/common/log.clj
eereiter: done

common-lib/src/cmr/common/log.clj
eereiter: done.

common-app-lib/src/cmr/common_app/api/log.clj
eereiter: done
I allowed :level, :ns-pattern-map, :ns-blacklist, and :ns-whitelist.

common-app-lib/src/cmr/common_app/api/log.clj
eereiter: There is no way to take out :all from the API - there is no way to remove any key from the API, I have added a reset function that resets the logging back to what it was before a user changed it. All configuration changes are "merged" - added or changed only, never removed.

common-app-lib/src/cmr/common_app/api/log.clj
eereiter: From a reply in chat, you were OK not adding ! since the new configuration is returned.  I have also documented the fact that the function also returns the new configuration.

system-int-test/src/cmr/system_int_test/utils/logging_util.clj
```diff
@@ -0,0 +1,52 @@
...
+  "This contains utilities for testing logging"
+  (:require
+   [clojure.test :refer [is]]
+   [cmr.common-app.api.routes :as cr]
+   [cmr.common.mime-types :as mt]
+   [cmr.mock-echo.client.echo-util :as e]
+   [cmr.system-int-test.system :as s]
+   [cmr.transmit.logging :as trans-logging]))
+
+(defn- process-response
```
chris-durbin: I don't understand why we have this function yet. Come back to this.

system-int-test/src/cmr/system_int_test/utils/logging_util.clj
```diff
@@ -0,0 +1,52 @@
+(ns cmr.system-int-test.utils.logging-util
+  "This contains utilities for testing logging"
+  (:require
+   [clojure.test :refer [is]]
+   [cmr.common-app.api.routes :as cr]
+   [cmr.common.mime-types :as mt]
+   [cmr.mock-echo.client.echo-util :as e]
+   [cmr.system-int-test.system :as s]
+   [cmr.transmit.logging :as trans-logging]))
```
chris-durbin: With this setup we aren't doing true system integration tests. The way the other tests work we hit a URL defined in a url-helper namespace. That way if a route changes in transmit our tests will fail (we broke backwards compatibility). The way this is written we're completely ignoring the API call and just using the source code to make the right call.

system-int-test/test/cmr/system_int_test/logging/runtime_change_test.clj
```diff
@@ -0,0 +1,119 @@
...
+   [cmr.system-int-test.system :as s]
+   [cmr.system-int-test.utils.logging-util :as log]
+   [cmr.system-int-test.utils.ingest-util :as ingest]))
+
+(use-fixtures :each (ingest/reset-fixture
+                     {"provguid1" "PROV1"}
+                     {:grant-all-search? false
+                      :grant-all-ingest? false}))
+
+;bootstrap is different and virtual-product I had problems - will address these later.
```
chris-durbin: Use two semicolons and a space for comments. We mark these types of TODOs with a ticket number.

system-int-test/test/cmr/system_int_test/logging/runtime_change_test.clj
```diff
@@ -0,0 +1,119 @@
...
+(def logging-apps-to-test
+  [:search :ingest :access-control :index-set :indexer :metadata-db :cubby])
+
+; Currently the integration tests logging level is set to debug - so set it to info for this test.
```
chris-durbin: same comment for comments

system-int-test/test/cmr/system_int_test/logging/runtime_change_test.clj
```diff
@@ -0,0 +1,119 @@
...
+  (str {:level :info
+        :ns-pattern-map {"cmr.common-app.api.log" :debug :all :info}}))
+
+(deftest logging-configuration-tests
+  (let [admin-update-group-concept-id (e/get-or-create-group (s/context) "admin-update-group")
+        admin-update-token (e/login (s/context) "admin" [admin-update-group-concept-id])
+        user-token (e/login (s/context) "user1")
+        _ (e/grant-group-admin (s/context) admin-update-group-concept-id :update)]
+    (doseq [app logging-apps-to-test]
+      (testing (str (name app) " tests: ")
```
chris-durbin: "logging tests"

transmit-lib/src/cmr/transmit/logging.clj
```diff
@@ -0,0 +1,85 @@
...
+     (h/request context application
+                {:url-fn reset-logging-url
+                 :method :get
+                 :raw? raw?
+                 :http-options (merge {:headers headers
+                                       :accept :json
+                                       ;; :decompress-body false is needed to
+                                       ;; workaround an issue with clj-http.client
+                                       ;; not handling gzip and unicode characters
+                                       :decompress-body false}
```
chris-durbin: I don't think you need to worry about special characters for this endpoint so setting :decompress-body false is not needed. Same for line 39 above.

common-app-lib/src/cmr/common_app/services/logging_config.clj
```diff
@@ -0,0 +1,134 @@
...
+   [cmr.common.log :as log]
+   [cmr.common-app.api.routes :as cr]
+   [cmr.common.mime-types :as mt]))
+
+(defn get-log-configuration-response
+  "Creates a successful log configuration response which is an output of the logging configuration map.
+   This function takes in three parameters, the first is a boolean - if true the output is in html format
+   otherwise the output will be a string representation of the configuration map. The second parameter is
+   the status code and the third is the actual configuration map."
+  [html? status-code data]
```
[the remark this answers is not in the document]
eereiter: done.

common-app-lib/src/cmr/common_app/services/logging_config.clj
```diff
@@ -0,0 +1,134 @@
...
+   or not set at all, then the compojure wrap-params will consume the body and it will be blank.
+   Therefore I will send back an error to correctly set the content-type. Otherwise I will have to
+   save the body in another key (like search is doing) in all services and use that key instead.
+   https://github.com/weavejester/compojure/issues/108"
+  [status-code content-type]
+  {:status status-code
+   :body (str "The content type of "
+              content-type
+              " in the request is either not provided or is not of the correct type. Please set the
+                request Content-Type to application/edn.")
```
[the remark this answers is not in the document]
eereiter: done

common-app-lib/src/cmr/common_app/services/logging_config.clj
```diff
@@ -0,0 +1,134 @@
...
+   :headers {cr/CONTENT_TYPE_HEADER "text/plain"}})
+
+(defn get-logging-config
+  "This function gets the logging configuration map and formats it into a proper response. The
+   passed in variable determins if the response is for an html accepting client or not."
```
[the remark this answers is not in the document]
eereiter: done

common-app-lib/src/cmr/common_app/services/logging_config.clj
```diff
@@ -0,0 +1,134 @@
...
+    ;; This tests to see if the content type is correct. If the user didn't set the correct
+    ;; content type or didn't set it at all then the compojure wrap-params will consume the
+    ;; body and it will be blank when the body gets here.
+    (if (string/blank? body-string)
+      (get-log-error-response 400 (mt/get-header headers "Content-Type"))
+      (merge-logging-config body-string (mt/mime-type-exists? mt/html "accept" headers)))))
+
+(defn reset-logging-config
+  "This function resets the logging configuration back as to when the application started. The
+   passed in variable determins if the response is for an html accepting client or not."
```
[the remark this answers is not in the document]
eereiter: done

common-lib/src/cmr/common/log.clj
```diff
@@ -2,11 +2,17 @@
...
+   [clojure.java.io :as io]
+   [clojure.string :as string]
+   [cmr.common.lifecycle :as lifecycle]
+   [taoensso.encore :as enc]
+   [taoensso.timbre :as t]
+   [taoensso.timbre.appenders.core :as a]))
+
+(def ^:const default-logging-level
+  "This is the default level to set if the starting application doesn't set it."
+  :warn)
```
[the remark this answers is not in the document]
eereiter: I have set it to :info

common-lib/src/cmr/common/log.clj
```diff
@@ -32,24 +38,207 @@
...
+(defn get-partial-logging-configuration
+  "This function returns only the logging configuration that the CMR is allowing a privileged
+   user to change."
+  []
+  (select-keys (get-logging-configuration) allowed-logging-configuration-changes))
+
+(def ^:const logging-levels-not-allowed
+  "This is the default logging levels not allowed in a normal CMR application because it would
+   harm metrics captured through logging."
+  #{:error :fatal})
```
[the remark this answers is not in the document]
eereiter: I have added :warn to the not allowed logging levels.

common-lib/src/cmr/common/log.clj
```diff
@@ -32,24 +38,207 @@
...
+
+   Here is a less verbose description for those that may understand it:
+   What it does precisely is: given the data structure to be logged (as the 1st arg) determines
+   whether the logging will actually happen or not by looking up in the configuration hashmap
+   ns-pattern-map, which has either the namespace string filters as specified in
+   taoensso.encore/ns-filter - or the keyword :all- as keys, and the level as a value (es. :debug)"
+  [& [{:keys [?ns-str config level] :as opts}]]
+  (let [ns-pattern-map (get config :ns-pattern-map)
+        namesp (get-namespace-from-pattern-map ns-pattern-map ?ns-str)
+        loglevel (get ns-pattern-map namesp (get config :level))]  ;(get map key not-found)
```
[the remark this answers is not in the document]
eereiter: done

system-int-test/test/cmr/system_int_test/logging/runtime_change_test.clj
eereiter: Wrote tickets for these and added a TODO comment for each.  Also added a space

system-int-test/test/cmr/system_int_test/logging/runtime_change_test.clj
eereiter: done

system-int-test/test/cmr/system_int_test/logging/runtime_change_test.clj
eereiter: done

common-lib/test/cmr/common/test/mime_types.clj
```diff
@@ -40,6 +40,20 @@
...
+    (is (= true (mt/mime-type-exists? mt/html "accept" {"accept" "text/html"})))
+    (is (= true (mt/mime-type-exists? mt/html "accept" {"accept" "text/csv, text/html"})))
+    (is (= true (mt/mime-type-exists? mt/html "accept" {"content-type" mt/json, "accept" "*/* , text/csv, text/html"})))
+    (is (= true (mt/mime-type-exists? mt/html "accept" {"connection" "keep-alive", "user-agent" "Apache-HttpClient/4.5 (Java/1.8.0_92)", "host" "localhost:3003", "accept" "*/*, text/html", "echo-token" "ABC-2"}))))
+
+  (testing "mime-types do not exist"
+    (is (= false (mt/mime-type-exists? mt/html "accept" {"accept" "application/xml; q=1"})))
+    (is (= false (mt/mime-type-exists? mt/html "accept" {"concept-type" "application/xml; q=1"})))
+    (is (= false (mt/mime-type-exists? mt/html "accept" {"connection" "keep-alive", "user-agent" "Apache-HttpClient/4.5 (Java/1.8.0_92)", "host" "localhost:3003", "accept" "*/*, text/csv", "echo-token" "ABC-2"})))))
+
```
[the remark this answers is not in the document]
eereiter: added blank lines to all of the tests in this namespace.  Also removed the non used required clojure.string namespace and replaced the are2 with are3 macro.

PR #287, opened 2017-09-22
CMR-4467 - Add a workaround for provider-id in services ingest until CMR-4468

oubiwann:
This hostfix is part of CMR-4467 (see PR #283). As soon as the tests pass, this can be merged, as it is an update to the already-merged PR.


oubiwann:
Build passed.

PR #289, opened 2017-09-26
CMR-4517: Fixed FIND_AND_REMOVE for bulkupdate.

sxu123:
FIND_AND_REMOVE should be able to remove all the found Platforms/Instruments/DataCenters/ScienceKeywords/LocationKeywords. unless it's required.
Before the fix, if you try to remove all of the platforms, it will give error saying the array is too short because it's empty. After the fix, it will just say that the Platforms is required.

Instruments and LocationKeywords are not required, so now we are able to remove all of the instruments and LocationKeywords.

### DIFF
--- a/system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
+++ b/system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
@@ -102,9 +102,9 @@
 
 (defn- ingest-collection-in-umm-json-format
   "Ingest a collection in UMM Json format and return a list of one concept-id.
-   This is used to test the FIND_AND_REMOVE on platforms/instruments. Since differenct
-   formats keep different parts of the platforms/instruments, can't combine them together
-   to test what's found and what's removed."
+   This is used to test the CMR-4517, on complete removal of platforms/instruments. 
+   Since it's only testing the bulk update part after the ingest, which format to use 
+   for ingest is irrelevant. So it's easier to just test with one format of ingest."
   [attribs]
   (let [collection (data-umm-c/collection-concept
                      (data-umm-c/collection 1 attribs)
--- a/system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
+++ b/system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
@@ -65,6 +65,25 @@
                       :VariableLevel1 "CLOUD MICROPHYSICS"
                       :VariableLevel2 "CLOUD LIQUID WATER/ICE"}]})
 
+(def find-remove-all-platforms-instruments-umm
+  {:Platforms [{:ShortName "a340-600-1"
+                :LongName "airbus a340-600-1"
+                :Type "Aircraft"}
+               {:ShortName "a340-600-2"
+                :LongName "airbus a340-600"
+                :Type "Aircraft"
+                :Instruments [{:ShortName "atm"
+                               :LongName "airborne topographic mapper"
+                               :Technique "testing"
+                               :NumberOfInstruments 0
+                               :OperationalModes ["mode1" "mode2"]}]}
+               {:ShortName "a340-600-3"
+                :LongName "airbus a340-600"
+                :Type "Aircraft"
+                :Instruments [{:ShortName "atm"
+                               :LongName "airborne topographic mapper"}]}]})
+
+
 (defn- generate-concept-id
   [index provider]
   (format "C120000000%s-%s" index provider))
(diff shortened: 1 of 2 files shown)


umm-spec-lib/src/cmr/umm_spec/field_update.clj
```diff
@@ -34,9 +34,12 @@
...
   (if (seq (get-in umm update-field))
-    (update-in umm update-field #(remove (fn [x]
-                                           (value-matches? find-value x))
-                                         %))
+    (let [umm-updated (update-in umm update-field #(remove (fn [x]
+                                                             (value-matches? find-value x))
+                                                     %))]
+      (if (seq (get-in umm-updated update-field))
+        umm-updated
+        (update-in umm-updated update-field seq)))
```
sxu123: I thought I could just do (remove-map-keys empty? umm-updated) to get rid of the empty platforms/instruments and so on. But when I did that, bulk update status returns things like "No matching clause" which I still don't know what exactly it means. The strange part is, even when I only remove partial of the platforms, it still gives me that status. I don't know if it's because I removed some other nil fields that somehow matter to this. Looks like it needs all the required elements there, so I just made them nil.

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
```diff
@@ -81,6 +100,18 @@
...
+   This is used to test the FIND_AND_REMOVE on platforms/instruments. Since differenct
+   formats keep different parts of the platforms/instruments, can't combine them together
+   to test what's found and what's removed."
+  [attribs]
+  (let [collection (data-umm-c/collection-concept
+                     (data-umm-c/collection 1 attribs)
+                     :umm-json)]
+    [(:concept-id (ingest/ingest-concept
+                   (assoc collection :concept-id (generate-concept-id 1 "PROV1"))))])) 
+
```
sxu123: I ran into a lot of issues with the tests.  First of all, (data/umm-c/collection-concept  collection format) doesn't work when format is iso* and collection contains list of Platforms/Instruments because it complains about smap-keyword-str not working.  Since it's not bulk-update related, I just skipped that. Then these other formats, they will remove different fields for platform/instrument, so you can't combine them in one test.  So I decided to just do one umm-json format.

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
```diff
@@ -217,6 +248,83 @@
                 :VariableLevel1 "EMISSIONS"}]
               (:ScienceKeywords (:umm concept)))))))
 
+(deftest bulk-update-remove-all-platforms-test
```
oubiwann: I think this test should make sure that what you remove later is present first. This will be redundant with other tests, for sure, but I think it's a good practice and will give more certainty to the test/setup/etc.

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
```diff
@@ -217,6 +248,83 @@
...
+  (let [concept-ids (ingest-collection-in-umm-json-format find-remove-all-platforms-instruments-umm)
+        _ (index/wait-until-indexed)
```
oubiwann: In light of the above comment, the assignment `concept-ids` can be in its own let, then you can just do a regular `(index/wait...)`, then a check or two for the presence of the data that you will be removing.

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
```diff
@@ -217,6 +248,83 @@
...
+        bulk-update-body {:concept-ids concept-ids
```
oubiwann: Then this can go in a nested `let` under that.

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
```diff
@@ -217,6 +248,83 @@
...
+                               :OperationalModes ["mode1" "mode2"]}]}
+               {:ShortName "a340-600-3"
+                :LongName "airbus a340-600"
+                :Type "Aircraft"
+                :Instruments [{:ShortName "atm"
+                               :LongName "airborne topographic mapper"}]}]
+              (:Platforms (:umm concept)))))))
+
+(deftest bulk-update-remove-all-instruments-test
+  (let [concept-ids (ingest-collection-in-umm-json-format find-remove-all-platforms-instruments-umm)
```
oubiwann: Same feedback for this test as the other.

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
oubiwann: I misread this -- sorry! You're right -- this is actual ingest with no removal.

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
oubiwann: Skip this feedback.

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
oubiwann: Ignore this comment.

PR #292, opened 2017-09-29
CMR-4223: Added created-at to bulk update status responses.

sxu123:


system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_flow_test.clj
```diff
@@ -101,7 +101,8 @@
         json-body (json/generate-string bulk-update-body)
         {:keys [task-id]} (ingest/bulk-update-collections "PROV1" bulk-update-body)
         _ (qb-side-api/wait-for-terminal-states)
-        status-response (ingest/bulk-update-task-status "PROV1" task-id)]
+        status-response (ingest/bulk-update-task-status "PROV1" task-id)
+        status-response (dissoc status-response :created-at)]
```
sxu123: Sorry running out of time to do sophisticated time comparisons since it involves time synchronization.   Added demo to demonstrate that created-at contains the timestamp when bulk update happens. Chris said maybe someone else could add this part next week.

ingest-app/src/cmr/ingest/data/bulk_update.clj
```diff
@@ -36,23 +37,31 @@
...
-                                            (su/from "bulk_update_task_status")
-                                            (su/where `(= :provider-id ~provider-id)))))
-          statuses (map util/map-keys->kebab-case statuses)]
-     (map #(update % :request-json-body util/gzip-blob->string)
-          statuses)))
+    (j/with-db-transaction
+     [conn db]
+     ;; Returns a list of bulk update statuses for the provider
+     (let [stmt (su/build (su/select [:created-at :task-id :status :status-message :request-json-body]
+                            (su/from "bulk_update_task_status")
```
chris-durbin: Minor: we should make the name of the table a constant defined in this file. The indentation looks off as well - this line should be directly below the start of the argument above.

oubiwann: Quick look at the Oracle schemas in a running local instance:
 * `CMR_INGEST.BULK_UPDATE_COLL_STATUS` doesn't have a `CREATED_AT` column
 * `CMR_INGEST.BULK_UPDATE_TASK_STATUS` does ...

sxu123: Correct. we only want to use the created_at from bulk_update_task_status table. Done.

PR #294, opened 2017-10-04
CMR-4524: Fix to only delete specific keys from cubby on calls to reset

chris-durbin:
I found that the previous change I made was still resulting in data being deleted from cubby because the one cache that it cleared was a consistent cache which uses cubby underneath to store a hash code. As a result I changed the design of cubby so that data can only be explicitly deleted from cubby by registering a list of keys that the caller wants to be deleted from cubby on any calls to reset on that cache.

### DIFF
--- a/common-app-lib/src/cmr/common_app/cache/consistent_cache.clj
+++ b/common-app-lib/src/cmr/common_app/cache/consistent_cache.clj
@@ -91,7 +91,7 @@
     [this key]
     (let [mem-value (c/get-value memory-cache key)]
       (when (and (not (nil? mem-value))
-                 (= (hash mem-value) 
+                 (= (hash mem-value)
                     (c/get-value hash-cache (key->hash-cache-key key))))
         mem-value)))
 
@@ -144,12 +144,9 @@
    (create-consistent-cache nil))
   ([options]
    (let [timeout (get options :hash-timeout-seconds (consistent-cache-default-hash-timeout-seconds))
-         hash-cache (fallback-with-timeout (cubby-cache/create-cubby-cache) timeout)
+         hash-cache (cubby-cache/create-cubby-cache options)
          main-cache (mem-cache/create-in-memory-cache)]
-     (create-consistent-cache main-cache hash-cache)))
+     (create-consistent-cache main-cache (fallback-with-timeout hash-cache timeout))))
   ([memory-cache hash-cache]
    (->ConsistentMemoryCache
     memory-cache hash-cache)))
-
-
-
--- a/common-app-lib/src/cmr/common_app/cache/cubby_cache.clj
+++ b/common-app-lib/src/cmr/common_app/cache/cubby_cache.clj
@@ -1,7 +1,11 @@
 (ns cmr.common-app.cache.cubby-cache
   "An implementation of the CMR cache protocol on top of the cubby application. Cubby only supports
   persistence and returning strings so this automatically serializes and deserializes the keys and
-  values with EDN. Any key or value serializable to EDN is supported."
+  values with EDN. Any key or value serializable to EDN is supported. Operationally cubby uses
+  Elasticsearch as a backend store, and multiple applications have multiple caches which use cubby.
+  Therefore calling reset on a cubby cache will not delete all of the data stored in the backend.
+  Instead when creating a cubby cache the caller must provide a list of keys which should be deleted
+  when calling reset on that particular cache."
   (:require
    [clojure.edn :as edn]
    [cmr.common.cache :as c]
@@ -21,7 +25,11 @@
 ;; Implements the CmrCache protocol by saving data in the cubby application
 (defrecord CubbyCache
   [;; A context containing a connection to cubby
-   context-with-conn]
+   context-with-conn
+
+   ;; A collection of keys used by this cache. Only these keys will be deleted from the backend
+   ;; store on calls to reset
+   keys-to-track]
 
   c/CmrCache
   (get-keys
@@ -43,7 +51,8 @@
 
   (reset
     [this]
-    (cubby/delete-all-values context-with-conn))
+    (doseq [the-key keys-to-track]
+      (cubby/delete-value context-with-conn (serialize the-key))))
 
   (set-value
     [this key value]
@@ -51,7 +60,8 @@
 
 (defn create-cubby-cache
   "Creates an instance of the cubby cache."
-  []
-  (->CubbyCache {:system (config/system-with-connections {} [:cubby])}))
-
-
+  ([]
+   (create-cubby-cache nil))
+  ([options]
+   (->CubbyCache {:system (config/system-with-connections {} [:cubby])}
+                 (:keys-to-track options))))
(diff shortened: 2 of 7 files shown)


common-app-lib/src/cmr/common_app/cache/consistent_cache.clj
```diff
@@ -144,12 +144,9 @@
    (create-consistent-cache nil))
   ([options]
    (let [timeout (get options :hash-timeout-seconds (consistent-cache-default-hash-timeout-seconds))
-         hash-cache (fallback-with-timeout (cubby-cache/create-cubby-cache) timeout)
+         hash-cache (cubby-cache/create-cubby-cache options)
          main-cache (mem-cache/create-in-memory-cache)]
-     (create-consistent-cache main-cache hash-cache)))
+     (create-consistent-cache main-cache (fallback-with-timeout hash-cache timeout))))
```
oubiwann: Nice little bit of cleaning; we should do more of this :-) Makes it easier to read.

common-app-lib/src/cmr/common_app/cache/cubby_cache.clj
```diff
@@ -1,7 +1,11 @@
 (ns cmr.common-app.cache.cubby-cache
   "An implementation of the CMR cache protocol on top of the cubby application. Cubby only supports
   persistence and returning strings so this automatically serializes and deserializes the keys and
-  values with EDN. Any key or value serializable to EDN is supported."
+  values with EDN. Any key or value serializable to EDN is supported. Operationally cubby uses
```
oubiwann: Excellent. Good find, and this feels like a good engineering solution to the old problem/convention.

common-app-lib/src/cmr/common_app/cache/cubby_cache.clj
```diff
@@ -21,7 +25,11 @@
 ;; Implements the CmrCache protocol by saving data in the cubby application
 (defrecord CubbyCache
   [;; A context containing a connection to cubby
-   context-with-conn]
+   context-with-conn
+
+   ;; A collection of keys used by this cache. Only these keys will be deleted from the backend
+   ;; store on calls to reset
+   keys-to-track]
```
oubiwann: My only question would be if this has any impact on existing usage ... I can only assume either "no" or you've fixed it below :-) I'll keep reading ...

indexer-app/src/cmr/indexer/system.clj
```diff
@@ -46,6 +46,11 @@
   {:default 5
    :type Long})
 
+(def index-set-mappings-cubby-keys
+  "The list of keys related to index-set-mappings which should be deleted when we want to clear
+  the index-set-mappings cache."
+  [":concept-mapping-types-hash-code" ":concept-indices-hash-code"])
```
oubiwann: Sidenote: yikes ... we use stringifyed keywords in CMR?

common-app-lib/src/cmr/common_app/cache/cubby_cache.clj
oubiwann: Had a quick pvt chat about this; looks good.

indexer-app/src/cmr/indexer/system.clj
chris-durbin: Yeah... I thought about cleaning that up, but since it is just cosmetic I didn't want to think through transition implications if we use two different keys for the same cached item while both the old and new version of apps are running.

chris-durbin:
I looked through all of our explicit uses of cubby:

KMS backup data
Collection granule aggregation backup data
Site content backup data
Elasticsearch scroll ID backup data
Humanizer report backup data

There is no use case for deleting that data (we only want to update it with set-value), so there are no mappings defined to delete that data.

We also use cubby for a hash code with the consistent cache implementation. I realized that clearing a consistent cache for an application with a single node will be fine even without deleting the hash code from cubby, but with multiple nodes, clearing a single node will not cause the others to be updated automatically. So I'm going to look at all of the consistent caches uses, and I will update those to register the keys to delete on a call to reset.

common-app-lib/src/cmr/common_app/cache/cubby_cache.clj
```diff
@@ -1,7 +1,11 @@
 (ns cmr.common-app.cache.cubby-cache
   "An implementation of the CMR cache protocol on top of the cubby application. Cubby only supports
   persistence and returning strings so this automatically serializes and deserializes the keys and
-  values with EDN. Any key or value serializable to EDN is supported."
+  values with EDN. Any key or value serializable to EDN is supported. Operationally cubby uses
+  Elasticsearch as a backend store, and multiple applications have multiple caches which use cubby.
+  Therefore calling reset on a cubby cache will not delete all of the data stored in the backend.
```
ygliuvt: Change "will not" to "should not".

ygliuvt:
I see you have partially answered my question while I am reviewing the code. Sometimes it is not possible to know all the cached keys beforehand. I think one way to deal with that is to introduce a cache namespace. e.g. all cached keys of indexer-app would be prefixed with something like "indexer.". This way the cache reset function within an app can just delete all the cached keys based on pattern matching. This will also eliminate the need to set up the keys-to-track. Just my 2 cents.

chris-durbin:
> I think one way to deal with that is to introduce a cache namespace. e.g. all cached keys of indexer-app would be prefixed with something like "indexer.". This way the cache reset function within an app can just delete all the cached keys based on pattern matching. This will also eliminate the need to set up the keys-to-track. Just my 2 cents.

I thought about that design too. I agree it is better to not have to register the keys that get deleted. The problem I have is that we only want to clear the cubby data for some of the cached values within a single app, not all of them (so again we're back to registering what should and should not be deleted by a call to clear). Then there's also some data cached in cubby shared by multiple applications making the application prefixed cache keys a problem.

chris-durbin:
> Except the indexer-app index-set-cache is modified to do that, none of the other caches are updated, this means the clear cache call will not properly clear the cubby caches that are used by the apps. @chris-durbin, have you thought this through?

Yep, this was the whole point. We've been deleting things from cubby unintentionally. Now we're going to be deliberate about what we delete. I've verified that we almost never call reset programmatically - the only calls are during deployments when we call update-indexes. For the manual clear cache calls that we may have to initiate from time to time I've tried to think about what we may need to clear. The ACLs hash code is the only thing that makes sense. Worst case scenario is we can use the cubby API to delete cached items, so we're still ok there.

ygliuvt:
>> I think one way to deal with that is to introduce a cache namespace. e.g. all cached keys of indexer-app would be prefixed with something like "indexer.". This way the cache reset function within an app can just delete all the cached keys based on pattern matching. This will also eliminate the need to set up the keys-to-track. Just my 2 cents.

> I thought about that design too. I agree it is better to not have to register the keys that get deleted. The problem I have is that we only want to clear the cubby data for some of the cached values within a single app, not all of them (so again we're back to registering what should and should not be deleted by a call to clear). Then there's also some data cached in cubby shared by multiple applications making the application prefixed cache keys a problem.

We can selectively add the prefix to cached keys, which means we will not do that for shared caches. I think it is OK to clear all caches for a single app. If we do have a use case that requires us to clear only a couple of them, we can either add a specific endpoint just for that or add a generic way to clear a particular key that is cached. I notice that it is more than you have planned in this PR, so feel free to push this off as future enhancement.

PR #295, opened 2017-10-05
CMR-4526 

daniel-zamora:
This PR removes "date in the past" validation for project's EndDate in umm-spec-lib and the test for it.

### DIFF
--- a/umm-spec-lib/test/cmr/umm_spec/test/validation/umm_spec_collection_validation_tests.clj
+++ b/umm-spec-lib/test/cmr/umm_spec/test/validation/umm_spec_collection_validation_tests.clj
@@ -202,8 +202,6 @@
           (h/assert-warnings-multiple-invalid
             coll
             [{:path [:Projects 0 :StartDate]
-              :errors ["Date should be in the past."]}
-             {:path [:Projects 0 :EndDate]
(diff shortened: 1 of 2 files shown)


umm-spec-lib/test/cmr/umm_spec/test/validation/umm_spec_collection_validation_tests.clj
```diff
@@ -202,8 +202,6 @@
           (h/assert-warnings-multiple-invalid
```
chris-durbin: This test is to verify multiple validation warnings can be returned. It is no longer testing that since there is only one condition. We should create another validation warning for this (maybe add a second project with a start date in the future).

daniel-zamora: I just looked at this and was writing a fix, I think the test isn't about multiple warnings, its for specifically testing Start and End date being in the past.

chris-durbin: Oh ok - we should change the assertion to use h/assert-invalid rather than the multiple-invalid one and the testing block to say "start date not in the past".

PR #298, opened 2017-10-05
CMR-4508 - Add stubbed collection and associated variables/services data

oubiwann:
## Summary
This is done in support of the EDSC team's need to develop against an API that won't be completed for several more weeks, but which they need access to sooner for their own development and testing.

## Stubbed Endpoints

The following endpoints offer an option of stubbed data:

* https://CMR-HOST/search/collections
* https://CMR-HOST/search/collections?concept_id=XXX
* https://CMR-HOST/search/concepts/XXX

Additionally:
* The responses from each are currently only in UMM-JSON format.
* The concept ID parameter may be either single-valued or multi-valued (usage for this is the same as the non-prototype API searches)
* The collection responses will always return the same collection, no matter how it is called (there is only one stubbed collection)
* The collection responses contain the new boolean fields `has-variables`, `has-transforms`, and `has-formats`

## Special Usage

In order to tell CMR you would like to use the endpoints against stubbed data, a custom HTTP header must be set: `CMR-Prototype-UMM: true`.

## Examples

Get the collection:

```
curl --silent \
  -H "Accept: application/json" \
  -H "CMR-Prototype-UMM: true" \
  "https://CMR-HOST/search/collections?concept_id\[\]=C123" | \
jq .
```

or

```
curl --silent \
  -H "CMR-Prototype-UMM: true" \
  "https://CMR-HOST/search/collections.json?concept_id\[\]=C123" | \
jq .
```

Get the collection associations:

```
curl --silent \
  -H "CMR-Prototype-UMM: true" \
  "https://CMR-HOST/search/collections.json?concept_id\[\]=C123" |  \
jq .feed.entry[0].associations
```

Note that for each of those calls we used the `application/json` content type; the remaining calls are all UMM-JSON instead.

Concept lookups for stubbed service and variables is also supported:

```
curl --silent \
  -H "Accept: application/vnd.nasa.cmr.umm+json;version=1.9" \
  -H "CMR-Prototype-UMM: true" \
  "https://CMR-HOST/search/concepts/S0001-GES_DISC" | \
jq .
```
```
curl --silent \
  -H "Accept: application/vnd.nasa.cmr.umm+json;version=1.9" \
  -H "CMR-Prototype-UMM: true" \
  "https://CMR-HOST/search/concepts/V00011-GES_DISC" | \
jq .
```

## Responses

As mentioned, the responses are in UMM-JSON format, so have the following form:

```json
{
  "hits": 1,
  "took": 42,
  "items": [
    {
      "meta": {
        "revision-id": 1,
        "deleted": false,
        "format": "application/vnd.nasa.cmr.umm+json",
        "provider-id": "GES_DISC",
        "user-id": "cmr-edsc-stubs",
        "native-id": "XXX",
        "concept-id": "XXX",
        "revision-date": "2017-09-27T18:20:46Z",
        "concept-type": "XXX"
      },
      "umm": {
          ...
      }
    }
  ]
}
```

## Questions

1. We have currently only provided stubs for UMM-JSON response data; from the notes it sounds like this is all that EDSC will need for now; please correct this assumption, if others are needed.
1. Based on my current understanding of EDSC's needs and the changes to the development/feature plans in the course of the related conversation, it seems that only the `/search/collections`, and `/search/concepts` endpoints are needed; is this correct?

## Notes

1. The code change in this PR is quite small due to the fact that it was all put in throw-away, third-party libraries; for those interested, the code that supports these changes is available in the following locations:
 * https://github.com/oubiwann/cmr-edsc-stubs/commits/master
 * https://github.com/oubiwann/cmr-sample-data/commits/master

### DIFF
--- a/dev-system/project.clj
+++ b/dev-system/project.clj
@@ -49,7 +49,25 @@
   :url "https://github.com/nasa/Common-Metadata-Repository/tree/master/dev-system"
   :license {:name "Eclipse Public License"
             :url "http://www.eclipse.org/legal/epl-v10.html"}
-  :dependencies ~(concat '[[org.clojure/clojure "1.8.0"]
+  :dependencies ~(concat '[[gov.nasa.earthdata/cmr-client "0.2.0-SNAPSHOT"
+                            :exclusions [org.clojure/clojurescript
+                                         org.clojure/data.xml
+                                         com.google.code.findbugs/jsr305
+                                         clj-http
+                                         cljs-http
+                                         org.clojure/core.async
+                                         org.clojure/tools.analyzer.jvm
+                                         org.clojure/tools.reader]]
+                           [gov.nasa.earthdata/cmr-edsc-stubs "0.1.0-SNAPSHOT"
+                            :exclusions [org.clojure/clojurescript
+                                         com.google.code.findbugs/jsr305
+                                         gov.nasa.earthdata/cmr-client
+                                         cljs-http
+                                         org.clojure/core.async
+                                         org.clojure/tools.analyzer.jvm
+                                         org.clojure/tools.reader
+                                         org.clojure/java.jdbc]]
+                           [org.clojure/clojure "1.8.0"]
                            [org.clojure/tools.nrepl "0.2.12"]
                            ;; Add groovy to support groovy scripting in elastic
                            [org.codehaus.groovy/groovy-all "2.4.0"]]
--- a/dev-system/project.clj
+++ b/dev-system/project.clj
@@ -49,8 +49,52 @@
   :url "https://github.com/nasa/Common-Metadata-Repository/tree/master/dev-system"
   :license {:name "Eclipse Public License"
             :url "http://www.eclipse.org/legal/epl-v10.html"}
-  :dependencies ~(concat '[[org.clojure/clojure "1.8.0"]
+  :dependencies ~(concat '[[gov.nasa.earthdata/cmr-client "0.2.0-SNAPSHOT"
+                            :exclusions [org.clojure/clojurescript
+                                         org.clojure/data.xml
+                                         com.google.code.findbugs/jsr305
+                                         clj-http
+                                         cljs-http
+                                         org.clojure/core.async
+                                         org.clojure/tools.analyzer.jvm
+                                         org.clojure/tools.reader]]
+                           [gov.nasa.earthdata/cmr-edsc-stubs "0.1.0-SNAPSHOT"
+                            :exclusions [org.clojure/clojurescript
+                                         com.google.code.findbugs/jsr305
+                                         gov.nasa.earthdata/cmr-client
+                                         cljs-http
+                                         org.clojure/core.async
+                                         org.clojure/tools.analyzer.jvm
+                                         org.clojure/tools.reader
+                                         org.clojure/java.jdbc]]
+                           [org.clojure/clojure "1.8.0"]
                            [org.clojure/tools.nrepl "0.2.12"]
+                           ;; XXX REMOVE the following deps when the stubbed
+                           ;;     responses are replaced with the real ones
+                           [gov.nasa.earthdata/cmr-client "0.2.0-SNAPSHOT"
+                            :exclusions [cljs-http
+                                         com.google.code.findbugs/jsr305
+                                         gov.nasa.earthdata/cmr-client
+                                         instaparse
+                                         org.clojure/clojurescript
+                                         org.clojure/core.async
+                                         org.clojure/data.xml
+                                         org.clojure/tools.analyzer.jvm
+                                         org.clojure/tools.reader
+                                         org.clojure/java.jdbc
+                                         ring/ring-codec]]
+                           [gov.nasa.earthdata/cmr-edsc-stubs "0.1.0-SNAPSHOT"
+                            :exclusions [cljs-http
+                                         com.google.code.findbugs/jsr305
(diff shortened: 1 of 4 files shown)


dev-system/project.clj
```diff
@@ -49,7 +49,18 @@
   :url "https://github.com/nasa/Common-Metadata-Repository/tree/master/dev-system"
   :license {:name "Eclipse Public License"
             :url "http://www.eclipse.org/legal/epl-v10.html"}
-  :dependencies ~(concat '[[org.clojure/clojure "1.8.0"]
+  :dependencies ~(concat '[;; XXX REMOVE the following dep when the stubbed
```
chris-durbin: I don't think we have to put these in dev-system since it already requires the search-app which will pull in the dependencies.

ygliuvt:
> Questions

> 1. We have currently only provided stubs for UMM-JSON response data; from the notes it sounds like this is all that EDSC will need for now; please correct this assumption, if others are needed.

EDSC originally said that the JSON response data is also used when the collections are first loaded. We should check with EDSC to make sure that they are OK with just use UMM JSON for the prototyping.  We should not merge this without approval from EDSC.

ygliuvt:
> curl --silent \
  -H "Accept: application/vnd.nasa.cmr.umm+json;version=1.9" \
  -H "CMR-Prototype-UMM: true" \
  "http://localhost:3003/variables" | \

In the example, the UMM JSON version of variables and services are set to 1.9. It should be 1.0.

oubiwann:
@ygliuvt Yeah, I've messaged the EDSC team on JIRA and pointed them here; awaiting their feedback.

dev-system/project.clj
oubiwann: Cool; I'll take them out (I needed them for testing in the REPL).

ygliuvt:
I think EDSC should use the search-app concept retrieval route for getting the variable and services, e.g.  curl -i  "https://cmr.sit.earthdata.nasa.gov/search/concepts/:concept-id"

This would be easier for EDSC and probably more performant than searching by concept-id. We have not planned any service search issues in PI7.

ygliuvt:
On stubbing out retrieving service concept by concept-id, we could actually move the issue (CMR-4479) up to this sprint and implement it for real now that we have finished the implementation of ingesting of services.

ygliuvt:
It is not clear from the writeup that we have associations field in the collection search UMM JSON response. The associations field should be in the format of:

"associations" : { "variables" : [ "V1200000007-PROV1", "V1200000008-PROV1"]
                              "services" : [ "S1200000009-PROV1", "S1200000012-PROV1"] }

I have not got a chance to look at the client code to double check this, but we should at least make it clear in the writeup.

oubiwann:
Adding a mention for Simon so he is included in the conversations ... @scantrel8277

oubiwann:
Chatted with @ygliuvt about the search, and even though support for service search is now landing, I think we still want to provide the stub so there is no need for ingest into different deployment environments, thus making for a consistent dev/testing experience for the EDSC team.

oubiwann:
Adding concept-id lookup stub support will be super-easy; we can confirm with EDSC which exact endpoints are needed when @mightynimble and @mreese84 are back ...

oubiwann:
@ygliuvt The only problem with using `/search/concepts/XXX` exclusively is that the responses for that only contain the what the regular search provides via the `umm` key and not the content of the `meta` key ... so at the very least, EDSC will need to use `/search/collections?XXX` to get the boolean flags they need as well as the associations.

oubiwann:
@ygliuvt Okay, I've added `/search/concepts/XXX` support. Commits that provided this:
 * https://github.com/oubiwann/cmr-edsc-stubs/commit/30ca570c941087d4117f27dcd0336f2b39b894e2
 * https://github.com/oubiwann/Common-Metadata-Repository/commit/df8bba2

ygliuvt:
> @ygliuvt The only problem with using /search/concepts/XXX exclusively is that the responses for that only contain the what the regular search provides via the umm key and not the content of the meta key ... so at the very least, EDSC will need to use /search/collections?XXX to get the boolean flags they need as well as the associations.

Yes, the idea is to use /search/collections?XXX to get the flags and associations, and then use the /search/concepts/XXX to load the associated variables/services one by one. Since variable/service metadata is in JSON format, just returning the metadata to EDSC should be enough. I think we have talked to EDSC on this during PI planning, but it is worth talking to them again to make sure. If EDSC prefers the UMM JSON representation, /search/concepts/XXX also supports .umm_json suffix and has been implemented for all concept types except service. We can update CMR-4479 accordingly.

ygliuvt:
We should update the writeup and examples to remove the reference to variable/service search. e.g.

https://CMR-HOST/search/services?concept_id=XXX

oubiwann:
Updated the description to remove variables/services from examples and list of supported stubs.

mightynimble:
Thank you @oubiwann and @ygliuvt. The curl commands are pretty handy to figure out what we need. Now to answer your questions:
1. We would like .json response from ```/search/collections```
2. Yes, I think you have covered all.

oubiwann:
@mightynimble & @mreese84 - I've updated the stubs to support plain JSON requests on the collections (both sending a header as well as using a file extension). Note that using `application/json` changes the data structure, and thus the `jq` command for extracting the associations has changed in the instructions.

mreese84:
Looks good to me.

oubiwann:
Thanks, all!

PR #300, opened 2017-10-09
CMR-4499: Bin community usage for relevancy sorting.

chris-durbin:
This ticket was written to attempt to improve on search relevancy.

The algorithm is to currently compare scores in the following order (we only go to the next level
in the scoring system in the case of a tie at the higher level):
   1. keyword boost
   2. temporal overlap
   3. **community usage**
   4. temporal end date of the collection
   5. processing level

We've updated community usage to consider collections with a similar number of accesses to be considered the same so that those collections will move on to the next tiebreaker (temporal end date for the collection). We used a bin size of 400 meaning anything from (0-199 will be considered equal then 200-599, etc). This configuration resulted in improving our provider anomaly test cases from 14 out of 34 passing to 18 out of 34 passing.

We can control the bin size via a configuration parameter (setting the value to 1.0 will effectively eliminate the binning).


search-app/resources/bin_community_usage.groovy
```diff
@@ -0,0 +1,7 @@
+/* Elastic groovy script to round the community usage to the nearest binSize.
+   Round to 2 decimal places to avoid slight differences from math, which would defeat
+   the purpose of the script.
+*/
+
+def binnedScore = binSize * (double)Math.round(doc['usage-relevancy-score'].value / binSize);
```
oubiwann: Dude, you had me there for a millisecond ... slow-motion table-flip stopped mid-flip when I saw "groovy" ;-)

PR #304, opened 2017-10-12
CMR-4457: Dockerize the CMR

mschmele:
Added Dockerfiles, Procfiles, and a few simple scripts to run the CMR as a dev-system, or as individual uberjars.

### DIFF
--- a/access-control-app/Dockerfile
+++ b/access-control-app/Dockerfile
@@ -1,0 +1,9 @@
+FROM centos:7
+
+RUN yum install -y java-1.8.0-openjdk
+
+COPY target/cmr-access-control-app-0.1.0-SNAPSHOT-standalone.jar /app/
+
+WORKDIR /app
+
+CMD "java" "-classpath" "/app/cmr-access-control-app-0.1.0-SNAPSHOT-standalone.jar" "cmr.access-control.runner"
--- a/access-control-app/Dockerfile
+++ b/access-control-app/Dockerfile
@@ -1,13 +1,13 @@
 FROM centos:7
 
 # Install Java and clean up.
-#RUN curl -LO --cookie 'oraclelicense=accept-securebackup-cookie' 'http://download.oracle.com/otn-pub/java/jdk/7u67-b01/jdk-7u67-linux-x64.rpm' \
 RUN curl -LO --cookie 'oraclelicense=accept-securebackup-cookie' 'http://download.oracle.com/otn-pub/java/jdk/8u45-b14/jdk-8u45-linux-x64.rpm' \
  && rpm -Uvh jdk-8u45-linux-x64.rpm \
  && rm jdk-8u45-linux-x64.rpm
 ENV JAVA_HOME /usr/java/jdk1.8.0_45/jre
 
 COPY target/cmr-access-control-app-0.1.0-SNAPSHOT-standalone.jar /app/
+EXPOSE 3011
 
 WORKDIR /app
 
--- a/access-control-app/Dockerfile
+++ b/access-control-app/Dockerfile
@@ -1,9 +1,14 @@
 FROM centos:7
 
-RUN yum install -y java-1.8.0-openjdk
+# Install Java and clean up.
+#RUN curl -LO --cookie 'oraclelicense=accept-securebackup-cookie' 'http://download.oracle.com/otn-pub/java/jdk/7u67-b01/jdk-7u67-linux-x64.rpm' \
+RUN curl -LO --cookie 'oraclelicense=accept-securebackup-cookie' 'http://download.oracle.com/otn-pub/java/jdk/8u45-b14/jdk-8u45-linux-x64.rpm' \
+ && rpm -Uvh jdk-8u45-linux-x64.rpm \
+ && rm jdk-8u45-linux-x64.rpm
+ENV JAVA_HOME /usr/java/jdk1.8.0_45/jre
 
 COPY target/cmr-access-control-app-0.1.0-SNAPSHOT-standalone.jar /app/
 
 WORKDIR /app
 
-CMD "java" "-classpath" "/app/cmr-access-control-app-0.1.0-SNAPSHOT-standalone.jar" "cmr.access-control.runner"
+CMD ["java", "-classpath", "/app/cmr-access-control-app-0.1.0-SNAPSHOT-standalone.jar", "cmr.access-control.runner"]
--- a/access-control-app/Procfile
+++ b/access-control-app/Procfile
@@ -1,1 +1,0 @@
-web: java -classpath /app/cmr-access-control-app-0.1.0-SNAPSHOT-standalone.jar cmr.access-control-app.runner
--- a/access-control-app/Procfile
+++ b/access-control-app/Procfile
@@ -1,0 +1,1 @@
+web: java -classpath /app/cmr-access-control-app-0.1.0-SNAPSHOT-standalone.jar cmr.access-control-app.runner
--- a/access-control-app/build-access-control-for-container.sh
+++ b/access-control-app/build-access-control-for-container.sh
@@ -1,4 +1,0 @@
-#!/bin/bash
-lein uberjar
-docker build -t access-control-app .
-docker run access-control-app
--- a/access-control-app/build-access-control-for-container.sh
+++ b/access-control-app/build-access-control-for-container.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+lein uberjar
+docker build -t access-control-app .
+docker run access-control-app
--- a/bootstrap-app/Dockerfile
+++ b/bootstrap-app/Dockerfile
@@ -1,6 +1,11 @@
 FROM centos:7
 
-RUN yum install -y java-1.8.0-openjdk
+# Install Java and clean up.
+#RUN curl -LO --cookie 'oraclelicense=accept-securebackup-cookie' 'http://download.oracle.com/otn-pub/java/jdk/7u67-b01/jdk-7u67-linux-x64.rpm' \
+RUN curl -LO --cookie 'oraclelicense=accept-securebackup-cookie' 'http://download.oracle.com/otn-pub/java/jdk/8u45-b14/jdk-8u45-linux-x64.rpm' \
+ && rpm -Uvh jdk-8u45-linux-x64.rpm \
+ && rm jdk-8u45-linux-x64.rpm
+ENV JAVA_HOME /usr/java/jdk1.8.0_45/jre
 
 COPY . target/cmr-bootstrap-app-0.1.0-SNAPSHOT-standalone.jar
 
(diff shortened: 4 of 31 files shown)


access-control-app/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM centos:7
+
+# Install Java and clean up.
+#RUN curl -LO --cookie 'oraclelicense=accept-securebackup-cookie' 'http://download.oracle.com/otn-pub/java/jdk/7u67-b01/jdk-7u67-linux-x64.rpm' \
```
chris-durbin: Remove this commented out line from all of the Dockerfiles.

access-control-app/Procfile
```diff
@@ -0,0 +1 @@
+web: java -classpath /app/cmr-access-control-app-0.1.0-SNAPSHOT-standalone.jar cmr.access-control-app.runner
```
chris-durbin: Our Procfile is currently:

```web: java \$JVM_OPTS -cp target/${app_filename} clojure.main -m cmr.${app_name}.runner```

I'm thinking we probably want to keep using the same Procfile when we deploy to NGAP from within Bamboo. I wouldn't put these individual Procfiles in yet, let's just commit the Dockerfiles.

File: access-control-app/build-access-control-for-container.sh [hunk 3]
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+lein uberjar
+docker build -t access-control-app .
+docker run access-control-app
```
chris-durbin: This is a build and **run** script. When would we use these shell scripts? I think some documentation in the README files for starting up the application using Docker is probably good enough for now.

access-control-app/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+RUN curl -LO --cookie 'oraclelicense=accept-securebackup-cookie' 'http://download.oracle.com/otn-pub/java/jdk/8u45-b14/jdk-8u45-linux-x64.rpm' \
+ && rpm -Uvh jdk-8u45-linux-x64.rpm \
+ && rm jdk-8u45-linux-x64.rpm
+ENV JAVA_HOME /usr/java/jdk1.8.0_45/jre
+
+COPY target/cmr-access-control-app-0.1.0-SNAPSHOT-standalone.jar /app/
+
```
chris-durbin: I was expecting to see an EXPOSE 3011 for the access-control port. How are we hitting the access control application running locally without exposing the port?

access-control-app/build-access-control-for-container.sh
chris-durbin: Don't we need a `-p 3011:3011` to make port 3011 from the container available at port 3011 on localhost?

dev-system/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM centos:7
+
+RUN yum install -y java-1.8.0-openjdk
```
chris-durbin: We should be using Oracle Java here as well.

dev-system/support/build-for-container.sh
```diff
@@ -0,0 +1,39 @@
...
+    cd ../ingest-app
+    sh build-ingest-for-container.sh &
+
+    cd ../dev-system
+fi
+
+if [[ $1 == "together" || -z $1 ]]; then
+  lein uberjar
+  docker build -t dev-system .
+  docker run -p 2999:2999 -p 3001:3001 -p 3002:3002 -p 3003:3003 -p 3004:3004 -p 3005:3005 -p 3006:3006 -p 3007:3007 -p 3008:3008 -p 3009:3009 -p 3010:3010 -p 3011:3011 -p 9210:9210 dev-system
```
chris-durbin: This doesn't start the container up in daemon mode (-d), so when you call exit 0 on the next line, doesn't it just shut down the container immediately?

dev-system/support/build-for-container.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
```
chris-durbin: Add a comment with a description of the script and any commandline variables it accepts.

access-control-app/Dockerfile
mschmele: Done

access-control-app/Dockerfile
mschmele: Done

dev-system/Dockerfile
mschmele: Done

dev-system/support/build-for-container.sh
mschmele: It doesn't, but I still think starting it in daemon mode is a good idea

bootstrap-app/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM centos:7
+
+# Install Java and clean up.
+RUN curl -LO --cookie 'oraclelicense=accept-securebackup-cookie' 'http://download.oracle.com/otn-pub/java/jdk/8u45-b14/jdk-8u45-linux-x64.rpm' \
+ && rpm -Uvh jdk-8u45-linux-x64.rpm \
+ && rm jdk-8u45-linux-x64.rpm
+ENV JAVA_HOME /usr/java/jdk1.8.0_45/jre
+
+COPY . target/cmr-bootstrap-app-0.1.0-SNAPSHOT-standalone.jar
+
```
chris-durbin: Only access-control was updated to add the port. We should update all of them to expose the default port. You can get the port numbers in `cmr.transmit.config`

dev-system/support/build-for-container.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+# Build and run the CMR as a single dev-system or as a set of microservices
+# Parameters:
+#  - separate: build and run each microservice separately
+#  - together: build and run dev-system in a container
+#  - default: pass no parameters, and the script will build and run dev-system
+
+build-and-run-container () {
```
chris-durbin: This should take one more argument - port number.

Then when you run you would say something like `docker run -d $1 -p $2:$2

dev-system/support/build-for-container.sh
```diff
@@ -0,0 +1,50 @@
...
+  lein uberjar
+  docker build -t $ .
```
chris-durbin: Doesn't this $ need to be $1?

dev-system/support/build-for-container.sh
mschmele: I thought `EXPOSE` statements in the Dockerfile did that for you, but I've updated that anyway for good measure

bootstrap-app/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM centos:7
+
+# Install Java and clean up.
+RUN curl -LO --cookie 'oraclelicense=accept-securebackup-cookie' 'http://download.oracle.com/otn-pub/java/jdk/8u45-b14/jdk-8u45-linux-x64.rpm' \
+ && rpm -Uvh jdk-8u45-linux-x64.rpm \
+ && rm jdk-8u45-linux-x64.rpm
+ENV JAVA_HOME /usr/java/jdk1.8.0_45/jre
+
+COPY . target/cmr-bootstrap-app-0.1.0-SNAPSHOT-standalone.jar
```
chris-durbin: This copy command should be copying the jar to /app/ like the other Dockerfiles

mschmele: Good catch - fixed

dev-system/support/build-for-container.sh
chris-durbin: From what I saw EXPOSE will force those ports to be mapped to a random port number on your localhost. (somewhere in the 3xxxx range). I had to explicitly set the ports I wanted mapped to my localhost port in order to make it work.

index-set-app/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM centos:7
+
+# Install Java and clean up.
+RUN curl -LO --cookie 'oraclelicense=accept-securebackup-cookie' 'http://download.oracle.com/otn-pub/java/jdk/8u45-b14/jdk-8u45-linux-x64.rpm' \
+ && rpm -Uvh jdk-8u45-linux-x64.rpm \
+ && rm jdk-8u45-linux-x64.rpm
+ENV JAVA_HOME /usr/java/jdk1.8.0_45/jre
+
+COPY . target/cmr-index-set-app-0.1.0-SNAPSHOT-standalone.jar
```
chris-durbin: Fix COPY.

indexer-app/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM centos:7
+
+# Install Java and clean up.
+RUN curl -LO --cookie 'oraclelicense=accept-securebackup-cookie' 'http://download.oracle.com/otn-pub/java/jdk/8u45-b14/jdk-8u45-linux-x64.rpm' \
+ && rpm -Uvh jdk-8u45-linux-x64.rpm \
+ && rm jdk-8u45-linux-x64.rpm
+ENV JAVA_HOME /usr/java/jdk1.8.0_45/jre
+
+COPY . target/cmr-indexer-app-0.1.0-SNAPSHOT-standalone.jar
```
chris-durbin: Fix COPY.

ingest-app/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM centos:7
+
+# Install Java and clean up.
+RUN curl -LO --cookie 'oraclelicense=accept-securebackup-cookie' 'http://download.oracle.com/otn-pub/java/jdk/8u45-b14/jdk-8u45-linux-x64.rpm' \
+ && rpm -Uvh jdk-8u45-linux-x64.rpm \
+ && rm jdk-8u45-linux-x64.rpm
+ENV JAVA_HOME /usr/java/jdk1.8.0_45/jre
+
+COPY . target/cmr-ingest-app-0.1.0-SNAPSHOT-standalone.jar
```
chris-durbin: Fix COPY.

virtual-product-app/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM centos:7
+
+# Install Java and clean up.
+RUN curl -LO --cookie 'oraclelicense=accept-securebackup-cookie' 'http://download.oracle.com/otn-pub/java/jdk/8u45-b14/jdk-8u45-linux-x64.rpm' \
+ && rpm -Uvh jdk-8u45-linux-x64.rpm \
+ && rm jdk-8u45-linux-x64.rpm
+ENV JAVA_HOME /usr/java/jdk1.8.0_45/jre
+
+COPY . target/cmr-virtual-product-app-0.1.0-SNAPSHOT-standalone.jar
```
chris-durbin: Fix COPY.

dev-system/support/build-for-container.sh
jonvel: Not exactly - that just makes it so that you _can_ expose the port to the outside world.  It works in conjunction with the, say `-P` flag in `docker run` or `-p <exposed port>` or `-p <exposed port>:<host opened port mapped>`

PR #314, opened 2017-10-17
CMR-4283

ygliuvt:
As a client, I want to dissociate a service from a list of collections by concept-ids.


system-int-test/src/cmr/system_int_test/system.clj
```diff
@@ -41,7 +41,7 @@
              ;; A map of the components (echo, elastic, db, and message queue) to whether they are
              ;; in-memory or external
              :component-type-map component-type-map}]
-    (transmit-config/system-with-connections sys [:echo-rest :search :access-control :urs :ingest])))
+    (transmit-config/system-with-connections sys [:echo-rest :search :access-control :metadata-db :urs :ingest])))
```
chris-durbin: I had a comment for Erich in his logging PR that we shouldn't really use the transmit library for our system integration tests. Someone could change an endpoint and no tests would break to let us know we broke backwards compatibility. Rather we should use black box HTTP calls (using the url-helper namespace) for our integration tests.

What do you think about making http requests rather than using the `cmr.transmit.association` and `cmr.transmit.services` namespaces?

oubiwann: This looks like the sort of change that might have taken a while to figure out why something wasn't working ... :-/

transmit-lib/src/cmr/transmit/association.clj
```diff
@@ -0,0 +1,75 @@
+(ns cmr.transmit.association
```
oubiwann: Nice. This let you get ride of a bunch of other code.

ygliuvt:
> Looks good. The Royal Academy of Refactoring approves.

Thanks! :)

system-int-test/test/cmr/system_int_test/search/service/service_association_test.clj
```diff
@@ -144,10 +144,221 @@
...
+    (index/wait-until-indexed)
+    ;; Associate the service with every collection
+    (au/associate-by-concept-ids
+     prov3-token
+     concept-id
+     (map #(hash-map :concept-id (:concept-id %)) all-colls))
+
+    (testing "Successfully dissociate service with collections"
+      (let [{:keys [status] :as resp} (au/dissociate-by-concept-ids
+                              token
```
chris-durbin: indentation

system-int-test/src/cmr/system_int_test/system.clj
ygliuvt: I agree. This is a temporary fix to use metadata-db rather than search-app for the verification. We will change it back to search when we added the search by service capability in search (probably in two weeks). I say we leave it as is for now. The existing search function is using clj-http directly. BTW, this is documented in the docstring of `cmr.system-int-test.utils.service-util/assert-service-associated-with-query`.

oubiwann: I smell a CMR client coming ... ;-)

Hey, so we should probably put a comment here linking to a ticket if this is going to be temporary ...

ygliuvt: OK. Changed the function to use clj-http directly and referenced CMR-4280 in comment.

system-int-test/test/cmr/system_int_test/search/service/service_association_test.clj
ygliuvt: Done.

system-int-test/src/cmr/system_int_test/utils/association_util.clj
```diff
@@ -17,7 +17,7 @@
    (associate-by-concept-ids token concept-id coll-concept-ids nil))
   ([token concept-id coll-concept-ids options]
    (let [options (merge {:raw? true :token token} options)
-         response (transmit-variable/associate-variable
+         response (transmit-assoc/associate-concept
```
ygliuvt: Notice this is still calling a transmit lib function, but this is only used in tests and I think it makes sense here as after all, transmit lib provides helper functions to make http request calls easy.

PR #318, opened 2017-10-19
CMR-4412 changes bulk update validation to prevent needless failures

daniel-zamora:
When bulk update is used on collections whose native format is not umm, sometimes the translation from XML->UMM causes schema validation failures, which prevents the bulk update from succeeding.
This PR prevents that schema validation from happening when bulk-update is used.


system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
```diff
@@ -363,3 +364,23 @@
                 :VariableLevel1 "EMISSIONS",
                 :VariableLevel2 "CLOUD LIQUID WATER/ICE"}]
               (:ScienceKeywords (:umm concept)))))))
+
+(deftest bulk-update-xml-to-umm-failure-test
+  (let [coll-metadata (slurp (io/resource "iso-samples/cmr-4125-iso-collection.xml"))
```
daniel-zamora: fixing test to use correct file.

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
```diff
@@ -141,9 +141,9 @@
       (let [updated-concept (update-collection-concept context concept bulk-update-params)
             warnings (validate-and-save-collection context updated-concept)]
         (data-bulk-update/update-bulk-update-task-collection-status context task-id
```
chris-durbin: Indentation nit:
`context` should be on the next line indented one space. If there end up being multiple lines of parameters they should line up at the same level as `context`.

PR #319, opened 2017-10-19
CMR-4480

ygliuvt:
As a Client User, I want the collection search JSON response to include has_variables field.

### DIFF
--- a/search-app/docs/api.md
+++ b/search-app/docs/api.md
@@ -778,7 +778,7 @@ __Example__
       "id" : "C1200000000-PROV1",
       "original_format" : "ECHO10",
       "browse_flag" : false,
-      "has-variables" : false,
+      "has_variables" : false,
       "online_access_flag" : false,
       "tags" : {"tag1": {"data": {"score": 85, "status": "reviewed"}},
                 "tag2": {"data" : "cloud cover > 80"}}
--- a/search-app/docs/api.md
+++ b/search-app/docs/api.md
@@ -342,6 +342,7 @@ The following fields are specific to the CMR output and most correspond to ECHO1
 | echo:browseFlag                            | true if the data contains browse imagery                                                                             |
 | echo:hasGranules (collections only)        | true if there are granules associated with the collection                                                            |
 | echo:granuleCount (collections only)       | granule count of the collection                                                                                      |
+| echo:hasVariables (collections only)       | true if there are variables associated with the collection                                                           |
 | relevance:score (collections only)         | relevance score of the collection to search parameters                                                               |
 | echo:tag (collections only)                | tags associated with the collection. It includes sub-elements of tagKey and optional data which is in embedded JSON. |
 | echo:dayNightFlag (granules only)          | day night flag of the granule                                                                                        |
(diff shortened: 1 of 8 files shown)


search-app/docs/api.md
```diff
@@ -777,6 +778,7 @@ __Example__
       "id" : "C1200000000-PROV1",
       "original_format" : "ECHO10",
       "browse_flag" : false,
+      "has-variables" : false,
```
oubiwann: All the others use underscores ... but the new `has-variables` uses a dash. Do you want underscore here?

The way I understood it (and the way the stubs were provided to EDSC): UMM-JSON has dashes, and regular JSON results have underscores ...

search-app/src/cmr/search/results_handlers/atom_json_results_handler.clj
```diff
@@ -73,6 +75,7 @@
                        :has_granules (when has-granules-map (or (< 0 granule-count)
                                                                 (get has-granules-map id false)))
                        :granule_count (when granule-counts-map granule-count)
+                       :has-variables has-variables
```
oubiwann: Again, not sure of the reason for the difference between under and dash here?

chris-durbin: Historically ECHO used underscores in all APIs. When transitioning to CMR a decision was made that all APIs should continue to use underscores rather than dashes.

search-app/docs/api.md
ygliuvt: Good catch. Done.

search-app/src/cmr/search/results_handlers/atom_json_results_handler.clj
ygliuvt: Good catch. Updated.

search-app/src/cmr/search/results_handlers/atom_results_handler.clj
```diff
@@ -427,6 +429,7 @@
                                                      (get has-granules-map id false))))
                (when granule-counts-map
                  (x/element :echo:granuleCount {} granule-count))
+               (x/element :echo:hasVariables {} has-variables)
```
chris-durbin: It looks like this is different than has-granules and granule-counts. For those we don't always return the field. I think this is controlled by a parameter `include_has_granules` and `include_granule_counts` (I might have the wrong names).

I'm not convinced that we need an include_has_variables flag, but it is not consistent.

ygliuvt: The short answer is we should not have added the `include_has_granules` parameter in our api doc. I guess the reason we added that parameter is we have thought that include has granules would be an expensive operation and we don't want to do it for all searches unless it is requested, but we have since introduced caching to has_granules and that makes the operation no longer expensive and unnecessary to have the `include_has_granules` parameter.

`has_variables` is totally different, it is directly indexed into the collection elasticsearch index and it is as simple as any other regular granule search field. There is definitely no reason to introduce an `include_has_variables` parameter.

oubiwann: I agree with @ygliuvt's assessment, and would like to propose something a bit further: the planned deprecation of the `include_has_*` (i.e., identifying when we want to deprecate them, updating the docs in that version to let users know it's going away, and identifying the version where they will actually be removed).

PR #320, opened 2017-10-20
CMR-4479 Add support for concept-id service lookup

oubiwann:
Summary:

This adds the service concept as a valid type when querying the `/search/concepts` CMR resource.

Previous comment, during the WIP PR for this branch:

@ygliuvt I'm getting a null exception when running the new tests against the `/search/concepts` resource with a service concept it. I haven't been able to track down where the error is coming from. I'm wondering if something is missing from our service work for this to be supported? (i.e., another ticket) or if there's some part of the CMR that needs to be updated that produces this sort of side effect (e.g., an ACL update somewhere ... I seem to remember we did something like that with UMM-Vars ...).

### DIFF
--- a/search-app/src/cmr/search/services/acl_service.clj
+++ b/search-app/src/cmr/search/services/acl_service.clj
@@ -13,18 +13,18 @@
   (fn [context acls concept]
     (:concept-type concept)))
 
+;; services currently have no ACLs, so return `true` for all ACL checks
+(defmethod acls-match-concept? :service
+  [context acls concept]
+  true)
+
 ;; tags have no acls so we always assume it matches
 (defmethod acls-match-concept? :tag
   [context acls concept]
   true)
 
-;; When plans solidify around acl checks for variable concepts, we'll update this.
+;; variables currently have no ACLs, so return `true` for all ACL checks
 (defmethod acls-match-concept? :variable
-  [context acls concept]
-  true)
-
-;; When plans solidify around acl checks for service concepts, we'll update this.
-(defmethod acls-match-concept? :service
   [context acls concept]
   true)
 
@@ -32,10 +32,16 @@
   [context acls concept]
   false)
 
-;; XXX This code is a point of contention for adding new concepts: on both
+;; XXX To be fixed with CMR-4394
+;;
+;;     See also:
+;;       * https://wiki.earthdata.nasa.gov/display/CMR/Towards+Improved+CMR+Concept+Creation
+;;       * https://wiki.earthdata.nasa.gov/display/CMR/CMR+Concept+Creation+API
+;;
+;;     This code is a point of contention for adding new concepts: on both
 ;;     occasions where we added varible and service support (e.g. to the
 ;;     `/search/concepts` route), tests failed with ACL errors due to this ns
-;;     not getting updated both in the function below as well as above in the
+;;     not getting updated, both in the function below as well as above in the
 ;;     `acls-match-concept?` multimethod.
 ;;
 ;;     This needs to be taken into account when we plan for refactoring the
@@ -45,7 +51,7 @@
 ;;     programming by conventions is a slow path to developer insanity and
 ;;     software project failure! Instead, all parts of the code that need to be
 ;;     touched should be replaced with APIs (e.g., protocols and their impl'ns)
-;;     with each method documented and, all toegher, being the only things a
+;;     with each method documented and, all together, being the only things a
 ;;     developer needs to add/change (in this instance) a new concept.
 (def concept-type->applicable-field
   "A mapping of concept type to the field in the ACL indicating if it is collection or granule
--- a/search-app/src/cmr/search/services/acl_service.clj
+++ b/search-app/src/cmr/search/services/acl_service.clj
@@ -23,15 +23,36 @@
   [context acls concept]
   true)
 
+;; When plans solidify around acl checks for service concepts, we'll update this.
+(defmethod acls-match-concept? :service
+  [context acls concept]
+  true)
+
 (defmethod acls-match-concept? :default
   [context acls concept]
   false)
 
+;; XXX This code is a point of contention for adding new concepts: on both
+;;     occasions where we added varible and service support (e.g. to the
+;;     `/search/concepts` route), tests failed with ACL errors due to this ns
+;;     not getting updated both in the function below as well as above in the
+;;     `acls-match-concept?` multimethod.
+;;
+;;     This needs to be taken into account when we plan for refactoring the
+;;     APIs and development process for adding concepts to the CMR. The fact
+;;     that this crucial change is tucked away in this part of the code with no
+;;     logical "pointers" or hints to it anywhere else is deeply problematic ...
+;;     programming by conventions is a slow path to developer insanity and
+;;     software project failure! Instead, all parts of the code that need to be
+;;     touched should be replaced with APIs (e.g., protocols and their impl'ns)
+;;     with each method documented and, all toegher, being the only things a
+;;     developer needs to add/change (in this instance) a new concept.
 (def concept-type->applicable-field
   "A mapping of concept type to the field in the ACL indicating if it is collection or granule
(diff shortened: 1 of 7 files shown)


search-app/src/cmr/search/services/acl_service.clj
```diff
@@ -23,15 +23,36 @@
...
+;; When plans solidify around acl checks for service concepts, we'll update this.
+(defmethod acls-match-concept? :service
+  [context acls concept]
+  true)
+
 (defmethod acls-match-concept? :default
   [context acls concept]
   false)
 
+;; XXX This code is a point of contention for adding new concepts: on both
```
oubiwann: This is the second time this has bitten us. We need to incorporate this into our future plans for refactoring the code for adding new concepts to the CMR.

system-int-test/test/cmr/system_int_test/search/service/concept_retrieval_test.clj
```diff
@@ -0,0 +1,180 @@
...
+          (testing "retrieval by service concept-id returns the latest revision."
+            (is (= svc1-name-new
+                   (:Name (json/parse-string (:body response) true)))))
+          (testing (str "retrieval by service concept-id and revision-id returns "
+                    "the specified service")
+            (is (= svc1-name
+                   (:Name (json/parse-string (:body response-v1) true))))
+            (is (= svc1-name-new
+                   (:Name (json/parse-string (:body response-v2) true)))))))
+    (testing "retrieval by service concept-id and incorrect revision-id returns error"
```
ygliuvt: Combine this test case and the next one into one with `are3`.

system-int-test/test/cmr/system_int_test/search/service/concept_retrieval_test.clj
```diff
@@ -0,0 +1,180 @@
...
+            response-v1 (search/retrieve-concept
+                         (:concept-id svc1-v1)
+                         1
+                         {:accept mt/any})
+            response-v2 (search/retrieve-concept
+                         (:concept-id svc1-v1)
+                         2
+                         {:accept mt/any})]
+          (testing "retrieval by service concept-id returns the latest revision."
+            (is (= svc1-name-new
```
ygliuvt: How about creating a helper function (e.g. `verify-retrieved-concept`) to do the whole retrieving concept, parsing JSON and comparing service name in just one function?

search-app/src/cmr/search/services/acl_service.clj
```diff
@@ -23,15 +23,36 @@
   [context acls concept]
   true)
 
+;; When plans solidify around acl checks for service concepts, we'll update this.
```
chris-durbin: Needs a ticket number. If we don't have a ticket number for ACLs for services we need to write one.

search-app/src/cmr/search/services/acl_service.clj
```diff
@@ -23,15 +23,36 @@
...
+;;     `acls-match-concept?` multimethod.
+;;
+;;     This needs to be taken into account when we plan for refactoring the
+;;     APIs and development process for adding concepts to the CMR. The fact
+;;     that this crucial change is tucked away in this part of the code with no
+;;     logical "pointers" or hints to it anywhere else is deeply problematic ...
+;;     programming by conventions is a slow path to developer insanity and
+;;     software project failure! Instead, all parts of the code that need to be
+;;     touched should be replaced with APIs (e.g., protocols and their impl'ns)
+;;     with each method documented and, all toegher, being the only things a
```
chris-durbin: Define _toegher_: Adjective, tougher and all together.

system-int-test/src/cmr/system_int_test/utils/service_util.clj
```diff
@@ -18,6 +18,10 @@
 (def default-opts {:accept-format :json
                    :content-type content-type})
 
+(defn token-opts
+  [token]
```
chris-durbin: Add docstring

system-int-test/test/cmr/system_int_test/search/service/concept_retrieval_test.clj
```diff
@@ -0,0 +1,180 @@
...
+  [(ingest/reset-fixture {"provguid1" "PROV1"})
+   service/grant-all-service-fixture]))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;
+;; Helper utility functions
+;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(defn- get-svc
```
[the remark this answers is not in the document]
oubiwann: Both this one and the next one were created (copied here; created in the variable concept retrieval test) because the `let` blocks that contained them were getting unruly: long, nested calls inside them, side-effect operations that returned no useful value (causing us to use the `[_ (call)]` anti-pattern), etc.

But I simultaneously harbour objections similar in nature to yours: does this mean that we then have little one-off functions scattered around our testing codebase whose sole purpose it to help us keep clean code? This would indicate that a symptom was addressed, not a root problem.

Which begs the question: what are we not taking into consideration with our APIs such that its usage is so awkward? A good API has very little of these usage issues; an excellent API has none of them.

In the mean time, I'll take another look and see if there's a nicer way to clean these up here...

search-app/src/cmr/search/services/acl_service.clj
oubiwann: Yeah, Liu and I looked at this comment this morning (it's a copy-paste) and where were wondering where it came from and what it might be referring to. I think I may have actually written the original in the variables version of this sys-int test ... I'll dig and try to remember, and then update the comment.

search-app/src/cmr/search/services/acl_service.clj
oubiwann: I'll also reference the epic ticket here ...

oubiwann: Done.

system-int-test/test/cmr/system_int_test/search/service/concept_retrieval_test.clj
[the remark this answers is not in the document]
ygliuvt: I find these helper functions make the test hard to follow. I don't want my `get-svc` call to actually ingest a service concept with the given service-name and returns the ingested service concept id and revision id. I would rather see something like `ingest-service-with-attrs` which at least tells me a service concept is ingested and that info is more important to understand the test.

system-int-test/src/cmr/system_int_test/utils/service_util.clj
oubiwann: Added.

search-app/src/cmr/search/services/acl_service.clj
oubiwann: I'm having vague memories of the comment origins; I think I made them after a pairing session with Liu (the first time we ran into the testing issues as a result of this code not being updated when variable concepts were added). I think that comment came from me poorly paraphrasing something Liu said on the call, so I've updated it to essentially say the same thing that the tag comment says: basically, there's no ACLs for variables and services right now.

In other words, done ;-)

system-int-test/test/cmr/system_int_test/search/service/concept_retrieval_test.clj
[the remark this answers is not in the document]
oubiwann: Yeah, that's a good point; those helper function names hide a benefit of the long names: self-documentation.  

We might be able to address my other concerns in the sys-int utility namespaces for now (and then, ultimately, a well-designed client library ... or utility library built on the client library).

PR #327, opened 2017-10-25
CMR-4455 As a client, I would like to specify and retrieve a name identifier for bulk updates

daniel-zamora:
This PR adds a name column to the bulk update status table.  That name is displayed for each task-id when status is queried.

Also, any task already in the database is migrated to have the name be the task-id.

### DIFF
--- a/ingest-app/docs/api.md
+++ b/ingest-app/docs/api.md
@@ -641,7 +641,7 @@ The following update types are supported:
 Bulk update post request takes the following parameters:
 
   * Concept-ids (required) - a list of concept ids to update
-  * Name (required) - a name used to identify a update task 
+  * Name (optional) - a name used to identify a bulk update task 
   * Update type (required) - choose from the enumeration: `ADD_TO_EXISTING`, `CLEAR_ALL_AND_REPLACE`, `FIND_AND_REPLACE`, `FIND_AND_REMOVE`, `FIND_AND_UPDATE`
   * Update field (required) - choose from the enumeration: `SCIENCE_KEYWORDS`, `LOCATION_KEYWORDS`, `DATA_CENTERS`, `PLATFORMS`, `INSTRUMENTS`
   * Update value (required for all update types except for `FIND_AND_REMOVE`) - UMM-JSON representation of the update to make
@@ -667,7 +667,8 @@ Example: Initiate a bulk update of 3 collections. Find platforms that have Type 
 ```
 curl -i -XPOST -H "Cmr-Pretty:true" -H "Content-Type: application/json" -H "Echo-Token: XXXX" %CMR-ENDPOINT%/providers/PROV1/bulk-update/collections -d
 '{"concept-ids": ["C1200000005-PROV1","C1200000006-PROV1","C1200000007-PROV1"],
-  "name": "TEST NAME"
+  "name": "TEST NAME",
+  "create-at": "2017-10-24T17:00:03.000Z",
   "update-type": "FIND_AND_UPDATE",
   "update-field": "PLATFORMS",
   "find-value": {"Type": "Aircraft"},
--- a/ingest-app/docs/api.md
+++ b/ingest-app/docs/api.md
@@ -668,7 +668,6 @@ Example: Initiate a bulk update of 3 collections. Find platforms that have Type 
 curl -i -XPOST -H "Cmr-Pretty:true" -H "Content-Type: application/json" -H "Echo-Token: XXXX" %CMR-ENDPOINT%/providers/PROV1/bulk-update/collections -d
 '{"concept-ids": ["C1200000005-PROV1","C1200000006-PROV1","C1200000007-PROV1"],
   "name": "TEST NAME",
-  "create-at": "2017-10-24T17:00:03.000Z",
   "update-type": "FIND_AND_UPDATE",
   "update-field": "PLATFORMS",
   "find-value": {"Type": "Aircraft"},
--- a/ingest-app/docs/api.md
+++ b/ingest-app/docs/api.md
@@ -690,7 +690,7 @@ curl -i -XPOST -H "Cmr-Pretty:true" -H "Content-Type: application/json" -H "Echo
 
 The task ids and status of all bulk update tasks for a provider can be queried by sending an HTTP GET request to `%CMR-ENDPOINT%/providers/<provider-id>/bulk-update/collections/status`
 
-This returns a list of: task id, status (IN_PROGRESS or COMPLETE), a status message, and the original request JSON body.
+This returns a list of: created-at, name, task id, status (IN_PROGRESS or COMPLETE), a status message, and the original request JSON body.
 
 Example
 ```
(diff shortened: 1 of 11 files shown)


ingest-app/src/cmr/ingest/api/bulk.clj
```diff
@@ -38,13 +38,17 @@
...
+ ([result status-list-key status-key id-key created-at-key name-key]
+  (generate-xml-status-list result status-list-key status-key id-key created-at-key name-key nil))
+ ([result status-list-key status-key id-key created-at-key name-key additional-keys]
   (xml/element status-list-key {}
     (for [status (get result status-list-key)
           :let [message (:status-message status)]]
      (xml/element status-key {}
       (when created-at-key
         (xml/element created-at-key {} (str (:created-at status))))
+      (when name-key
```
chris-durbin: I think we will always want to return a name. If there is no name we should return some generated default.

ingest-app/src/cmr/ingest/data/bulk_update.clj
```diff
@@ -115,34 +117,34 @@
...
-                                            [:concept-id :status]
-                                            (su/from "bulk_update_coll_status")
-                                            (su/where `(= :task-id ~task-id)))))
-               pending-collections (filter #(= "PENDING" (:status %)) task-collections)
-               failed-collections (filter #(= "FAILED" (:status %)) task-collections)]
-           (when-not (seq pending-collections)
-             (update-bulk-update-task-status db task-id "COMPLETE"
-               (generate-task-status-message
-                 (count failed-collections) (count task-collections)))))))
+        [conn db]
```
chris-durbin: This is one place the original indentation matched our style guide (all your other changes to indentation looked good). If there are no arguments on the function call line we indent the first argument a single space on the next line.

https://github.com/bbatsov/clojure-style-guide#one-space-indent

ingest-app/src/migrations/008_add_name_column_bulk_update_status.clj
```diff
@@ -0,0 +1,22 @@
...
+(defn up
+  "Migrates the database up to version 8."
+  []
+  (println "migrations.008-add-name-column-bulk-update-status up...")
+  (j/db-do-commands (config/db)
+                    "ALTER TABLE CMR_INGEST.bulk_update_task_status ADD
+                     name VARCHAR(255) DEFAULT '' NOT NULL")
+  (j/db-do-commands (config/db)
+                    "UPDATE table SET name = task_id
+                     WHERE name = ''"))
```
chris-durbin: We should either alter the column definition to DEFAULT to the TASK_ID or add one more alter table change to get rid of the DEFAULT value.

I think the latter is the better choice, so you'd add:
```
(j/db-do-commands (config/db)
                  "ALTER TABLE CMR_INGEST.bulk_update_task_status
                   MODIFY name VARCHAR(255) NOT NULL")
```
as the last step after setting all the values in the table.

system-int-test/src/cmr/system_int_test/utils/ingest_util.clj
```diff
@@ -545,6 +546,7 @@
  ([provider-id]
   (bulk-update-provider-status provider-id nil))
  ([provider-id options]
+  (println "dbg2:" )
```
chris-durbin: Remove debug

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_endpoint_test.clj
```diff
@@ -119,11 +126,20 @@
...
             {:concept-ids ["C1", "C2", "C3"]
+             :name "TEST NAME"
              :update-field "Science keywords"
              :update-type "ADD_TO_EXISTING"
              :update-value {:Category "EARTH SCIENCE"}}
             400
-            ["/update-field instance value (\"Science keywords\") not found in enum (possible values: [\"SCIENCE_KEYWORDS\",\"LOCATION_KEYWORDS\",\"DATA_CENTERS\",\"PLATFORMS\",\"INSTRUMENTS\"])"]))))
+            ["/update-field instance value (\"Science keywords\") not found in enum (possible values: [\"SCIENCE_KEYWORDS\",\"LOCATION_KEYWORDS\",\"DATA_CENTERS\",\"PLATFORMS\",\"INSTRUMENTS\"])"]
+
+            "Missing name value"
```
chris-durbin: Change this to check the value for name defaults to the task ID instead of causing an error.

ingest-app/docs/api.md
```diff
@@ -641,6 +641,7 @@ The following update types are supported:
 Bulk update post request takes the following parameters:
 
   * Concept-ids (required) - a list of concept ids to update
+  * Name (required) - a name used to identify a update task
```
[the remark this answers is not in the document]
daniel-zamora: was suppose to be bulk, hmm wonder how it got changed.

ingest-app/src/cmr/ingest/api/bulk.clj
```diff
@@ -38,13 +38,17 @@
   (generate-xml-status-list result status-list-key status-key id-key nil))
  ([result status-list-key status-key id-key created-at-key]
   (generate-xml-status-list result status-list-key status-key id-key created-at-key nil))
- ([result status-list-key status-key id-key created-at-key additional-keys]
+ ([result status-list-key status-key id-key created-at-key name-key]
+  (generate-xml-status-list result status-list-key status-key id-key created-at-key name-key nil))
+ ([result status-list-key status-key id-key created-at-key name-key additional-keys]
```
mschmele: Nitpick: I feel like the arity of this function is a little out of control. I'm not sure how we would fix it, but I think it's worth establishing how to break something like this up. I'm thinking @oubiwann might have some thoughts here

chris-durbin: I looked at the full namespace. It is a private function that is used in exactly two places. So there definitely should not be more than 2 arrities.

There is not that much code that would be repeated, so it makes sense to just break into two simple functions that just take the result as an argument.

ingest-app/src/cmr/ingest/data/bulk_update.clj
```diff
@@ -41,70 +42,71 @@
...
-      (errors/throw-service-error :invalid-data
-       [(str "Error creating creating bulk update status "
-             (.getMessage e))]))))
+      (j/with-db-transaction
+        [conn db]
+        (let [task-id (:nextval (first (su/query db ["SELECT task_id_seq.NEXTVAL FROM DUAL"])))
+              statement (str "INSERT INTO bulk_update_task_status "
+                             "(task_id, provider_id, name, request_json_body, status)"
+                             "VALUES (?, ?, ?, ?, ?)")
+              name (or (get (json/parse-string json-body) "name") task-id)
```
chris-durbin: `get` accepts an additional parameter of a default value when not found in the map. We should use that rather than or.

```
(get (json/parse-string json-body) "name" task-id)
```

ingest-app/src/migrations/008_add_name_column_bulk_update_status.clj
```diff
@@ -0,0 +1,22 @@
...
+(defn up
+  "Migrates the database up to version 8."
+  []
+  (println "migrations.008-add-name-column-bulk-update-status up...")
+  (j/db-do-commands (config/db)
+                    "ALTER TABLE bulk_update_task_status ADD
+                     name VARCHAR(255) DEFAULT NULL")
+  (j/db-do-commands (config/db)
+                    "UPDATE bulk_update_task_status SET name = task_id
+                     WHERE name IS NULL"))
```
chris-durbin: I think this comment was missed when changing the migration. We should not allow inserting NULLs after this migration has run. So we need to add a 3rd step after updating everything that had a NULL value to no longer allow NULLs:

```
(j/db-do-commands (config/db)
                  "ALTER TABLE CMR_INGEST.bulk_update_task_status
                   MODIFY name VARCHAR(255) NOT NULL")
```

ingest-app/docs/api.md
```diff
@@ -666,6 +667,8 @@ Example: Initiate a bulk update of 3 collections. Find platforms that have Type
 ```
 curl -i -XPOST -H "Cmr-Pretty:true" -H "Content-Type: application/json" -H "Echo-Token: XXXX" %CMR-ENDPOINT%/providers/PROV1/bulk-update/collections -d
 '{"concept-ids": ["C1200000005-PROV1","C1200000006-PROV1","C1200000007-PROV1"],
+  "name": "TEST NAME",
+  "create-at": "2017-10-24T17:00:03.000Z",
```
sxu123: Did you meant to add created-at?  Created-at is automatically populated in the database table. Why is it included in the request?

chris-durbin: This probably came from a (wrong) comment from me. I thought this was a JSON response, didn't really this was a request body. Agree that it should not be in the request.

PR #330, opened 2017-10-26
CMR-4523: Updated test to use functions for better reporting when there is a test failure.

chris-durbin:
I ran the test locally more than 50 times and could not reproduce the failure. I checked every build on every branch that is still in Bamboo for everyone's fork. This particular test does not have a single failure.

It looks like the test is not failing at all anymore or failing extremely infrequently. I changed the test code to provide more information if the test does fail again.

### DIFF
--- a/system-int-test/test/cmr/system_int_test/search/acls/collection.clj
+++ b/system-int-test/test/cmr/system_int_test/search/acls/collection.clj
@@ -1,12 +1,12 @@
 (ns cmr.system-int-test.search.acls.collection
   "Tests searching for collections with ACLs in place."
   (:require
-   [cmr.common-app.test.side-api :as side]
-   [clojure.test :refer :all]
    [clojure.string :as str]
+   [clojure.test :refer :all]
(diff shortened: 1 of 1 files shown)


system-int-test/test/cmr/system_int_test/search/acls/collection.clj
```diff
@@ -295,8 +305,8 @@
           (od/assert-collection-opendata-results-match guest-permitted-collections actual-od))))
 
     (testing "all_revisions"
-      (are2 [collections params]
-        (d/refs-match? collections (search/find-refs :collection params))
+      (util/are3 [collections params]
```
daniel-zamora: are we not referring are3 anymore?

chris-durbin: Our goal is to use refer rarely so it is always clear where a function comes from. We still use it for things like logging, clojure.test functions, and compojure functions.

PR #332, opened 2017-10-27
Cmr 4223 doc update

sxu123:


sxu123:
It was updated by Daniel's PR already

PR #334, opened 2017-10-31
CMR-4553 Status detail message for Bulk Update is truncated

daniel-zamora:
This PR modifies bulk update to allow for larger status messages by removing the truncation, and migrating the database column to 4000 (a size that is subject to change).  Before status messages were limited to 255 characters.

### DIFF
--- a/ingest-app/src/cmr/ingest/data/bulk_update.clj
+++ b/ingest-app/src/cmr/ingest/data/bulk_update.clj
@@ -128,7 +128,7 @@
        (let [statement (str "UPDATE bulk_update_coll_status "
                             "SET status = ?, status_message = ?"
                             "WHERE task_id = ? AND concept_id = ?")
-             status-message (util/trunc status-message 255)]
+             status-message (util/trunc status-message 4000)]
          (j/db-do-prepared db statement [status status-message task-id concept-id])
          (let [task-collections (su/query db
                                           (su/build (su/select
--- a/ingest-app/src/migrations/009_update_bulk_status_message.clj
+++ b/ingest-app/src/migrations/009_update_bulk_status_message.clj
@@ -8,8 +8,8 @@
   []
   (println "migrations.009-update-bulk-status-message up...")
   (j/db-do-commands (config/db)
-                    "ALTER TABLE bulk_update_task_status MODIFY
-                     (status_message VARCHAR(1030))"))
+                    "ALTER TABLE bulk_update_coll_status MODIFY
+                     (status_message VARCHAR(4000))"))
 
 (defn down
   "Migrates the database down to version 8."
--- a/ingest-app/src/migrations/009_update_bulk_status_message.clj
+++ b/ingest-app/src/migrations/009_update_bulk_status_message.clj
@@ -9,7 +9,7 @@
   (println "migrations.009-update-bulk-status-message up...")
   (j/db-do-commands (config/db)
                     "ALTER TABLE bulk_update_coll_status MODIFY
-                     (status_message VARCHAR(4000))"))
+                     (status_message VARCHAR2(4000))"))
 
 (defn down
   "Migrates the database down to version 8."
@@ -17,4 +17,4 @@
   (println "migrations.009-update-bulk-status-message down...")
   (j/db-do-commands (config/db) "TRUNCATE TABLE bulk_update_coll_status")
   (j/db-do-commands (config/db) "ALTER TABLE bulk_update_coll_status MODIFY
-                                 (status_message VARCHAR(255))"))
+                                 (status_message VARCHAR2(255))"))
--- a/ingest-app/src/migrations/009_update_bulk_status_message.clj
+++ b/ingest-app/src/migrations/009_update_bulk_status_message.clj
@@ -14,4 +14,7 @@
 (defn down
   "Migrates the database down to version 8."
   []
-  (println "migrations.009-update-bulk-status-message down..."))
+  (println "migrations.009-update-bulk-status-message down...")
+  (j/db-do-commands (config/db) "TRUNCATE TABLE bulk_update_coll_status")
+  (j/db-do-commands (config/db) "ALTER TABLE bulk_update_coll_status MODIFY
+                                 (status_message VARCHAR(255))"))
--- a/ingest-app/src/migrations/009_update_bulk_status_message.clj
+++ b/ingest-app/src/migrations/009_update_bulk_status_message.clj
@@ -14,7 +14,4 @@
 (defn down
   "Migrates the database down to version 8."
   []
-  (println "migrations.009-update-bulk-status-message down...")
-  (j/db-do-commands (config/db) "TRUNCATE TABLE bulk_update_coll_status")
-  (j/db-do-commands (config/db) "ALTER TABLE bulk_update_coll_status MODIFY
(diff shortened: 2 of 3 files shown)


ingest-app/src/cmr/ingest/data/bulk_update.clj
```diff
@@ -128,7 +128,7 @@
        (let [statement (str "UPDATE bulk_update_coll_status "
                             "SET status = ?, status_message = ?"
                             "WHERE task_id = ? AND concept_id = ?")
-             status-message (util/trunc status-message 255)]
+             status-message (util/trunc status-message 4000)]
```
chris-durbin: I'm wondering if we'll have a byte vs. character issue. Some UTF-8 chars can take 2 bytes and so util/trunc needs to be consistent with the way Oracle works.

What I think can happen is that we'll truncate to 4000 chars which ends up being 4008 bytes and get an error trying to save in Oracle.

ingest-app/src/migrations/009_update_bulk_status_message.clj
```diff
@@ -0,0 +1,17 @@
...
+
+(defn up
+  "Migrates the database up to version 9."
+  []
+  (println "migrations.009-update-bulk-status-message up...")
+  (j/db-do-commands (config/db)
+                    "ALTER TABLE bulk_update_coll_status MODIFY
+                     (status_message VARCHAR(4000))"))
+
+(defn down
```
oubiwann: Won't you need to truncate here? I'm imagining running migrations locally (up and down) when large status messages have been inserted into the DB ... I think there'd be a migration failure in that case?

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
```diff
@@ -349,3 +377,23 @@
...
+                                         :Term "ENVIRONMENTAL IMPACTS"
+                                         :VariableLevel1 "HEAVY METALS CONCENTRATION"}}
+        task-id (:task-id (ingest/bulk-update-collections "PROV1" bulk-update-body))]
+    (index/wait-until-indexed)
+    (let [collection-response (ingest/bulk-update-task-status "PROV1" task-id)
+          collection-status (first (:collection-statuses collection-response))]
+      (is (= "COMPLETE" (:task-status collection-response)))
+      (is (= "COMPLETE" (:status collection-status)))
+      (is (< 255 (count (:status-message collection-status)))))))
+
```
oubiwann: I may be missing some code/context here that already exists elsewhere ... I would think you'd need to test a message of under 4000 chars, 4000 char exactly, and over 4000 chars ...

daniel-zamora: I'm not sure, 4000 is just as arbitrary as 255.  Its not like the number 4000 is tied to a specific requirement.  MMT just doesn't want their messages to be truncated.  It was truncated before at 255.

chris-durbin: 4000 is the limit for VARCHAR2 sizes in Oracle 11g and earlier and Oracle 12c when not running in extended mode. I forgot to ask the MMT guys, but seeing that it is tied to a single collection, truncating at 4000 characters seems fine.

Note that with VARCHAR types we will only allocate as many bytes as are actually used for each row so oversizing the column is not a problem.

chris-durbin: Also I'm ok with this system integration test since we don't know of a way to generate a condition that would generate more than 4000 characters and we are at least verifying that we are not truncating at 255 anymore.

Ideally we'd have a test showing we truncate in a way that the data is successfully saved in Oracle even with special characters that take more than 1 byte in UTF-8 (like π). Is it possible to do this as a "unit with database test" where we manually call an appropriate function which sets the status-message to something like 5000 π characters?

ingest-app/src/migrations/009_update_bulk_status_message.clj
```diff
@@ -0,0 +1,20 @@
...
+  (println "migrations.009-update-bulk-status-message up...")
+  (j/db-do-commands (config/db)
+                    "ALTER TABLE bulk_update_coll_status MODIFY
+                     (status_message VARCHAR2(4000))"))
+
+(defn down
+  "Migrates the database down to version 8."
+  []
+  (println "migrations.009-update-bulk-status-message down...")
+  (j/db-do-commands (config/db) "TRUNCATE TABLE bulk_update_coll_status")
```
chris-durbin: I just saw this statement. We shouldn't delete data from the tables.

What we might have wanted to do was an update statement to every row to replace the status_message with up to the first 255 characters of the status message. But you cannot reduce the length of a column so this would need to be done by creating a new temporary column, dropping the old column and then renaming the temporary column to the original column.

After thinking that through It would be a crazy amount of work to do for a down migration. In this particular case we shouldn't do anything in the down migration (leave the column size at 4000).

ingest-app/src/migrations/009_update_bulk_status_message.clj
```diff
@@ -0,0 +1,20 @@
...
+   [clojure.java.jdbc :as j]
+   [config.migrate-config :as config]))
+
+(defn up
+  "Migrates the database up to version 9."
+  []
+  (println "migrations.009-update-bulk-status-message up...")
+  (j/db-do-commands (config/db)
+                    "ALTER TABLE bulk_update_coll_status MODIFY
+                     (status_message VARCHAR2(4000))"))
```
sxu123: I forgot the details why I did it the way I did in 006_update_bulk_update_tables.clj. I just remember that whenever you change the column size, it's a non-trivial thing.  So I'm not sure a simple alter table will do the trick.

chris-durbin: Increasing the size of a column with the same type should be fine. Looking at the other migration you were changing the task_id column from a number to a varchar, which cannot be done in place.

When I read that migration I see the all of the bulk update tables are using VARCHAR rather than VARCHAR2. I'm not sure if you'll have an issue changing to VARCHAR2 so you can just leave it as VARCHAR.

ingest-app/src/migrations/009_update_bulk_status_message.clj
daniel-zamora: Cool, thats they way I had it before.

ingest-app/src/cmr/ingest/data/bulk_update.clj
daniel-zamora: You are correct, tested locally, larger UTF-8 chars will cause the update to fail if there are more bytes than allocated in the column.  This was true with 255 as well.

chris-durbin: I would just throw in a comment for now that with special UTF-8 characters that take 2 bytes and a status message larger than 4000 characters we will have an error updating.

We do not expect that we will have status messages this large though, and so it is not worth the effort to figure out how to truncate to 4000 bytes (4000 characters is fine). If we find problems due to this we will open a new ticket and address.

oubiwann: I'm cool with that, then :-)

PR #339, opened 2017-11-02
CMR-4549 Granule links need to search document mime-types case insensitively.

daniel-zamora:
This PR implements case-insensitivity for granule documentation urls in OnlineResource field.

### DIFF
--- a/umm-lib/src/cmr/umm/related_url_helper.clj
+++ b/umm-lib/src/cmr/umm/related_url_helper.clj
@@ -32,9 +32,8 @@
 (defn documentation-url?
   "Returns true if the related-url is documentation url"
   [related-url]
-  (let [mime-type (:mime-type related-url)]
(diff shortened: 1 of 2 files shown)


umm-lib/src/cmr/umm/related_url_helper.clj
```diff
@@ -30,7 +32,9 @@
 (defn documentation-url?
   "Returns true if the related-url is documentation url"
   [related-url]
-  (some #{(:mime-type related-url)} DOCUMENTATION_MIME_TYPES))
+  (let [mime-type (:mime-type related-url)]
+    (when (seq mime-type)
```
chris-durbin: I can't think of another place where we use `seq` on a string. I think it would work, but we should make this `(when mime-type` or really as suggested above `(when-let [mime-type (:mime-type related-url)]`

PR #340, opened 2017-11-02
CMR-4569: Added special update case for the data center to only updat…

sxu123:
…e home page url inside ContactInformation.


sxu123:
This PR adds a special FIND_AND_UPDATE case:  FIND_AND_UPDATE_HOME_PAGE_URL for the DATA_CENTERS update.  

It treats ContactInformation in update-value differently.  Rather than replacing the whole ContactInformation with the new ContactInformation, it only replaces(or adds if doesn't exists) the home page url part of the RelatedUrls inside the ContactInformation, while leaving everything else untouched. 

If the user passes anything other than the home page url under ContactInformation in update-value, they will be ignored.  

Example:
(curl+ -i -XPOST
 -H "Cmr-Pretty:true"
 -H "Content-Type: application/json"
 -H "Echo-Token: mock-echo-system-token"
 "http://localhost:3002/providers/PROV1/bulk-update/collections"
 -d {:concept-ids ["C1200000005-PROV1"]
     :update-type "FIND_AND_UPDATE_HOME_PAGE_URL"
     :update-field "DATA_CENTERS"
     :find-value {:ShortName "AARHUS-HYDRO-1"}
     :update-value {:LongName "New Long Name" :ShortName "New ShortName"
                    :ContactInformation {:RelatedUrls [{:URLContentType "DataCenterURL"
                                                                                 :Type "HOME PAGE"
                                                                                 :URL "http://test.org/daac/index.html"}
                                                                                 {:OtherURLs "Ignored"}] 
                                                        :OtherStuff "Ignored"}}})

Still need to add integration tests. Just put the PR out so that we can point it to MMT people for the example and documentation.

chris-durbin:
The example above is in EDN - I assume MMT is actually posting JSON, correct?

ingest-app/docs/api.md
```diff
@@ -642,15 +642,17 @@ Bulk update post request takes the following parameters:
...
 
 The difference between `FIND_AND_UPDATE` and `FIND_AND_REPLACE` is `FIND_AND_REPLACE` will remove the matches and replace them entirely with the values specified in update value, while with `FIND_AND_UPDATE`, only the field(s) specified in the update value will be replaced, with the rest of the original value retained. For example, with a platform update value of `{"ShortName": "A340-600"}`, only the short name will be updated during a find and update, while the long name, instruments, and other fields retain their values. If a field specified in the update value doesn't exist in the matches, the field will be added.
 
+`FIND_AND_UPDATE_HOME_PAGE_URL` is a special case for `FIND_AND_UPDATE`. It can only be used with Update field being DATA_CENTERS. It is the same as FIND_AND_UPDATE except that when update value contains :ContactInformation, it doesn't replace the ContactInformation, instead it only replaces the datacenter HOME PAGE URL part  with the value specified in the RelatedUrls, if it exists, and leave everything else in the ContactInformation untouched.
```
chris-durbin: "ContactInformation" rather than :ContactInformation.

extra space "part  with"

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
```diff
@@ -49,18 +49,32 @@
...
     (when (and (not= FIND_AND_REMOVE update-type)
                (nil? update-value))
       (errors/throw-service-errors :bad-request
                                    [(format "An update value must be supplied when the update is of type %s"
                                             update-type)]))
+    (when (and (= FIND_AND_UPDATE_HOME_PAGE_URL update-type)
+               (not= DATA_CENTERS update-field))
+      (errors/throw-service-errors 
+       :bad-request
+      [(str (format "FIND_AND_UPDATE_HOME_PAGE_URL update type can not be used for the [%s] update field. " update-field)
```
chris-durbin: It looks like you are using `str` and `format` together to avoid a long line? In this case I think I would just use `str`.

umm-spec-lib/src/cmr/umm_spec/field_update.clj
```diff
@@ -6,11 +6,60 @@
...
+
 (defn field-update-functions
   "Partial update functions for handling specific update cases"
   [umm]
   {[:Instruments] (partial util/update-in-each umm [:Platforms])})
 
+(defn- get-home-page-url
+  "Get data center home page url from a list of related urls."
+  [related-urls]
+  (when (seq related-urls)
```
chris-durbin: Calling `first` on an empty list will return nil so the when line is unnecessary.

umm-spec-lib/src/cmr/umm_spec/field_update.clj
```diff
@@ -6,11 +6,60 @@
...
+  "Data center URL's type."
+  "HOME PAGE")
+
 (defn field-update-functions
   "Partial update functions for handling specific update cases"
   [umm]
   {[:Instruments] (partial util/update-in-each umm [:Platforms])})
 
+(defn- get-home-page-url
+  "Get data center home page url from a list of related urls."
```
chris-durbin: The code is only choosing one home page URL. We should update the docstring to note that if there are more than one home page URLs we are arbitrarily choosing the first one.

umm-spec-lib/src/cmr/umm_spec/field_update.clj
```diff
@@ -6,11 +6,60 @@
...
+                            (= HomePage Type))]
+             {:URLContentType DataCenterURL
+              :Type HomePage
+              :URL URL})))) 
+
+(defn- home-page-url?
+  "Check if a related url is a data center home page url."
+  [related-url]
+  (let [{:keys [URLContentType Type URL]} related-url]
+    (if (and (= DataCenterURL URLContentType)
```
chris-durbin: I've seen this logic a few times in the CMR where we do an `if` then `true` or `false`. I would just remove all that and return:

`(and (= DataCenterURL URLContentType)
          (= HomePage Type))`

Anyone have a strong style preference?

umm-spec-lib/src/cmr/umm_spec/field_update.clj
```diff
@@ -6,11 +6,60 @@
...
+             (= HomePage Type))
+      true
+      false)))
+
+(defn- home-page-url-update
+  "Apply special homepageurl update to data-center using the update-value.
```
chris-durbin: Add dashes between all words in the docstring and the bindings in the code to make it easier to read:
e.g.
homepageurl -> home-page-url
update-value-non-contactinfo-part -> update-value-non-contact-info-part

umm-spec-lib/src/cmr/umm_spec/field_update.clj
```diff
@@ -6,11 +6,60 @@
...
+      a. If data-center-relatedurls contains homepageurl, update it with the new homepageurl.
+      b. If data-center-relatedurls doesn't contain homepageurl, add the new homepageurl to it.
+      Note: Anything other than the homepageurl in update-value are ignored. 
+   2.  Update the non ContactInformation part for the data center normally(replace or add)."
+  [data-center update-value]
+  (let [update-value-non-contactinfo-part (dissoc update-value :ContactInformation)
+        update-value-relatedurls (get-in update-value [:ContactInformation :RelatedUrls])
+        update-value-homepageurl (get-home-page-url update-value-relatedurls)
+        data-center-relatedurls (get-in data-center [:ContactInformation :RelatedUrls])]
+    (-> (if (seq update-value-homepageurl)
```
chris-durbin: Thread macros are generally for a series of 3 or more function composition (only 2 here).

chris-durbin: This logic is complicated enough it probably deserves its own function with a docstring.

ingest-app/src/cmr/ingest/services/bulk_update_service.clj

ingest-app/docs/api.md
sxu123: Done.

umm-spec-lib/src/cmr/umm_spec/field_update.clj
sxu123: Done.

umm-spec-lib/src/cmr/umm_spec/field_update.clj
sxu123: Done.

umm-spec-lib/src/cmr/umm_spec/field_update.clj
sxu123: I could go either way. The if then one is more intuitive, this one is more compact. I will change it. Done.

umm-spec-lib/src/cmr/umm_spec/field_update.clj
sxu123: Done.

umm-spec-lib/src/cmr/umm_spec/field_update.clj
sxu123: Done.

sxu123: Done.

sxu123:
Added Integration tests and changed the example in the comment.

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
```diff
@@ -40,6 +40,34 @@
                   :ContactPersons [{:Roles ["Data Center Contact"]
                                     :LastName "Smith"}]}]})
 
+(def data-centers-url
```
chris-durbin: Add a docstring and rename. This is not just a data center URL. Maybe `data-centers-with-urls`?

umm-spec-lib/src/cmr/umm_spec/field_update.clj
```diff
@@ -6,11 +6,63 @@
   [cmr.umm-spec.date-util :as date-util]
   [cmr.umm-spec.umm-spec-core :as spec-core]))
 
+(def DataCenterURL
```
chris-durbin: Nit: Clojure style is generally `data-center-url` and `home-page` for constants. I've seen us also use screaming snake case for constants: `DATA_CENTER_URL`. I'm good with either of those, but we probably shouldn't add CamelCase. I think the only place we end up using camel case is from the UMM JSON models.

sxu123: Thanks. Done.

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
sxu123: Done. Changed it to data-centers-umm-with-home-page-url.

umm-spec-lib/src/cmr/umm_spec/field_update.clj
oubiwann: Yeah, just a quick additional note on Camel case: it's also used in protocol and record/type names. 

In some open source Clojure projects, constants used in more than one module are put into a `const` ns. We should avoid global vars, but the classic pattern there in Clojure (inherited from Common Lisp) is earmuffs around kebab-case, e.g., `*some-global-const*`, but that's less and less common in the Clojure ecosystem.

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
```diff
@@ -49,23 +49,38 @@
   "Represents FIND_AND_UPDATE update type"
   "FIND_AND_UPDATE")
 
+(def FIND_AND_UPDATE_HOME_PAGE_URL
```
oubiwann: I think you're already covering this, but I'll make a note where these should be changed anyway ;-) This should be changed to kebab-case.

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
```diff
@@ -49,23 +49,38 @@
...
+  "Represents FIND_AND_UPDATE_HOME_PAGE_URL update type"
+  "FIND_AND_UPDATE_HOME_PAGE_URL")
+
+(def DATA_CENTERS
```
oubiwann: This should be changed to kebab-case.

umm-spec-lib/src/cmr/umm_spec/field_update.clj
```diff
@@ -6,11 +6,63 @@
...
+  "Data center URL's content type."
+  "DataCenterURL")
+
+(def HomePage
```
oubiwann: This needs to be changed to kebab-case.

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
sxu123: Thanks. Done.

ingest-app/src/cmr/ingest/services/bulk_update_service.clj

PR #344, opened 2017-11-02
CMR-4581

ygliuvt:
As a Client User, I want collection JSON search response to include variable and service associations.

### DIFF
--- a/system-int-test/test/cmr/system_int_test/search/service/collection_service_search_test.clj
+++ b/system-int-test/test/cmr/system_int_test/search/service/collection_service_search_test.clj
@@ -12,17 +12,19 @@
    [cmr.system-int-test.utils.index-util :as index]
    [cmr.system-int-test.utils.ingest-util :as ingest]
    [cmr.system-int-test.utils.search-util :as search]
-   [cmr.system-int-test.utils.service-util :as service-util]))
+   [cmr.system-int-test.utils.service-util :as service-util]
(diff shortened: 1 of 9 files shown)


system-int-test/test/cmr/system_int_test/search/service/collection_service_search_test.clj
```diff
@@ -47,11 +52,13 @@
...
-  "Verify the collection in ATOM and JSON response has-formats field has the given value"
-  [coll has-formats]
-  (verify-collection-atom-has-formats coll has-formats)
-  (verify-collection-json-has-formats coll has-formats))
+(defn- assert-collection-atom-json-result
+  "Verify collection in ATOM and JSON response has-formats, has-variables and associations fields"
+  ([coll has-formats serv-concept-ids]
+   (assert-collection-atom-json-result coll has-formats serv-concept-ids nil))
+  ([coll has-formats serv-concept-ids var-concept-ids]
+   (assert-collection-atom-result coll has-formats (boolean (seq var-concept-ids)))
```
chris-durbin: Not sure if `some?` or `boolean` makes more sense. I guess the only time it would behave different is if the value was literally `false`.

ygliuvt: `some?` may be more readable in this case. I will change it in my ongoing work and include it in my next PR.

PR #349, opened 2017-11-08
CMR-4582 Add support for has-spatial-subsetting

oubiwann:
Not 100% sure if this is what @ygliuvt and @chris-durbin had in mind, but seems like it should be close :-)

### DIFF
--- a/indexer-app/src/cmr/indexer/data/concepts/service.clj
+++ b/indexer-app/src/cmr/indexer/data/concepts/service.clj
@@ -23,6 +23,15 @@
         supported-formats (get-in service [:ServiceOptions :SupportedFormats])]
     (> (count supported-formats) 1)))
 
+(defn- has-spatial-subsetting?
+  "Returns true if the given service has a defined SubsetType with one of its
+  values being 'spatial'."
+  [context service-concept]
+  (let [service (concept-parser/parse-concept context service-concept)
+        {{subset-type :SubsetType} :ServiceOptions} service]
+    (and (seq subset-type)
+         (contains? (set subset-type) "Spatial"))))
+
 (defn- has-transforms?
   "Returns true if the given service has a defined SubsetType or InterpolationType,
   or multiple SupportedProjections values."
@@ -43,4 +52,5 @@
                                  (map #(service-association->service-concept context %)
                                       service-associations))]
     {:has-formats (boolean (some #(has-formats? context %) service-concepts))
+     :has-spatial-subsetting (boolean (some #(has-spatial-subsetting? context %) service-concepts))
      :has-transforms (boolean (some #(has-transforms? context %) service-concepts))}))
--- a/indexer-app/src/cmr/indexer/data/concepts/service.clj
+++ b/indexer-app/src/cmr/indexer/data/concepts/service.clj
@@ -52,5 +52,5 @@
                                  (map #(service-association->service-concept context %)
                                       service-associations))]
     {:has-formats (boolean (some #(has-formats? context %) service-concepts))
-     :has-spatial-subsetting (some #(has-spatial-subsetting? context %) service-concepts)
+     :has-spatial-subsetting (boolean (some #(has-spatial-subsetting? context %) service-concepts))
      :has-transforms (boolean (some #(has-transforms? context %) service-concepts))}))
--- a/indexer-app/src/cmr/indexer/data/concepts/service.clj
+++ b/indexer-app/src/cmr/indexer/data/concepts/service.clj
@@ -52,5 +52,5 @@
                                  (map #(service-association->service-concept context %)
                                       service-associations))]
     {:has-formats (boolean (some #(has-formats? context %) service-concepts))
-     :has-spatial-subsetting (boolean (some #(has-spatial-subsetting? context %) service-concepts))
+     :has-spatial-subsetting (some #(has-spatial-subsetting? context %) service-concepts)
      :has-transforms (boolean (some #(has-transforms? context %) service-concepts))}))
--- a/search-app/docs/api.md
+++ b/search-app/docs/api.md
@@ -346,7 +346,7 @@ The following fields are specific to the CMR output and most correspond to ECHO1
 | echo:hasVariables (collections only)       | true if there are variables associated with the collection                                                           |
 | echo:hasFormats (collections only)         | true if there are multiple supported formats for any services associated with the collection                         |
 | echo:hasTransforms (collections only)      | true if there are transformations (subset, interpolation or projection) in any of its associated services            |
-| echo:hasSpatialSubsetting (collections only)| true if the setset contains "Spatial" in any of its associated services            |
+| echo:hasSpatialSubsetting (collections only)| true if any of its associated services support spatial subsetting            |
 | relevance:score (collections only)         | relevance score of the collection to search parameters                                                               |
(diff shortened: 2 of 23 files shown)


search-app/docs/api.md
```diff
@@ -346,6 +346,7 @@ The following fields are specific to the CMR output and most correspond to ECHO1
 | echo:hasVariables (collections only)       | true if there are variables associated with the collection                                                           |
 | echo:hasFormats (collections only)         | true if there are multiple supported formats for any services associated with the collection                         |
 | echo:hasTransforms (collections only)      | true if there are transformations (subset, interpolation or projection) in any of its associated services            |
+| echo:hasSpatialSubsetting (collections only)| true if the setset contains "Spatial" in any of its associated services            |
```
chris-durbin: "true if any of its associated services support spatial subsetting"

search-app/docs/api.md
```diff
@@ -787,6 +788,7 @@ __Example__
       "has_variables" : true,
       "has_formats" : false,
       "has_transforms" : false,
+      "has_spatial_subsetting": false
```
chris-durbin: Needs a comma at the end of the line and probably should put a space before the colon to be consistent with the rest of the fields in the example (and our pretty format apparently has a space).

search-app/src/cmr/search/results_handlers/atom_json_results_handler.clj
```diff
@@ -78,6 +78,7 @@
                        :has_variables has-variables
                        :has_formats has-formats
                        :has_transforms has-transforms
+                       :has-spatial-subsetting has-spatial-subsetting
```
chris-durbin: We use underscores on our API so the keyword should have an underscore here.

indexer-app/src/cmr/indexer/data/concepts/service.clj
```diff
@@ -23,6 +23,15 @@
...
     (> (count supported-formats) 1)))
 
+(defn- has-spatial-subsetting?
+  "Returns true if the given service has a defined SubsetType with one of its
+  values being 'spatial'."
+  [context service-concept]
+  (let [service (concept-parser/parse-concept context service-concept)
+        {{subset-type :SubsetType} :ServiceOptions} service]
+    (and (seq subset-type)
+         (contains? (set subset-type) "Spatial"))))
```
chris-durbin: This is still one of the most confusing parts of Clojure for me - should I be using `contains?` or `some` for finding a value in a collection? `contains?` needs to build a set, but has constant or logarithmic lookup time, while `some` doesn't have to build a set but needs to perform a linear search.

I assume that `contains?` should almost always be preferred then, so why does the documentation for `some` specifically list "One common idiom... `(some #{:fred} coll)`"?

search-app/src/cmr/search/results_handlers/atom_json_results_handler.clj
oubiwann: Whoops! Fixed.

search-app/docs/api.md
oubiwann: Fixed.

search-app/docs/api.md
oubiwann: Changed.

indexer-app/src/cmr/indexer/data/concepts/service.clj
oubiwann: Note that `(contains? (set subset-type) "Spatial"))` returns a boolean; `(some #{"Spatial"} subset-type)` returns a matching value. One way to make the latter return a bool is `(some? (some #{"spatial"} subset-type))` ... awkward. Another way to do this would be `(boolean (some #(= % "spatial") subset-type))` (the `boolean` is needed to convert the non-matching `nil` to `false`).

For my own view of readability, `(contains? (set subset-type) "Spatial"))` beats the others ...

For performance, let's see ... 

On a per-function call basis:

`contains?`: ~0.06 msecs
`some?`/`some`: ~ 0.05 msecs
`boolean`/`some`: ~0.15-0.4 msecs

For the last one I gave a range because its performance fluctuates pretty widely.

10 millions calls to the following with small data sets gives:

`contains?`: ~4.5 seconds
`some?`/`some`: ~3.2 seconds
`boolean`/`some`: ~3.5 seconds

Given the similarity of performance (especially on the per-single-call basis), I'd probably lean towards `contains?` ... but it's up to you guys.

indexer-app/src/cmr/indexer/data/concepts/service.clj
```diff
@@ -43,4 +52,5 @@
                                  (map #(service-association->service-concept context %)
                                       service-associations))]
     {:has-formats (boolean (some #(has-formats? context %) service-concepts))
+     :has-spatial-subsetting (some #(has-spatial-subsetting? context %) service-concepts)
```
oubiwann: @ygliuvt I didn't use `boolean` here, since the predicate function I added above always returns `true` or `false`, never `nil`. However, it looks like the others also will only return `true` or `false`, since they are using logic operators. Do you know why `boolean` is wrapping those here?

system-int-test/test/cmr/system_int_test/search/service/collection_service_search_test.clj
```diff
@@ -269,4 +281,15 @@
       (au/associate-by-concept-ids token serv9-concept-id [{:concept-id (:concept-id coll4)}])
       (index/wait-until-indexed)
       ;; after service association is made, has-transforms is true
-      (assert-collection-atom-json-result coll4 {:has-transforms true} [serv9-concept-id]))))
+      (assert-collection-atom-json-result coll4 {:has-transforms true} [serv9-concept-id]))
+
+    (testing "SubsetType with no 'Spatial' results in spatial-subsetting being false"
```
ygliuvt: Move this check up to line 270 above, i.e. associate service 10 with coll3 and check `has-spatial-subsetting` is false; then associated service8 with coll3, verify that `has-spatial-subsetting` would be true after that.

indexer-app/src/cmr/indexer/data/concepts/service.clj
ygliuvt: We only have three possible values in the set, I don't think we care about performance in this case. `contains?` returns true, so a good choice.

indexer-app/src/cmr/indexer/data/concepts/service.clj
ygliuvt: This is just defensive coding. While `has-formats` always returns true or false, `has-transforms` does not. So I feel safer to just wrap a boolean around the call.

oubiwann:
Note: I think I might want to merge this after @ygliuvt's recent PR ...

ygliuvt:
> Note: I think I might want to merge this after @ygliuvt's recent PR ...

I think it is OK for you to merge first. I can resolve the merge conflict later.

indexer-app/src/cmr/indexer/data/concepts/service.clj
ygliuvt: @oubiwann Actually, use `boolean` is a must here as `some` returns either true or nil and in this case, we want to convert nil to `false` for indexing. I bet once you add the `boolean` call, most (if not all) of the failed tests should be fixed.

oubiwann:
I've rebased on @ygliuvt's work, and a whole series of tests that had been failing before (other concept types) are now passing, so that's a relief. Running the full suite of tests locally, and when I've confirmed those are passing, I'll merge.

PR #350, opened 2017-11-08
CMR-4486; CMR-4483

ygliuvt:
As a Client User, I want collection UMM JSON search response to include variable and service associations.
As a Client User, I want to collection search UMM JSON response to include has_variables, has_transforms, has_formats fields.

### DIFF
--- a/umm-spec-lib/resources/json-schemas/collection/umm/v1.0/umm-search-results-json-schema.json
+++ b/umm-spec-lib/resources/json-schemas/collection/umm/v1.0/umm-search-results-json-schema.json
@@ -38,9 +38,7 @@
         },
         "concept-id": {
           "description": "The concept id of the item found.",
-          "type": "string",
-          "minLength": 4,
-          "pattern": "[A-Z]+\\d+-[A-Z0-9_]+"
+          "$ref": "#/definitions/ConceptIdType"
         },
         "revision-id": {
           "description": "A number >= 1 that indicates which revision of the item.",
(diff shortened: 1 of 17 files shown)


umm-spec-lib/resources/json-schemas/collection/umm/v1.0/umm-search-results-json-schema.json
```diff
@@ -64,9 +62,54 @@
           "description": "The mime type of the original metadata",
           "type": "string",
           "minLength": 1
+        },
```
mschmele: I feel like having this in versions 1.0 through 1.8 defeats the purpose of a versioned schema. Is there a particular reason for adding it to all versions?

ygliuvt: First of all, this schema is used only in our testing code. Secondly, as we are not really versioning the search result schema in CMR, rather just link to different versions of UMM-C schema here, there is no way to add new fields in the search results without updating all the old schemas.

umm-spec-lib/src/cmr/umm_spec/models/umm_collection_models.clj
```diff
@@ -70,9 +70,9 @@
    ;; This element stores the DOI (Digital Object Identifier) that identifies the collection. Note:
    ;; The values should start with the directory indicator which in ESDIS' case is 10. If the DOI
    ;; was registered through ESDIS, the beginning of the string should be 10.5067. The DOI URL is
-   ;; not stored here; it should be stored as a RelatedURL. The DOI organization that is responsible
-   ;; for creating the DOI is described in the Authority element. For ESDIS records the value of
-   ;; http://dx.doi.org/ should be used.
+   ;; not stored here; it should be stored as a RelatedURL. Authority is the registry that registers
```
chris-durbin: Out of curiosity - how did you know the model needed to be updated? We should be catching this when we make the actual ticket change for things like this.

umm-spec-lib/resources/json-schemas/collection/umm/v1.0/umm-search-results-json-schema.json
oubiwann: Which is another point backing up your question as to the usefulness of having this versioned ...

oubiwann:
> It would be really nice to get rid of the JSON schema for UMM JSON search results from the versioned UMM directories and put in the resources directory unversioned. Any reason not to do that?

Yeah, I think we're all now asking this same question ... (which @ygliuvt started asked this morning in chat, iirc)

umm-spec-lib/src/cmr/umm_spec/models/umm_collection_models.clj
oubiwann: 🤔

ygliuvt: I don't. I just run the `generate-umm-records` task and this is changed as a result. 

I guess this is a result of the person who made the original schema change forgot to run the task to update the model before merging it to master. Since this only changed some comments, no tests failed. So everybody is happy.

umm-spec-lib/resources/json-schemas/collection/umm/v1.0/umm-search-results-json-schema.json
ygliuvt: The reason this is versioned is not for versioning the fields in the meta field, but for pointing to the versioned UMM-C schema. I am OK with this as it is only for testing, but we should probably remove the search result schema from the EMFD repo. I would like to know if that is OK.

umm-spec-lib/src/cmr/umm_spec/models/umm_collection_models.clj
oubiwann: The humor ... it's so dry, I think I need skin lotion.

PR #352, opened 2017-11-09
CMR-4570 Bulk Update 'Add to Existing' should not duplicate an existing Science or Location keyword

daniel-zamora:
This PR removes any duplicate values on bulk update :add-to-existing for location keywords and science keywords.

### DIFF
--- a/system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
+++ b/system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
@@ -204,9 +204,9 @@
     (side/eval-form `(ingest-config/set-bulk-update-enabled! true))
     ;; Kick off bulk update
     (let [response (ingest/bulk-update-collections "PROV1" bulk-update-body)]
-          ;; Initiate bulk update that shouldn't add anything, including duplicates.
-      (ingest/bulk-update-collections "PROV1" duplicate-body)
       (is (= 200 (:status response)))
+      ;; Initiate bulk update that shouldn't add anything, including duplicates.
+      (ingest/bulk-update-collections "PROV1" duplicate-body)
       ;; Wait for queueing/indexing to catch up
       (index/wait-until-indexed)
       (let [collection-response (ingest/bulk-update-task-status "PROV1" (:task-id response))]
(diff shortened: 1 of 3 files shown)


umm-spec-lib/src/cmr/umm_spec/field_update.clj
```diff
@@ -77,9 +77,12 @@
 
 (defmethod apply-umm-list-update :add-to-existing
   [update-type umm update-field update-value find-value]
-  (if (sequential? update-value)
-    (update-in umm update-field #(concat % update-value))
-    (update-in umm update-field #(conj % update-value))))
+  (let [umm (if (sequential? update-value)
```
mschmele: Nitpick: you could do this function as one thread if you wanted to:

```
(as-> umm umm
  (if (sequential? update-value)
    (update-in umm update-field #(concat % update-value))
    (update-in umm update-field #(conj % update-value)))
  (util/update-in-each umm update-field util/remove-nil-keys)
  (update-in umm update-field distinct))
```

Both work, but this is a little clearer to me
https://clojuredocs.org/clojure.core/as-%3E

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
```diff
@@ -181,50 +181,61 @@
...
-         (is (= 200 (:status response)))
-         ;; Wait for queueing/indexing to catch up
-         (index/wait-until-indexed)
-         (let [collection-response (ingest/bulk-update-task-status "PROV1" (:task-id response))]
-           (is (= "COMPLETE" (:task-status collection-response))))
+    (side/eval-form `(ingest-config/set-bulk-update-enabled! true))
+    ;; Kick off bulk update
+    (let [response (ingest/bulk-update-collections "PROV1" bulk-update-body)
+          ;; Initiate bulk update that shouldn't add anything, including duplicates.
+          _ (index/wait-until-indexed)
```
oubiwann: These two non-assignments don't need to be (and shouldn't be) in the `let` block; they can accomplish their tasks just fine before the `(is ...)` call.

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
```diff
@@ -243,9 +254,23 @@
...
+                                        :Topic "OCEANS"}
+                                       {:Category "EARTH SCIENCE2"
+                                        :Topic "HUMAN DIMENSIONS2"
+                                        :Term "ENVIRONMENTAL IMPACTS2"
+                                        :VariableLevel1 "HEAVY METALS CONCENTRATION2"}]}]
        ;; Kick off bulk update
-       (let [response (ingest/bulk-update-collections "PROV1" bulk-update-body)]
+       (let [response (ingest/bulk-update-collections "PROV1" bulk-update-body)
+             ;; Initiate bulk update that shouldn't add anything, including duplicates.
+             _ (ingest/bulk-update-collections "PROV1" duplicate-body)]
```
oubiwann: Move this non-assignment bulk update call out of the `let` block.

PR #355, opened 2017-11-10
CMR-4504: Add Virtual Directories Prototype

chris-durbin:
This is prototype code. I've added a feature toggle that is disabled by default for an environment. When disabled the virtual directory routes are not exposed. I'll plan to turn it on in at least the SIT environment and potentially UAT pending customer feedback.

Being a prototype that is likely to change substantially or entirely, no documentation or tests have been added. Once we've started to build up requirements we'll write the appropriate stories and add tests and documentation as the stories are implemented.

### DIFF
--- a/dev-system/dev/user.clj
+++ b/dev-system/dev/user.clj
@@ -251,7 +251,7 @@
 (defn stop
   "Shuts down and destroys the current development system."
   []
-  (alter-var-root #'system
+  (alter-var-root #'user/system
                   (fn [s]
                     (when s (system/stop s)))))
 
@@ -284,7 +284,7 @@
   ;; Stop the system integration test system
   (sit-sys/stop)
   ; Stops the running code
-  (stop)
+  (user/stop)
   ; Refreshes all of the code and then restarts the system
   (refresh :after 'user/start))
 
--- a/search-app/src/cmr/search/services/virtual_directory.clj
+++ b/search-app/src/cmr/search/services/virtual_directory.clj
@@ -94,13 +94,6 @@
         parent-value-node (build-title-type-links-to-remove context concept-id parent-value
                                                             group-node link-level)]
     parent-value-node))
-    ; (build-filter-node parent-value group-node)))
-
-(defn build-group-and-filter-node-orig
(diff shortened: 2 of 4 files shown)


oubiwann:
Ah, gotta love the prototype ;-)

dev-system/dev/user.clj
```diff
@@ -251,7 +251,7 @@
 (defn stop
   "Shuts down and destroys the current development system."
   []
-  (alter-var-root #'system
+  (alter-var-root #'user/system
```
oubiwann: Good catch.

search-app/src/cmr/search/services/virtual_directory.clj
```diff
@@ -0,0 +1,225 @@
...
+
+(def num-urls
+  "Number of download URLs to return"
+  100)
+
+(defn- build-top-level-response
+  "Builds the final JSON response."
+  [context concept-id {:keys [year month day hour] :as time-ranges} children download-urls]
+  (let [interval-granularity (time-ranges->next-interval time-ranges)
+        hour-node (if (= :hour interval-granularity)
```
oubiwann: I'd split all of these (`hour-node`, `day-node`, etc.) into their own private functions and just call them here to get the assignments. Too much nested code in the `let`, etc.

chris-durbin: I wrote the huge prototype disclaimer to allow horrible ugly code... But yes this whole function needs to be rewritten and there's probably a lot of commonality that can be pulled out.

dev-system/dev/user.clj
ygliuvt: Not sure why we have to make this change.

chris-durbin: There's a require using the alias `system`, so that var was ambiguous. The problem I saw that was fixed by changing this is after calling user/stop directly I could not perform a refresh because my system was not actually stopped. After this change, calling stop did successfully stop my system.

chris-durbin:
I'm going to hold off on merging this in until we get more feedback from the OPeNDAP team.

PR #360, opened 2017-11-15
CMR-4528: Fixed the bug that caused the deleted-granule search respon…

sxu123:
…se header to miss the Content-Type

          which was why the pretty-print wouldn't work.

### DIFF
--- a/system-int-test/test/cmr/system_int_test/ingest/deleted_granules_index_test.clj
+++ b/system-int-test/test/cmr/system_int_test/ingest/deleted_granules_index_test.clj
@@ -108,7 +108,7 @@
       (are3 [params granules]
         (is (= (set (map :concept-id granules))
                (set (find-deleted-granules params))))
-
+        
         "after tomorrow"
         {:revision_date (t/plus- (t/now) (t/days 1))}
         []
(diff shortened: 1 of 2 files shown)


system-int-test/test/cmr/system_int_test/ingest/deleted_granules_index_test.clj
```diff
@@ -117,6 +117,14 @@
         {:revision_date (t/minus- (t/now) (t/days 1))}
         [granule1-prov1 granule2-prov1 granule3-prov1-2 granule4-prov2]))
 
+    (testing "content-type in search header"
+      (are3 [params]
```
chris-durbin: Usually only use are3 when there are multiple similar test scenarios with different parameters. Since this is just a single test you can just:

```
(testing "Deleted granule search header contains application/json content-type"
  (let [results (search/find-deleted-granules {:revision_date (t/minus- (t/now) (t/days 1))}
    (is (= "application/json; charset=utf-8"
              (get-in results [:headers :content-type]))))
```

sxu123: Sure. Thanks. Done.

PR #362, opened 2017-11-15
CMR-4401: Remove bootstrap, force responsive tables

mschmele:
Bootstrap had a bunch of conflicting styles that messed with margins and positioning - this just fixes that while keeping tables relatively mobile-friendly. Behavior is the same as with bootstrap, but it's conflict free

### DIFF
--- a/common-app-lib/resources/templates/structure.html
+++ b/common-app-lib/resources/templates/structure.html
@@ -14,7 +14,6 @@
     <!-- EUI -->
     <link href="https://cdn.earthdata.nasa.gov/eui/1.1.3/stylesheets/application.css" rel="stylesheet" />
     <link href="https://fonts.googleapis.com/css?family=Open+Sans:400,700,300" rel="stylesheet" type="text/css">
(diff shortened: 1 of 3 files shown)


common-app-lib/resources/templates/structure.html
```diff
@@ -14,7 +14,6 @@
     <!-- EUI -->
     <link href="https://cdn.earthdata.nasa.gov/eui/1.1.3/stylesheets/application.css" rel="stylesheet" />
     <link href="https://fonts.googleapis.com/css?family=Open+Sans:400,700,300" rel="stylesheet" type="text/css">
-    <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0-beta.2/css/bootstrap.min.css" integrity="sha384-PsH8R72JQ3SOdhVi3uxftmaW6Vc51MKb0q5P2rRUpPvrszuE4W1povHYgTpBfshb" crossorigin="anonymous">
```
chris-durbin: Does this also fix the issue reported by the Mobile-Friendly Test tool?
Page partially loaded
Not all page resources could be loaded. This can affect how Google sees and understands your page. Fix availability problems for any resources that can affect how Google understands your page.

mschmele:
screen shot 2017-11-15 at 3 13 36 pm

Screenshot of fixed margins with tables that don't extend past page-width

common-app-lib/resources/templates/structure.html
mschmele: No - those were issues with Google analytics. I'll look into that as well, but I think it's a separate issue

chris-durbin: Mark and I found that those issues are gone in SIT.

PR #365, opened 2017-11-16
CMR-4483: Updated the concept search extension details in the Search …

hulk92:
…API docs

### DIFF
--- a/search-app/docs/api.md
+++ b/search-app/docs/api.md
@@ -272,8 +272,8 @@ Here is a list of supported extensions and their corresponding MimeTypes:
   * `opendata`  "application/opendata+json" (only supported for collections)
   * `kml`       "application/vnd.google-earth.kml+xml"
   * `native`    "application/metadata+xml" (Returns search results in their individual native formats)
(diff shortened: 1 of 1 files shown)


search-app/docs/api.md
```diff
@@ -272,8 +272,8 @@ Here is a list of supported extensions and their corresponding MimeTypes:
   * `opendata`  "application/opendata+json" (only supported for collections)
   * `kml`       "application/vnd.google-earth.kml+xml"
   * `native`    "application/metadata+xml" (Returns search results in their individual native formats)
-  * `umm_json`   "application/vnd.nasa.cmr.legacy_umm_results+json" (only supported for collections)
```
chris-durbin: We do still have umm-json as an extension which is what this is actually documenting (just with the wrong extension - should be a dash instead of underscore). I think we still should leave this documentation in with umm-json until we've actually removed it.

So document umm-json and umm_json.

hulk92: I added the umm-json extension to the list of supported formats.

PR #366, opened 2017-11-16
CMR-4593: Replace str/lower-case with util/safe-lowercase to avoid nu…

sxu123:
…ll exception.    
The ticket shows the exception happened in
cmr.indexer.data.concepts.collection.platform/platform-short-name->elastic-do
clojure.string/lower-case,  java.lang.NullPointerException

1.  Replaced str/lower-case with util/safe-lowercase in platform.clj and a bunch of other files.  
2. Modified util/safe-lowercase so that it could handle boolean the same way as str/lower-case.

### DIFF
--- a/common-lib/src/cmr/common/util.clj
+++ b/common-lib/src/cmr/common/util.clj
@@ -115,7 +115,7 @@
 (defn safe-lowercase
   "Returns the given string in lower case safely."
   [v]
-  (when v (string/lower-case v)))
+  (when-not (nil? v) (string/lower-case v)))
 
 (defn safe-uppercase
   "Returns the given string in upper case safely."
--- a/indexer-app/src/cmr/indexer/data/concepts/tag.clj
+++ b/indexer-app/src/cmr/indexer/data/concepts/tag.clj
@@ -15,7 +15,7 @@
       {:concept-id concept-id
        :deleted deleted}
       {:concept-id concept-id
-       :tag-key.lowercase (str/lower-case tag-key)
+       :tag-key.lowercase (util/safe-lowercase tag-key)
        :description description
        :originator-id.lowercase  (util/safe-lowercase originator-id)})))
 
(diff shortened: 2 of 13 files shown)


common-lib/src/cmr/common/util.clj
```diff
@@ -115,7 +115,7 @@
 (defn safe-lowercase
   "Returns the given string in lower case safely."
   [v]
-  (when v (string/lower-case v)))
+  (when-not (nil? v) (string/lower-case v)))
```
ygliuvt: Change this to `(when (some? v) (string/lower-case v)))`.

indexer-app/src/cmr/indexer/data/concepts/tag.clj
```diff
@@ -15,15 +15,14 @@
       {:concept-id concept-id
        :deleted deleted}
       {:concept-id concept-id
-       :tag-key.lowercase (str/lower-case tag-key)
+       :tag-key.lowercase (util/safe-lowercase tag-key)
```
ygliuvt: Not entirely sold on why we have to make these changes. It is not possible to have a nil tag-key and when this happens, I would rather see an exception with the indexing failed rather than our code silently ignores the problem and index bad stuff in elasticsearch.

oubiwann: This is a good point ...

sxu123: I got your point. It crossed my mind too.  However, my thinking is that if something that can not be nil becomes nil, it will fail in many places rather than just the indexing...  Also, seems that when we added these lower-case ones, we didn't really know clearly if the field could be nil, so there could be many cases that it would cause trouble.  So I thought unless we have a clear list on which fields could be nil, which I don't have that list, maybe it's better to use safe-lowercase. Just my thought.

PR #368, opened 2017-11-17
CMR-4542: Add support for DOIs to JSON query

mschmele:


system-int-test/test/cmr/system_int_test/search/collection_doi_search_test.clj
```diff
@@ -50,4 +58,15 @@
        [coll1 coll2] ["Doi*"] {:pattern true}
 
        "search for collections with both doi1 and doi2 returns nothing"
-       [] ["Doi1" "doI2"] {:and true}))))
+       [] ["Doi1" "doI2"] {:and true}))
+    (testing "search doi with json query"
+      (let [not-provided-results (search/find-refs-with-json-query
```
ygliuvt: Change this to use `are3`. Add more tests similar to what doi search param above.

system-int-test/test/cmr/system_int_test/search/collection_doi_search_test.clj
```diff
@@ -50,4 +58,19 @@
...
 
        "search for collections with both doi1 and doi2 returns nothing"
-       [] ["Doi1" "doI2"] {:and true}))))
+       [] ["Doi1" "doI2"] {:and true}))
+    (testing "search doi with json query"
+      (are3 [items json-search]
+            (d/refs-match? items (search/find-refs-with-json-query
+                                  :collection {} json-search))
+            "Search for empty values"
+            [coll3] {:doi {:value "Not provided" :pattern true}}
```
chris-durbin: Change this test to not use pattern. Just `{:doi "Not provided"}` should be the search.

system-int-test/test/cmr/system_int_test/search/collection_doi_search_test.clj
```diff
@@ -50,4 +58,19 @@
...
+
+            "Wildcard search"
+            [coll1 coll2 coll3] {:doi {:value "*" :pattern true}}
+
+            "Search for collections with doi1 or doi2"
+            [coll1 coll2] {:doi {:value "Doi*" :pattern true}}
+
+            "Search for collections with doi1"
```
chris-durbin: The main purpose of the ticket is to be able to search for collections without DOI values. We should have some collections that do not have a DOI field at all (rather than "Not provided") and add two more tests:

1. `{:not {:doi {:value "*" :pattern true}}}` returns any collections without a DOI
2. `{:or [{:not {:doi {:value "*" :pattern true}}} {:not {:doi "Not provided"}}]` returns any collections without a DOI or with a value of Not provided.

Side note - the tests with Not provided probably don't make sense for DOI, they should be added for another field which is required in UMM JSON.

system-int-test/test/cmr/system_int_test/search/collection_doi_search_test.clj
```diff
@@ -50,4 +58,19 @@
...
+            [coll1] {:doi {:value "doi1" :pattern true}}))))
```
chris-durbin: `{:doi "doi1"}` makes more sense since you are not testing pattern here.

chris-durbin: We should also add a test for case-sensitive DOI searching in addition to the pattern test.

system-int-test/test/cmr/system_int_test/search/collection_doi_search_test.clj
mschmele: Done

system-int-test/test/cmr/system_int_test/search/collection_doi_search_test.clj
```diff
@@ -50,4 +64,29 @@
...
+      (are3 [items json-search]
+            (d/refs-match? items (search/find-refs-with-json-query
+                                  :collection {} json-search))
+            "Search for empty values"
+            [coll3] {:doi {:value "Not provided"}}
+
+            "Wildcard search"
+            [coll1 coll2 coll3] {:doi {:value "*" :pattern true}}
+
+            "Search for collections with doi1 or doi2"
```
ygliuvt: For completeness, add a test case for pattern false.

mschmele: Done

system-int-test/test/cmr/system_int_test/search/collection_doi_search_test.clj
```diff
@@ -50,4 +64,32 @@
...
+            [coll1 coll2] {:doi {:value "Doi*" :pattern true}}
+
+            "Search for collections with doi1 or doi2"
+            [] {:doi {:value "Doi*" :pattern false}}
+
+            "Search for collections with doi1"
```
chris-durbin: Can you add one more test as just a string rather than a map?
```
[coll1] {:doi "doi1"}
```
should be supported.

mschmele: Done

PR #369, opened 2017-11-20
CMR-4593: Platform parsing error during index.

sxu123:
1. The index error is caused by platform's short name being nil in an old collection C179003298-ORNL_DAAC revision 27.
2. Verified that the bad collection couldn't be ingested by the current code because of the nil platform short name.
3. Modified safe-lowercase, safe-uppercase because they don't work on the case when the parameter is boolean false.  Added unit tests for it.
4. Modified the search api doc to correct a grammar mistake from an earlier ticket.

### DIFF
--- a/common-lib/test/cmr/common/test/util.clj
+++ b/common-lib/test/cmr/common/test/util.clj
@@ -773,3 +773,15 @@
 
     "Set of numbers"
     #{1 7 52 3} 52))
+
+(deftest safe-lower-upper-case-test
+  (testing "safe-lowercase"
+    (is (= (util/safe-lowercase false) (str/lower-case false)))
+    (is (= (util/safe-lowercase true) (str/lower-case true)))
+    (is (= (util/safe-lowercase "StRing") (str/lower-case "StRing")))
+    (is (= (util/safe-lowercase nil) nil)))
+  (testing "safe-upperrcase"
+    (is (= (util/safe-uppercase false) (str/upper-case false)))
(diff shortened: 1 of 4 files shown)


common-lib/test/cmr/common/test/util.clj
```diff
@@ -773,3 +773,15 @@
 
     "Set of numbers"
     #{1 7 52 3} 52))
+
```
oubiwann: Nice!  And thanks :-) 

(this gets us one step closer to the goal of putting as much testing as possible in unit tests, and working to keep the sys-int tests as tightly focused on integration-tests-only as possible :-) )

sxu123: Thanks for suggesting it.

common-lib/test/cmr/common/test/util.clj
```diff
@@ -773,3 +773,15 @@
...
+(deftest safe-lower-upper-case-test
+  (testing "safe-lowercase"
+    (is (= (util/safe-lowercase false) (str/lower-case false)))
+    (is (= (util/safe-lowercase true) (str/lower-case true)))
+    (is (= (util/safe-lowercase "StRing") (str/lower-case "StRing")))
+    (is (= (util/safe-lowercase nil) nil)))
+  (testing "safe-upperrcase"
```
chris-durbin: uppercase typo

sxu123: Sorry didn't see the comment. Will fix this in my next PR.

PR #380, opened 2017-11-27
CMR-4570 

daniel-zamora:
This PR fixes issue where duplicates were still being added, after bulk-update was modified to remove duplicates on ADD_TO_EXISTING update-type.

### DIFF
--- a/system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
+++ b/system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
@@ -189,11 +189,11 @@
                         :update-value {:Category "EARTH SCIENCE"
                                        :Term "MARINE SEDIMENTS"
                                        :Topic "OCEANS"}}]
-    
+
     ;; Initiate bulk update that shouldn't add any duplicates.
     (let [response (ingest/bulk-update-collections "PROV1" duplicate-body)
           _ (index/wait-until-indexed)
-          collection-response (ingest/bulk-update-task-status "PROV1" response)]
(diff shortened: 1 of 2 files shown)


system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
```diff
@@ -189,38 +189,38 @@
                         :update-value {:Category "EARTH SCIENCE"
                                        :Term "MARINE SEDIMENTS"
                                        :Topic "OCEANS"}}]
+    ;; Initiate bulk update that shouldn't add anything, including duplicates.
```
sxu123: My only confusion is at this comment here - This bulk update won't add because the update-value already exists right? so it shouldn't add because it's a duplicate...  What do you mean by "including duplicates"?   Are the anything else the prevents the bulkupdate from happening?

sxu123: I meant it shouldn't add duplicates, is there anything else it shouldn't add?

daniel-zamora: the bulk update completes, but no keywords are added.

daniel-zamora: That specific update, before this ticket, would have added a duplicate keyword

sxu123: I understand that :)

PR #381, opened 2017-11-28
CMR-4685: Modified bulk update to allow data center home page url to …

sxu123:
…be removed when the new value is not present.

1. When doing FIND_AND_UPDATE_HOME_PAGE_URL on DataCenter, if the update-value doesn't contain a new home page url,  the original home page url of the found DataCenters should be removed.

2. Any empty RelatedUrls and ContactInformation will be removed following the home page url update/removal. 

3. Added the unit tests and Integration tests.

4. Modified the api doc.

### DIFF
--- a/ingest-app/docs/api.md
+++ b/ingest-app/docs/api.md
@@ -640,6 +640,7 @@ The following update types are supported:
   * Clear all and replace - clear the list and replace with the update value.
   * Find and replace - replace any instance in the list that matches the find value with the update value.
   * Find and update - merge update value into any instance in the list that matches the find value.
+  * Find and update home page url - A special case for Find and update that only applies to datacenter update. 
   * Find and remove - remove any instance from the list that matches the find value.
 
 Bulk update post request takes the following parameters:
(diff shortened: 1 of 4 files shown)


umm-spec-lib/test/cmr/umm_spec/test/field_update.clj
```diff
@@ -216,6 +216,133 @@
...
+               (field-update/apply-update update-type umm [:DataCenters] update-value find-value)))
+
+        "Find and update home page url" 
+        :find-and-update-home-page-url
+        {:ShortName "NewShortName" :LongName "NewLongName" :ContactInformation {:RelatedUrls [{:URLContentType "DataCenterURL"
+                                                                                               :Type "HOME PAGE"
+                                                                                               :URL "http://nsidc.org/daac/newindex.html"}]}}
+        {:ShortName "ShortName"}
+        {:DataCenters [{:ShortName "NewShortName"
+                              :LongName "NewLongName"
```
laurenfrederick: Indentation is off here

ingest-app/docs/api.md
```diff
@@ -655,7 +656,7 @@ Update types that include a FIND will match on the fields supplied in the find v
...
 
-`FIND_AND_UPDATE_HOME_PAGE_URL` is a special case for `FIND_AND_UPDATE`. It can only be used with Update field being DATA_CENTERS. It is the same as FIND_AND_UPDATE except that when update value contains ContactInformation, it doesn't replace the ContactInformation, instead it only replaces the datacenter HOME PAGE URL part with the value specified in the RelatedUrls, if it exists, and leave everything else in the ContactInformation untouched. 
+`FIND_AND_UPDATE_HOME_PAGE_URL` is a special case for `FIND_AND_UPDATE`. It can only be used with Update field being `DATA_CENTERS`. It is the same as `FIND_AND_UPDATE` except that when update value contains ContactInformation, it doesn't replace the ContactInformation, instead it only replaces the datacenter HOME PAGE URL part with the new datacenter HOME PAGE URL specified in the RelatedUrls, if it exists, and leave everything else in the ContactInformation untouched. If the new datacenter HOME PAGE URL is not present in the update value, the HOME PAGE URL of the found data centers will be removed.
```
laurenfrederick: Should be "leaves everything else...". Also, change datacenter to data center in this paragraph and above.

ingest-app/docs/api.md
```diff
@@ -640,6 +640,7 @@ The following update types are supported:
   * Clear all and replace - clear the list and replace with the update value.
   * Find and replace - replace any instance in the list that matches the find value with the update value.
   * Find and update - merge update value into any instance in the list that matches the find value.
+  * Find and update home page url - A special case for Find and update that only applies to datacenter update.
```
laurenfrederick: Maybe change to: a special case for Find and update that applies specifically to data center home page url update

ingest-app/docs/api.md
sxu123: Done.

ingest-app/docs/api.md
sxu123: Per our discussion, we will only say "A special case for Find and update." to avoid confusion.  Detailed explanation could be found in the home page.

PR #386, opened 2017-11-30
Cmr 4589: Make user-facing error messages friendlier

mschmele:
Get them nasty regexes outta my sight!

### DIFF
--- a/common-lib/src/cmr/common/validations/json_schema.clj
+++ b/common-lib/src/cmr/common/validations/json_schema.clj
@@ -41,12 +41,18 @@
   com.github.fge.jsonschema.core.report.ListProcessingReport.
 
   For details, see:
-  * http://fge.github.io/json-schema-validator/2.2.x/index.html
-  * http://java-json-tools.github.io/json-schema-core/1.2.x/index.html"
+  * [ProcessingMessage](http://java-json-tools.github.io/json-schema-core/1.2.x/com/github/fge/jsonschema/core/report/ProcessingMessage.html)
+  * [ListProcessingReport](http://java-json-tools.github.io/json-schema-core/1.2.x/com/github/fge/jsonschema/core/report/ListProcessingReport.html)"
   [^ListProcessingReport report]
   (flatten
     (for [error-report (seq (.asJson report))
           :let [json-error-report (json/decode (str ^Object error-report) true)]]
+      ;; XXX DEBUG
+      (warn "DEBUG: error-report")
+      (warn error-report)
+      (warn "DEBUG: error-report (JSON)")
+      (warn json-error-report)
+      ;; XXX END DEBUG
       (conj (parse-nested-error-report (:reports json-error-report))
             (parse-error-report json-error-report)))))
 
@@ -91,7 +97,10 @@
   json string to validate. Uses com.github.fge.jsonschema to perform the
   validation. The JSON schema must be provided as a
   com.github.fge.jsonschema.main.JsonSchema object and the json-to-validate must
-  be a string. Returns a list of the errors found."
+  be a string. Returns a list of the errors found.
+
+  For details, see:
+  * [JsonSchema](http://java-json-tools.github.io/json-schema-validator/2.2.x/com/github/fge/jsonschema/main/JsonSchema.html)"
   [^JsonSchema json-schema json-to-validate]
   (let [validation-report (.validate json-schema (json-string->JsonNode json-to-validate))]
     (parse-validation-report validation-report)))
--- a/common-lib/src/cmr/common/validations/json_schema.clj
+++ b/common-lib/src/cmr/common/validations/json_schema.clj
@@ -1,16 +1,17 @@
 (ns cmr.common.validations.json-schema
   "Functions used to perform JSON schema validation. See http://json-schema.org/
   for more details."
-  (:require [cheshire.core :as json]
-            [cmr.common.services.errors :as errors]
-            [cmr.common.log :as log :refer (warn)]
-            [clojure.string :as str])
-  (:import [com.github.fge.jsonschema.main
-            JsonSchema
-            JsonSchemaFactory]
-           com.github.fge.jackson.JsonLoader
-           com.github.fge.jsonschema.core.report.ListProcessingReport
-           com.fasterxml.jackson.core.JsonParseException))
+  (:require
+   [cheshire.core :as json]
+   [cmr.common.services.errors :as errors]
+   [cmr.common.log :as log :refer (warn)]
+   [clojure.string :as str])
+  (:import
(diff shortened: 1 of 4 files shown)


ingest-app/src/cmr/ingest/validation/validation.clj
```diff
@@ -94,16 +94,41 @@
...
+       (first friendly-error))
+      error)))
+
+(defn- sanitize-error-messages
+  "Remove nasty error messages from the collection to be returned.
+   Group errors by path and ensure that end-users don't have to see regexes. The horror!"
+  [error-messages]
+  (->> error-messages
+       (map create-user-friendly-error-message)
+       (remove #(re-find #"(ECMA|regex)" %))))
```
chris-durbin: I'm not understanding the changes. Why are we throwing away any part of a message after a semi-colon, and then it looks like we're also throwing away any error message that still has ECMA in it?

This seems like it is going to have some bad side effects of not returning errors that we do want to return. I think the fix needs to be much more targeted to recognize a regex message and remove that part of the message from the overall error.

ingest-app/src/cmr/ingest/validation/validation.clj
```diff
@@ -94,16 +94,41 @@
...
+   The original message should be logged."
+  [error]
+  (let [friendly-error (string/split error #";")]
+    (if (>= (count friendly-error) 1)
+      (do
+       (log-full-error-message error)
+       (first friendly-error))
+      error)))
+
+(defn- sanitize-error-messages
```
ygliuvt: I couldn't tell what this function does to the error message. It would be nice to have an example of the before and after message in the docstring or test to make sure that we indeed made the error message more informative.

system-int-test/test/cmr/system_int_test/ingest/collection_misc_validation_test.clj
```diff
@@ -77,6 +78,43 @@
...
+       "Invalid platforms"
+       platform-collection
+
+       "Invalid Data Centers"
+       data-center-collection
+
+       "Invalid Science Keywords"
+       science-keyword-collection))))
+
+
```
chris-durbin: I want an explicit test added that verifies the error message returned is:

`/Platforms/1/LongName string "" is too short (length: 0, required minimum: 1)`

rather than:
`/Platforms/1/LongName string "" is too short (length: 0, required minimum: 1); /Platforms/1/LongName ECMA 262 regex "[\w\-&'()\[\]/."#$%\^@!*+=,][\w-&'”)`

ingest-app/src/cmr/ingest/validation/validation.clj
mschmele: The error message in MMT as I saw it was semi-colon separated, and wanted to account for that as well as for when the messages were separated into `required: x actual: y`. This felt like the simplest way to decide whether or not to truncate or drop a message entirely

chris-durbin: The MMT message was created by this function:

```
(defn- create-success-status-message
  "If there are warnings, create a status string with warnings, otherwise status
  is nil"
  [warnings]
  (when (seq warnings)
    (str "Collection was updated successfully, but translating the collection "
         "to UMM-C had the following issues: "
         (string/join "; " warnings))))
```
It puts a semi-colon between each individual warning. So the semi-colon was not part of the original message in this case.

The warning message we wanted to remove from the example was:
/Platforms/1/LongName ECMA 262 regex "[\w\-&'()\[\]/."#$%\^@!*+=,][\w-&'”)

in this case I think the best thing to do would be to change the error message to:
/Platforms/1/LongName is in an invalid format rather than dropping it entirely. So it failed two validations, too short and invalid format.

We could probably do this with a string substitution of s/ECMA.*/is in an invalid format.

system-int-test/test/cmr/system_int_test/ingest/collection_misc_validation_test.clj
chris-durbin: So now I'm saying instead:
```
/Platforms/1/LongName string "" is too short (length: 0, required minimum: 1);
/Platforms/1/LongName string is in an invalid format.
```

ingest-app/src/cmr/ingest/validation/validation.clj
chris-durbin: I forgot to clarify:

The reason I don't want to drop it entirely is there may be cases where the only error returned is a regex error, so we definitely want to let them know which field was in a format that is considered invalid by the schema.

mschmele: Ohhhh okay I follow now. Thanks for clearing that up

system-int-test/test/cmr/system_int_test/ingest/collection_misc_validation_test.clj
```diff
@@ -77,6 +78,44 @@
...
```
chris-durbin: I think these tests are good. Could you add one more that verifies the full message string explicitly as well rather than just looking for certain substrings in the message? That way we can easily see whether it is human friendly.

common-lib/src/cmr/common/validations/json_schema.clj
```diff
@@ -20,10 +21,15 @@
...
+  (let [{:keys [instance message]} error-report
+        pointer (:pointer instance)]
     (str (when-not (str/blank? pointer)
-           (str pointer " "))
-         (:message error-report))))
+           (str pointer " "
+            (if (re-find #"ECMA" message)
+             (do
+              (info (str "UMM Validation error. Full message: " pointer " " message))
+              (str/replace message #"ECMA.*" "is an invalid format"))
```
chris-durbin: change to "is in an invalid format" to make more sense

system-int-test/test/cmr/system_int_test/ingest/collection_misc_validation_test.clj
```diff
@@ -77,6 +78,53 @@
...
+                                          :Category ""
+                                          :Term ""
+                                          :DetailedVariable ""
+                                          :Topic ""})]})]
+      (are3 [collection]
+       (let [ingest-response (d/ingest-umm-spec-collection "PROV1" collection {:format :umm-json :allow-failure? true})
+             validation-response (ingest/validate-concept (data-umm-c/collection-concept collection :umm-json))]
+         (is (nil? (first (map #(re-find #"(ECMA|regex)" %) (:errors ingest-response)))))
+         (is (nil? (first (map #(re-find #"(ECMA|regex)" %) (:errors validation-response)))))
+         (is (some? (map #(re-find #"is an invalid format" %) (:errors ingest-response))))
```
chris-durbin: These tests don't look like they've been updated to include the word "in". Are they failing?

PR #387, opened 2017-11-30
CMR-4621: Make MMT tests faster

chris-durbin:
Matthew and I found that there was an MMT test that was creating some invalid bulk update requests for CMR that resulted in a message on the CMR message queue that could not be processed. Found a bug in our local test broker that we weren't marking a request as failed after it had failed all retries. With this change the MMT tests went from:

Before:
```
Finished in 69 minutes 45 seconds (files took 4.31 seconds to load)
1550 examples, 0 failures
```
After:
```
Finished in 22 minutes 7 seconds (files took 4.08 seconds to load)
1550 examples, 0 failures
```

Now they only spend 11 seconds total in their calls to waiting for CMR indexing to complete:

```
|    11.370s |   0.025s |   0.147s |   452 | Helpers::CmrHelper#wait_for_cmr 
```

### DIFF
--- a/message-queue-lib/src/cmr/message_queue/test/queue_broker_wrapper.clj
+++ b/message-queue-lib/src/cmr/message_queue/test/queue_broker_wrapper.clj
@@ -183,7 +183,9 @@
         (handler msg)
         (update-message-queue-history broker-wrapper queue-name :process msg :success)
         (catch Exception e
-          (update-message-queue-history broker-wrapper queue-name :process msg :retry)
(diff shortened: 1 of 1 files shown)


chris-durbin:
> Nice catch. Is there a test we can write to mimic that scenario?
Sorry I missed this. Yes I think I can add a test. We might even want to add a validation to bulk update for the passed in concept-ids before we create a message on the queue. I'll think about this tomorrow, add a test, and/or write a ticket.

PR #388, opened 2017-12-01
Cmr 4320

laurenfrederick:
Added validation to give a warning when the ends at present flag is true and the end date is set on a collection.

Moved temporal extent validation and tests into separate files.

### DIFF
--- a/umm-spec-lib/src/cmr/umm_spec/validation/temporal_extent.clj
+++ b/umm-spec-lib/src/cmr/umm_spec/validation/temporal_extent.clj
@@ -1,0 +1,30 @@
+(ns cmr.umm-spec.validation.temporal-extent
+  "Defines validations for UMM temporal extents."
+  (:require
+   [clj-time.core :as time]
+   [cmr.common.validations.core :as validations-core]
+   [cmr.umm-spec.date-util :as date]
+   [cmr.umm-spec.validation.umm-spec-validation-utils :as util]))
+
+(defn- range-date-time-validation
+  "Defines range-date-time validation"
+  [field-path value]
+  (let [{:keys [BeginningDateTime EndingDateTime]} value]
+    (when (and BeginningDateTime EndingDateTime (time/after? BeginningDateTime EndingDateTime))
+      {field-path [(format "BeginningDateTime [%s] must be no later than EndingDateTime [%s]"
+                           (str BeginningDateTime) (str EndingDateTime))]})))
+
+(defn- temporal-end-date-in-past-validator
+  "Validate that the end date in TemporalExtents is in the past"
+  [field-path value]
+  (when (and value (not (date/is-in-past? value)))
+    {field-path [(str "Ending date should be in the past. Either set ending date to a date "
+                      "in the past or remove end date and set the ends at present flag to true.")]}))
+
+(def temporal-extent-warning-validation
+  {:RangeDateTimes (validations-core/every {:BeginningDateTime util/date-in-past-validator
+                                            :EndingDateTime temporal-end-date-in-past-validator})
+   :SingleDateTimes (validations-core/every util/date-in-past-validator)})
+
+(def temporal-extent-validation
+  {:RangeDateTimes (validations-core/every range-date-time-validation)})
--- a/umm-spec-lib/src/cmr/umm_spec/validation/temporal_extent.clj
+++ b/umm-spec-lib/src/cmr/umm_spec/validation/temporal_extent.clj
@@ -25,7 +25,6 @@
   "Validate that the collection ends at present flag and end date time are
   not both set"
   [field-path value]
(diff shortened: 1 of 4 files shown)


umm-spec-lib/src/cmr/umm_spec/validation/temporal_extent.clj
```diff
@@ -0,0 +1,42 @@
+(ns cmr.umm-spec.validation.temporal-extent
```
chris-durbin: 👍

umm-spec-lib/src/cmr/umm_spec/validation/temporal_extent.clj
```diff
@@ -0,0 +1,42 @@
...
+    {field-path [(str "Ending date should be in the past. Either set ending date to a date "
+                      "in the past or remove end date and set the ends at present flag to true.")]}))
+
+(defn- ends-at-present-validation
+  "Validate that the collection ends at present flag and end date time are
+  not both set"
+  [field-path value]
+  (when (and (= true (:EndsAtPresentFlag value))
+             (seq (:RangeDateTimes value))
+             (not (some nil? (map :EndingDateTime (:RangeDateTimes value)))))
```
chris-durbin: Nit: Maybe easier to read as:
```(seq (keep :EndingDateTime (:RangeDateTimes value)))```
I just have trouble with double-negatives I think, nothing really wrong with the code.

File: umm-spec-lib/src/cmr/umm_spec/validation/temporal_extent.clj [hunk 3]
```diff
@@ -0,0 +1,42 @@
...
+    {field-path [(str "Ends at present flag is set to true, but an ending date is "
+                      "specified. Remove the latest ending date or set the ends at "
+                      "present flag to false.")]}))
+
+(def temporal-extent-warning-validation
```
chris-durbin: Missing docstrings for the next two vars.

umm-spec-lib/src/cmr/umm_spec/validation/umm_spec_collection_validation.clj
```diff
@@ -63,24 +51,12 @@
...
-                             :EndingDateTime temporal-end-date-in-past-validator})
-   :SingleDateTimes (v/every vu/date-in-past-validator)})
-
 (def collection-validation-warnings
  "Defines validations for collections that we want to return as warnings and not
  as failures"
  {:RelatedUrls (v/every url/related-url-validations)
   :Projects project/projects-warning-validation
-  :TemporalExtents (v/every temporal-extent-warning-validation)
+  :TemporalExtents (v/every temporal-extent/temporal-extent-warning-validation)
```
chris-durbin: If you pass in the strict validation header do we return failures if there is any warning? I don't see a test for it, which we should probably add if that's how it works.

laurenfrederick: Are you talking about the validate-umm-c header? If so, there is a test in collection-umm-spec-validation-test (cmr.system-int-test.ingest.collection-umm-spec-validation-test) that tests that the warnings are returned as errors when the header is true.

umm-spec-lib/src/cmr/umm_spec/validation/temporal_extent.clj
laurenfrederick: I don't think that is the same logic? I am looking to make sure that there is at least one date range with an ending date time of nil.

umm-spec-lib/src/cmr/umm_spec/validation/temporal_extent.clj
laurenfrederick: Fixed, thanks.

umm-spec-lib/src/cmr/umm_spec/validation/umm_spec_collection_validation.clj
chris-durbin: I think that's the header - so this warning would be treated as an error in that case? I'm ok with it if there's existing tests, I just wasn't sure of the behavior.

umm-spec-lib/src/cmr/umm_spec/validation/temporal_extent.clj
chris-durbin: Oh, I retract my comment and code.

umm-spec-lib/src/cmr/umm_spec/validation/umm_spec_collection_validation.clj
laurenfrederick: Yes, in that case the warning would be treated as an error. There is a test for the header, but not specific to every validation warning.

PR #391, opened 2017-12-04
CMR-4625

ygliuvt:
Fixed the erroneous error messages when running lein modules commands.

### DIFF
--- a/dev-system/project.clj
+++ b/dev-system/project.clj
@@ -59,9 +59,9 @@
   :repl-options {
     :init-ns user
     :timeout 300000
-    :welcome ~(do
-               (println (slurp "resources/text/banner.txt"))
-               (println (slurp "resources/text/loading.txt")))}
+    :welcome (do
+              (println (slurp "resources/text/banner.txt"))
+              (println (slurp "resources/text/loading.txt")))}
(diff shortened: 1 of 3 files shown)


dev-system/project.clj
```diff
@@ -59,9 +59,9 @@
   :repl-options {
     :init-ns user
     :timeout 300000
-    :welcome ~(do
-               (println (slurp "resources/text/banner.txt"))
-               (println (slurp "resources/text/loading.txt")))}
+    :welcome (do
```
chris-durbin: Out of curiosity - what did this fix?

oubiwann: @ygliuvt Mentioned this to me on a pairing session we just had ... seems that this was the cause for some of the error messages (about project.clj files) we've seen when running `lein install!`.

I learned something here, though -- I though that all function calls in a project.clj file had to be escaped, since lein treats the contents of project.clj as literals (which is why you don't have to escape lists in project.clj file, e.g. you do `(:a :b)` not `'(:a :b)`).

Of course, I'll have to test this to be totally convinced the fix works :-)

ygliuvt: yes, what @oubiwann said.

PR #392, opened 2017-12-05
CMR-4400 changes platform longname and shortname validation

daniel-zamora:
In this PR, we want to prevent KMS validation on Platform LongName when the LongName is nil due to some restrictions with ECHO-10.  We still want to validate ShortName in that case.  If both LongName and ShortName are present, then the validation should work as before.

### DIFF
--- a/common-app-lib/src/cmr/common_app/services/kms_lookup.clj
+++ b/common-app-lib/src/cmr/common_app/services/kms_lookup.clj
@@ -46,6 +46,7 @@
     field-to-compare m
   }"
   [m fields-to-compare]
+  (proto-repl.saved-values/save 2)
   (if (map? m)
     (->> (select-keys m fields-to-compare)
          util/remove-nil-keys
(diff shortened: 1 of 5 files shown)


common-app-lib/test/cmr/common_app/test/services/kms_lookup.clj
```diff
@@ -109,7 +109,13 @@
                        :subregion-2 "CHAD" :subregion-3 "AOUZOU"} "location1-uuid"
 
     "Lookup works without all matching fields present"
-    :spatial-keywords {:category "CONTINENT" :type "AFRICA"} "location2-uuid"))
+    :spatial-keywords {:category "CONTINENT" :type "AFRICA"} "location2-uuid")
+
+  ;; CMR-4400
+  (testing "Platform Shortname exists but Longname is nil"
```
chris-durbin: We should have four tests total.
1) Long name in KMS, and long name in the metadata - matches.
2) Nil long name in KMS and nil long name in the metadata - matches.
3) Nil long name in KMS, but a long name in the metadata - does not match.
4) Long name in KMS, but a nil long name in the metadata - does not match.

This looks to be test case #4 and it finds a match when it should not (if I'm reading this correctly).

daniel-zamora: 4 should match, 3 should not match, 2 should match, and 1 should match.

chris-durbin: Oh ok I read through the comments in the tickets and the acceptance criteria. I agree 4 should match now.

common-app-lib/src/cmr/common_app/services/kms_lookup.clj
```diff
@@ -140,16 +140,37 @@
   [kms-index location-string]
   (get-in kms-index [:locations-index (str/upper-case location-string)]))
 
-(defn lookup-by-umm-c-keyword
+(defn- remove-long-name-from-kms-index
```
chris-durbin: Add a docstring. Besides just explaining what it does, why we have this function would be useful to document.

PR #393, opened 2017-12-06
CMR-4622:  Added functionality to set user-id in bulk update.

sxu123:
Turns out bulk update and regular ingest/update are not going through the same ingest-collection function, in which the user-id is set.  
Bulk update doesn't contain the functionality to set the user id. It just makes use of the user-id from the previous revision.  This is why the regular update will set the user-id, while the bulk update just carries  on whatever the user-id is from the previous revision.

### DIFF
--- a/ingest-app/src/cmr/ingest/api/bulk.clj
+++ b/ingest-app/src/cmr/ingest/api/bulk.clj
@@ -25,7 +25,7 @@
           content (api-core/read-body! body)]
       (api-core/verify-provider-exists request-context provider-id)
       (acl/verify-ingest-management-permission request-context :update :provider-object provider-id)
-      (let [task-id (bulk-update/validate-and-save-bulk-update request-context provider-id content)]
+      (let [task-id (bulk-update/validate-and-save-bulk-update request-context provider-id content headers)]
         (api-core/generate-ingest-response
           headers
           {:status 200
--- a/ingest-app/src/cmr/ingest/api/bulk.clj
+++ b/ingest-app/src/cmr/ingest/api/bulk.clj
@@ -4,6 +4,7 @@
    [clojure.data.xml :as xml]
    [clojure.string :as string]
    [cmr.acl.core :as acl]
+   [cmr.common.cache :as cache]
    [cmr.common.log :refer [debug info warn error]]
    [cmr.common.mime-types :as mt]
    [cmr.common.services.errors :as srvc-errors]
@@ -12,7 +13,22 @@
    [cmr.ingest.config :as ingest-config]
    [cmr.ingest.data.bulk-update :as data-bulk-update]
    [cmr.ingest.services.bulk-update-service :as bulk-update]
-   [cmr.ingest.services.ingest-service :as ingest]))
+   [cmr.ingest.services.ingest-service :as ingest]
+   [cmr.transmit.echo.tokens :as tokens]))
+
+(def user-id-cache-key
+  "The cache key for the token user id cache"
+  :token-user-ids)
+
+(defn- get-user-id
+  "Get user id based on context and headers"
+  [context headers]
+  (if-let [user-id (get headers "user-id")]
+    user-id
+    (when-let [token (get headers "echo-token")]
+      (cache/get-value (cache/context->cache context user-id-cache-key)
+                       token
+                       (partial tokens/get-user-id context token)))))
 
 (defn bulk-update-collections
   "Bulk update collections. Validate provider exists, check ACLs, and validate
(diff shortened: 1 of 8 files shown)


system-int-test/src/cmr/system_int_test/utils/ingest_util.clj
```diff
@@ -495,13 +495,18 @@
  ([provider-id request-body options]
   (let [accept-format (get options :accept-format :xml)
         token (:token options)
+        user-id (:user-id options)
+        headers (when (or user-id token)
+                  (util/remove-nil-keys
+                    {transmit-config/token-header token
+                     "user-id" user-id}))
```
sxu123: Even though the keys here are "echo-token" and "user-id", but when it gets to handle-bulk-update-event in the bulk_update_service.clj,  these keys are changed to :echo-token and :user-id.   Not sure how it happened. I can't modify api/core  set-user-id to retrieve these using the :echo-token :user-id because it will break existing tests.

ingest-app/src/cmr/ingest/api/core.clj
```diff
@@ -181,6 +181,20 @@
...
+   at handle-bulk-update-event"
+  [concept context headers]
+  (assoc concept :user-id
+         (if-let [user-id (get headers :user-id)]
+           user-id
+           (when-let [token (get headers :echo-token)]
+             (cache/get-value (cache/context->cache context user-id-cache-key)
+                              token
+                              (partial tokens/get-user-id context token))))))
+
```
sxu123: See my comment below under ingest_util.clj

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
```diff
@@ -7,6 +7,7 @@
    [cmr.common.services.errors :as errors]
    [cmr.common.time-keeper :as time-keeper]
    [cmr.common.validations.json-schema :as js]
+   [cmr.ingest.api.core :as api-core]
```
chris-durbin: Services namespaces should not require API namespaces (only the other direction).

I'd move the `set-user-id-for-bulk-update` function into the `ingest-app/src/cmr/ingest/api/bulk.clj` namespace since it is not a core function - only related to the bulk update API. Instead of passing headers into all the service functions you should pass in the user-id.

sxu123: Thank you. I just focused on getting it to work and forgot about the big picture.

ingest-app/src/cmr/ingest/api/bulk.clj
```diff
@@ -12,7 +13,22 @@
    [cmr.ingest.config :as ingest-config]
    [cmr.ingest.data.bulk-update :as data-bulk-update]
    [cmr.ingest.services.bulk-update-service :as bulk-update]
-   [cmr.ingest.services.ingest-service :as ingest]))
+   [cmr.ingest.services.ingest-service :as ingest]
+   [cmr.transmit.echo.tokens :as tokens]))
+
+(def user-id-cache-key
```
chris-durbin: This one is used outside of bulk ingest - it should stay in core.

ingest-app/src/cmr/ingest/api/bulk.clj
```diff
@@ -12,7 +13,22 @@
...
+
+(defn- get-user-id
+  "Get user id based on context and headers"
+  [context headers]
+  (if-let [user-id (get headers "user-id")]
+    user-id
+    (when-let [token (get headers "echo-token")]
+      (cache/get-value (cache/context->cache context user-id-cache-key)
+                       token
+                       (partial tokens/get-user-id context token)))))
```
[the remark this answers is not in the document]
chris-durbin: `user-id-cache-key` definitely belongs in core. If you want to move `get-user-id` there as well and refactor the existing `set-user-id` to use it that makes sense to me.

ingest-app/src/cmr/ingest/api/bulk.clj
sxu123: Oh, I just made a copy, I forgot bulk.clj is in api and can reference the api core :)  Done.

sxu123: Done. Removed user-id-cache-key and moved get-user-id to api core.  Refactored the set-user-id there.

ingest-app/src/cmr/ingest/api/core.clj
```diff
@@ -170,16 +170,20 @@
     {}
     {:threshold 1000}))
 
+(defn get-user-id
+  "Get user id based on context and headers"
+  [context headers]
+  (if-let [user-id (get headers "user-id")]
+    user-id
+    (when-let [token (get headers "echo-token")]
```
ygliuvt: I prefer `transmit-config/token-header` over `"echo-token"`. Not sure why you changed it.

sxu123: Done. Added user-id-header in transmit-config.

PR #394, opened 2017-12-12
CMR-4640: Allow list in update-value when doing CLEAR_ALL_AND_REPLACE…

sxu123:
… bulk update operation.


umm-spec-lib/src/cmr/umm_spec/field_update.clj
```diff
@@ -107,7 +107,14 @@
 
 (defmethod apply-umm-list-update :clear-all-and-replace
   [update-type umm update-field update-value find-value]
-  (assoc-in umm update-field [update-value]))
+  (as-> umm umm
+        (if (sequential? update-value)
+          (assoc-in umm update-field update-value)
+          (assoc-in umm update-field [update-value]))
+        (util/update-in-each umm update-field util/remove-nil-keys)
```
laurenfrederick: Lines 114-117 in this function are the same as the lines 103-106 of `apply-umm-list-update :add-to-existing`. Create a common function to be used by both functions.

sxu123: Done.

PR #395, opened 2017-12-12
CMR-4533 implements DOI warning validation for DOI formatting

daniel-zamora:
This PR adds validation on DOI formatting, the DOI should have a normal format of prefix/suffix where the prefix normally is dd.dddd, where d is a digit and the suffix is a string .

### DIFF
--- a/umm-spec-lib/src/cmr/umm_spec/validation/umm_spec_collection_validation.clj
+++ b/umm-spec-lib/src/cmr/umm_spec/validation/umm_spec_collection_validation.clj
@@ -27,8 +27,9 @@
 (defn doi-format-warning-validation
   "Validates that DOI is properly formatted."
   [field-path value]
-  (when-not (re-matches #"\d\d\.\d\d\d\d(\.\d+)*\/.+" value)
-    {field-path [(format "DOI [%s] is improperly formatted." value)]}))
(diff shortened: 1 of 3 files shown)


umm-spec-lib/src/cmr/umm_spec/validation/umm_spec_collection_validation.clj
```diff
@@ -24,6 +24,12 @@
       {field-path [(format "%%s minimum [%s] must be less than or equal to the maximum [%s]."
                            (str MinimumValue) (str MaximumValue))]})))
 
+(defn doi-format-warning-validation
+  "Validates that DOI is properly formatted."
+  [field-path value]
+  (when-not (re-matches #"\d\d\.\d\d\d\d(\.\d+)*\/.+" value)
+    {field-path [(format "DOI [%s] is improperly formatted." value)]}))
```
sxu123: Do we want to say what the proper format is?

daniel-zamora: I doubt it, there is another ticket out there for removing regular expressions from error messages because they aren't user friendly enough.  Thats the reason I left it out.

PR #397, opened 2017-12-13
CMR-4641: Allow list in update-value for FIND_AND_REPLACE operation i…

sxu123:
…n bulk update.


umm-spec-lib/src/cmr/umm_spec/field_update.clj
```diff
@@ -132,15 +132,19 @@
...
-                                                      update-value
-                                                      x))
-                                                  %))))
+  (if (seq (get-in umm update-field)) 
+    (let [update-value (if (sequential? update-value)
+                         (map #(util/remove-nil-keys %) update-value)
+                         (util/remove-nil-keys update-value))
+          ;; For each entry in update-field, if we find it using the find params,
+          ;; completely replace with update value
+          updated-umm (update-in umm update-field #(flatten
```
daniel-zamora: will mapcat work here instead of flatten?

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
```diff
@@ -29,6 +29,15 @@
                       :Topic "OCEANS"
                       :Term "MARINE SEDIMENTS"}]})
 
+(def science-keywords-umm-1
```
daniel-zamora: my personal preference here is to make a more descriptive name, instead of adding a number to the end.
perhaps ```multiple-science-keywords-umm``` or something like.  Also, looking through the file, it seems there are already many defs for multiple science keywords, used in many different tests.  Maybe we should consolidate this in some way so we don't have an ever increasing list of refs used in random tests.

umm-spec-lib/src/cmr/umm_spec/field_update.clj
daniel-zamora:
```
          updated-umm (update-in umm update-field #(mapcat (fn [x]
                                                                                                         (if (value-matches? find-value x)
                                                                                                           update-value
                                                                                                           x))
                                                                                                      %)))]
```

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
sxu123: I will change it to a more descriptive name. 
Consolidating will be more difficult because different tests have different needs. This particular one needs to contain a ScienceKeyword that exists in the update-value but not part of the keywords that's replaced.  
Personally I like to see different tests using different umm unless they need the exact same content, rather than keep adding more features to make a giant umm. It's easier to see what is expected.

umm-spec-lib/src/cmr/umm_spec/field_update.clj
sxu123: I got 22 failures after changing it to mapcat from bulk_update_test.clj.

sxu123: user=> (mapcat #(if % {:a 1}) [{:a 1} {:b 2}])
([:a 1] [:a 1])
user=> (flatten (map #(if % {:a 1}) [{:a 1} {:b 2}]))
({:a 1} {:a 1})
They are not the same.

sxu123: Since I replace one object with possibly an array of objects, update-field could become [object [objects]] and I need to flatten it to [object1 object2... objectn]

daniel-zamora: so I fiddled around and found a way to do it without flatten.
If anyone else thinks you should replace it with this then you can otherwise I'm going to set it to approved.
```
          updated-umm (update-in umm update-field (fn [values]
                                                    (reduce (fn [coll x]
                                                              (if (value-matches? find-value x)
                                                                (if (sequential? update-value)
                                                                  (concat update-value coll)
                                                                  (conj coll update-value))
                                                                (conj coll x)))
                                                            []
                                                            values)))]
```

PR #398, opened 2017-12-13
CMR-4128

ygliuvt:
Make the native id of granule tombstone reusable to reference a different parent collection.

### DIFF
--- a/metadata-db-app/src/cmr/metadata_db/data/memory_db.clj
+++ b/metadata-db-app/src/cmr/metadata_db/data/memory_db.clj
@@ -202,15 +202,16 @@
   (get-granule-concept-ids
    [this provider native-id]
    (let [provider-id (:provider-id provider)
-         concept-ids-fn (fn [gran]
-                          [(:concept-id gran) (get-in gran [:extra-fields :parent-collection-id])])]
-     (->> @concepts-atom
-          (filter (fn [c]
-                    (and (= :granule (:concept-type c))
-                         (= provider-id (:provider-id c))
-                         (= native-id (:native-id c)))))
-          first
-          concept-ids-fn)))
+         matched-gran (->> @concepts-atom
+                           (filter (fn [c]
+                                     (and (= :granule (:concept-type c))
+                                          (= provider-id (:provider-id c))
+                                          (= native-id (:native-id c)))))
+                           (sort-by :revisoin-id)
+                           last)
+         {:keys [concept-id deleted]} matched-gran
+         parent-collection-id (get-in matched-gran [:extra-fields :parent-collection-id])]
+     [concept-id parent-collection-id deleted]))
 
   (get-concept
    [db concept-type provider concept-id]
--- a/metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
+++ b/metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
@@ -294,11 +294,27 @@
   (get-granule-concept-ids
    [db provider native-id]
    (let [table (tables/get-table-name provider :granule)
-         result (su/find-one
-                 db (select [:concept-id :parent-collection-id]
-                            (from table)
-                            (where (by-provider :granule provider `(= :native-id ~native-id)))))]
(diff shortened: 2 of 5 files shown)


metadata-db-app/src/cmr/metadata_db/data/memory_db.clj
```diff
@@ -202,15 +202,16 @@
...
-         concept-ids-fn (fn [gran]
-                          [(:concept-id gran) (get-in gran [:extra-fields :parent-collection-id])])]
-     (->> @concepts-atom
-          (filter (fn [c]
-                    (and (= :granule (:concept-type c))
-                         (= provider-id (:provider-id c))
-                         (= native-id (:native-id c)))))
-          first
-          concept-ids-fn)))
+         matched-gran (->> @concepts-atom
```
mschmele: Nitpick: I think indentation is a little off here

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
```diff
@@ -294,11 +294,27 @@
   (get-granule-concept-ids
    [db provider native-id]
    (let [table (tables/get-table-name provider :granule)
-         result (su/find-one
-                 db (select [:concept-id :parent-collection-id]
-                            (from table)
-                            (where (by-provider :granule provider `(= :native-id ~native-id)))))]
-     [(:concept_id result) (:parent_collection_id result)]))
+         {:keys [provider-id small]} provider
+         stmt (if small
```
mschmele: Consider breaking this `if` statement out into its own function. Something like `stmt (prepare-granule-concept-id-query <args>)` would feel much easier to read to me

ygliuvt: I intentionally keep the sql statements together, so we can see what is different between the regular provider query and the small provider query. It is done the same way in other functions too in the same namespace. I think we should keep it this way.

metadata-db-app/src/cmr/metadata_db/data/memory_db.clj
ygliuvt: It is indented correctly based on my ATOM editor.

system-int-test/test/cmr/system_int_test/ingest/granule_ingest_test.clj
```diff
@@ -48,34 +56,53 @@
...
+          coll2 (d/ingest-umm-spec-collection
+                 "PROV1" (data-umm-c/collection {:EntryTitle "coll2"
+                                                 :ShortName "short2"
+                                                 :Version "V2"
+                                                 :native-id "native2"}))
+          gran1 (d/item->concept
+                 (dg/granule-with-umm-spec-collection
+                  coll1
+                  (:concept-id coll1)
+                  {:granule-ur "gran1"
```
daniel-zamora: is specifying the granule-ur necessary?

daniel-zamora: when I change gran2 ur to gran2 instead of gran1, the test passes.  When I remove the :granule-ur there is a unique one created by the granule-with-umm-spec-collection.  Either of these scenarios allows the test to pass.  Below is the relevant code in granule-with-umm-spec-collection:

```
         minimal-gran {:granule-ur (d/unique-str "ur")
                       :collection-ref coll-ref
                       ;; Including the parent collection concept id so it's available later.
                       :collection-concept-id collection-concept-id}
```

ygliuvt: Specifying granule-ur is unnecessary in this test. I am going to remove it. Thanks.

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
mschmele: That's fair, I just like to try keeping actual logic out of `let` blocks. Functions feel much easier to reason about if there are more, simpler functions.

PR #399, opened 2017-12-14
CMR-4627-4628: Corrected some typos in api doc and dif_util.clj.

sxu123:


umm-spec-lib/src/cmr/umm_spec/dif_util.clj
```diff
@@ -46,7 +46,7 @@
   ["GET DATA" "ECHO"] {:URLContentType "DistributionURL" :Type "GET DATA" :Subtype "ECHO"}
   ["GET DATA" "EDG"] {:URLContentType "DistributionURL" :Type "GET DATA" :Subtype "EDG"}
   ["GET DATA" "EOSDIS DATA POOL"] {:URLContentType "DistributionURL" :Type "GET DATA" :Subtype "EOSDIS DATA POOL"}
-  ["GET DATA" "GDS"] {:URLContentType "DistributionURL" :Type "GET DATA" :Subtype "EOSDIS DATA POOL"}
+  ["GET DATA" "GDS"] {:URLContentType "DistributionURL" :Type "GET DATA" :Subtype "GDS"}
```
daniel-zamora: If this wasn't failing cmr.umm-spec.test.validation.related-url/dif-valid-url-types test, we need to figure out why it wasn't and fix the test as well.

daniel-zamora: Looking at the test, it doesn't check if the keys correspond to the values in the map.  So the reason it didn't fail in the test is because EOSDIS DATA POOL was valid for DistributionURL URLContentType and GET DATA Type.  The test should probably also check the keys against the values of the map.

sxu123: Agreed. The reason I didn't add the tests was because I have some questions for Erich and I noticed that we didn't have any GDS subtypes for our example data.

sxu123: But overall, I think it's good to "double check", just in case someone accidentally change the mapping, that might be the only purpose, I think.

sxu123: However, I've identified one risky area: umm-url-type->dif-umm-content-type and dif-url-content-type->umm-url-types, they are independently defined. They shouldn't be. After changing that, and add the GDS subtype dif data, I think we are good to go.

eereiter: I don't think it is worth adding a test to validate that the map is correct.  Then we will just be creating another map to test the map and then the map is encoded twice.  We could add tests to make sure our validation is correct, to test if a sub-type is a valid value under the specific type, if we don't have these already.  Even if we had these tests, they would have all passed.

sxu123: The round-trip testing in the generate-and-parse.clj are supposed to cover the testing of this case, given the right data.  However it is not because the two conversion maps:
umm-url-type->dif-umm-content-type and dif-url-content-type->umm-url-types 
are not exact same maps with keys and values switched. Because of this, the round trip won't work in all cases, to make them work, I see the "expected" url-value is actually obtained from the round trip too. This made the tests pass, but won't detect the inconsistencies in the two conversion maps. Talk to Erich about it, he said there are other cases like this one too. Since the maps could change over time, it's hard to change the "expected" to be the true expected because we will have to handle the special cases when the entries in one map are missing from the other.

So, I'm thinking to just add a unit test to check the discrepancies of the two maps.

daniel-zamora: I was talking specifically about this one map.  That the keys are not mismatched against the values.

daniel-zamora: @eereiter it won't take another map to test. You just need to check that the type and subtype in the key is matched to the mapped subtype and type in the value.

sxu123: I figured out a way of testing this. The reason "expected" is obtained from round-trip is when there are entries present in one map and missing in the other, which will result in a default-url-value. 
But when the round-trip doesn't return a default-url-value, we should expect the "expected" to be the same as the original url-value from the original related-url.  This way, if there is any mis-match in the two maps mentioned above, it could still detect it.  Actually, after doing this, I found another discrepancy of the same nature: 
1. Both "DATA SET LANDING PAGE" and "DOI" map to the same ["VIEW DATA SET LANDING PAGE" nil] in umm-url-type->dif-umm-content-type
2. In dif-url-content-type->umm-url-types
["VIEW DATA SET LANDING PAGE" nil]  only maps to {:URLContentType "CollectionURL" :Type "DATA SET LANDING PAGE"}.

It's okay to map to something that doesn't exist in the other map, but not okay to map to something that exists in the other map, but is not mapped back to the same url type. @eereiter

umm-spec-lib/src/cmr/umm_spec/dif_util.clj
```diff
@@ -106,7 +106,7 @@
  "Map a combination of UMM URLContentType, Type, and Subtype (optional) to a dif url content type
  type and subtype. This is not dif->umm list flipped."
  {{:URLContentType "CollectionURL" :Type "DATA SET LANDING PAGE"} ["VIEW DATA SET LANDING PAGE" nil]
-  {:URLContentType "CollectionURL" :Type "DOI"} ["VIEW DATA SET LANDING PAGE" nil]
+  {:URLContentType "CollectionURL" :Type "DOI"} ["Erich needs to change this. VIEW DATA SET LANDING PAGE" nil]
```
sxu123: @eereiter, please see my comments above. we can't have both "DOI" and "DATA SET LANDING PAGE" map to the same thing - that's how I understand it at least.

umm-spec-lib/src/cmr/umm_spec/dif_util.clj
eereiter: Yes, it is.  Because DOI doesn't exist in DIF as a URL sub-type we mapped it to VIEW DATA SET LANDING PAGE.  To map it back if you only have one type VIEW DATA SET LANDING PAGE and it is mapped to DATA SET LANDING PAGE that is where it is mapped. There is no way to knows that this map used to be DOI - that knowledge doesn't exist in the record.

umm-spec-lib/src/cmr/umm_spec/dif_util.clj
eereiter: Yes we can, otherwise we would be dropping the URL data.

umm-spec-lib/src/cmr/umm_spec/dif_util.clj
sxu123: Thanks @eereiter for the quick response. I was afraid of this. In that case, there is really no way to use the two maps to cross check the "discrepancy", therefore there's no reference we can follow to verify that which key should map to which value in the map - the map itself is the rule. @chris-durbin, can we skip the testing for this?

sxu123: All I can do is to add a unit test to check
["GET DATA" "GDS"] matches {:URLContentType "DistributionURL" :Type "GET DATA" :Subtype "GDS"}, maybe that's all what Daniel is asking :)

PR #404, opened 2017-12-20
CMR-4632: When making bulk update status calls, the task-id and provi…

sxu123:
…der-id need to match in bulk_update_task_status table.

### DIFF
--- a/ingest-app/src/cmr/ingest/api/bulk.clj
+++ b/ingest-app/src/cmr/ingest/api/bulk.clj
@@ -123,7 +123,7 @@
   (let [{:keys [headers request-context]} request]
     (api-core/verify-provider-exists request-context provider-id)
     (acl/verify-ingest-management-permission request-context :read :provider-object provider-id)
-    (let [task-status (data-bulk-update/get-bulk-update-task-status-for-provider request-context task-id)
+    (let [task-status (data-bulk-update/get-bulk-update-task-status-for-provider request-context task-id provider-id)
           collection-statuses (data-bulk-update/get-bulk-update-collection-statuses-for-task request-context task-id)]
       (when (or (nil? task-status) (nil? (:status task-status)))
         (srvc-errors/throw-service-error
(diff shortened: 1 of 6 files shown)


ygliuvt:
From reading the issue description, the caller should only get the status of the task if he/she has READ permission on the provider that is associated with the task. This is not implemented in this PR. As a result, any one can see that task status of any provider's task as long as the caller has permission to the provider that is specified in the route. 

Given the requirement in this issue, we should reconsider the route definition of the API. We should remove the `provider-id` from the get task status route as we no longer mandate the connection between the task id and the provider id in the route.

ygliuvt:
I think we can change the get bulk update status route to something like: `%CMR-ENDPOINT%/bulk-update/status/<task-id>`. The checking of READ permission should be performed on the provider associated with the given task-id. If the task does not exist for the given task-id, return 404; if caller has no permission, return 401. 

I will leave it to @sxu123 if she wants to file a separate issue for this. I don't think there is a need to merge this PR as it is.

sxu123:
@ygliuvt. This PR only attempts to fix the mismatch of provider-id and task-id.  If yeou want to change the design to task-id driven rather than provider-id driven, you could write a new ticket for it.  The permission is not bypassed, it's checking the permission at the beginning, on the provider-id provided, just not on the provider-id associated with the task-id you suggested. it's a different approach, but not wrong.

sxu123:
I will change the error message to sth like "Bulk update task with task id [2] could not be found for provider id [PROV1]." instead of only saying the task is could not be found, to be more clear.

ygliuvt:
@sxu123, OK, I filed CMR-4669 to revisit the route for getting bulk update task status. If you fix the error message (which is still not ideal as it doesn't distinguish a totally non-existent task from an existing task that just not matching the given provider), I will approve.

sxu123:
@ygliuvt, Thanks. I know it's still not ideal, but I really don't want to have two db calls, one for task-id alone, and one for provider-id and task-id...  Once we change the design, there'll be no need for the "mismatch" check.

ygliuvt:
> @ygliuvt I think this is the right way to handle this. I don't think this is something that would be run into often at all, mostly when those internal to EED are debugging, which is how I found it in the first place. MMT would not run into this so it'd have to be someone with permissions for more than one provider directly calling the API. I think the message that is in there is good.

From an API design point of view, the current API is not ideal. It is not an issue of how often people will run into problems, but people could run into problems. I think the API suggested in CMR-4669 will eliminate these problems.

PR #405, opened 2017-12-22
CMR-4279, CMR-4674

ygliuvt:
As a client, I want to search collections by its associated services through the service unique identifier.
Fixed bulk-update-science-keywords test which fails intermittently in CI and local dev env.

### DIFF
--- a/system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
+++ b/system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
@@ -206,31 +206,30 @@
     ;; Initiate bulk update that shouldn't add any duplicates.
     (let [response (ingest/bulk-update-collections "PROV1" duplicate-body)
           _ (index/wait-until-indexed)
-          collection-response (ingest/bulk-update-task-status "PROV1" (:task-id response))]
-      (is (= "COMPLETE" (:task-status collection-response))))
+          {:keys [task-id]} response
+          collection-response (ingest/bulk-update-task-status "PROV1" task-id)]
+      (is (= "COMPLETE" (:task-status collection-response)))
 
-    (side/eval-form `(ingest-config/set-bulk-update-enabled! false))
-    ;; Kick off bulk update
-    (let [response (ingest/bulk-update-collections "PROV1" bulk-update-body)]
-      (is (= 400 (:status response)))
-      (is (= ["Bulk update is disabled."] (:errors response))))
-    ;; Wait for queueing/indexing to catch up
-    (index/wait-until-indexed)
-    (let [collection-response (ingest/bulk-update-task-status "PROV1" 2)]
-      (is (= 404 (:status collection-response)))
-      (is (= ["Bulk update task with task id [2] could not be found for provider id [PROV1]."]
-             (:errors collection-response))))
-    (let [collection-response (ingest/bulk-update-task-status "PROV1" 1)]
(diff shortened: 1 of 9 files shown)


system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
```diff
@@ -206,31 +206,30 @@
...
+          collection-response (ingest/bulk-update-task-status "PROV1" task-id)]
+      (is (= "COMPLETE" (:task-status collection-response)))
 
-    (side/eval-form `(ingest-config/set-bulk-update-enabled! false))
-    ;; Kick off bulk update
-    (let [response (ingest/bulk-update-collections "PROV1" bulk-update-body)]
-      (is (= 400 (:status response)))
-      (is (= ["Bulk update is disabled."] (:errors response))))
-    ;; Wait for queueing/indexing to catch up
-    (index/wait-until-indexed)
```
sxu123: Thanks for finding this cut and paste error.

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
```diff
@@ -206,31 +206,30 @@
...
-             (:errors collection-response))))
-    (side/eval-form `(ingest-config/set-bulk-update-enabled! true))
+      (side/eval-form `(ingest-config/set-bulk-update-enabled! false))
+      ;; Kick off bulk update when bulk update is disabled is not allowed
+      (let [response (ingest/bulk-update-collections "PROV1" bulk-update-body)]
+        (is (= 400 (:status response)))
+        (is (= ["Bulk update is disabled."] (:errors response))))
+      (let [collection-response (ingest/bulk-update-task-status "PROV1" 2)]
+        (is (= 404 (:status collection-response)))
+        (is (= ["Bulk update task with task id [2] could not be found for provider id [PROV1]."]
```
sxu123: Could you fix this hard-coded task id as well?

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
```diff
@@ -206,31 +206,30 @@
...
-      (is (= 404 (:status collection-response)))
-      (is (= ["Bulk update task with task id [1] could not be found for provider id [PROV2]."]
-             (:errors collection-response))))
-    (side/eval-form `(ingest-config/set-bulk-update-enabled! true))
+      (side/eval-form `(ingest-config/set-bulk-update-enabled! false))
+      ;; Kick off bulk update when bulk update is disabled is not allowed
+      (let [response (ingest/bulk-update-collections "PROV1" bulk-update-body)]
+        (is (= 400 (:status response)))
+        (is (= ["Bulk update is disabled."] (:errors response))))
+      (let [collection-response (ingest/bulk-update-task-status "PROV1" 2)]
```
sxu123: Since it's not guaranteed that the task-id generated in the previous bulkupdate to be 1, then it's possible that it's being 2. So after the disabling of the bulk update, querying task-id being 2 could actually return the previous task's status.

ygliuvt: @sxu123, I didn't write the test. I don't see much point in this check here. But just to blindly follow the setup, I think you are right, we should check on task status for task-id equal to `(+ 1 task-id)`. I will update the test.

PR #407, opened 2018-01-06
CMR-4607-4608.

sxu123:
This PR fixed the issues in both CMR-4607 and 4608.

### DIFF
--- a/system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
+++ b/system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
@@ -687,6 +687,45 @@
                 :VariableLevel1 "EMISSIONS"}]
               (:ScienceKeywords (:umm concept)))))))
 
+(deftest bulk-update-replace-with-identical-update-value-test
+  (let [concept-ids (ingest-collection-in-each-format find-replace-keywords-umm)
+        _ (index/wait-until-indexed)
+        bulk-update-body {:concept-ids concept-ids
+                          :name "TEST NAME"
+                          :update-type "FIND_AND_REPLACE"
+                          :update-field "SCIENCE_KEYWORDS"
+                          :find-value {:VariableLevel1 "CARBON MONOXIDE"}
+                          :update-value {:Category "EARTH SCIENCE"
+                                         :Topic "ATMOSPHERE"
+                                         :Term "AIR QUALITY"
+                                         :VariableLevel1 "CARBON MONOXIDE"}}
+        task-id (:task-id (ingest/bulk-update-collections "PROV1" bulk-update-body))]
+      (index/wait-until-indexed)
+      (let [collection-response (ingest/bulk-update-task-status "PROV1" task-id)]
+        (is (= "COMPLETE" (:task-status collection-response))))
+
+      ;; Check that each concept was updated
+      (doseq [concept-id concept-ids
+              :let [concept (-> (search/find-concepts-umm-json :collection
+                                                               {:concept-id concept-id})
+                                :results
+                                :items
+                                first)]]
+       (is (= 2
+              (:revision-id (:meta concept))))
+       (is (= "application/vnd.nasa.cmr.umm+json"
+              (:format (:meta concept))))
+       (is (= [{:Category "EARTH SCIENCE"
+                :Topic "ATMOSPHERE"
+                :Term "AIR QUALITY"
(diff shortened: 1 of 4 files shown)


umm-spec-lib/src/cmr/umm_spec/field_update.clj
```diff
@@ -167,38 +179,55 @@
...
+                                                (if (value-matches? find-value x)
+                                                  (data-center-update x update-value)
+                                                  x))
+                                              %))))
+
+
+(defn- apply-list-update-nested
+  "Apply the umm list update when the update-field is a supported nested field."
+  [update-type umm update-field update-value find-value partial-update-fn]
+  (if (re-matches #":find-and.*" (str update-type))
```
chris-durbin: We should have a var that contains a collection of the update-types that support a find value rather than using a regex. Then check if that update-type is in the list. Same for the other function below.

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
```diff
@@ -760,6 +762,44 @@
...
+              :let [concept (-> (search/find-concepts-umm-json :collection
+                                                               {:concept-id concept-id})
+                                :results
+                                :items
+                                first)]]
+       (is (= 1
+              (:revision-id (:meta concept))))
+       (is (not= "application/vnd.nasa.cmr.umm+json"
+              (:format (:meta concept)))))))
+
```
chris-durbin: I would add one more test that documents your comment in CMR-4608. If performing a find and update with the same value and that value is found in the collection metadata (so the UMM ends up being identical), we perform the update so that there is a new revision created and saved in UMM JSON format.

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
```diff
@@ -760,6 +762,44 @@
...
+                                         :VariableLevel1 "EMISSIONS"}}
+        task-id (:task-id (ingest/bulk-update-collections "PROV1" bulk-update-body))]
+      (index/wait-until-indexed)
+      (let [collection-response (ingest/bulk-update-task-status "PROV1" task-id)
+            concept-id (first concept-ids)
+            status-message (str "Collection with concept-id [" concept-id
+                                "] is not updated because no find-value found.")]
+        (is (= "COMPLETE" (:task-status collection-response)))
+        (is (= [{:concept-id concept-id,
+                 :status "FAILED",
```
chris-durbin: I see this is what the acceptance criteria states, but I don't think "FAILED" is the best status for the scenario where the find value is not in the metadata. I think something like "SKIPPED" would be more informative. If you agree we should discuss changing the original acceptance criteria with Kathy.

chris-durbin:
I think the most common cause of this will be case sensitivity in the find lookup. Could you also add a test that if the find value is in a different case, but the same when compared case insensitively it is still skipped?

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
sxu123: Yes. David wrote the ticket. He is fine with "SKIPPED". Done.

umm-spec-lib/src/cmr/umm_spec/field_update.clj
sxu123: Thanks. I thought about that too. Done.

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
```diff
@@ -760,6 +762,44 @@
...
+      ;; Check that each concept was not updated, no new revision-id, format is not changed to umm json.
+      (doseq [concept-id concept-ids
+              :let [concept (-> (search/find-concepts-umm-json :collection
+                                                               {:concept-id concept-id})
+                                :results
+                                :items
+                                first)]]
+       (is (= 1
+              (:revision-id (:meta concept))))
+       (is (not= "application/vnd.nasa.cmr.umm+json"
```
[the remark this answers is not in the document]

sxu123:
@chris-durbin  I incorporated the case sensitive case you asked for in the bulk-update-update-not-found-test

sxu123:
Double checked with MMT. They are fine with the new status.

PR #409, opened 2018-01-08
CMR-4537 - Resolve jar conflicts

oubiwann:
This is Part 2 of the version maintenance tasks. Before updating versions of library dependencies, I needed to fix the issues with the deps where the JVM was incapable of deterministically resolving a particular version of a given dep to include (e.g., when starting up the CMR).

Next up, Part 3: remaining deps version changes.


chris-durbin:
Can you help me understand this better:

1. How did you figure out that there were version conflicts (why do we want not have any problems with the way things are today)?
2. For every exclusion we add are we basically saying, I don't care what versions my dependencies want to use, they should use the explicit version I say in my dependencies?
3. When you update the versions in a subsequent PR, do you expect the number of exclusions to go up or down?

oubiwann:
1) Running `lein deps :tree` shows you where lein/Maven/the JVM detect ambiguous dependencies; often these prevent the JVM from predictably selecting a particular .jar file to load. As such, it could mean that in production, we don't really know which version is being used. Sometimes this is the source for non-deterministic and/or irregularly recurring bugs. In worst case scenarios, the JVM chooses a .jar file that doesn't contain a class file that the correct version does contain, at which point it bails. This is what I came across when trying to use chunks of the CMR from a third-party, open source project. Some CMR jar files were unusable :-(

2) Yeah, this is correct. What is also correct is "some projects don't maintain their dependencies and include buggy or aot-compiled code that messes with my project." There is no perfect path here ...

3) Neither, I hope. Part 2 of this work (this PR) is intended to definitively set the exclusions and explicit inclusions such that lein/Maven/the JVM do not detect any ambiguity when trying to resolve jar files. In this PR, I tried to change as few of the versions in our core deps as possible, to minimize potential issues. (This paid off, since these changes were fewer than when I'd previously attempted this for the CMR.) That being said, there's no way to tell if a newer version of one of our deps includes another, new dep of its own. So .jar file conflicts will still need to be checked. That being said, short of an unforeseen worse-than-worst-case scenario, I would expect any changes in exclusions to be fairly minimal in the next PR.

General note: good dependency management is one of the holy grails in software engineering; it doesn't exist in reality. It's a historically big and troubling problem in the JVM ecosystem, one that larger projects (like the CMR) suffer from more than others (for obvious reasons: due to the greater number of potentially conflicting dependencies). Even though tools like `lein` and `mvn` let you easily ignore conflicts (up to the point of JVM startup failures), we really shouldn't. Deps maintenance is something we should keep in the forefront of our minds when planning our own resources.

oubiwann:
Oh, I should also note (for others; I did already mention this to @chris-durbin) that the Clojure core dev team is trying to address some of the dependency hell in the JVM (at least as it impacts Clojure libs/apps) with its significant focus on related tooling work that's happening now.  This is being greeted with a great deal of enthusiasm by the rest of the Clojure community :-)

chris-durbin:
Thanks for all the notes. There is a strange legacy-services startup problem that occurs on really rare occasion where it can't find classes. If this is really decided at JVM start time and is non-deterministic that could explain that failure.

PR #410, opened 2018-01-09
CMR-4680: Add initial support for the include_facets=v2 parameter

chris-durbin:
This is the first step to supporting granule facets. Returns an empty facets object for the JSON format, and a 400 error when requesting v2 facets in any other format.

Subsequent tickets CMR-4681 through CMR-4687 will add the remaining functionality.

In addition to adding support for the parameter for granule searches, this PR separated out collection and granule specific logic from the main V2 facets capability so that we can reuse as much as possible from the collection V2 facets implementation when adding similar support for granules.

### DIFF
--- a/search-app/src/cmr/search/services/query_execution/facets/granule_v2_facets.clj
+++ b/search-app/src/cmr/search/services/query_execution/facets/granule_v2_facets.clj
@@ -1,0 +1,22 @@
+(ns cmr.search.services.query-execution.facets.granule-v2-facets
+  "Functions for generating v2 granule facets. Similar structure as v2 collection facets, but
+  granule fields. First major use case is supporting OPeNDAP virutal directories capability.")
+
+(def granule-facet-params->elastic-fields
+  {})
+
+(def granule-facet-fields
+  "Faceted fields for granules."
+  (keys granule-facet-params->elastic-fields))
+
+(def granule-fields->aggregation-fields
+  "Defines the mapping of granule parameter names to the aggregation parameter names."
+  {})
+
+(def granule-v2-facets-root
+  "Root element for the facet response"
+  {:title "Browse Granules"})
+
+(def v2-facets-result-field-in-order
+  "Defines the v2 facets result field in order"
+  [])
--- a/system-int-test/test/cmr/system_int_test/search/facets/granule_facets_v2_search_test.clj
+++ b/system-int-test/test/cmr/system_int_test/search/facets/granule_facets_v2_search_test.clj
@@ -22,12 +22,13 @@
       (is (= "Browse Granules" (:title facets)))))
 
   (testing "Only the json format supports V2 granule facets."
-    (let [unsupported-formats (remove #(or (= :timeline %) (= :json %))
-                                      (cqv/supported-result-formats :granule))
-          xml-error-string (str "<?xml version=\"1.0\" encoding=\"UTF-8\"?><errors><error>"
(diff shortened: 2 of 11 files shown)


oubiwann:
Thank you soooo much for separating the granule and collection logic ... been wanting to do that for ages.

system-int-test/test/cmr/system_int_test/search/facets/granule_facets_v2_search_test.clj
```diff
@@ -0,0 +1,56 @@
...
+   (let [query-params (merge search-params {:page-size 0 :include-facets "v2"})]
+     (get-in (search/find-concepts-json :granule query-params) [:results :facets]))))
+
+(deftest granule-facet-tests
+  (testing "Granule facets are returned when requesting V2 facets in JSON format."
+    (let [facets (search-and-return-v2-facets)]
+      (is (= "Browse Granules" (:title facets)))))
+
+  (testing "Only the json format supports V2 granule facets."
+    (let [unsupported-formats (remove #(or (= :timeline %) (= :json %))
```
ygliuvt: Why do we remove `:timeline` format from the test? Does it return a different error message?
The remove pred check can be done as `#(#{:timeline :json} %)` which is more readable than `or`.

sxu123: I was gonna ask the same.

chris-durbin: Don't tell me #(#{:timeline :json} %) is more readable than anything! 😆 

I removed timeline because we don't support .timeline as an extension and I couldn't be bothered to figure out how to test that it also returns the correct message. Now that I've been called out, I'll take a look to see if I can test it, or worst case add a comment.

chris-durbin:
My plan for documentation was to start documenting as part of CMR-4682 where we actually start returning non-empty facets.

search-app/src/cmr/search/services/query_execution/facets/granule_v2_facets.clj
```diff
@@ -0,0 +1,22 @@
...
+  "Faceted fields for granules."
+  (keys granule-facet-params->elastic-fields))
+
+(def granule-fields->aggregation-fields
+  "Defines the mapping of granule parameter names to the aggregation parameter names."
+  {})
+
+(def granule-v2-facets-root
+  "Root element for the facet response"
+  {:title "Browse Granules"})
```
sxu123: New to facet code. Why is it that granule-v2-facets-root doesn't need a type while collection-v2-facets-root has it?  Is it because collection includes granules? Is there any relationship between granule facets and collection facets?

system-int-test/test/cmr/system_int_test/search/facets/granule_facets_v2_search_test.clj
sxu123: What is #(#{:timeline :json} %)? Stay with "or" please :)

search-app/src/cmr/search/services/query_execution/facets/granule_v2_facets.clj
chris-durbin: Granules will eventually - this is just the first ticket that adds the minimal support.

https://bugs.earthdata.nasa.gov/browse/CMR-4684 is when it will get added in. The whole capability requires CMR-4680 through CMR-4687 to be implemented.

PR #412, opened 2018-01-10
CMR-4537: Version bumps

oubiwann:
This is the final PR in the series of tasks that aimed to bring our dependencies up to date and in an explicit, non-conflicted state (prevent non-determinate .jar file loads).

This PR aimed to only upgrade to safe, minor-version bumps, no major-version bumps. Even though Clojure has bumped its minor version from 1.8 to 1.9, I did not try to upgrade Clojure. The biggest impact will likely be clj-http when we migrate from 2.x to 3.x; I believe, however, that @ygliuvt has tackled this a bit in a branch of his own, and it might not be as big a deal as I had anticipated, if we're only making minimal use of the client lib. Regardless, that's not a change in this PR.

Some libs with minor-version bumps actually had breaking API changes; these I had to revert (and did so in their own commits, so we have a record of the problem versions).

Additionally, @chris-durbin has recently uncovered an issue with our previous version of the AWS Java SDK, and this branch should address that issue by upgrading to the latest.


oubiwann:
Even as I wrote that description, I've found another issue in a version upgrade (the matrix library). I'll have to make a commit that reverts that version, too ...

access-control-app/project.clj
```diff
@@ -1,89 +1,85 @@
...
+        [lein-ancient "0.6.15"]
+        [lein-bikeshed "0.5.0"]
+        [lein-kibit "0.1.6"]
+        [venantius/yagni "0.1.4"]]}}
   :test-paths ["test" "int_test"]
   :aliases {"generate-static" ["with-profile" "static"
                                "run" "-m" "cmr.access-control.site.static" "all"]
             ;; Prints out documentation on configuration environment variables.
             "env-config-docs" ["exec" "-ep" "(do (use 'cmr.common.config) (print-all-configs-docs))"]
-            ;; Creates the checkouts directory to the local projects
```
chris-durbin: This is useful for anyone try to work on just the access-control project. I think we need to put this back in (and the cmr-deps and dev-cmr-deps need to be added back to support it).

oubiwann: Ah, I guess the `git grep` I did wasn't enough :-/

We can discuss this later, but this seems like an odd place to but some crucial functionality ... is there are reason it's just on `access-control`? Is this app worked on alone, more than any others?

Maybe this would find a better home somewhere more generally useful?

oubiwann: Also, by putting this change back in (which I'm doing right now!), we loose the ability to check the deps versions using `lein ancient`, due to the way they are dynamically constructed ...

oubiwann: Okay, I added the checkouts alias back in, but I did it a bit differently -- I used the approach that `dev-system` uses (it's cleaner).

oubiwann:
Okay, build is now successful.

oubiwann:
@chris-durbin This is ready for your re-review ...

PR #414, opened 2018-01-11
CMR-4681: Granule facets can only be returned when limiting search to a single collection

chris-durbin:


search-app/src/cmr/search/services/query_execution/facets/granule_v2_facets.clj
```diff
@@ -20,3 +22,22 @@
...
   [])
+
+(defn single-collection-validation
+  "Validates that the provided query is limited to a single collection. We do this to prevent
+  expensive aggregation queries that would have to run against more than one granule index."
+  [context]
+  (let [collection-ids (:query-collection-ids context)
+        collection-count (count collection-ids)]
+    (when-not (= 1 collection-count)
+      [(str "Granule V2 facets are limited to a single collection, but query matched "
```
ygliuvt: I think the following would be more readable, and there is no need to call `zero?`.
`[(format "Granule V2 facets are limited to a single collection, but query matched %s collections."
    (if (= 0 collection-count) "an undetermined number of" collection-count))...`

search-app/src/cmr/search/services/query_execution/facets/facets_v2_results_feature.clj
```diff
@@ -140,15 +140,20 @@
       (assoc (v2-facets-root concept-type) :has_children true :children v2-facets)
       (assoc (v2-facets-root concept-type) :has_children false))))
 
+(def facets-validation
+  "Mapping of concept type to the validator to run for that concept type."
+  {:granule granule-v2-facets/validator})
+
 (defmethod query-execution/pre-process-query-result-feature :facets-v2
-  [{:keys [query-string]} query _]
+  [{:keys [query-string] :as context} {:keys [concept-type facet-fields] :as query} _]
```
ygliuvt: This destructing in argument list is getting too complicated, move them inside the function as a `let` block.

search-app/src/cmr/search/services/query_execution/facets/granule_v2_facets.clj
chris-durbin: Me too. I only used `str` to avoid the long line or 😱  using `(format (str`. I'll go ahead and use `format` and the long line.

PR #419, opened 2018-01-16
CMR-4583: Removed EDSC stubbed data support for previous PI testing tasks

oubiwann:
This is also blocking some changes we want to make to support the local nexus jar repo in CI/CD (it can't currently handle SNAPSHOTS from Clojars).


oubiwann:
Thanks, guys!

PR #423, opened 2018-01-17
CMR-4630-4698: Added functionality to bulkupdate all the collections …

sxu123:
4630: Add functionality to bulkupdate all the collections for a provider.

 4689: Add the check to make sure all the concept-ids are associated with the provider.

### DIFF
--- a/ingest-app/docs/api.md
+++ b/ingest-app/docs/api.md
@@ -645,7 +645,7 @@ The following update types are supported:
 
 Bulk update post request takes the following parameters:
 
-  * Concept-ids (required) - a list of concept ids to update
+  * Concept-ids (optional) - a list of concept ids to update, which need to be associated with the provider the bulk update is initiated with. If not provided, all the collections for the provider will be updated.
   * Name (optional) - a name used to identify a bulk update task 
   * Update type (required) - choose from the enumeration: `ADD_TO_EXISTING`, `CLEAR_ALL_AND_REPLACE`, `FIND_AND_REPLACE`, `FIND_AND_REMOVE`, `FIND_AND_UPDATE`, `FIND_AND_UPDATE_HOME_PAGE_URL`
   * Update field (required) - choose from the enumeration: `SCIENCE_KEYWORDS`, `LOCATION_KEYWORDS`, `DATA_CENTERS`, `PLATFORMS`, `INSTRUMENTS`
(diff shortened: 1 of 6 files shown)


ingest-app/docs/api.md
```diff
@@ -645,7 +645,7 @@ The following update types are supported:
 
 Bulk update post request takes the following parameters:
 
-  * Concept-ids (required) - a list of concept ids to update
+  * Concept-ids (optional) - a list of concept ids to update, which need to be associated with the provider the bulk update is initiated with. If not provided, all the collections for the provider will be updated.
```
oubiwann: Hurray for docs updates!

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
```diff
@@ -113,6 +121,15 @@
   (validate-bulk-update-post-params json)
   (let [bulk-update-params (json/parse-string json true)
         {:keys [concept-ids]} bulk-update-params
+        concept-ids (if (seq concept-ids)
```
mschmele: I think this could be its own function, and would throw the service errors to avoid having the `_ (when-not` in the `let`

Something like:
```
(defn- filter-concept-ids
  [concept-ids]
  (let [concept-ids (if (seq concept-ids)...])
    (if (seq concept-ids)
      concept-ids
      (errors/throw-service-errors ...))))
```

mschmele:
Siwei answered my questions via our Slack - all looks good to me

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
```diff
@@ -105,6 +105,14 @@
                                    [(format "A find value must be supplied when the update is of type %s"
                                             update-type)]))))
 
+(defn- get-collection-concept-ids
+  "Returns a list of collection concept ids for a given provider-id"
+  [context provider-id]
+  (let [concepts (mdb/find-concepts context 
+                                    {:provider-id provider-id} 
+                                    :collection)]
+    (into [] (distinct (map :concept-id concepts)))))
```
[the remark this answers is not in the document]
sxu123: The only reason is to make the concept-ids obtained through provider-id having the same structure as the concept-ids when they pass in from the bulk update body.  But I've noticed that without it, the code worked too. so I will remove it and maybe use mapv just to be consistent with the structure.

PR #427, opened 2018-01-20
CMR-4609: Bulkupdate name is unique within provider.

sxu123:


ingest-app/src/migrations/010_add_unique_constraint_bulk_update_status.clj
```diff
@@ -0,0 +1,21 @@
+(ns migrations.010-add-unique-constraint-bulk-update-status
+  (:require
+   [clojure.java.jdbc :as j]
+   [config.migrate-config :as config]))
+
+(defn up
+  "Migrates the database up to version 10."
+  []
+  (println "migrations.010-add-unique-constraint-bulk-update-status up...")
+  (j/db-do-commands (config/db)
```
chris-durbin: This will fail if there is any existing data in the system which violates the constraint. Have you verified the names are unique in all environments?

I think it's fine to leave the migration like this, but we should try to address any data issues just prior to deployment.

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
```diff
@@ -893,13 +879,31 @@
...
+                          :find-value {:Topic "ATMOSPHERE"}
+                          :update-value {:Category "EARTH SCIENCE"
+                                         :Topic "ATMOSPHERE"
+                                         :Term "AIR QUALITY"
+                                         :VariableLevel1 "EMISSIONS"}}
+        response (ingest/bulk-update-collections "PROV1" bulk-update-body)
+        response2 (ingest/bulk-update-collections "PROV1" bulk-update-body)]
+    (is (= 200 (:status response)))
+    (is (= 422 (:status response2)))
+    (is (= ["Error creating bulk update task: Bulkupdate name needs to be unique within the provider."]
```
chris-durbin: Change Bulkupdate to Bulk update.

ingest-app/src/cmr/ingest/data/bulk_update.clj
```diff
@@ -136,9 +136,7 @@
                  (concat (map #(vector task-id % "PENDING") concept-ids) [:transaction? false]))
           task-id))
       (catch Exception e
```
ygliuvt: We don't have to catch and throw the exception if we are not going to do anything with it.

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj

PR #428, opened 2018-01-23
CMR-4630: Modified the indication of all collection update.

sxu123:
In bulk update body, pass in :concept-ids ["ALL"] to indicate all the collections for the provider will be updated, rather than not passing any concept-ids in the previous implementation.

### DIFF
--- a/ingest-app/src/cmr/ingest/services/bulk_update_service.clj
+++ b/ingest-app/src/cmr/ingest/services/bulk_update_service.clj
@@ -134,15 +134,14 @@
     err-msg))
 
 (defn- validate-concept-ids
-  "Throws exception if there exists invalid
-   concept-ids."
+  "Validate concept-ids. Raise error if there exist invalid concept-ids."
   [concept-ids]
   (when-not (= ["ALL"] (map string/upper-case concept-ids))
     (let [err-msgs (get-collection-concept-id-validation-err-msgs concept-ids)]
       (when (seq err-msgs)
         (errors/throw-service-errors
           :bad-request
-          [(string/join "; " err-msgs)])))))
+          [(string/join ", " err-msgs)])))))
 
 (defn- get-concept-ids
   "Get the concept-ids from either the concept-ids passed in or
--- a/ingest-app/src/cmr/ingest/services/bulk_update_service.clj
+++ b/ingest-app/src/cmr/ingest/services/bulk_update_service.clj
@@ -107,20 +107,23 @@
                                             update-type)]))))
 
 (defn- get-provider-concept-ids
-  "Returns a list of collection concept ids for a given provider-id"
+  "Returns a list of collection concept ids for a given provider-id.
+   Throws exception if non are returned."
   [context provider-id]
   (let [concepts (mdb/find-concepts context 
                                     {:provider-id provider-id} 
                                     :collection)]
-    (when (seq concepts)
-      (distinct (map :concept-id concepts)))))
+    (if (seq concepts)
+      (distinct (map :concept-id concepts))
+      (errors/throw-service-errors
+        :bad-request
+        ["No concept-ids are associated with the provider-id."]))))
 
-(defn- get-validated-concept-ids
-  "Returns the concept-ids that are valid. Throws exception if there exists invalid
+(defn- validate-concept-ids
+  "Throws exception if there exists invalid
    concept-ids."
-  [context concept-ids provider-id]
-  (if (= ["ALL"] (map string/upper-case concept-ids)) 
-    (get-provider-concept-ids context provider-id)
+  [concept-ids]
+  (when-not (= ["ALL"] (map string/upper-case concept-ids))
     (let [err-msgs 
            (for [id concept-ids
                  :let [err-msg (concepts/concept-id-validation id)
@@ -130,22 +133,18 @@
                                    [(str "Collection concept-id " id " does not start with C")]))]
                  :when err-msg]
              err-msg)]
-      (if (seq err-msgs)
+      (when (seq err-msgs)
         (errors/throw-service-errors
           :bad-request
-          [(string/join ", " err-msgs)])
-        concept-ids))))
+          [(string/join ", " err-msgs)])))))
 
 (defn- get-concept-ids
-  "Get the concept-ids from either the concept-ids passed in or 
-   from the provider. If both are empty, throws exception." 
+  "Get the concept-ids from either the concept-ids passed in or
+   from the provider."
   [context concept-ids provider-id]
-  (if-let [concept-ids (get-validated-concept-ids context concept-ids provider-id)]
-    concept-ids
-    (errors/throw-service-errors
-      :bad-request
-      [(str "Concept-ids not found - None provided in the request,"
-            " and none are associated with the provider-id either.")]))) 
+  (if (= ["ALL"] (map string/upper-case concept-ids))
+    (get-provider-concept-ids context provider-id)
+    concept-ids))
 
 (defn validate-and-save-bulk-update
   "Validate the bulk update POST parameters, save rows to the db for task
@@ -155,6 +154,7 @@
   (validate-bulk-update-post-params json)
   (let [bulk-update-params (json/parse-string json true)
         {:keys [concept-ids]} bulk-update-params
+        _ (validate-concept-ids concept-ids)
         concept-ids (get-concept-ids context concept-ids provider-id)
         bulk-update-params (assoc bulk-update-params :concept-ids concept-ids) 
         ;; Write db rows - one for overall status, one for each concept id
--- a/ingest-app/src/cmr/ingest/services/bulk_update_service.clj
+++ b/ingest-app/src/cmr/ingest/services/bulk_update_service.clj
@@ -108,18 +108,17 @@
 
 (defn- get-provider-collection-concept-ids
   "Returns a list of collection concept ids for a given provider-id.
-   Throws exception if non are returned."
+   Raise error if no collections are found."
   [context provider-id]
-  (let [concepts (mdb/find-concepts context 
-                                    {:provider-id provider-id
-                                     :latest "true"} 
-                                    :collection)
-        concepts (remove :deleted concepts)]
-    (if (seq concepts)
-      (distinct (map :concept-id concepts))
+  (let [collections (mdb/find-collections context 
+                                          {:provider-id provider-id
+                                           :latest true})
+        collections (remove :deleted collections)]
+    (if (seq collections)
+      (map :concept-id collections)
       (errors/throw-service-errors
         :bad-request
-        ["No concept-ids are associated with the provider-id."]))))
+        [(format "There are no un-deleted collections for provider-id [%s]." provider-id)]))))
 
 (defn- get-collection-concept-id-validation-err-msgs
   "Returns the concept-id validation msgs"
--- a/ingest-app/src/cmr/ingest/services/bulk_update_service.clj
+++ b/ingest-app/src/cmr/ingest/services/bulk_update_service.clj
@@ -118,9 +118,9 @@
   "Get the concept-ids from either the concept-ids passed in or 
    from the provider. If both are empty, throws exception." 
   [context concept-ids provider-id]
-  (if-let [concept-ids (if (seq concept-ids)
-                         concept-ids
-                         (get-provider-concept-ids context provider-id))]
+  (if-let [concept-ids (if (= "[\"ALL\"]" (string/trim (string/upper-case concept-ids)))
+                         (get-provider-concept-ids context provider-id)
+                         concept-ids)]
     concept-ids
     (errors/throw-service-errors
       :bad-request
--- a/ingest-app/src/cmr/ingest/services/bulk_update_service.clj
+++ b/ingest-app/src/cmr/ingest/services/bulk_update_service.clj
@@ -106,46 +106,51 @@
                                    [(format "A find value must be supplied when the update is of type %s"
                                             update-type)]))))
 
-(defn- get-provider-concept-ids
+(defn- get-provider-collection-concept-ids
   "Returns a list of collection concept ids for a given provider-id.
    Throws exception if non are returned."
   [context provider-id]
   (let [concepts (mdb/find-concepts context 
-                                    {:provider-id provider-id} 
-                                    :collection)]
+                                    {:provider-id provider-id
+                                     :latest "true"} 
+                                    :collection)
+        concepts (remove :deleted concepts)]
     (if (seq concepts)
       (distinct (map :concept-id concepts))
       (errors/throw-service-errors
         :bad-request
         ["No concept-ids are associated with the provider-id."]))))
 
+(defn- get-collection-concept-id-validation-err-msgs
+  "Returns the concept-id validation msgs"
+  [concept-ids]
+  (for [id concept-ids
+        :let [err-msg (concepts/concept-id-validation id)
+              err-msg 
+               (if err-msg
+                 err-msg
+                 (when-not (string/starts-with? id "C")
+                   [(format "Collection concept-id [%s] is invalid, must start with C" id)]))] 
+        :when err-msg]
+    err-msg))
+
 (defn- validate-concept-ids
   "Throws exception if there exists invalid
    concept-ids."
   [concept-ids]
   (when-not (= ["ALL"] (map string/upper-case concept-ids))
-    (let [err-msgs 
-           (for [id concept-ids
-                 :let [err-msg (concepts/concept-id-validation id)
-                       err-msg (if err-msg
-                                 err-msg
-                                 (when-not (string/starts-with? id "C")
-                                   [(str "Concept-id [" id 
-                                         "] is not a valid collection concept id, "
-                                         "must start with C")]))]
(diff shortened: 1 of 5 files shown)


system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_endpoint_test.clj
```diff
@@ -75,6 +75,18 @@
...
+            {:concept-ids ["all" ]
+             :name "TEST NAME 1"
+             :update-field "SCIENCE_KEYWORDS"
+             :update-type "ADD_TO_EXISTING"
+             :update-value {:Category "EARTH SCIENCE"
+                            :Topic "LAND SURFACE"
+                            :Term "SURFACE RADIATIVE PROPERTIES"
+                            :VariableLevel1 "REFLECTANCE"}}
+            400
             ["Concept-ids not found - None provided in the request, and none are associated with the provider-id either."]
```
chris-durbin: Is this test saying the provider has no collections so specifying "ALL" returns an error?

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_endpoint_test.clj
```diff
@@ -75,6 +75,18 @@
                             :Term "SURFACE RADIATIVE PROPERTIES"
                             :VariableLevel1 "REFLECTANCE"}}
             400
+            ["object has missing required properties ([\"concept-ids\"])"]
+
+            "All concept-ids"
+            {:concept-ids ["all" ]
```
chris-durbin: You test the blank spaces elsewhere. Should get rid of the blank space here.

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
```diff
@@ -188,14 +188,16 @@
         bulk-update-options1 {:token (e/login (s/context) "user1") :user-id "user2"}
         bulk-update-options2 {:token (e/login (s/context) "user1")}
         bulk-update-options3 {:user-id "user2"}
-        bulk-update-body {:update-type "ADD_TO_EXISTING"
+        bulk-update-body {:concept-ids ["ALL" ]
```
chris-durbin: You test the blank spaces on line 199. Should get rid of the blank space here.

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
```diff
@@ -118,9 +118,9 @@
   "Get the concept-ids from either the concept-ids passed in or 
    from the provider. If both are empty, throws exception." 
   [context concept-ids provider-id]
-  (if-let [concept-ids (if (seq concept-ids)
-                         concept-ids
-                         (get-provider-concept-ids context provider-id))]
+  (if-let [concept-ids (if (= "[\"ALL\"]" (string/trim (string/upper-case concept-ids)))
```
oubiwann: I'm not the biggest fan of nested conditionals if there are obvious, idiomatic alternatives, but this is fairly small ... and straight-forward.

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_endpoint_test.clj
sxu123: Yes.

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_endpoint_test.clj
sxu123: ok. done.

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
sxu123: Thanks. I will see if it looks better to have a separate function.

ygliuvt: This check looks off. It should just check `(map string/upper-case concept-ids)` equals `["ALL"]`. Also what if concept-ids is `["C1111-PROV1" "ALL"]`? I don't see this edge case being addressed.

chris-durbin: We should add a test for this case.

sxu123: I don't need to.  Honestly I thought about the case and when the user is mixing "ALL" with other concept-ids, it will be treated as the regular concept-ids. Only when the concept-ids = ["ALL"], it's treated as all collections.

sxu123: In which case, it will complain "ALL" concept-id doesn't exist.

ygliuvt: I think we should add validation of the concept-ids, any invalid concept-id will fail the request (including the mixed case). Only `["ALL"]` will be accepted as valid option for all collections.

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
```diff
@@ -114,13 +115,32 @@
...
+        concept-ids))))
+
 (defn- get-concept-ids
   "Get the concept-ids from either the concept-ids passed in or 
    from the provider. If both are empty, throws exception." 
   [context concept-ids provider-id]
-  (if-let [concept-ids (if (seq concept-ids)
-                         concept-ids
-                         (get-provider-concept-ids context provider-id))]
+  (if-let [concept-ids (get-validated-concept-ids context concept-ids provider-id)]
```
ygliuvt: I think we should separate the validation from getting concept ids function. This way, we would have a definitive error message as to if there is no collection for the given provider or the provided concept ids list is empty.

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
```diff
@@ -114,13 +115,32 @@
...
+  [context concept-ids provider-id]
+  (if (= ["ALL"] (map string/upper-case concept-ids)) 
+    (get-provider-concept-ids context provider-id)
+    (let [err-msgs 
+           (for [id concept-ids
+                 :let [err-msg (concepts/concept-id-validation id)
+                       err-msg (if err-msg
+                                 err-msg
+                                 (when-not (string/starts-with? id "C")
+                                   [(str "Collection concept-id " id " does not start with C")]))]
```
ygliuvt: Perhaps the error message should read something like "Concept id [S1111-PROV1] is not a valid collection concept id, must start with C."

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
```diff
@@ -114,13 +115,32 @@
...
+                 :when err-msg]
+             err-msg)]
+      (if (seq err-msgs)
+        (errors/throw-service-errors
+          :bad-request
+          [(string/join ", " err-msgs)])
```
ygliuvt: use a `;` to separate the error messages.

ingest-app/src/cmr/ingest/services/bulk_update_service.clj

ingest-app/src/cmr/ingest/services/bulk_update_service.clj

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
```diff
@@ -106,26 +107,46 @@
...
+   concept-ids."
+  [concept-ids]
+  (when-not (= ["ALL"] (map string/upper-case concept-ids))
+    (let [err-msgs 
+           (for [id concept-ids
+                 :let [err-msg (concepts/concept-id-validation id)
+                       err-msg (if err-msg
+                                 err-msg
+                                 (when-not (string/starts-with? id "C")
+                                   [(str "Concept-id [" id
```
ygliuvt: use `format` here would be better than `str`.

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
```diff
@@ -106,26 +107,46 @@
                                             update-type)]))))
 
 (defn- get-provider-concept-ids
```
ygliuvt: Rename this function to something like `get-provider-collection-concept-ids`.

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
```diff
@@ -106,26 +107,46 @@
...
-  "Returns a list of collection concept ids for a given provider-id"
+  "Returns a list of collection concept ids for a given provider-id.
+   Throws exception if non are returned."
   [context provider-id]
   (let [concepts (mdb/find-concepts context
```
ygliuvt: Since we can only update the latest revision of a collection, we should use `latest = true` to  only get the latest revision back. Also, I think this call will return tombstones (not certain) which we will want to skip.

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
```diff
@@ -106,26 +107,46 @@
...
+                                 (when-not (string/starts-with? id "C")
+                                   [(str "Concept-id [" id 
+                                         "] is not a valid collection concept id, "
+                                         "must start with C")]))]
+                 :when err-msg]
+             err-msg)]
+      (when (seq err-msgs)
+        (errors/throw-service-errors
+          :bad-request
+          [(string/join ", " err-msgs)])))))
```
ygliuvt: We should use `;` as separate as there is `,` in some of the error messages.

ingest-app/src/cmr/ingest/services/bulk_update_service.clj

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
sxu123: Used str because the line is getting too long...   Will use a separate function then *Sign*

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
sxu123: Done. I checked the metadata db call, it didn't exclude any deleted ones.

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
```diff
@@ -105,27 +106,52 @@
                                    [(format "A find value must be supplied when the update is of type %s"
                                             update-type)]))))
 
-(defn- get-provider-concept-ids
-  "Returns a list of collection concept ids for a given provider-id"
+(defn- get-provider-collection-concept-ids
+  "Returns a list of collection concept ids for a given provider-id.
+   Throws exception if non are returned."
```
ygliuvt: Change this line to "Raise error if no collections are found."

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
```diff
@@ -105,27 +106,52 @@
...
   [context provider-id]
   (let [concepts (mdb/find-concepts context
```
ygliuvt: Change this to call `mdb/find-collections`.

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
```diff
@@ -105,27 +106,52 @@
...
+  "Returns a list of collection concept ids for a given provider-id.
+   Throws exception if non are returned."
   [context provider-id]
   (let [concepts (mdb/find-concepts context 
-                                    {:provider-id provider-id} 
-                                    :collection)]
-    (when (seq concepts)
-      (distinct (map :concept-id concepts)))))
+                                    {:provider-id provider-id
+                                     :latest "true"}
```
ygliuvt: nit: `:latest true`.

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
```diff
@@ -105,27 +106,52 @@
...
-                                    {:provider-id provider-id} 
-                                    :collection)]
-    (when (seq concepts)
-      (distinct (map :concept-id concepts)))))
+                                    {:provider-id provider-id
+                                     :latest "true"} 
+                                    :collection)
+        concepts (remove :deleted concepts)]
+    (if (seq concepts)
+      (distinct (map :concept-id concepts))
```
ygliuvt: `distinct` can be removed now.

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
```diff
@@ -105,27 +106,52 @@
...
+      (errors/throw-service-errors
+        :bad-request
+        ["No concept-ids are associated with the provider-id."]))))
```
ygliuvt: Change the error message to something like "There is no collections for provider xyz."

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
```diff
@@ -105,27 +106,52 @@
...
+              err-msg 
+               (if err-msg
+                 err-msg
+                 (when-not (string/starts-with? id "C")
+                   [(format "Collection concept-id [%s] is invalid, must start with C" id)]))] 
+        :when err-msg]
+    err-msg))
+
+(defn- validate-concept-ids
+  "Throws exception if there exists invalid
```
ygliuvt: nit: put the docstring in one line.

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
```diff
@@ -105,27 +106,52 @@
...
+   concept-ids."
+  [concept-ids]
+  (when-not (= ["ALL"] (map string/upper-case concept-ids))
+    (let [err-msgs (get-collection-concept-id-validation-err-msgs concept-ids)]
+      (when (seq err-msgs)
+        (errors/throw-service-errors
+          :bad-request
+          [(string/join "; " err-msgs)])))))
```
ygliuvt: Sorry, we should change the `;` back to `,` as the individual errors are wrapped within brackets. Didn't see this earlier.

ingest-app/src/cmr/ingest/services/bulk_update_service.clj

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
sxu123: Both true and "true" work.  Done.

ingest-app/src/cmr/ingest/services/bulk_update_service.clj

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
sxu123: Changed to "There are no un-deleted collections for provider xyz."

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
```diff
@@ -860,6 +864,36 @@
       ;; revision-id not changed, format not changed, it's identical to the original-concepts.
       (is (= new-concepts original-concepts)))) 
 
+(deftest bulk-update-update-all-tomb-stone-test
```
ygliuvt: nit: `tombstone` instead of `tomb-stone`.

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
```diff
@@ -860,6 +864,36 @@
...
+                          :update-value {:Category "EARTH SCIENCE"
+                                         :Topic "ATMOSPHERE"
+                                         :Term "AIR QUALITY"
+                                         :VariableLevel1 "EMISSIONS"}}]
+    ;; perform bulk update, verify that one collection is skipped because find-value is not found.
+    (let [response (ingest/bulk-update-collections "PROV1" bulk-update-body)
+          _ (index/wait-until-indexed)
+          collection-response (ingest/bulk-update-task-status "PROV1" (:task-id response))]
+      (is (= "COMPLETE" (:task-status collection-response)))
+      (is (= "Task completed with 1 SKIPPED out of 1 total collection update(s)." (:status-message collection-response))))
```
ygliuvt: nit: it would be better to have this update succeed without any special messages. `skipped` is not as nice as total success.

sxu123: I only need to distinguish the case when there is a collection to be processed or not.

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
sxu123: Changed to "Raise error if there exists invalid concept-ids."

ingest-app/src/cmr/ingest/services/bulk_update_service.clj
sxu123: Okay...

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
sxu123: I'm confused. bulkupdate needs be separated, but tomb-stone need to be combined? Done.

PR #429, opened 2018-01-26
CMR-4708: Exclude deleted collections for bulk update.

sxu123:
1. Verified that all the collections that failed the bulkupdate are the deleted collections. 
2. The error happens at spec-core/parse-metadata (xpath/context metadata) 
3. Since the deleted collections  are already excluded in the update "ALL" case, we should exclude them in individual concept-id case as well, which will get rid of the error.  Still don't understand why the error only happens on the deleted collections.

### DIFF
--- a/system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
+++ b/system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
@@ -190,7 +190,7 @@
         bulk-update-options1 {:token (e/login (s/context) "user1") :user-id "user2"}
         bulk-update-options2 {:token (e/login (s/context) "user1")}
         bulk-update-options3 {:user-id "user2"}
-        bulk-update-body {:concept-ids concept-ids 
+        bulk-update-body {:concept-ids ["all"] 
                           :update-type "ADD_TO_EXISTING"
                           :update-field "SCIENCE_KEYWORDS"
                           :update-value {:Category "EARTH SCIENCE"
@@ -198,7 +198,7 @@
                                          :Term "ENVIRONMENTAL IMPACTS"
                                          :VariableLevel1 "HEAVY METALS CONCENTRATION"}}
         ;; CMR-4570 tests that no duplicate science keywords are created.
-        duplicate-body {:concept-ids concept-ids 
+        duplicate-body {:concept-ids ["ALL"] 
                         :update-type "ADD_TO_EXISTING"
                         :update-field "SCIENCE_KEYWORDS"
                         :update-value {:Category "EARTH SCIENCE"
--- a/system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
+++ b/system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
@@ -864,9 +864,11 @@
       ;; revision-id not changed, format not changed, it's identical to the original-concepts.
       (is (= new-concepts original-concepts)))) 
 
-(deftest bulk-update-update-all-tomb-stone-test
+(deftest bulk-update-update-all-tombstone-test
   (let [coll1 (data2-core/ingest-umm-spec-collection "PROV1" (data-umm-c/collection {:EntryTitle "E1"
                                                                                      :ShortName "S1"}))
+        coll2 (data2-core/ingest-umm-spec-collection "PROV1" (data-umm-c/collection {:EntryTitle "E2"
+                                                                                     :ShortName "S2"}))
         _ (index/wait-until-indexed)
         bulk-update-body1 {:concept-ids ["ALL"] 
                            :update-type "FIND_AND_UPDATE"
@@ -877,7 +879,7 @@
                                           :Topic "ATMOSPHERE"
                                           :Term "AIR QUALITY"
                                           :VariableLevel1 "EMISSIONS"}}
-        bulk-update-body2 {:concept-ids [(:concept-id coll1)] 
+        bulk-update-body2 {:concept-ids [(:concept-id coll1) (:concept-id coll2)] 
                            :update-type "FIND_AND_UPDATE"
                            :update-field "SCIENCE_KEYWORDS"
                            ;; Note: find-value is case-sensitive.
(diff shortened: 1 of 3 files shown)


system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
```diff
@@ -864,36 +864,58 @@
       ;; revision-id not changed, format not changed, it's identical to the original-concepts.
       (is (= new-concepts original-concepts)))) 
 
-(deftest bulk-update-update-all-tombstone-test
```
daniel-zamora: pretty sure tombstone is one word

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
```diff
@@ -864,36 +864,58 @@
...
     ;; collections from the provider. 
-    (let [response (ingest/bulk-update-collections "PROV1" bulk-update-body)
+    (let [response (ingest/bulk-update-collections "PROV1" bulk-update-body1)
           _ (index/wait-until-indexed)]  
       (is (= ["There are no un-deleted collections for provider-id [PROV1]."]
-             (:errors response))))))
-              
+             (:errors response))))
+
+    ;; perform a bulk update with the deleted collection's concept-id. It doesn't exclude tomb-stone one.
```
daniel-zamora: Can we include a collection that isn't deleted with the list of concept-ids?

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
```diff
@@ -864,36 +864,58 @@
...
     (is (= 200 (:status (ingest/delete-concept (data2-core/umm-c-collection->concept coll1 :echo10)))))
     (index/wait-until-indexed)
-    
+   
     ;; perform another bulk update, verify that the deleted collection is not included when getting all 
     ;; collections from the provider. 
-    (let [response (ingest/bulk-update-collections "PROV1" bulk-update-body)
+    (let [response (ingest/bulk-update-collections "PROV1" bulk-update-body1)
           _ (index/wait-until-indexed)]  
       (is (= ["There are no un-deleted collections for provider-id [PROV1]."]
```
daniel-zamora: I just want to note that this a strange message, I can foresee people being confused about it.

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
```diff
@@ -864,36 +864,58 @@
...
+    ;; Looks like when updating the tomb stone one it will cause error
+    ;; https://bugs.earthdata.nasa.gov/browse/CMR-4708 
+    (let [response (ingest/bulk-update-collections "PROV1" bulk-update-body2)
+          _ (index/wait-until-indexed)
+          collection-response (ingest/bulk-update-task-status "PROV1" (:task-id response))
+          collection-statuses (:collection-statuses collection-response)]
+      (is (= "COMPLETE" (:task-status collection-response)))
+      (is (= "Collection with concept-id [C1200000009-PROV1] is deleted. Can not be updated."  
+             (get (first collection-statuses) :status-message)))
+      (is (= "Task completed with 1 FAILED out of 1 total collection update(s)." (:status-message collection-response))))))
```
daniel-zamora: Why is the status FAILED instead of SKIPPED?

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
sxu123: SKIPPED is reserved for the good collections that are not touched because find-value is not found.

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
mschmele: Rather than un-deleted, maybe just use "restored" or "reconstructed"? Or is this meant to mean that the collection was never deleted in the first place?

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
daniel-zamora: ok

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
sxu123: There are regular collections, and collections that have been deleted(deleted column = 1 in database).  We don't want to return the deleted collections. I just feel like I have to let people know that just because we don't return any collections doesn't necessarily mean there aren't any collections for the provider.

sxu123: Change it to "There are no collections that are not deleted for provider-id [PROV1]."

mschmele: Ah okay. I think "There are no collections that have not been deleted for provider [provider-id]" sounds better, but that makes perfect sense.

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
sxu123: Sure. Done.

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
sxu123: Okay. Change it back :)

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
```diff
@@ -867,33 +867,67 @@
...
     (is (= 200 (:status (ingest/delete-concept (data2-core/umm-c-collection->concept coll1 :echo10)))))
     (index/wait-until-indexed)
-    
+   
     ;; perform another bulk update, verify that the deleted collection is not included when getting all 
     ;; collections from the provider. 
-    (let [response (ingest/bulk-update-collections "PROV1" bulk-update-body)
-          _ (index/wait-until-indexed)]  
-      (is (= ["There are no un-deleted collections for provider-id [PROV1]."]
+    (let [response (ingest/bulk-update-collections "PROV1" bulk-update-body1)
```
mschmele: I think it would make debugging easier if you put these let blocks inside a `testing` block

sxu123: Thanks. Done.

PR #431, opened 2018-01-26
CMR-4691: Update UMM-S

oubiwann:
This change pulls in the newly released v1.0 of the UMM-S schema.


system-int-test/src/cmr/system_int_test/data2/umm_spec_service.clj
```diff
@@ -11,12 +11,21 @@
 
 (def ^:private sample-umm-service
   {:Name "AIRX3STD"
-   :Type "OPeNDAP"
+   :LongName "OPeNDAP Service for AIRS Level-3 retrieval products"
+   :Type "WEB SERVICES"
```
chris-durbin: Shouldn't the Type still be "OPeNDAP"? I know this is just a sample, but with that LongName and URL and OPeNDAP being in the list of enums it seems like that's the right Type to use.

umm-spec-lib/src/cmr/umm_spec/json_schema.clj
```diff
@@ -12,7 +12,8 @@
    [cmr.umm-spec.versioning :as ver]))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
-;; Defined schema files
+;;; Defined schema files
```
chris-durbin: When should we use 3 semi-colons?

Our coding style right now has:
https://wiki.earthdata.nasa.gov/display/CMR/Coding+Style+Guidelines

Use two semicolons for a comment on it's own line.
Use one semicolon for a comment at the end of a line.

Most of the places I see changed here seem like they should be 2 to stay consistent.

system-int-test/src/cmr/system_int_test/data2/umm_spec_service.clj
oubiwann: No, that will actually fail validation. The type is now one of a specific set of enums.

umm-spec-lib/src/cmr/umm_spec/json_schema.clj
oubiwann: * 4 semi-colons is for the top of a file for whole module/ns notes
* 3 is for module/ns-level notes pertinent to a section
* 2 is for function-level notes
* 1 is for end of line notes

Common Lisp set this pattern and when Google bought out one of the biggest Lisp companies (air line reservation software) their new Lispers started publishing some nice Lisp docs. Semi-colons are addressed here:
 * https://google.github.io/styleguide/lispguide.xml#Comment_semicolons

The Clojure style guide borrows from this long-standing Lisp tradition:
 * https://github.com/bbatsov/clojure-style-guide

system-int-test/src/cmr/system_int_test/data2/umm_spec_service.clj
chris-durbin: Right I saw there was an enum, but OPeNDAP is in that list. If it's not in the list there is something wrong with our error message.
```
(str "/Type instance value (\"\") not found in enum "
                    "(possible values: [\"OPeNDAP\",\"THREDDS\",\"WEB SERVICES\","
                    "\"WCS\",\"WMS\",\"SOFTWARE PACKAGE\",\"TOOL\",\"WEB PORTAL\","
                    "\"International Web Portal\",\"MODEL\",\"NOT PROVIDED\"])")```

umm-spec-lib/src/cmr/umm_spec/json_schema.clj
chris-durbin: Sounds good to me, can you update our wiki page?

system-int-test/src/cmr/system_int_test/data2/umm_spec_service.clj
oubiwann: You are too right! I had missed that when I looked at the schema changes and the (non-OPeNDAP) example files from Simon. My statement about the failure of validation is incorrect!

I will change back to OPeNDAP ...

umm-spec-lib/src/cmr/umm_spec/json_schema.clj
oubiwann: Done.

oubiwann:
Build is green; merging ...

PR #433, opened 2018-01-30
CMR-4720: Don't use alias for building Ruby gems.

oubiwann:
This is to fix an issue that Chris as seen in the Heritage build.


oubiwann:
Quick background on this: the source of the issue hasn't been fully discovered; this alias works when called directly from the terminal (maybe there's an issue with the bang when called from a shell script?). Anyway, I made this same change to the `build-internal.sh` script and it fixed failing builds in Bamboo for CMR.

oubiwann:
Thanks, guys!

PR #434, opened 2018-01-30
CMR-4691: Additional UMM-S Updates

oubiwann:
Before we cut 1.0, I was asked to add one more change -- this PR is that change.

### DIFF
--- a/umm-spec-lib/src/cmr/umm_spec/test/umm_record_sanitizer.clj
+++ b/umm-spec-lib/src/cmr/umm_spec/test/umm_record_sanitizer.clj
@@ -104,7 +104,9 @@
   [records]
   (seq (for [record records]
          (if-let [contact-info (:ContactInformation record)]
-           (update-in record [:ContactInformation :RelatedUrls] sanitize-related-urls)
+           (update-in record
+                      [:ContactInformation :RelatedUrls]
+                      sanitize-related-urls)
            record))))
 
 (defn- sanitize-data-centers
@@ -160,9 +162,7 @@
       (sanitize-umm-record-contacts :ContactPersons)
       (sanitize-umm-record-contacts :ContactGroups)
       (sanitize-umm-record-related-url :CollectionCitations)
-      (sanitize-umm-record-related-url :PublicationReferences)
-      (update-in-each [:PublicationReferences] sanitize-online-resource)
-      (update-in-each [:CollectionCitations] sanitize-online-resource)))
+      (sanitize-umm-record-related-url :PublicationReferences)))
 
 (defn- sanitize-umm-number-of-instruments
   "Sanitize all the instruments with the :NumberOfInstruments for its child instruments"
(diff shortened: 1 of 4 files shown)


oubiwann:
@ygliuvt I'm getting a weird fail on a test after this change. Basically, it looks like the test expects an object reference but instead gets realized data ... here's an excerpt:

```
Test: cmr.umm-spec.test.umm-json/all-umm-s-records
File: umm_json.clj
Line number: 127
Expected: cmr.umm_spec.models.umm_service_models.UMM-S@da7444b
Actual: (#cmr.umm_spec.models.umm_service_models.UMM-S{:RelatedURL  ... [more record data]}
```

Have you seen something like this happen with generated UMM before?

oubiwann:
Hrm, seems to be a local issue ... it's building and passing just fine on Bamboo. Weird.

oubiwann:
I diffed the two results in separate files and got this (expected vs. actual):

```diff
        :Size -1.0,
        :Unit "GB",
        :Checksum ".ZL3Fd4\"r4"}}
-     {:_type RelatedUrlType,
+     {:Description nil,
       :URLContentType "PublicationURL",
       :Type "EXTENDED METADATA",
       :Subtype "NOMADS",
```

oubiwann:
@eereiter You made the changes to the source docs, right? Does this diff make any sense to you?

oubiwann:
Liu and I had a pairing session about this tonight; there's a chance that `minItems` needs to be added to these 2 new types, but it's still too early to make a genuine recommendation yet. I'll pick this up again in the morning.

oubiwann:
Liu, Erich, and I paired this morning and we tracked the issue down to some changes needed in our auto-generated tests. In particular, the following needed to be updated/created:
 * `sanitize-umm-record-urls`
 * `sanitize-umm-collection-citations`
 * `sanitized-umm-c-record`
 * `sanitized-umm-s-record`

oubiwann:
@mschmele Yeah,  some of the functions were originally written to be general UMM but actually had logic that was particular to UMM-C and was causing the UMM-S tests to fail as a result. Had to pull that out and put in new functions.

mschmele:
Gotcha - just wanted to clarify. Thanks!

umm-spec-lib/src/cmr/umm_spec/test/umm_record_sanitizer.clj
```diff
@@ -160,9 +162,7 @@
       (sanitize-umm-record-contacts :ContactPersons)
       (sanitize-umm-record-contacts :ContactGroups)
       (sanitize-umm-record-related-url :CollectionCitations)
-      (sanitize-umm-record-related-url :PublicationReferences)
-      (update-in-each [:PublicationReferences] sanitize-online-resource)
-      (update-in-each [:CollectionCitations] sanitize-online-resource)))
+      (sanitize-umm-record-related-url :PublicationReferences)))
```
sxu123: I see you removed sanitize-online-resource for PublicationReferences and CollectionCitation in sanitize-umm-record-urls, then add them back later in sanitized-umm-c-record which calls sanitize-umm-record-urls. That's good because I was wondering why you need to change sanitized-umm-c-record with the service schema update...  
The question is:  Did you do this to make the code look better or sanitize-umm-record-urls can't contain the sanitize-online-record for sanitized-umm-s-record to work? I'm guessing you want to categorize it better but leaving sanitize-online-record there won't hurt? so the real change is only on ContactInformation and umm-s-record sanitization only?

PR #435, opened 2018-01-31
CMR-4679: Added analyze-test task.

sxu123:
1. Added a comment to the ticket to explain why the keyword score is higher for TOMS products over OMI products: https://bugs.earthdata.nasa.gov/browse/CMR-4679

2. Added analyze-test task and modify the existing code to make it work so that we could use it to help analyzing the test failures.

3. Changed the default boost value of project to 1.0.  This change will make 2 more tests pass, including the one in this ticket, without failing any of the tests that passed previously.

### DIFF
--- a/search-relevancy-test/README.md
+++ b/search-relevancy-test/README.md
@@ -14,6 +14,7 @@ The tasks supported by this project are:
 * download-collections
 * relevancy-tests
 * boost-tests
+* analyze-test
 
 ### Download collections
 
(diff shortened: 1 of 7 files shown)


search-relevancy-test/README.md
```diff
@@ -50,6 +51,17 @@ Arguments:
 * -min-value (optional) - the smallest boost value, defaults to 1.0
 * -max-value (optional) - the highest boost value, defaults to 4.0
 
+### analyze-test
```
chris-durbin: Thanks for adding this, it is really useful.

search-relevancy-test/src/search_relevancy_test/anomaly_analyzer.clj
```diff
@@ -11,8 +11,8 @@
   "Get the test data from the CSV file - need the concept ids we want to look at
   and the search params. If the anomaly has multiple tests, we get the data for
   all of the tests."
-  [anomaly-number]
-  (let [csv-data (core/read-anomaly-test-csv)
+  [anomaly-number file-name]
```
chris-durbin: This is why every project should have tests. I didn't realize I broke this when adding the EDSC metrics based anomalies.

sxu123: Yeah. Maybe we could address the testing issue after moving it to its own repository.

PR #437, opened 2018-01-31
CMR-4700 - update gem version to leverage browse image markup for sch…

littleidiot40:


collection-renderer-lib/project.clj
```diff
@@ -14,7 +14,7 @@
    the hardcoded commit id during dev integration with cmr_metadata_preview project.
    The hardcoded commit id should be updated when MMT releases a new version of the gem."
   (or (System/getenv "CMR_METADATA_PREVIEW_COMMIT")
-      "a931085"))
+      "59dd7e0"))
```
mschmele: I think in the past, we generally have used the merge commit (`1ae7059fcbc` in this case), but I'm not sure that it actually matters. If it doesn't, then 👍

PR #441, opened 2018-02-05
CMR-4690: Reorg'ed plugin code and fix failing build

oubiwann:
As part of the work on removing services data in elasticsearch, I'm testing locally ... but I couldn't get the local build to complete successfully. In trying to troubleshoot this, I found the following minor issues:

* The namespaces were very awkwardly named with lots of hypens and/or underscores (depending upon whether the context was calling Java code or Clojure code)
* Awkward functions/function generators
* Ambiguous AOU compilation (I'm pretty sure this was the actual source of the problem)

To fix these, I:

* created a more standard hierarchical namespace
* used the `:name` option in `:gen-class` to compile the plugin code to more managable Java names
* created named functions to replay anonymous function usage/function generation
* cleaned up the `project.clj` file (in particular, the `:aot` settings)

All of this has been tested on a local elasticsearch with the newly generated plugin installed.

Note that these changes did not require a change in the project name, group id, or artifact id.


es-spatial-plugin/project.clj
```diff
@@ -86,18 +92,31 @@
...
-            "install-workload" ["do"
-                                "package,"
-                                "shell" "install_plugin_into_workload.sh" ~plugin-zip-name "spatialsearch-plugin"]
+            "install-aws" [
+              "do"
+              "package,"
+              ;; IP address is hard coded for now
+              "shell"
+                "../cmr-vms/elastic_aws/install_plugin.sh"
+                "54.193.23.62"
```
chris-durbin: We should look into this - I don't remember elastic_aws at all - I don't think it was related to moving CMR to the cloud. Was this part of prototyping a really long time ago (i.e. prior to the CMR)? We should ideally get rid of this.

es-spatial-plugin/src/cmr/elasticsearch/plugins/spatial/script/helper.clj
```diff
@@ -1,13 +1,14 @@
...
-           org.elasticsearch.search.lookup.FieldsLookup
-           org.elasticsearch.search.lookup.FieldLookup
-           org.elasticsearch.common.logging.ESLogger))
-
+(ns cmr.elasticsearch.plugins.spatial.script.helper
+  (:require
+   [clojure.string :as s]
+   [cmr.common.util :as u]
+   [cmr.spatial.serialize :as srl])
+  (:import
```
chris-durbin: We have a different style for imports now. I think something like 
`(org.elasticsearch.search.lookup DocLookup FieldLookup FieldsLookup)`

oubiwann:
@chris-durbin wrote:

> With JIRA down I can't look at the ticket, what is the driver for this change?

Not sure what I can say other than what I put in the PR description ... the build was failing, and I was having a hard time identifying the cause(s); the error was the infamous class not found one, and that has any number of potential causes (including miss use of underscores/dashes in mixed Clojure/Java code). I did the clean-up to use names without hyphens to better troubleshoot the build issues.

es-spatial-plugin/src/cmr/elasticsearch/plugins/spatial/script/helper.clj
oubiwann: Done.

es-spatial-plugin/project.clj
oubiwann: Yeah, looks like the last change in `crm-vms/elastic_aws` was in 2014-01-29. The IP doesn't respond to an HTTP `GET` ... anyway, seeing how our infrastructure has completely changed (probably several times), I'm 👍 for removal ...

oubiwann:
Update: ran lots of tests locally using this branch (while working on other stuff -- in particular, testing against an external Elasticsearch instance, running marvel, etc.) and it's looking good -- no issues. Bamboo build is nice and green.

Chatted with Chris, and we're going to put this on hold, since there's so much stuff happening this sprint. I'll rebase next sprint, and then we can put this branch into workload and see if there is any performance impact of the changes.

PR #443, opened 2018-02-06
CMR-4404: Cache the search-params to improve granule scroll search performance.

sxu123:
1. Issue:
If you do the following granule scroll search:
curl -i "http://localhost:3003/granules?provider=PROV&page_size=1&scroll=true"
and then do a subsequent granule scroll search using the scroll-id returned:
curl -i -H "CMR-Scroll-Id: -1261131662" "http://localhost:3003/granules?scroll=true"
The search-params with the provider=PROV info is lost for the subsequent scroll search, which results in the query search against all the indexes.  

2. Solution:
Cache the search-params together with the scroll-id in the cache and use the cached search-params if it's not nil.

3. Integration test issue:
Since the search-params doesn't affect the subsequent scroll search result, only impacts the performance, not sure how to test this. The only thing I could think of is to add the search-params to the search response so that the end-user can use the info to test.

### DIFF
--- a/search-app/src/cmr/search/api/concepts_search.clj
+++ b/search-app/src/cmr/search/api/concepts_search.clj
@@ -64,7 +64,8 @@
                         (rfh/printable-result-format result-format) (pr-str params)))
         search-params (lp/process-legacy-psa params)
         search-params (if cached-search-params
-                        (assoc cached-search-params :result-format (:result-format search-params))
+                        (assoc cached-search-params :result-format (:result-format search-params)
+                                                    :include_facets (:include_facets search-params))
                         search-params)
         results (query-svc/find-concepts-by-parameters ctx concept-type search-params)]
     (if (:scroll-id results)    
--- a/search-app/src/cmr/search/api/concepts_search.clj
+++ b/search-app/src/cmr/search/api/concepts_search.clj
@@ -53,16 +53,22 @@
   [ctx path-w-extension params headers body]
   (let [concept-type (concept-type-path-w-extension->concept-type path-w-extension)
         short-scroll-id (get headers (string/lower-case common-routes/SCROLL_ID_HEADER))
-        scroll-id (core-api/get-scroll-id-from-cache ctx short-scroll-id)
+        scroll-id-and-search-params (core-api/get-scroll-id-and-search-params-from-cache ctx short-scroll-id)
+        scroll-id (:scroll-id scroll-id-and-search-params)
+        cached-search-params (:search-params scroll-id-and-search-params)
         ctx (assoc ctx :query-string body :scroll-id scroll-id)
         params (core-api/process-params concept-type params path-w-extension headers mt/xml)
         result-format (:result-format params)
         _ (info (format "Searching for %ss from client %s in format %s with params %s."
                         (name concept-type) (:client-id ctx)
                         (rfh/printable-result-format result-format) (pr-str params)))
-        search-params (lp/process-legacy-psa params)
+        search-params (if cached-search-params
+                        cached-search-params 
+                        (lp/process-legacy-psa params))
         results (query-svc/find-concepts-by-parameters ctx concept-type search-params)]
-    (core-api/search-response ctx results)))
+    (if (:scroll-id results)    
+      (core-api/search-response ctx results search-params)
+      (core-api/search-response ctx results))))
 
 (defn- find-concepts
   "Invokes query service to find results and returns the response.
--- a/search-app/src/cmr/search/api/concepts_search.clj
+++ b/search-app/src/cmr/search/api/concepts_search.clj
@@ -62,10 +62,9 @@
         _ (info (format "Searching for %ss from client %s in format %s with params %s."
                         (name concept-type) (:client-id ctx)
                         (rfh/printable-result-format result-format) (pr-str params)))
-        search-params (lp/process-legacy-psa params)
         search-params (if cached-search-params
                         cached-search-params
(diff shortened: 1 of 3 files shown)


search-app/src/cmr/search/api/concepts_search.clj
```diff
@@ -53,16 +53,22 @@
...
+        cached-search-params (:search-params scroll-id-and-search-params)
         ctx (assoc ctx :query-string body :scroll-id scroll-id)
         params (core-api/process-params concept-type params path-w-extension headers mt/xml)
         result-format (:result-format params)
         _ (info (format "Searching for %ss from client %s in format %s with params %s."
                         (name concept-type) (:client-id ctx)
                         (rfh/printable-result-format result-format) (pr-str params)))
-        search-params (lp/process-legacy-psa params)
+        search-params (if cached-search-params
+                        cached-search-params
```
sxu123: @chris-durbin, I need to combine cache-search-params with the params, in some way. but not sure what. I need to at least preserve the result format from params, I remember you said in the subsequent scroll search, people can change the format for the result. Anything else?   Also, I tested in the baseline and found that changing the format in the subsequent scroll search doesn't work.  It works after at least one subsequent scroll search with the original format.

search-app/src/cmr/search/api/concepts_search.clj
```diff
@@ -53,16 +53,23 @@
...
+        cached-search-params (:search-params scroll-id-and-search-params)
         ctx (assoc ctx :query-string body :scroll-id scroll-id)
         params (core-api/process-params concept-type params path-w-extension headers mt/xml)
         result-format (:result-format params)
         _ (info (format "Searching for %ss from client %s in format %s with params %s."
                         (name concept-type) (:client-id ctx)
                         (rfh/printable-result-format result-format) (pr-str params)))
         search-params (lp/process-legacy-psa params)
+        search-params (if cached-search-params
+                        (assoc cached-search-params :result-format (:result-format search-params))
```
sxu123: Anything else needs to be kept other than :result-format?   I wrote the ticket https://bugs.earthdata.nasa.gov/browse/CMR-4737 for the baseline, but somehow after the change of this ticket, it worked fine.

search-app/src/cmr/search/api/concepts_search.clj
```diff
@@ -53,16 +53,24 @@
...
         ctx (assoc ctx :query-string body :scroll-id scroll-id)
         params (core-api/process-params concept-type params path-w-extension headers mt/xml)
         result-format (:result-format params)
         _ (info (format "Searching for %ss from client %s in format %s with params %s."
                         (name concept-type) (:client-id ctx)
                         (rfh/printable-result-format result-format) (pr-str params)))
         search-params (lp/process-legacy-psa params)
+        search-params (if cached-search-params
+                        (assoc cached-search-params :result-format (:result-format search-params)
+                                                    :include_facets (:include_facets search-params))
```
sxu123: I'm not sure if I should check if (:include_facets search-params) is not nil before assigning it.  Need to test in the baseline to see if all these result related parameters take effect in the subsequent scroll searches.

daniel-zamora: whats going on with the kebab and snake case mixing in the search params

chris-durbin: You should just overwrite the search params with the cached ones. We didn't actually want to allow people to change the search parameters after initiating the scrolling session, it was an unintended behavior of the original code.

sxu123: I noticed that too. It's just the way it is in the code...   They will be removed. I will only keep the cached-search-params here and ignore the search-params from the subsequent scroll search like Chris said. The feature of changing the result related params in the subsequent scroll search doesn't work in the baseline anyway.

sxu123: Thanks. Done.

system-int-test/test/cmr/system_int_test/search/scrolling_search_test.clj
```diff
@@ -79,6 +80,26 @@
...
+            ;; Do a subsequent scroll search with different format, provider, scroll and page-size.
+            ;; Verify the new search params are ignored. It still returns the same xml format.
+            subsequent-result (search/find-concepts-in-format
+                                mime-types/json
+                                :granule
+                                {:provider "PROV2" :scroll false :page-size 10}
+                                {:headers {routes/SCROLL_ID_HEADER scroll-id}})
+            subsequent-format (get-in subsequent-result [:headers :Content-Type])
+            subsequent-hits (get-in subsequent-result [:headers :CMR-Hits])]
+         (is (= hits subsequent-hits))
```
chris-durbin: Can you add in one more verification that the response body contains 2 items and not 10? Also for the test setup do we have more than 4 granules that would have matched the original query?

sxu123: The previous existing test already covers this: 
(testing "Subsequent searches gets original page-size results"
          (let [result (search/find-refs :granule
                                         {:scroll true :page-size 10}
                                         {:headers {routes/SCROLL_ID_HEADER scroll-id}})]
            (is (= (count all-grans) hits))
            (is (data2-core/refs-match? [gran3 gran4] result))))

sxu123: And yes. we have 5 granules that match the original query.

PR #445, opened 2018-02-09
CMR-4705: Add constraint to all granule queries in the api doc.

sxu123:
1. s/granules?/granules?collection_concept_id=%CMR-EXAMPLE-COLLECTION-ID%\&/g
2. Excluded the following cases:
    a. provider search
    b. short name search
    c. entry title search.
    d. cases where generic collection_concept_id doesn't make sense.
3. Modified static.clj to customize the collection_concept_id for each environment.

### DIFF
--- a/common-app-lib/src/cmr/common_app/static.clj
+++ b/common-app-lib/src/cmr/common_app/static.clj
@@ -204,11 +204,20 @@
         ;; Static HTML resources, typically API documentation which needs endpoint URLs replaced
         (GET ["/:page", :page #".*\.html$"] {headers :headers, {page :page} :params}
           (when-let [resource (site-resource page)]
-            (let [cmr-root (str public-protocol "://" (headers "host") relative-root-url)]
+            (let [cmr-root (str public-protocol "://" (headers "host") relative-root-url)
+                  provider (let [host-info (headers "host")]
+                             (case host-info
+                                   "cmr.earthdata.nasa.gov" "PODAAC"
+                                   "cmr.uat.earthdata.nasa.gov" "DEMO_89"
+                                   "cmr.sit.earthdata.nasa.gov" "DEMO_PROV"
+                                   "cmr.wl.earthdata.nasa.gov" "LAADS"
+                                   "PROV1"))
+                  cmr-example-collection-id (str "C1234567-" provider)]
               {:status 200
                :body (-> resource
                          slurp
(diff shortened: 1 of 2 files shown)


search-app/docs/api.md
```diff
@@ -1694,64 +1694,64 @@ __Sample response__
 
 ### <a name="granule-search-by-parameters"></a> Granule Search By Parameters
```
chris-durbin: We should add a short paragraph here to explain search performance for granule searches is significantly improved by including an identifier that limits the search to a certain collection or subset of collections. Examples of parameters which limit the scope of the search include collection_concept_id, short_name, entry_title, or provider.

common-app-lib/src/cmr/common_app/static.clj
```diff
@@ -204,11 +211,14 @@
         ;; Static HTML resources, typically API documentation which needs endpoint URLs replaced
         (GET ["/:page", :page #".*\.html$"] {headers :headers, {page :page} :params}
           (when-let [resource (site-resource page)]
-            (let [cmr-root (str public-protocol "://" (headers "host") relative-root-url)]
+            (let [cmr-root (str public-protocol "://" (headers "host") relative-root-url)
+                  site-example-provider (get site-provider-map (headers "host") "PROV1") 
+                  cmr-example-collection-id (str "C1234567-" site-example-provider)]
```
chris-durbin: I was really hoping we could provide a real collection concept ID for each environment so that for the most part they could execute the example query and get back results. It seems like we should be able to read from an environment variable for the collection ID.

Could we add
`(defconfig example-collection-id ...)`, or do you see an issue with that?

search-app/docs/api.md
sxu123: Thanks, Done.

common-app-lib/src/cmr/common_app/static.clj
sxu123: The issue is we have many dummy values and ids in the existing example queries. Only changing the new constraint to use the real values is not going to get the results back. Also, with defconfig, someone at the sites has to set it to a real value...    Communicated this to Chris. He said it's no big deal.  Since we can't make the whole collection concept id real, I could just use a dummy one instead of using the dummy C1234567- with the real provider name. But I will just leave it as is. In case we want to change all the values to real values in the examples.

PR #448, opened 2018-02-15
CMR-4746: Fix services to use transaction ID instead of revision ID when creating and deleting services.

chris-durbin:
Since services have associations we need to use the transaction ID when specifying the version for the Elasticsearch document that gets created. We do the same for collections, variables, and tags.

We had existing tests which failed when switching to transaction ID in one place, and then succeeded when switching to use transaction IDs everywhere.

### DIFF
--- a/indexer-app/src/cmr/indexer/data/elasticsearch.clj
+++ b/indexer-app/src/cmr/indexer/data/elasticsearch.clj
@@ -51,6 +51,12 @@
          (:transaction-id concept)
          (map :transaction-id (:variable-associations concept))))
 
+(defmethod get-elastic-version :service
+ [concept]
+ (apply max
+        (:transaction-id concept)
+        (map :transaction-id (:service-associations concept))))
+
 (defmethod get-elastic-version :default
   [concept]
   (:revision-id concept))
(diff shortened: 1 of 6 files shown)


message-queue-lib/src/cmr/message_queue/test/queue_broker_wrapper.clj
```diff
@@ -9,7 +9,7 @@
    [clojure.set :as set]
    [cmr.common.dev.record-pretty-printer :as record-pretty-printer]
    [cmr.common.lifecycle :as lifecycle]
-   [cmr.common.log :as log :refer (debug info warn error)]
+   [cmr.common.log :as log :refer (debug trace info warn error)]
```
oubiwann: Ha! Now I'm not the only one to have introduced a `trace` debug :evil-grin:

indexer-app/src/cmr/indexer/data/elasticsearch.clj
```diff
@@ -51,6 +51,12 @@
          (:transaction-id concept)
          (map :transaction-id (:variable-associations concept))))
 
+(defmethod get-elastic-version :service
+ [concept]
+ (apply max
+        (:transaction-id concept)
+        (map :transaction-id (:service-associations concept))))
+
```
oubiwann: So I do have a question related to this: do we need to change the `or` clause in the private function `get-elastic-id` below to also include a check for `:service`?

chris-durbin: We'd need to once we start indexing to an "all-revisions-index" for services. It looks like that will be next sprint with ticket https://bugs.earthdata.nasa.gov/browse/CMR-4657.

oubiwann: Got it -- that makes sense. I've been running into the all-revisions-indexing code this whole week and was wondering what we were going to do about that ... ;-)

PR #449, opened 2018-02-15
CMR-4291: Index UMM-S keywords

oubiwann:
For those (like me!) not familiar with  updated concepts to be indexed, I've created a wiki page that covers the basic process and the areas of CMR that need to be touched:
 * https://wiki.earthdata.nasa.gov/display/CMR/Updating+the+CMR+to+Make+Schema+Fields+Searchable

The code for this has landed, however I'm still working my way through the system integration tests:

* [x] Science Keywords
* [x] RelatedURL Keywords
* [x] Simple Keywords
* [x] ContactGroups Keywords
* [x] ContactPersons Keywords
* [x] Platforms Keywords
* [x] Service Keywords
* [x] Service Organizations Keywords

I'll check these off as they are completed,  and add a comment to the PR when they are all done.


oubiwann:
Okay, all the tests have been added now!

indexer-app/src/cmr/indexer/data/concepts/keyword_util.clj
```diff
@@ -20,5 +20,152 @@
...
+   ;; Simple multi-valued data
+   :AncillaryKeywords :AncillaryKeywords
+   :ContactGroups #(mapcat contact-group->keywords (:ContactGroups %))
+   :ContactPersons #(mapcat contact-person->keywords (:ContactPersons %))
+   :Platforms #(mapcat platform->keywords (:Platforms %))
+   :ScienceKeywords #(mapcat science-keyword->keywords (:ScienceKeywords %))
+   :ServiceKeywords #(mapcat service-keyword->keywords (:ServiceKeywords %))
+   :ServiceOrganizations #(mapcat service-organization->keywords (:ServiceOrganizations %))})
+
+(defn flatten-collections
```
chris-durbin: Could you add a docstring for this? The function name doesn't make a lot of sense to me - in one case you are flattening, in the other case you are creating a collection.

indexer-app/test/cmr/indexer/test/data/concepts/keyword_util.clj
```diff
@@ -0,0 +1,231 @@
+(ns cmr.indexer.test.data.concepts.keyword-util
+  "Functions for testing cmr.indexer.test.data.concepts.keyword-util namespace."
```
chris-durbin: testing `cmr.indexer.data.concepts.keyword-util`

indexer-app/test/cmr/indexer/test/data/concepts/keyword_util.clj
```diff
@@ -0,0 +1,231 @@
...
+           (sort
+            (keyword-util/concept-keys->keywords
+             sample-umm-service-concept schema-keys))))))
+
+(deftest concept-keys->keyword-text
+  (let [schema-keys [:LongName
+                     :Name
+                     :Version]]
+    (is (= "1 1.9 3 9 airs airx3std for level opendap opendap service for airs level-3 retrieval products products retrieval service"
```
chris-durbin: Does order matter when we perform keyword searches? I'm thinking for fields that are a string with multiple words. So someone might want to search for "AIRS level-3" for example. It looks like you are sorting alphabetically the full strings for a field so that if someone searched with multiple words it would be found?

system-int-test/src/cmr/system_int_test/data2/umm_spec_service.clj
```diff
@@ -71,3 +72,111 @@
          (service attribs)
          (assoc :provider-id provider-id :native-id native-id)
          umm-s->concept))))
+
+(defn contact-person
+  "Returns a ContactPersonType suitavke as an element in a
```
chris-durbin: You got lucky that was a real word.

system-int-test/test/cmr/system_int_test/search/service/service_search_test.clj
```diff
@@ -216,6 +218,326 @@
...
+
+      "Combination of keywords"
+      [svc1]
+      "Service Name-1"
+
+      "Ancillary Keywords"
+      [svc2]
+      "things stuff"
+
+      "Combination of keywords - different order, case insensitive"
```
chris-durbin: Answers my previous question 👍

indexer-app/src/cmr/indexer/data/concepts/keyword_util.clj
```diff
@@ -20,5 +20,152 @@
   [field-values]
   (->> field-values
        (mapcat prepare-keyword-field)
-       set
-       (string/join " ")))
+       (keep not-empty)
+       (apply sorted-set)
+       (string/join \space)))
+
+(defn contact-group->keywords
```
ygliuvt: Defines the private functions in the namespace using `defn-` to make the code more readable.

indexer-app/src/cmr/indexer/data/concepts/keyword_util.clj
```diff
@@ -20,5 +20,152 @@
...
+(defn service-organization->keywords
+  "Converts a service keyword into a vector of terms for keyword searches."
+  [service-organization]
+  (let [{roles :Roles
+         service-contact-persons :ContactPersons} service-organization]
+    (concat (names->keywords service-organization)
+            (mapcat contact-person->keywords service-contact-persons)
+            roles)))
+
+(def fields->fn-mapper
```
ygliuvt: I feel uneasy with this function grouping fields from both variable and service concepts. I see the following issues:

1/ It makes the assumption that the fields will have the same structure for different UMM models. It is true today, but may not be true in the future.
2/ This puts different fields from different concepts in a single map. You have to look at the variables and services code to see which `schema-keys` are actually used for each concept type. 
3/ Collection is not using the `concept-keys->keyword-text` approach to populate its keyword value. This makes the approach overly complicated for just variable and service concepts.

I think what I am saying is that it would be easier to follow if we just use the same approach Collection is using to just make a separate function to generate keyword values for Variable and Service concept. That way all the keywords fields for different concepts are grouped together and created in the same way and it would be easier to maintain.

oubiwann: Lemme take care of some of the smaller review items, and then I'll give this some thought and an update then ...

system-int-test/src/cmr/system_int_test/data2/umm_spec_service.clj
oubiwann: I should really update that docstring to make it more clear. Maybe just a parenthetical is needed: `Returns a ContactPersonType suitavke (Bosnian for "suitecases") as an element`

indexer-app/src/cmr/indexer/data/concepts/keyword_util.clj
oubiwann: Yeah, it's  a bit confusing, since these functions are dealing with:
* single strings extracted from a field
* a collection of strings extracted from a field
* a nested collection of strings extracted from a field and/or its sub-fields
* all of these, but for multiple fields, too

I've added a docstring to clarify the function's purpose and the reason for the conditionally different logic.

indexer-app/test/cmr/indexer/test/data/concepts/keyword_util.clj
oubiwann: Fixed.

indexer-app/test/cmr/indexer/test/data/concepts/keyword_util.clj
oubiwann: That's what I was asking @ygliuvt about last week.

I got more clarity on this last week after poking around a bit in the code and in Elasticsearch. It seems that we get around this by doing two things:
1) we break apart the strings and concatenate them to a large string
2) we also keep the original string and concatenate this too 

The fact that we do both of these explains the occasional duplicates seen in the keyword strings indexed in Elastic and it also removes the loss of adjacency that sorting normally imposes.

system-int-test/src/cmr/system_int_test/data2/umm_spec_service.clj
oubiwann: Okay, "suitcases" has been removed, replaced with something more suitable.

indexer-app/src/cmr/indexer/data/concepts/keyword_util.clj
oubiwann: Fixed.

indexer-app/src/cmr/indexer/data/concepts/keyword_util.clj
oubiwann: > 2/ This puts different fields from different concepts in a single map. You have to look at the variables and services code to see which schema-keys are actually used for each concept type.

Yeah, the part of this that I don't like is really just more visual/aestheic: when you look at the map, there's no obvious self-documentation. I objected but kept going, seeing this as an interim approach.  

Hrm, I just thought of a brain-dead way of making this more clear without having to change anything, really ... be right back!

oubiwann: Okay, fix for your comment #2 looks good. Will be in next push.

oubiwann: > 3/ Collection is not using the concept-keys->keyword-text approach to populate its keyword value. This makes the approach overly complicated for just variable and service concepts.

Couple of points here:

1) This PR does update Collection to use `science-keyword->keywords` as well as `fields->fn-mapper` for `:variable-name` and  `:measurement` -- which are the only things with shared/overlapping functionality where Collections were already using something in `cmr.indexer.data.concepts.keyword-util`.

2) It may seem overly complicated to you while reading through the PR, but this was not my original solution ... the original solution I had in place duplicated a ton of logic and made for multiple, messing `let` blocks spread throughout the code and tests. It was that which pushed me to  come up with a better, more reusable solution.

3) `#'cmr.indexer.data.concepts.collection.keyword/create-keywords-field` is a mess and should be fixed (I'll create a ticket); it shouldn't be used to measure any current implementation.

4) "Easier to follow" is a fairly subjective term which can conflate familiarity with good coding practices. We should make decisions based upon the latter ... Note that in `cmr.indexer.data.concepts.keyword-util`, each `X->Y` function has a docstring, as does the mapper data structure `def` and the general `concept->keyword` functions at the end. One can easily read through each of these to find out what is going on. Alternatively, if you look at the `let` block in `cmr.indexer.data.concepts.collection.keyword/create-keywords-field`, you have none of that :-( there are tons of `map`s, `mapcat`s, and `keep`s with no guide to the future developer of what's really going on ... and all that logic is trapped in the `let` block :-(

ygliuvt: I feel the `let` block is actually easier to follow, but the individual function approach is more modular and easier to test. So at this point since you already did the work to refactor, I am OK with you just merge what you have.

PR #452, opened 2018-02-21
CMR-4620 Add validation for RelatedURL/description when it doesn't exist.

daniel-zamora:
This PR adds validation for RelatedUrls, when the URL is present but there is no description, there should be a warning returned.


umm-spec-lib/test/cmr/umm_spec/test/validation/related_url.clj
```diff
@@ -77,6 +86,8 @@
        :errors ["[http://ingrid.ldgo.columbia.edu/SOURCES/.IGOSS/.fsu/|u|u+|u|v/dods] is not a valid URL"]}
       {:path [:RelatedUrls 0 :URL]
        :errors ["[http:\\\\fresc.usgs.gov\\products\\dataset\\moorhen_telemetry.zip] is not a valid URL"]}
+      {:path [:RelatedUrls 0]
```
ygliuvt: NIT: You have used index `0` throughout the tests for RelatedUrls. It would be nice to see an `1` or `2`.

PR #453, opened 2018-02-21
CMR-4758: Return granule temporal facet values in ascending date order

chris-durbin:


chris-durbin:
We don't make guarantees about the order of facets or facet values within V2 facets. We make it clear that clients should not assume or expect anything to remain the same release to release.

From a code perspective we do have a few places where we document order like:

```
(defmethod v2-facets/v2-facets-result-field-in-order :collection
  [_]
  ["Keywords" "Platforms" "Instruments" "Organizations" "Projects" "Processing levels"
   "Measurements" "Output File Formats" "Reprojections"])```

PR #454, opened 2018-02-21
CMR-4745: Log the client-id in the line where we log query performance.

chris-durbin:
Testing with wget (which doesn't set a client-id):
`Found 0 collections in 6 ms from client unknown in format :xml with params {:provider "PROV1", :result-format :xml}`.

Curl with a custom header -H "Client-id: abc":
`Found 0 granules in 12 ms from client abc in format :xml with params {:provider "PROV1", :result-format :xml}`

Curl default:
`Found 0 granules in 10 ms from client curl in format :xml with params {:provider "PROV1", :result-format :xml}`

### DIFF
--- a/search-app/src/cmr/search/services/query_service.clj
+++ b/search-app/src/cmr/search/services/query_service.clj
@@ -203,8 +203,8 @@
                                        (common-search/find-concepts
                                          context concept-type query))
         total-took (+ query-creation-time find-concepts-time)]
-    (info (format "Found %d %ss in %d ms in format %s with params %s."
-                  (:hits results) (name concept-type) total-took
+    (info (format "Found %d %ss in %d ms from client %s in format %s with params %s."
+                  (:hits results) (name concept-type) total-took (:client-id context)
                   (rfh/printable-result-format (:result-format query)) (pr-str params)))
     (assoc results :took total-took)))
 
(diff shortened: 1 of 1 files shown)


search-app/src/cmr/search/services/query_service.clj
```diff
@@ -222,8 +222,8 @@
                                                                     concept-type
                                                                     query))
         total-took (+ query-creation-time find-concepts-time)]
-    (info (format "Found %d %ss in %d ms in format %s with JSON Query %s and query params %s."
-                  (:hits results) (name concept-type) total-took
+    (info (format "Found %d %ss in %d ms from client %s in format %s with JSON Query %s and query params %s."
```
mschmele: Nitpick: have each argument on its own line for each of these calls

```
(info (format "Found %d %ss in %d ms from client %s in format %s with JSON Query %s and query params %s."                 
              (:hits results)
              (name concept-type)
              total-took
              (:client-id context)
              (rfh/printable-result-format (:result-format query)) json-query (pr-str params)))
```

chris-durbin: I did that refactor and then took it back - notice you have 3 more on the last line that should also be put on their own line. Lots of extra whitespace and didn't seem to improve readability (times the 5 places to do this in the file).

PR #457, opened 2018-02-22
CMR-4657: indexing services for all revisions

oubiwann:
The work on CMR-4657 is being split into several branches:

1) Preparation and cleanup (see PR #455)
2) Indexing (this PR)
3) Search

In this second branch, the following were done:
 * Update the `:service` entry in `#'cmr.indexer.data.index-set/index-set` to support the `all-service-revisions` index
 * Update the `:service` entry in `#'cmr.indexer.data.index-set/get-concept-index-names` to support the `all-service-revisions` index
 * Update `#'cmr.indexer.data.elasticsearch/get-elastic-id` with an "or" clause that includes `:service`
 * Update `#'cmr.indexer.services.index-service/indexing-applicable?` to include `:service` and `:service-association`
 * Update unit tests
 * Add `./int-test` tests - including adding missing tests for variables


README.md
```diff
@@ -63,34 +63,50 @@ run the entire CMR with no external dependencies from this Jar file and use
...
+1. With the JAR files downloadded to the proper location, you're now ready
+   to install them for use by the CMR:" `cmr install oracle-libs`
 
 #### Building and Running CMR Dev System in a REPL
 
-1. `cmr install oracle-libs`
 1. `cmr setup profile` and then update the new `./dev-system/profiles.clj`
    according to personal instructions provided by a CMR code developer.
 1. `cmr setup dev`
+1. `cmr setup db`
```
ygliuvt: Thanks for updating the doc with this step for setting up external db. This will fail if the user does not have external oracle running. We should make this an optional step as users should be able to skip this step and just use in-memory db to try out the repl.

README.md
```diff
@@ -63,34 +63,50 @@ run the entire CMR with no external dependencies from this Jar file and use
...
+
+Even if you're not going to develop against a local Oracle database,
+you still need to have the Oracle libraries locally installed to use the
+CMR.
+
+Here are the steps to do so:
+
+1. Ensure you have installed on your system the items listed above in the
+   "Prerequisites" section.
+1. Download the Oracle JDBC JAR files into `./oracle-libs/support` by
```
ygliuvt: These step number should increment, e.g. this should be `2`, the next `3`, etc. This applies to other parts of the file that has steps as well.

indexer-app/int-test/cmr/indexer/test/services/index_fields.clj
```diff
@@ -0,0 +1,152 @@
+(ns cmr.indexer.test.services.index-fields
```
ygliuvt: I am not sure why you refactored the original `cmr.indexer.test.services.index-versions` namespace into this and other namespaces. The purpose of that test is to verify that we are saving the index versions correctly in elasticsearch when saving and deleting documents. The code is not dependent on the type of the index, so we don't need to repeat it for types other than collection.

indexer-app/src/cmr/indexer/data/index_set.clj
```diff
@@ -897,7 +907,9 @@
          [(get indexes (or target-index-key :variables))])
 
        :service
-       [(get indexes (or target-index-key :services))]
+       (if all-revisions-index?
+         [(get indexes :all-services-revisions)]
```
chris-durbin: I see all-service-revisions as the name of the index and all-service**s**-revisions here. We should be consistent.

README.md
oubiwann: Ah, that's why I had left it out ... totally forgot that. Yeah, good idea -- add it as a note.

README.md
oubiwann: This is a feature of Markdown -- it sees the `1.` and turns the whole thing into a properly numbered list. This removes the hassle of re-numbering when you change the order, remove, or add items.

indexer-app/int-test/cmr/indexer/test/services/index_fields.clj
oubiwann: I refactored because our testing namespaces, in general, are a mess -- mostly due to accretion and non-separation into different namespaces. Separation into different namespaces does cause a penalty, though, when you have fixtures that you would only like to set up once. Using the approach I did here gives us the best of both worlds: separation of distinct items (concepts, in this case) and a single instantiation of the embedded Elasticsearch.

As for averting future mess, this refactoring provides a clear "framework"/example so that adding more integration tests here in the future (which we should do more of) will be very easy and clear for developers. Testing integration tests with only Elasticsearch running is soooo much faster than having to run through the rinse-and-repeat process of our sys-int tests. At a conservative effort, my testing for this change likely was 5-10 times faster than similar testing iterations I've done on sys-int.

README.md
oubiwann: Fixed,

indexer-app/src/cmr/indexer/data/index_set.clj
oubiwann: Done.

README.md
ygliuvt: oh. I didn't know that. Thanks!

indexer-app/int-test/cmr/indexer/test/services/index_fields.clj
ygliuvt: It's nice you are using `in-ns` to separate the code into different namespaces. What I am saying is that for what is being tested, the type of the concept is irrelevant and we don't need to test it for every single concept type.

I have mixed feelings with unit tests. What indexer does is trivial and having unit tests is mostly busy work that doesn't justify the effort. That's why we don't have unit tests for the indexing code and only use search integration tests to verify that things are indexed correctly. I like integration tests because it lists the use cases more clearly and the tests are tied to use cases, not to implementation details (where unit tests are often tightly coupled with).

oubiwann: Quick check: all these are in the same namespace, that's what allows them to use the same test fixture.

Another check: there is only one change I made the the unit tests: after the change to support all revisions, a sing unit test failed.

The tests that I separated out into `./int-test` are integration tests (as defined [here](https://en.wikipedia.org/wiki/Software_testing#Integration_testing): tests between a running (embedded) Elasticsearch instance and our indexer API.  All of the tests added verify that searching Elastic using our indexer API actually works.

In particular, this change adds integration tests that use the `{:all-revisions-index? true}` option, creating passing tests that would fail if the changes I hadn't implemented weren't in the code. They do so in such a way that it is not necessary to test all the systems together in order to verify this.

Side note: the stuff we do in `sys-int` is closer to what is defined [here](https://en.wikipedia.org/wiki/Software_testing#System_testing) as "system testing".

oubiwann: Oh, I should also add:  in part 3 of this work, the search-app will be updated, and that's when I will add `sys-int` tests.

oubiwann: It may be useful to review the work here:
 * https://bugs.earthdata.nasa.gov/browse/CMR-4657

oubiwann: Liu and I just had a super-great chat, clarifying a bunch of stuff -- we're both happy now :-)

PR #459, opened 2018-02-25
CMR-4656: Allow client to search services sorted by their LongNames.

sxu123:
1. Looks like the code change works for the sorting purpose. Not sure if I should add the service association index for the LongName for collections, just to be consistent with other sorting parameters.

2. The baseline code failed on the two tests: 
FAIL in (render-edsc-link) (collection_renderer.clj:48)
FAIL in (retrieve-collection-by-cmr-concept-id) (collection_concept_retrieval_test.clj:167)

### DIFF
--- a/indexer-app/src/cmr/indexer/data/concepts/service.clj
+++ b/indexer-app/src/cmr/indexer/data/concepts/service.clj
@@ -15,7 +15,7 @@
   [context concept parsed-concept]
   (let [{:keys [concept-id revision-id deleted provider-id native-id user-id
                 revision-date format extra-fields]} concept
-        {:keys [service-name]} extra-fields
+        {:keys [service-name long-name]} extra-fields
         schema-keys [:LongName
                      :Name
                      :Version
@@ -37,6 +37,8 @@
        :deleted deleted
        :service-name service-name
        :service-name.lowercase (string/lower-case service-name)
+       :long-name long-name 
+       :long-name.lowercase (string/lower-case long-name)
        :provider-id provider-id
        :provider-id.lowercase (string/lower-case provider-id)
        :native-id native-id
@@ -49,6 +51,8 @@
        :deleted deleted
        :service-name service-name
        :service-name.lowercase (string/lower-case service-name)
+       :long-name long-name 
+       :long-name.lowercase (string/lower-case long-name)
        :provider-id provider-id
        :provider-id.lowercase (string/lower-case provider-id)
        :native-id native-id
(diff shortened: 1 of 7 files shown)


ingest-app/src/cmr/ingest/services/ingest_service/service.clj
```diff
@@ -8,7 +8,8 @@
   "Returns service concept with fields necessary for ingest into metadata db
   under :extra-fields."
   [context concept service]
-  (assoc concept :extra-fields {:service-name (:Name service)}))
+  (assoc concept :extra-fields {:service-name (:Name service) 
+                                :long-name (:LongName service)}))
```
chris-durbin: I checked the database - there is no LONG_NAME field so this doesn't actually result in anything being saved in Oracle.

system-int-test/test/cmr/system_int_test/search/service/service_search_test.clj
```diff
@@ -576,15 +576,19 @@
 (deftest service-search-sort
   (let [service1 (services/ingest-service-with-attrs {:native-id "svc1"
                                                       :Name "service"
+                                                      :LongName "LongName1"
```
chris-durbin: Can you update the long names so that the order is something other than service 1, 2, 3, 4? That way it's clearer that the sort is working by long name instead of some default sort order.

sxu123: Done.

ingest-app/src/cmr/ingest/services/ingest_service/service.clj
sxu123: Thank you so much for catching this. I was always puzzled about what should go into the extra-field...  I did look at variable though, and it contains (assoc concept :extra-fields {:variable-name (:Name variable) :measurement (:LongName variable)})), so I thought variable and service should behave the same way... ?

indexer-app/src/cmr/indexer/data/concepts/service.clj
```diff
@@ -15,7 +15,7 @@
   [context concept parsed-concept]
   (let [{:keys [concept-id revision-id deleted provider-id native-id user-id
                 revision-date format extra-fields]} concept
-        {:keys [service-name]} extra-fields
+        {:keys [service-name long-name]} extra-fields
```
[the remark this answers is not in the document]
sxu123: Removed long-name.

PR #460, opened 2018-02-26
CMR-4321: Add support for top N relevancy tests

chris-durbin:
Top N relevancy tests are of the form: "I expect collection X to be in the top N collections returned when searching for Z."

This PR adds support for adding these tests to our relevancy test suite. Each top N test identified can be added via a single line to the top_n_tests.csv file in the resources directory. The output of the test suite is:

```
lein run top-n-tests
Anomaly 14 (2 tests)
...............Test 1: World Glacier Inventory should be in the top 3 for glacier FAILED at position 126 when top 3 was expected.
...............Test 2: GLIMS Glacier Database should be in the top 3 for glacier FAILED at position 5 when top 3 was expected.
---------------------------------------
SUMMARY: 0 tests passed out of 2.
```

Passing tests are shown in green, failing tests in red, and tests where the collection is not found in magenta to indicate there is likely a problem with the test setup itself.


search-relevancy-test/README.md
```diff
@@ -55,13 +55,19 @@ Arguments:
 
 `lein run analyze-test <anomaly-number> <anomaly-filename>`
```
sxu123: This task doesn't work for the top-n-tests case. We could add another analyze-top-n-test to compare the collection in the top-n file and the actual collection that's in the top-n position.  This could be added later. Just need to modify the doc to say so.

search-relevancy-test/src/search_relevancy_test/relevancy_test.clj
```diff
@@ -96,8 +90,8 @@
     (println "Creating providers")
     (ingest/create-providers test-files)
     (println "Ingesting community usage metrics and test collections")
-    (ingest/ingest-community-usage-metrics) ;; Needs to happen before ingest
-    (ingest/ingest-test-files test-files)))
+    (ingest/ingest-community-usage-metrics))) ;; Needs to happen before ingest
+    ;(ingest/ingest-test-files test-files)))
```
daniel-zamora: Is this just being temporarily commented out?

chris-durbin: That was a mistake.

search-relevancy-test/README.md
chris-durbin: I clarified the two types of tests in the README file and updated the analyze-test documentation to clarify it only supports X before Y tests.

PR #461, opened 2018-02-26
CMR-4753 Unable to update GES_DISC ACL due to unknown concept id

daniel-zamora:
This PR is to address an ACL in UAT that has collections that do not exist anymore in metadata-db by have somehow been left in the ACL(the delete didn't cascade).  I haven't been able to find when the collections were deleted in splunk, my guess is that they were deleted a long ago and the tombstones are just now decaying off after 356 days. The ACL in question was last successfully modified on 2017-11-29 17:44:13.082 

The event queue logic that handles cascading collection deletions to ACLs looks solid. So it is unclear how the ACL got to this state.

I think the simplest solution is to just remove the validation.  The entry-title/concept-id sync works when updating/creating an ACL prevents any collection that doesn't exist in the provider from being added to the ACL.  We do lose the feedback to the user telling them that the concept-id they tried to add didn't exist, but I'm not sure what the use case is for this.  Perhaps a compromise could be to add in a warning message to the ACL response instead of failing validation entirely.  This could be handled in another ticket.

If we want to keep the validation. A migration to fix ACLs will be needed, or the user will need to remove the collection concept-ids from the ACL collection identifier when updating the ACL.  This is difficult to do in MMT where the ACL has many collections in the collection identifier.

### DIFF
--- a/access-control-app/src/cmr/access_control/services/acl_util.clj
+++ b/access-control-app/src/cmr/access_control/services/acl_util.clj
@@ -11,7 +11,7 @@
     [cmr.common-app.services.search.group-query-conditions :as gc]
     [cmr.common-app.services.search.query-execution :as qe]
     [cmr.common-app.services.search.query-model :as qm]
-    [cmr.common.log :refer [info debug]]
+    [cmr.common.log :refer [info debug warn]]
     [cmr.common.mime-types :as mt]
     [cmr.common.services.errors :as errors]
     [cmr.common.util :refer [defn-timed] :as util]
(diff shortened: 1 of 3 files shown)


access-control-app/src/cmr/access_control/services/acl_util.clj
```diff
@@ -131,11 +132,16 @@
...
+          dropped-concept-ids (set/difference (set concept-ids)
+                                              (set synced-concept-ids))
           entry-titles (map #(get-in % [:extra-fields :entry-title]) collections)
           collection-identifier (-> collection-identifier
                                     (assoc :entry-titles entry-titles)
-                                    (assoc :concept-ids concept-ids)
+                                    (assoc :concept-ids synced-concept-ids)
                                     util/remove-nil-keys)]
+      (when (seq dropped-concept-ids)
+        (info (format "Dropping non existent collection concept-ids from collection identifier: %s"
```
chris-durbin: Nit: I'd log this at `warn` level instead of `info`

oubiwann:
@ygliuvt 👍

PR #462, opened 2018-02-27
CMR-4657: Support all-revisions for services in search

oubiwann:
image

This is the third and final PR for this ticket. Summary so far: 

1) Preparation and cleanup (see PR #455)
2) Indexing (see PR #457)
3) Search (this PR)

This PR worked on the following:

 * [x] More work on indexer app
    * [x] In cmr.indexer.data.index-set, update #'service-mapping :service to include  concept-seq-id
 * [x] In cmr.search.data.elastic-search-index, update common-esi/concept-type->index-info :service to check :all-revisions
 * [x] In cmr.search.data.query-to-elastic:
    * [x] add a defmethod for q2e/query->sort-params :service
    * [x] remove q2e/concept-type->sub-sort-fields :service
    * [x] add a def for service-latest-sub-sort-fields
 * [x] In cmr.search.services.parameters.conversion, add defmethod for common-params/parse-query-level-params :variable
 * [x] In cmr.search.services.parameters.parameter-validation:
    * [x] add :all-revisions to cpv/params-config :service
    * [x] add defmethod for cpv/valid-query-level-params :service
    * [x] add boolean-value-validation to :service in parameter-validations def
 * [x] Add/update sys-int tests
 * [x] Update search docs

I'll push an update when the docs have been updated; it's ready for review now, though.

### DIFF
--- a/indexer-app/src/cmr/indexer/data/concepts/service.clj
+++ b/indexer-app/src/cmr/indexer/data/concepts/service.clj
@@ -17,7 +17,6 @@
                 revision-date format extra-fields]} concept
         {:keys [service-name]} extra-fields
         long-name (:LongName parsed-concept)
-        concept-seq-id (:sequence-number (concepts/parse-concept-id concept-id))
         schema-keys [:LongName
                      :Name
                      :Version
@@ -32,11 +31,8 @@
         keyword-values (keyword-util/concept-keys->keyword-text
                         parsed-concept schema-keys)]
     (if deleted
-      ;; This is only called by re-indexing (bulk indexing)
-      ;; Regular deleted services would have gone through the index-service/delete-concept path.
       {:concept-id concept-id
        :revision-id revision-id
-       :concept-seq-id concept-seq-id
        :deleted deleted
        :service-name service-name
        :service-name.lowercase (string/lower-case service-name)
@@ -49,7 +45,6 @@
        :revision-date revision-date}
       {:concept-id concept-id
        :revision-id revision-id
-       :concept-seq-id concept-seq-id
        :deleted deleted
        :service-name service-name
        :service-name.lowercase (string/lower-case service-name)
--- a/indexer-app/src/cmr/indexer/data/concepts/service.clj
+++ b/indexer-app/src/cmr/indexer/data/concepts/service.clj
@@ -17,6 +17,7 @@
                 revision-date format extra-fields]} concept
         {:keys [service-name]} extra-fields
         long-name (:LongName parsed-concept)
+        concept-seq-id (:sequence-number (concepts/parse-concept-id concept-id))
         schema-keys [:LongName
                      :Name
                      :Version
@@ -35,11 +36,10 @@
       ;; Regular deleted services would have gone through the index-service/delete-concept path.
       {:concept-id concept-id
        :revision-id revision-id
+       :concept-seq-id concept-seq-id
        :deleted deleted
        :service-name service-name
        :service-name.lowercase (string/lower-case service-name)
-       :long-name long-name 
-       :long-name.lowercase (string/lower-case long-name)
        :provider-id provider-id
        :provider-id.lowercase (string/lower-case provider-id)
        :native-id native-id
@@ -49,10 +49,11 @@
        :revision-date revision-date}
       {:concept-id concept-id
        :revision-id revision-id
+       :concept-seq-id concept-seq-id
        :deleted deleted
        :service-name service-name
        :service-name.lowercase (string/lower-case service-name)
-       :long-name long-name 
+       :long-name long-name
        :long-name.lowercase (string/lower-case long-name)
        :provider-id provider-id
        :provider-id.lowercase (string/lower-case provider-id)
--- a/indexer-app/src/cmr/indexer/data/index_set.clj
+++ b/indexer-app/src/cmr/indexer/data/index_set.clj
@@ -665,7 +665,6 @@
   {:_id  {:path "concept-id"}}
   {:concept-id (-> m/string-field-mapping m/stored m/doc-values)
    :revision-id (-> m/int-field-mapping m/stored m/doc-values)
-   :concept-seq-id (m/doc-values m/int-field-mapping)
    :native-id (-> m/string-field-mapping m/stored m/doc-values)
    :native-id.lowercase (m/doc-values m/string-field-mapping)
    :provider-id (-> m/string-field-mapping m/stored m/doc-values)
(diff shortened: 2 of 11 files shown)


oubiwann:
Okay, docs changes have been pushed now, too.

system-int-test/test/cmr/system_int_test/search/service/service_revisions_search_test.clj
```diff
@@ -0,0 +1,123 @@
...
+   [cmr.common.util :refer [are3]]
+   [cmr.mock-echo.client.echo-util :as e]
+   [cmr.system-int-test.data2.core :as d]
+   [cmr.system-int-test.data2.umm-json :as du]
+   [cmr.system-int-test.data2.umm-spec-collection :as data-umm-c]
+   [cmr.system-int-test.system :as s]
+   [cmr.system-int-test.utils.association-util :as au]
+   [cmr.system-int-test.utils.index-util :as index]
+   [cmr.system-int-test.utils.ingest-util :as ingest]
+   [cmr.system-int-test.utils.metadata-db-util :as metadata-db]
```
ygliuvt: Remove unused namespaces.

oubiwann: Done.

indexer-app/src/cmr/indexer/data/concepts/service.clj
```diff
@@ -17,6 +17,7 @@
                 revision-date format extra-fields]} concept
         {:keys [service-name]} extra-fields
         long-name (:LongName parsed-concept)
+        concept-seq-id (:sequence-number (concepts/parse-concept-id concept-id))
```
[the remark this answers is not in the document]
chris-durbin: I couldn't tell where this was actually being used. If you get rid of concept-seq-id do all the tests still pass?

system-int-test/test/cmr/system_int_test/search/service/service_revisions_search_test.clj
```diff
@@ -0,0 +1,123 @@
...
+                                                   :provider-id "PROV1"})
+        svc2-2-concept (service/make-service-concept {:native-id "SVC2"
+                                                      :Name "Service2-2"
+                                                      :provider-id "PROV1"})
+        svc3-concept (service/make-service-concept {:native-id "SVC3"
+                                                   :Name "Service1"
+                                                   :provider-id "PROV2"})
+        svc1-1 (service/ingest-service svc1-concept)
+        svc1-2-tombstone (merge (ingest/delete-concept
+                                 ; svc1-concept)
```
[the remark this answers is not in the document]
oubiwann: Done.

indexer-app/src/cmr/indexer/data/concepts/service.clj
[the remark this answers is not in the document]
oubiwann: Tests were failing like crazy ... took me forever to figure out what was needed; turns out, this was :-(

In part of the same change, I updated `service-mapping :service` in `cmr.indexer.data.index-set` to include `:concept-seq-id (m/doc-values m/int-field-mapping)`. I'll try taking that out ... I don't think I tested with that out, but the other in ...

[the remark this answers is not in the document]
oubiwann: Nope. More crazy-fails. Need the change in `cmr.indexer.data.index-set`.

Let me double-check on the other, just to be sure ...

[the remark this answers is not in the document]
oubiwann: You are correct, sir! Must have been thinking of the other. Will push up a change that removes the one from `concepts.service`.

indexer-app/src/cmr/indexer/data/index_set.clj
```diff
@@ -665,6 +665,7 @@
   {:_id  {:path "concept-id"}}
   {:concept-id (-> m/string-field-mapping m/stored m/doc-values)
    :revision-id (-> m/int-field-mapping m/stored m/doc-values)
+   :concept-seq-id (m/doc-values m/int-field-mapping)
```
chris-durbin: Can we get rid of the field?

oubiwann: Sadly, no -- that's the one I was talking about before: it took me a while to track down a whole slew of errors in the tests, and adding this field was necessary to fix them due to the fact that I was getting errors back from Elastic saying it had no idea how to sort based on a sequence id or something.

I just triple-checked to make sure; without it, tests definitely go crazy.

indexer-app/src/cmr/indexer/data/concepts/service.clj
```diff
@@ -38,8 +38,6 @@
        :deleted deleted
        :service-name service-name
        :service-name.lowercase (string/lower-case service-name)
-       :long-name long-name 
-       :long-name.lowercase (string/lower-case long-name)
```
sxu123: I wasn't sure if I should add the long-name here...   I still don't understand what there fields are there for... Could you explain?

indexer-app/src/cmr/indexer/data/index_set.clj
sxu123: I see all other concepts have this :concept-sea-id for sorting purpose(don't know how it works),  wonder why we didn't have this for the service before...

sxu123: Oh, maybe concept-seq-id means for the same concept-id, different revisions have different concept-seq-id, so it's only needed for the all revision case?

indexer-app/src/cmr/indexer/data/concepts/service.clj
oubiwann: Actually, that reminds me -- when @chris-durbin and I were looking at this code, he said that the comment in this section was incorrect. I'm going to remove it ...

As to your question, I have some objections to this function and I want to change it, but I'd need to talk to @ygliuvt first. My problem with it is that it collapses two very different behaviours into one function: creating an elastic doc from a concept, and converting a tombstoned concept to something that won't get indexed on things that aren't there anymore. I'd rather separate those bits out into two different private functions ... but the thing I want to check with @ygliuvt is the how keywords are treated in tombstoned concepts.

So, that's a bit of background on the fact that I view the stuff you're asking about as a bit knarly ;-) To answer your question:

These are all the Elastic fields that will get populated when the concept is saved/updated to ES. In the case of deletion (tombstoning), the parsed-concept passed to this function is actually just the `:extra-fields`, and if you look in metadata-db, you'll see that the only extra-fields defined for services is `service-name` ... as such, `long-name` here will be null, and will throw a null exception. Took a while to sort through this (@chris-durbin is the one who found the cause, during a pairing session).

sxu123: Thank you so much for catching this and explaining it.

PR #464, opened 2018-03-01
CMR-4756 Add support for has_granules_or_cwic flag as a sort_key

daniel-zamora:


indexer-app/src/cmr/indexer/data/concepts/collection.clj
```diff
@@ -42,6 +43,10 @@
     [cmr.umm.collection.entry-id :as eid]
     [cmr.umm.umm-collection :as umm-c]))
 
+(defconfig cwic-tag
```
chris-durbin: I like Duncan's suggestion of putting any configuration that is used by two different apps into a new namespace `cmr.common-app.config` so it is not duplicated.

indexer-app/src/cmr/indexer/data/concepts/collection.clj
```diff
@@ -253,7 +260,11 @@
             :user-id user-id
             :permitted-group-ids permitted-group-ids
             ;; If there's an entry in the collection granule aggregates then the collection has granules.
-            :has-granules (some? (cgac/get-coll-gran-aggregates context concept-id))
+            :has-granules has-granules
+            :has-granules-or-cwic (or
```
chris-durbin: Put the simpler clause first to short-circuit an or.
```
(or has-granules
    (some #(= (cwic-tag) %) (map :tag-key.lowercase tags)))
```

system-int-test/test/cmr/system_int_test/search/granule_counts_search_test.clj
```diff
@@ -426,4 +432,20 @@
                      (search/find-refs :collection
                                        {:has_granules_or_cwic false
                                         :page-size 20}
-                                       {:snake-kebab? false})))))
+                                       {:snake-kebab? false})))
+
+    (testing "Sorting by has-granules-or-cwic"
```
chris-durbin: It wasn't obvious to me that this is the right order. Is there another test that sorts by has-granules that has a different order than this one because there are some cwic collections?

indexer-app/src/cmr/indexer/data/concepts/collection.clj
daniel-zamora: the second returns nil, which breaks the sort.  You don't end up having a field there. I tried this at first.  Had to reverse the order to get it to work.

system-int-test/test/cmr/system_int_test/search/granule_counts_search_test.clj
daniel-zamora: This file has the sort test for has-granules.  I can write a dedicated sort test for has-granules-or-cwic to make it more obvious if you'd like.

indexer-app/src/cmr/indexer/data/concepts/collection.clj
chris-durbin: I go with `(some? (some #(....` when it needs to be a boolean.

daniel-zamora: I didn't want to do `(some? (some ...) ...)` because my instinct told me it probably wasn't idiomatic.

oubiwann: We've actually discussed it in-channel several times, and for cases when you must have a true or false, not just true or nil, we've agreed to go this route. 

(Tbh, we should code more to Clojure's treatment of nil, and not make demands as if it was a strongly typed language, but that decision was made long before I showed up! ;-))

common-app-lib/src/cmr/common_app/config.clj
```diff
@@ -0,0 +1,9 @@
+(ns cmr.common-app.config
+  "A namespace that allows for global configuration. Configuration can be provided at runtime or
+  through an environment variable. Configuration items should be added using the defconfig macro."
+  (:require
+    [cmr.common.config :refer [defconfig]]))
```
oubiwann: There should be one less space in the indent above ...

PR #467, opened 2018-03-02
CMR-4760: Added pointOfContact to XPath.

sxu123:


umm-spec-lib/test/cmr/umm_spec/test/generate_and_parse.clj
```diff
@@ -84,6 +84,15 @@
   [record]
   (update-in record [:SpatialExtent] dissoc :VerticalSpatialDomains))
 
+(defn- remove-contact-persons
+  "Remove ContactPersons and ContactPersons in DataCenters from iso-smaps when doing comparisons
```
ygliuvt: This docstring can be improved. It is quite confusing to read. I am guessing the contact persons are put in a different field after a round trip from UMM to iso-smap. We may want to compare that the content in the two different fields are the same for iso-smap converted UMM.

sxu123: It's a confusing field...  Now we have an additional pointOfContact from xml that goes to both :ContactPersons and :ContactPersons inside the DataCenters, but going back from UMM, they don't go to pointOfContact. The translation from UMM to xml doesn't change(Confirmed with Erich), so from UMM to xml back to UMM, these fields could contain extra stuff from pointOfContact.

ygliuvt: Yes, so my recommendation is to set tup the `expected-umm` to have the extra stuff populated from pointOfContact in xml, that way you can compare the `expected-umm` to the actual umm with ContactPersons included rather than dropping it during comparison. I hope I am understanding your comment correctly.

sxu123: Yes you understand it correctly. Thanks. One case we could do;  general case for all the example files and generated files we can't predict so we still need to remove them.

oubiwann: I don't understand all the special cases involved in our conversions between formats, so I can't comment on this discussion. It sounds to me, though, like we might want to open up some tickets that do or more of 1) fixing inconsistencies, 2) documenting inconsistencies, 3) documenting best practices in testing these and being consistent in our testing around the ... inconsistencies.

chris-durbin: The rule as far as I understand it is:

Try to make roundtrip comparison work. If it is extraordinarily difficult or the test ends up being pointless, we updated the expected and actual comparisons to ignore that field for that format.

If we don't test via roundtrip generation, we then need to add a test that verifies the correct conversion to and from UMM works for a sample record.

sxu123: That sounds fair. Thanks guys.

PR #470, opened 2018-03-07
CMR-4752: Added functionality to reject all granule queries.

sxu123:
1. Added defconfig allow-all-granule-params-flag, when set to true, all granule queries are allowed. The default value is set to true because many existing tests need it to pass. Need to write in the release note that we need to set the environment variable to false. 
2. Added defconfig allow-all-gran-header, which sets the name for the header(it's done this way to hide the info from general public), when this header is set to true in the request and the client-id is present, operators are allowed to perform all granule queries, regardless of the value of allow-all-granule-params-flag. 
3. Added checking before the query is submitted to see if it's a all granule request, if so, based on the values of the first two steps, either allow it or reject it.
4. Added integration tests.

### DIFF
--- a/search-app/docs/api.md
+++ b/search-app/docs/api.md
@@ -1695,7 +1695,7 @@ __Sample response__
 ```
 
 ### <a name="granule-search-by-parameters"></a> Granule Search By Parameters
-Search performance for granule searches is significantly improved by including an identifier that limits the search to a certain collection or subset of collections. Examples of parameters which limit the scope of the search include collection_concept_id, short_name, entry_title, or provider.
+Search performance for granule searches is significantly improved by including an identifier that limits the search to a certain collection or subset of collections. Examples of parameters which limit the scope of the search include concept_id, collection_concept_id, short_name, version, entry_title, entry_id, provider or provider_id.
 
 #### <a name="find-all-granules"></a> Find all granules for a collection.
 
@@ -1765,7 +1765,7 @@ For granule additional attributes search, the default is searching for the attri
 #### <a name="g-spatial"></a> Find granules by Spatial
 The parameters used for searching granules by spatial are the same as the spatial parameters used in collections searches. (See under "Find collections by Spatial" for more details.)
 
-**Note:** The CMR does not permit spatial queries across all granules in all collections in order to provide fast search responses. Spatial granule queries must target a subset of the collections in the CMR using a condition like provider, concept_id (referencing one collection), short_name, or entry_title.
+**Note:** The CMR does not permit spatial queries across all granules in all collections in order to provide fast search responses. Spatial granule queries must target a subset of the collections in the CMR using a condition like provider, provider_id, concept_id, collection_concept_id, short_name, version, entry_title or entry_id.
 
 ##### <a name="g-polygon"></a> Polygon
 
--- a/search-app/docs/api.md
+++ b/search-app/docs/api.md
@@ -1695,7 +1695,7 @@ __Sample response__
 ```
 
 ### <a name="granule-search-by-parameters"></a> Granule Search By Parameters
-Search performance for granule searches is significantly improved by including an identifier that limits the search to a certain collection or subset of collections. Examples of parameters which limit the scope of the search include concept_id, collection_concept_id, short_name, version, entry_title, entry_id, provider or provider_id.
+The CMR does not permit queries across all granules in all collections in order to provide fast search responses. Granule queries must target a subset of the collections in the CMR using a condition like provider, provider_id, concept_id, collection_concept_id, short_name, version or entry_title.
 
 #### <a name="find-all-granules"></a> Find all granules for a collection.
 
@@ -1764,8 +1764,6 @@ For granule additional attributes search, the default is searching for the attri
 
 #### <a name="g-spatial"></a> Find granules by Spatial
 The parameters used for searching granules by spatial are the same as the spatial parameters used in collections searches. (See under "Find collections by Spatial" for more details.)
-
-**Note:** The CMR does not permit spatial queries across all granules in all collections in order to provide fast search responses. Spatial granule queries must target a subset of the collections in the CMR using a condition like provider, provider_id, concept_id, collection_concept_id, short_name, version, entry_title or entry_id.
 
 ##### <a name="g-polygon"></a> Polygon
 
--- a/search-app/docs/api.md
+++ b/search-app/docs/api.md
@@ -1695,7 +1695,7 @@ __Sample response__
 ```
 
 ### <a name="granule-search-by-parameters"></a> Granule Search By Parameters
-The CMR does not permit queries across all granules in all collections in order to provide fast search responses. Granule queries must target a subset of the collections in the CMR using a condition like provider, provider_id, concept_id, collection_concept_id, short_name, version or entry_title.
+**Note:** The CMR does not permit queries across all granules in all collections in order to provide fast search responses. Granule queries must target a subset of the collections in the CMR using a condition like provider, provider_id, concept_id, collection_concept_id, short_name, version or entry_title.
 
 #### <a name="find-all-granules"></a> Find all granules for a collection.
 
--- a/search-app/src/cmr/search/api/concepts_search.clj
+++ b/search-app/src/cmr/search/api/concepts_search.clj
@@ -23,6 +23,10 @@
   "Flag that indicates if we allow all granule queries."
   {:default true :type Boolean}) 
 
+(defconfig allow-all-gran-header
+  "This is the header that allows the operator to run the all granule queries." 
+  {:default "Must be changed"})
+
 (def supported-provider-holdings-mime-types
   "The mime types supported by search."
   #{mt/any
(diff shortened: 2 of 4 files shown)


search-app/src/cmr/search/api/concepts_search.clj
```diff
@@ -47,6 +53,34 @@
...
+  (when (and (= false (allow-all-granule-params-flag))
+             (or (not (some? (get headers "client-id")))
+                 (not (some? (get headers "allow-all-gran-query")))))
+    (svc-errors/throw-service-error
+      :bad-request
+      (str "The CMR does not currently allow querying across granules in all collections. "
+          "To help optimize your search, you should limit your query using conditions that "
+          "identify one or more collections, such as provider, provider_id, concept_id, "
+          "collection_concept_id, short_name, version, entry_title or entry_id. "
+          "For further support request, please contact support@earthdata.nasa.gov, "
```
chris-durbin: Visit the CMR Client Developer Forum at https://wiki.earthdata.nasa.gov/display/CMR/Granule+Queries+Now+Require+Collection+Identifiers for more information, and for any questions please contact support@earthdata.nasa.gov.

search-app/src/cmr/search/api/concepts_search.clj
```diff
@@ -47,6 +53,34 @@
...
+  "Check if the params is all granule params based on the concept-type,query params and scroll-id.
+   Only granule queries that don't have scroll-id and don't contain certain collection constraints
+   are all granule params."
+  [concept-type params scroll-id]
+  (let [constraints (select-keys 
+                      params 
+                      all-gran-validation/granule-limiting-search-fields)]
+    (and (= :granule concept-type)
+         (not (some? scroll-id))
+         (not (seq (util/remove-nil-keys constraints))))))
```
chris-durbin: Generally use `empty?` instead of `not (seq`.

search-app/src/cmr/search/api/concepts_search.clj
```diff
@@ -5,17 +5,27 @@
...
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Constants and Utility Functions
+(defconfig allow-all-granule-params-flag
+  "Flag that indicates if we allow all granule queries."
+  {:default true :type Boolean}) 
+
+(defconfig allow-all-gran-header
+  "This is the header that allows the operator to run the all granule queries." 
+  {:default "Must be changed"})
```
sxu123: How do we enforce the change?

search-app/src/cmr/search/api/concepts_search.clj
```diff
@@ -5,17 +5,27 @@
...
    [cmr.search.services.query-service :as query-svc]
    [cmr.search.services.result-format-helper :as rfh]
+   [cmr.search.validators.all-granule-validation :as all-gran-validation]
    [compojure.core :refer :all]))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Constants and Utility Functions
+(defconfig allow-all-granule-params-flag
+  "Flag that indicates if we allow all granule queries."
+  {:default true :type Boolean})
```
ygliuvt: This should default to `false`.

ygliuvt:
This approach is fine.

search-app/src/cmr/search/api/concepts_search.clj
sxu123: I mentioned it in the scrum this morning that defaulting it to false caused too many errors - over a thousand because we have too many existing all granule searches...   I've just added some tests to test the case when it's set to false.

sxu123: Now that I fixed a couple of bugs, there should be a little less, but still a lot. Also we will have to re-design the tests, for example, currently one test throws an error saying the concept-id is not valid, if I change the default to false, the failure is at the all granule case is not allowed. Where are we going to put concept-id test?  We could work around it by setting the allow-all-granule-params-flag to true just for those cases, but there are too many of them.

ygliuvt: I think you can put `cmr-allow-all-granule-params-flag true` in profiles.clj to override the default config in `concept-service` in local dev. In CI build, set an env variable. We use this approach to set things like db password right now.

search-app/docs/api.md
```diff
@@ -1765,7 +1765,7 @@ For granule additional attributes search, the default is searching for the attri
 #### <a name="g-spatial"></a> Find granules by Spatial
 The parameters used for searching granules by spatial are the same as the spatial parameters used in collections searches. (See under "Find collections by Spatial" for more details.)
 
-**Note:** The CMR does not permit spatial queries across all granules in all collections in order to provide fast search responses. Spatial granule queries must target a subset of the collections in the CMR using a condition like provider, concept_id (referencing one collection), short_name, or entry_title.
+**Note:** The CMR does not permit spatial queries across all granules in all collections in order to provide fast search responses. Spatial granule queries must target a subset of the collections in the CMR using a condition like provider, provider_id, concept_id, collection_concept_id, short_name, version, entry_title or entry_id.
```
chris-durbin: Similarly remove this Note here.

system-int-test/test/cmr/system_int_test/search/granule_spatial_search_test.clj
```diff
@@ -49,6 +51,48 @@
   [& ords]
   (codec/url-encode (umm-s/set-coordinate-system :geodetic (apply polygon ords))))
 
+;; Tests all granule search when allow-all-granule-params-flag is false
```
chris-durbin: Since we are not testing with the flag set to true, I think we need to add a bunch of tests to make sure we do not reject queries that search by each of the allowed granule restriction parameters (provider, collection concept ID, ...).

I'd also want a test that verifies setting most of the other parameters (like spatial, temporal, page_size, page_num, etc) still get rejected.

We can do most of the testing via unit tests of the function and just add a couple of integration tests for both rejection and non-rejection cases with the parameter set to false.

sxu123: Chris, great catch on this!  I was thinking about this and thought about setting the flag to true in the test file to safeguard it but then forgot all about it.  I think I might have to do that...

sxu123: Sorry, please disregard what I said. I didn't know what I was talking about. :)

search-app/docs/api.md
```diff
@@ -1695,7 +1695,7 @@ __Sample response__
 ```
 
 ### <a name="granule-search-by-parameters"></a> Granule Search By Parameters
-Search performance for granule searches is significantly improved by including an identifier that limits the search to a certain collection or subset of collections. Examples of parameters which limit the scope of the search include collection_concept_id, short_name, entry_title, or provider.
+Search performance for granule searches is significantly improved by including an identifier that limits the search to a certain collection or subset of collections. Examples of parameters which limit the scope of the search include concept_id, collection_concept_id, short_name, version, entry_title, entry_id, provider or provider_id.
```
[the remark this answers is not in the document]

search-app/docs/api.md

system-int-test/test/cmr/system_int_test/search/granule_spatial_search_test.clj
sxu123: Done. Added all in integration tests.

system-int-test/test/cmr/system_int_test/search/granule_spatial_search_test.clj
```diff
@@ -49,6 +51,86 @@
...
+        _ (side/eval-form `(concepts-search/set-allow-all-gran-header! "allow-all-gran"))
+        header1 {"client-id" "testing", "allow-all-gran" true} 
+        header2 {"client-id" "testing", "allow-all-gran" false} 
+        header3 {"client-id" "testing"}
+        header4 {"allow-all-gran" true} 
+        header5 {"allow-all-gran" false}
+        header6 {}
+        params1 {:provider "PROV1"}
+        params2 {:provider_id "PROV1"}
+        params3 {:concept_id "C1234-PROV1"}
```
chris-durbin: This should be a granule concept-id so start with "G" instead of "C"

system-int-test/test/cmr/system_int_test/search/granule_spatial_search_test.clj
```diff
@@ -49,6 +51,86 @@
...
+        params4 {:collection_concept_id "C1234-PROV1"}
```
chris-durbin: Can you also had tests for echo_granule_id, dataset_id, echo_collection_id, and shortName?

These are aliases we support to parameter names concept_id, entry title, collection_concept_id, and short_name, and I've seen all of these used by clients in production. They should be allowed.

sxu123: Done.

system-int-test/test/cmr/system_int_test/search/granule_spatial_search_test.clj
sxu123: I added a test for concept-id being the "G" case since concept-id could be both granule concept id and collection concept id here.

system-int-test/test/cmr/system_int_test/search/granule_spatial_search_test.clj
sxu123: Done.

PR #471, opened 2018-03-07
CMR-4781: Temporarily block a very specific query

chris-durbin:
too_many_requests


oubiwann:
For explicit clarification: this is a hard-coded change against a particular set of queries that are frequent offenders; actual intelligence will be done later.

chris-durbin:
We've verified by looking at Splunk logs that no other client in the past day has sent queries with blank params, and added in the JSON and short_name check to be even more restrictive.

PR #474, opened 2018-03-10
Cmr 4752 reject all granule parameters

sxu123:
For the most part, this has been reviewed and approved by Chris. Started a new branch for the following reasons:
1. Reorganized the code into smaller modules.
2. Added a validation step to reject queries with concept_id not being a granule or collection concept id.
This is an existing bug that will result in a true all granule query.
3. Added code and tests to handle the situations when parameter is passed in with a equal sign but missing a value, in both the single parameter case, and array of same parameter case.
4. Pulled the integration test out of the granule_spatial_test and created its own test.
5. Merge conflict.

Below is the original changes for the ticket:
1. Added defconfig allow-all-granule-params-flag, when set to true, all granule queries are allowed. The default value is set to true because many existing tests need it to pass. Need to write in the release note that we need to set the environment variable to false.
2. Added defconfig allow-all-gran-header, which sets the name for the header(it's done this way to hide the info from general public), when this header is set to true in the request and the client-id is present, operators are allowed to perform all granule queries, regardless of the value of allow-all-granule-params-flag. We need to set the environment variable for the header as well.
3. Added checking before the query is submitted to see if it's a all granule request, if so, based on the values of the first two steps, either allow it or reject it.
4. Added integration tests.
5. Updated the API doc.

### DIFF
--- a/search-app/src/cmr/search/api/concepts_search.clj
+++ b/search-app/src/cmr/search/api/concepts_search.clj
@@ -81,7 +81,7 @@
       "Excessive query rate. Please contact support@earthdata.nasa.gov.")))
 
 (defn- reject-all-granule-query?
-  "Returns true if the all granule query will be rejected."
+  "Return true if the all granule query will be rejected."
   [headers]
   (and (false? (allow-all-granule-params-flag))
        (or (not (some? (get headers "client-id")))
@@ -119,9 +119,9 @@
    filtered out before getting to the params; with the = sign, no value, it becomes empty 
    string."
   [v]
-  (or (= "" v)
-      (when (sequential? v)
-        (= [] (remove #(= "" %) v)))))
+  (if (sequential? v)
+    (empty? (remove #(= "" %) v))
+    (= "" v)))
 
 (defn- all-granule-params?
   "Returns true if it's a all granule query params.
@@ -137,12 +137,15 @@
   (when (= :granule concept-type)
     ;; Check to see if any concept-id(s) are not starting with C or G. If so, bad request.
     ;; otherwise, check to see if it's all-granule-params?, if so, handle it.
-    (let [params (lp/replace-parameter-aliases (util/map-keys->kebab-case params))
-          params (util/remove-map-keys empty-string-values? params)
+    (let [params (->> params
+                      util/map-keys->kebab-case
+                      lp/replace-parameter-aliases
+                      (util/remove-map-keys empty-string-values?)) 
           constraints (select-keys params all-gran-validation/granule-limiting-search-fields)
           concept-id-param (:concept-id constraints)
-          illegal-concept-id-msg (str "Granule query concept_id param [" concept-id-param
-                                      "] contains concept ids not starting with G or C.")]
+          illegal-concept-id-msg (str "Invalid concept_id [" concept-id-param
+                                      "]. For granule queries concept_id must be"
+                                      " either a granule or collection concept ID.")]
       (if (illegal-concept-id-in-granule-query? concept-id-param)
         (svc-errors/throw-service-error :bad-request illegal-concept-id-msg) 
         (when (all-granule-params? scroll-id constraints) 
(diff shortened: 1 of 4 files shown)


search-app/src/cmr/search/api/concepts_search.clj
```diff
@@ -70,6 +80,74 @@
...
+
+(defn empty-string-values? 
+  "This function is used by remove-map-keys function.
+   The keys will be removed if their values are empty strings, or if it's sequential and
+   only contain empty strings.
+   Note:  When you pass a parameter without = sign, it is considered a nil value and get  
+   filtered out before getting to the params; with the = sign, no value, it becomes empty 
+   string."
+  [v]
+  (or (= "" v)
```
chris-durbin:
```
(if (sequential? v)
  (empty? (remove string/blank? v))
  (string/blank? v)
```

search-app/src/cmr/search/api/concepts_search.clj
```diff
@@ -70,6 +80,74 @@
...
+  [headers concept-type params scroll-id]
+  (when (= :granule concept-type)
+    ;; Check to see if any concept-id(s) are not starting with C or G. If so, bad request.
+    ;; otherwise, check to see if it's all-granule-params?, if so, handle it.
+    (let [params (lp/replace-parameter-aliases (util/map-keys->kebab-case params))
+          params (util/remove-map-keys empty-string-values? params)
+          constraints (select-keys params all-gran-validation/granule-limiting-search-fields)
+          concept-id-param (:concept-id constraints)
+          illegal-concept-id-msg (str "Granule query concept_id param [" concept-id-param
+                                      "] contains concept ids not starting with G or C.")]
```
chris-durbin: Invalid concept_id [concept-id-param]. For granule queries concept_id must be either a granule or collection concept ID.

search-app/src/cmr/search/api/concepts_search.clj
```diff
@@ -70,6 +80,74 @@
...
+   Note:  When you pass a parameter without = sign, it is considered a nil value and get  
+   filtered out before getting to the params; with the = sign, no value, it becomes empty 
+   string."
+  [v]
+  (or (= "" v)
+      (when (sequential? v)
+        (= [] (remove #(= "" %) v)))))
+
+(defn- all-granule-params?
+  "Returns true if it's a all granule query params.
```
mschmele: Nitpick: "Return true if it's an all-granule query"

search-app/src/cmr/search/api/concepts_search.clj
```diff
@@ -70,6 +80,74 @@
...
+  (and (not (some? scroll-id))
+       (empty? coll-constraints)))
+
+(defn- handle-granule-search-params
+  "Check the params when it is a granule search query."
+  [headers concept-type params scroll-id]
+  (when (= :granule concept-type)
+    ;; Check to see if any concept-id(s) are not starting with C or G. If so, bad request.
+    ;; otherwise, check to see if it's all-granule-params?, if so, handle it.
+    (let [params (lp/replace-parameter-aliases (util/map-keys->kebab-case params))
```
mschmele: Nitpick:

```
(let [params (->> params
                  util/map-keys->kebab-case
                  lp/replace-parameter-aliases
                  (util/remove-map-keys empty-string-values?))
```

search-app/src/cmr/search/api/concepts_search.clj
sxu123: Done.

search-app/src/cmr/search/api/concepts_search.clj
sxu123: Done.

search-app/src/cmr/search/api/concepts_search.clj
sxu123: Done.

search-app/src/cmr/search/api/concepts_search.clj
sxu123: Just verified that it's possible that v could be non string. Changed to:
(if (sequential? v)
    (empty? (remove #(= "" %) v))
    (= "" v)))

PR #475, opened 2018-03-12
CMR-4650: Change JSON schema for DOI to support two different definitions

chris-durbin:
oneofus

This change allows the UMM for the DOI field to be either DOI and Authority or MissingReason and Explanation. We needed to add support for field level oneOf definitions in JSON schema in order to make this work.

### DIFF
--- a/umm-spec-lib/src/cmr/umm_spec/test/echo10_expected_conversion.clj
+++ b/umm-spec-lib/src/cmr/umm_spec/test/echo10_expected_conversion.clj
@@ -259,6 +259,13 @@
       [(cmn/map->ResourceCitationType
         {:OtherCitationDetails other-citation-details})])))
 
+(defn- expected-doi
+  "Updates the expected DOI for ECHO10."
+  [doi]
+  (let [updated-doi (util/remove-nil-keys (dissoc doi :MissingReason :Explanation))]
+    (when (seq updated-doi)
+      updated-doi)))
+
 (defn umm-expected-conversion-echo10
   [umm-coll]
   (-> umm-coll
@@ -300,4 +307,5 @@
       (assoc :MetadataDates (expected-metadata-dates umm-coll))
       (update :ScienceKeywords expected-science-keywords)
       (update :AccessConstraints conversion-util/expected-access-constraints)
-      (assoc :CollectionProgress (conversion-util/expected-coll-progress umm-coll))))
+      (assoc :CollectionProgress (conversion-util/expected-coll-progress umm-coll))
+      (update :DOI expected-doi)))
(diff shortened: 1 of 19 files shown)


umm-spec-lib/src/cmr/umm_spec/test/echo10_expected_conversion.clj
```diff
@@ -259,6 +259,13 @@
       [(cmn/map->ResourceCitationType
         {:OtherCitationDetails other-citation-details})])))
 
+(defn- expected-doi
+  "Updates the expected DOI for ECHO10."
+  [doi]
+  (let [updated-doi (util/remove-nil-keys (dissoc doi :MissingReason :Explanation))]
```
mschmele: Nitpick: Would a `when-let` work here? I see this `let -> when -> seq` in a few places

umm-spec-lib/src/cmr/umm_spec/xml_to_umm_mappings/iso_shared/doi.clj
```diff
@@ -0,0 +1,40 @@
...
+  "gov.nasa.esdis.umm.doi")
+
+(defn- is-doi-field?
+  "Returns true if the given gmd-id is for a DOI field."
+  [gmd-id]
+  (and (= (value-of gmd-id "gmd:MD_Identifier/gmd:description/gco:CharacterString") "DOI")
+       (= (value-of gmd-id "gmd:MD_Identifier/gmd:codeSpace/gco:CharacterString") doi-namespace)))
+
+(defn parse-doi
+  "There could be multiple CI_Citations. Each CI_Citation could contain multiple gmd:identifiers.
```
mschmele: Super helpful docstring 👍 

Might be worth adding a sentence about why we're picking the first DOI for now

eereiter:
I didn't comment on the clojure style stuff, just looked at the basic logic of the code and it looks good.

umm-spec-lib/src/cmr/umm_spec/test/echo10_expected_conversion.clj
chris-durbin: `updated-doi` is a map, not a collection. I was checking if it was empty (had no keys) with the `seq`. I don't actually want to return the seq of a map, so I couldn't use a `when-let`.

umm-spec-lib/src/cmr/umm_spec/xml_to_umm_mappings/iso_shared/doi.clj
chris-durbin: I can't take credit for this docstring. I don't think there's anything special about the first, that's just the pattern used when there are multiple of something in one format, but we can only use one in another format.

PR #477, opened 2018-03-14
CMR-4785: Prelinimary refactoring for Metadata-DB API cleanup

oubiwann:
There will be several branches/PRs for this work, so this is the first in a series of what might be 3 or 4 PRs.

Tasks completed so far:
* Evaluate current implementation of metadata-cmr API
* Document proposed changes to the API and update development tasks here
* Implement non-breaking changes, test in workload, and merge
  * Convert all extend-* to just extend
  * Separate methods out from records into independent functions
  * Update implicit contextual references to record fields to explicit record field lookups
  * Do this for in-memory DB:
     * cmr.metadata-db.data.memory-db)
  * Do this for Oracle:
     * cmr.metadata-db.data.oracle.concepts
     * cmr.metadata-db.data.oracle.providers
     * cmr.metadata-db.data.oracle.search

A little more explanation about the motivation here:

There are several ways in which implementations for protocols can be done in Clojure ... the way the CMR has done them so far is quite common: bury all your methods in a record and use `extend-type`. This can be nice for isolated code, since it neatly encapsulates everything. However, if you want to share functions between implementations or do other forms of refactoring, you're out of luck.

This PR switches the metadata-db over to splitting out methods into stand-alone functions, gathering them together in a behaviour data structure, and then using `extend` directly instead of the `extend-*` wrapping macros like we have been.

Once this PR lands, we will be in a position to easily refactor the rest of the code in `metadata-db/data`.

### DIFF
--- a/metadata-db-app/src/cmr/metadata_db/data/oracle/search.clj
+++ b/metadata-db-app/src/cmr/metadata_db/data/oracle/search.clj
@@ -149,7 +149,7 @@
 
 (defn find-concepts-in-batches
   ([db provider params batch-size]
-   (c/find-concepts-in-batches db provider params batch-size 0))
+   (find-concepts-in-batches db provider params batch-size 0))
   ([db provider params batch-size requested-start-index]
    (let [{:keys [concept-type]} params
          provider-id (:provider-id provider)
--- a/metadata-db-app/src/cmr/metadata_db/data/oracle/search.clj
+++ b/metadata-db-app/src/cmr/metadata_db/data/oracle/search.clj
@@ -136,75 +136,83 @@
                                                result))
                    (su/query conn stmt))))))
 
-(extend-protocol c/ConceptSearch
-  OracleStore
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
-  (find-concepts
-    [db providers params]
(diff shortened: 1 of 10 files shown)


oubiwann:
Ah, while working in the next branch, I saw a bunch more of the weird semi-circular references the impl is making back up into the API (one of the big downsides to embedding methods in fields instead of having regular functions). I'll push a commit that cleans these up to this branch ...

metadata-db-app/src/cmr/metadata_db/data/oracle/search.clj
```diff
@@ -136,75 +136,83 @@
...
+(defn find-concepts
+  [db providers params]
+  {:pre [(coll? providers)]}
+  (let [concept-type (:concept-type params)
+        tables-to-providers (group-by #(tables/get-table-name % concept-type) providers)]
+    (util/mapcatv (fn [[table provider-list]]
+                    (find-concepts-in-table db table concept-type provider-list params))
+                  tables-to-providers)))
+
+(defn find-concepts-in-batches
```
daniel-zamora: This is a big function

oubiwann: Man, tell me about it.. You would not believe the problems I am discovering as I've been digging into this codebase ... had to create a new Google doc just to track them all ...

PR #479, opened 2018-03-15
CMR-4785: MemoryStore/OracleStore Parity for Metadata-DB 

oubiwann:
This is a continuation of the work started in PR #477.

This change set adjusts the following:
* Brings the in-memory data into  naming, code-layout, and conceptual parity with the oracle data store
* Moves utility functions and constants into places that can be used not only by any metadata-db implementation, but also any other part of the CMR -- without having to call down into the implementation
* Additionally, there was another in-memory db in the CMR used for ACL hash maps; this has been renamed to be more specific and to avoid confusion.


common-lib/src/cmr/common/memorydb/connection.clj
```diff
@@ -0,0 +1,88 @@
+(ns cmr.common.memorydb.connection
```
chris-durbin: I think I'm on board with all the changes except renaming memory-db to memorydb. Can we put that back?

oubiwann:
Looking at a test failure in CI/CD now ...

common-lib/src/cmr/common/memorydb/connection.clj
```diff
@@ -0,0 +1,88 @@
...
+  "Initialize the providers lookup table to the given hash map or set it to a
+  default of an empty hash map."
+  [opts]
+  (or (:providers opts) {}))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; A stoage record for use in parity with cmr.oracle.connection
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(defrecord MemoryStore
```
chris-durbin: These are all required to be atoms in order to work with the code written against this record. It's not obvious in the code that later uses these that they need to be deref'ed. For example the old name of concepts-atom was more clear for a maintainer. I'm not sure what I'm suggesting to do, I think at a minimum documenting they must be atoms here would help - maybe changing the names to atom would help too?

chris-durbin: Oh I actually see they used to all be called -atom. We should put that change back to make it clear.

metadata-db-app/src/cmr/metadata_db/data/const.clj
```diff
@@ -0,0 +1,13 @@
+(ns cmr.metadata-db.data.const
```
chris-durbin: Is it possible to include these vars in the new data/util.clj namespace instead so all the common data functionality is  in one place?

metadata-db-app/src/cmr/metadata_db/services/concept_service.clj
```diff
@@ -22,6 +22,15 @@
...
   ;; Required to get code loaded
+  ;; XXX This is really awful, and we do it a lot in the CMR. What we've got
+  ;;     as a result of this is a leak from implementation into a separate part
+  ;;     of the code ... not to mention that whatever is happing is is 100%
+  ;;     implicit and chock-a-block with side effects. I believe the reason we
+  ;;     do this is to work around issues with multimethods. We really need to
+  ;;     refactor our multimethod code -- and this isn't the only reason
+  ;;     (multimethods use slower code under the hood). If we really do need
+  ;;     the flexible polymorphism that multimethods provide, then let's
+  ;;     rethink our code reuse stragegy around multimethods.
```
chris-durbin: strategy

common-lib/src/cmr/common/memorydb/connection.clj
oubiwann: Sure.

common-lib/src/cmr/common/memorydb/connection.clj
oubiwann: Really? That's like naming all string vars something-string, floats something-float ...

The code that actually creates the instance generates the atoms, and all the references to the memory db's attributes use atom operations (`set!` and `@`), so it's clear everywhere that these are atoms ... maybe we need to have a chat so you can clarify for me what might be unclear to a maintainer? And maybe we can solve that with docs?

metadata-db-app/src/cmr/metadata_db/data/const.clj
oubiwann: Sure, if you want. My habit is to put constants in a const file, but I don't mind another approach.

oubiwann:
Update on CI/CD: branch tests all pass. Just to confirm, I re-ran them locally, too.

Heritage Release passed last night.

Heritage Feature Branches failed last night, with error "No failed tests found". Looks like the issue was WAR files:
```
error	15-Mar-2018 18:04:50	ls: cannot access browse-scaler/echo/echo-browse-scaler/dist/echo-browse-scaler.war: No such file or directory
build	15-Mar-2018 18:04:50	-rw-r--r-- 1 bamboo bamboo 108317434 Mar 15 20:30 legacy-services/echo/echo-rest-rails/dist/echo-rest-rails.war
build	15-Mar-2018 18:04:50	-rw-r--r-- 1 bamboo bamboo 163327773 Mar 15 21:33 catalog-rest/echo/echo-catalog-rest/dist/echo-catalog-rest.war
error	15-Mar-2018 18:04:50	Not all expected war files were built.  Failing!
```

oubiwann:
The CI/CD team has indicated that this shouldn't be a problem (because reasons).

metadata-db-app/src/cmr/metadata_db/services/concept_service.clj
oubiwann: Done.

metadata-db-app/src/cmr/metadata_db/data/const.clj
oubiwann: Done.

common-lib/src/cmr/common/memorydb/connection.clj
oubiwann: Done.

common-lib/src/cmr/common/memorydb/connection.clj
oubiwann: Taking a closer look at this, yeah -- at the very least I'll update the comments in this record to indicate the need for these to be atom. We can chat about the rest of it more, too ...

oubiwann: Okay, I've updated the record as well as its attributes with code comments. Also updated the constructor docstring.

chris-durbin: I don't see any benefit to removing -atom from the vars. I do see a benefit to having -atom in there. I think self documenting vars are better (and by naming them explicitly we don't need documentation).

It's the principle of least surprise (I might be making up that name), but when I see something called concepts I expect it to be a collection of concepts like everywhere else the name concepts is used in the code. When I see concepts-atom I think of an atom holding a collection of concepts.

oubiwann:
@chris-durbin Btw, I've updated the record fields to put `-atom` back in ;-)

chris-durbin:
> @chris-durbin Btw, I've updated the record fields to put -atom back in ;-)
What if I changed my mind?

PR #482, opened 2018-03-16
CMR-4732: Made CMR changes to check on NAMS CMR Ingest group.

ygliuvt:
This PR should be reviewed together with the Legacy Services PR. This added checking of NAMS CMR Ingest group during ingest.

### DIFF
--- a/common-app-lib/src/cmr/common_app/api/request_context_user_augmenter.clj
+++ b/common-app-lib/src/cmr/common_app/api/request_context_user_augmenter.clj
@@ -73,7 +73,7 @@
      (errors/throw-service-error :unauthorized token-required-message)))))
 
 (defn add-user-id-and-sids-to-context
-  "Adds information to the context including the user is and sids. Lazy assoc with a delay so we don't
+  "Adds information to the context including the user id and sids. Lazy assoc with a delay so we don't
   do the expensive work until we need the sids or user id. This is called for every search api
   call, so don't want this to affect performance."
   [context]
--- a/common-app-lib/src/cmr/common_app/api/request_context_user_augmenter.clj
+++ b/common-app-lib/src/cmr/common_app/api/request_context_user_augmenter.clj
@@ -72,11 +72,11 @@
     (when token-required-message
      (errors/throw-service-error :unauthorized token-required-message)))))
 
-(defn- add-user-id-and-sids-to-context
+(defn add-user-id-and-sids-to-context
   "Adds information to the context including the user is and sids. Lazy assoc with a delay so we don't
   do the expensive work until we need the sids or user id. This is called for every search api
   call, so don't want this to affect performance."
-  [context params headers]
+  [context]
   (-> context
       (util/lazy-assoc :sids (context->sids context))
       (util/lazy-assoc :user-id (context->user-id context))))
@@ -86,7 +86,7 @@
   It expects the request context is already associated with the request."
   [handler]
   (fn [request]
-    (let [{:keys [request-context params headers]} request]
+    (let [{:keys [request-context]} request]
       (-> request
-          (update-in [:request-context] add-user-id-and-sids-to-context params headers)
+          (update-in [:request-context] add-user-id-and-sids-to-context)
           (handler)))))
--- a/ingest-app/src/cmr/ingest/api/routes.clj
+++ b/ingest-app/src/cmr/ingest/api/routes.clj
@@ -24,7 +24,7 @@
    [drift.execute :as drift]))
 
 (def URS_TOKEN_MAX_LENGTH 100)
-(def CMR_INGEST_SEPARATOR "CMR_INGEST:")
+(def WRITE_ACCESS_SEPARATOR "WRITE_ACCESS:")
 
 (def db-migration-routes
   (POST "/db-migrate"
@@ -73,8 +73,14 @@
             ctx)
            {:status 200}))))
 
-(defn- add-cmr-ingest-to-context
-  "Add CMR Ingest prefix to token in the request context of the given request.
+(defn- is-launchpad-token?
+  "Returns true if the given token is a launchpad token.
+   Currently we only check the length of the token to decide."
+  [token]
+  (> (count token) URS_TOKEN_MAX_LENGTH))
+
+(defn- add-write-access-to-request
+  "Add Write Access prefix to token in the request context of the given request.
    This function should be called on routes that will ingest into CMR.
    The CMR Ingest prefix is used to indicates to legacy services that when
    validating the Launchpad token, the NAMS CMR Ingest group should also be checked.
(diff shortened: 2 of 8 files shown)


common-app-lib/src/cmr/common_app/api/request_context_user_augmenter.clj
```diff
@@ -72,11 +72,11 @@
     (when token-required-message
      (errors/throw-service-error :unauthorized token-required-message)))))
 
-(defn- add-user-id-and-sids-to-context
+(defn add-user-id-and-sids-to-context
   "Adds information to the context including the user is and sids. Lazy assoc with a delay so we don't
```
chris-durbin: "user id"

ingest-app/src/cmr/ingest/api/routes.clj
```diff
@@ -69,6 +73,24 @@
...
+   This function should be called on routes that will ingest into CMR.
+   The CMR Ingest prefix is used to indicates to legacy services that when
+   validating the Launchpad token, the NAMS CMR Ingest group should also be checked.
+   Ingest will only be allowed if the user is in the NAMS CMR Ingest group and
+   also has the right ACLs which is based on Earthdata Login uid."
+  [request]
+  (let [token (-> request :request-context :token)]
+    (if (> (count token) URS_TOKEN_MAX_LENGTH)
+      ;; for Launchpad token add CMR_INGEST: prefix so that legacy service
+      ;; can do extra vaidation to check if the user has signed up for
```
chris-durbin: validation

chris-durbin: > signed up for

been approved for the

ingest-app/src/cmr/ingest/api/routes.clj
```diff
@@ -69,6 +73,24 @@
...
+(defn- add-cmr-ingest-to-context
+  "Add CMR Ingest prefix to token in the request context of the given request.
+   This function should be called on routes that will ingest into CMR.
+   The CMR Ingest prefix is used to indicates to legacy services that when
+   validating the Launchpad token, the NAMS CMR Ingest group should also be checked.
+   Ingest will only be allowed if the user is in the NAMS CMR Ingest group and
+   also has the right ACLs which is based on Earthdata Login uid."
+  [request]
+  (let [token (-> request :request-context :token)]
+    (if (> (count token) URS_TOKEN_MAX_LENGTH)
```
chris-durbin: Even though it is just one line it might be worth pulling this out into its own function `is-launchpad-token?` to make it obvious this is the check we're performing to decide if a token is a Launchpad token.

ingest-app/src/cmr/ingest/api/routes.clj
```diff
@@ -69,6 +73,24 @@
...
+      ;; for Launchpad token add CMR_INGEST: prefix so that legacy service
+      ;; can do extra vaidation to check if the user has signed up for
+      ;; CMR Ingest workflow.
+      (-> request
+          (update-in [:request-context :token] #(str CMR_INGEST_SEPARATOR %))
+          (update-in [:request-context] augmenter/add-user-id-and-sids-to-context))
```
chris-durbin: Why did we need to explicitly call this? Isn't it already added in a ring handler?

ingest-app/src/cmr/ingest/api/routes.clj
```diff
@@ -22,6 +23,9 @@
    [compojure.core :refer [DELETE GET POST PUT context routes]]
    [drift.execute :as drift]))
 
+(def URS_TOKEN_MAX_LENGTH 100)
```
chris-durbin: Both of these vars should have docstrings.

ingest-app/src/cmr/ingest/api/routes.clj
```diff
@@ -22,6 +23,9 @@
...
+(def CMR_INGEST_SEPARATOR "CMR_INGEST:")
```
chris-durbin: I think this is supposed to signify we need to check whether someone should be allowed to write to NASA systems (broader than just ingest access). So I think the name should be something like WRITE_ACCESS rather than CMR_INGEST.

ingest-app/src/cmr/ingest/api/routes.clj
```diff
@@ -69,6 +73,24 @@
             ctx)
            {:status 200}))))
 
+(defn- add-cmr-ingest-to-context
```
chris-durbin: Similarly I'd change the name of this function.

ingest-app/src/cmr/ingest/api/routes.clj
```diff
@@ -83,10 +105,10 @@
        (context ["/collections/:native-id" :native-id #".*$"] [native-id]
          (PUT "/"
            request
-           (collections/ingest-collection provider-id native-id request))
+           (collections/ingest-collection provider-id native-id (add-cmr-ingest-to-context request)))
```
chris-durbin: It seems like this is the kind of thing that should be put into middleware. Ideally we would have a way to identify routes that require Launchpad write access rather than explicitly calling this function on every route. I'm not exactly sure how to accomplish that, maybe someone else has an idea.

ingest-app/src/cmr/ingest/api/routes.clj
chris-durbin: Also note we'll need to do this in other applications, so `common-app-lib` is the place to add this code rather than `ingest.api.routes`.

transmit-lib/src/cmr/transmit/echo/tokens.clj
```diff
@@ -48,9 +48,9 @@
                                              :form-params {:id token}})]
       (case (int status)
         200 (get-in parsed [:token_info :user_name])
-        401 (errors/throw-service-error
+        401 (errors/throw-service-errors
              :unauthorized
-             (format "Token %s does not exist" token))
+             (:errors (json/decode body true)))
```
chris-durbin: If someone tries to use a Launchpad token (or anything longer than 100 characters) in one of the environments will they get a 401 back? I just want to make sure we don't throw a 500 error because we're not actually set up to use Launchpad anywhere without the manual cacerts install.

common-app-lib/src/cmr/common_app/api/request_context_user_augmenter.clj
ygliuvt: Done.

transmit-lib/src/cmr/transmit/echo/tokens.clj
oubiwann: Sounds like we need an explicit test for this ...

ingest-app/src/cmr/ingest/api/routes.clj
ygliuvt: Done.

ygliuvt: Done.

ingest-app/src/cmr/ingest/api/routes.clj
ygliuvt: Good idea. Done.

ingest-app/src/cmr/ingest/api/routes.clj
ygliuvt: When the ring handler is constructed, the request context does not have the "CMR_INGEST:" prefix added to the token. Since we added the prefix, we need to update the handler to use the correct token.

ingest-app/src/cmr/ingest/api/routes.clj
ygliuvt: Renaming done. The move to common-app-lib will be done later in real implementation (not prototyping).

ingest-app/src/cmr/ingest/api/routes.clj
ygliuvt: Done.

ingest-app/src/cmr/ingest/api/routes.clj
ygliuvt: Added comment to explain why this line is added here.

transmit-lib/src/cmr/transmit/echo/tokens.clj
ygliuvt: We will return 401 back with "Token [xxx] does not exist." in this case.

PR #483, opened 2018-03-19
CMR-4651: OnlineResourceType and UseConstraintsType changes.

sxu123:
There are two parts of schema changes involved
Part 1: OnlineResourceType change: a. Added MimeType, b. Removed required Description and Name. There is no translation of MimeType. So no new data is needed. 
Part2: UseConstraintsType Change, which will make use of the new OnlineResouceType:  It is changed from a string to an object containing objects. Existing dif/dif10 data covers it. Echo10 has no change. Added new ISO19115 data file. 
Note: 
1. the schema change recommended for this in the ticket doesn't work. I had to use an alternative way to accomplish the same:
a. oneOf [ {Description LicenseUrl} {Description LicenseText}] doesn't seem to work with the same "Description" in both. when validating, it doesn't know which schema to use.
b. oneOf [ {required "Description"} {required "LicenseUrl"}] doesn't work the way it looks. It doesn't allow
both "Description" and "LicenseUrl" to appear together.  This doesn't work even when you have the simplest structure without the nested oneOf. 

2. We don't have a mechanism to really test the schema currently.  In generate-and-parse, we just validate the umm AFTER it's converted from the xml file, which will never have all "Description" "LicenseUrl" and "LicenseText".  I hard-coded one there(removed now) to see if my schema will allow the different combinations and reject the case when both LicenseUrl and LicenseText are present. So I know the schema works as expected and that's how I found out the oneOf doesn't work.

3. I'm not familiar with how the umm are generated based on the schema. One strange thing is that even though I've tested that LicenseUrl and LicenseText can not appear together to pass the validation. But when the collection is generated, it could still contain both(with nil values).  Had to sanitize it. 

4. Found a bug in ingest validation that that is supposed to limit the ingest version - have to fix it to resolve other test failures because it depends on this to work.

### DIFF
--- a/ingest-app/src/cmr/ingest/validation/validation.clj
+++ b/ingest-app/src/cmr/ingest/validation/validation.clj
@@ -100,7 +100,10 @@
                    (if (mt/umm-json? (:format concept))
                      (let [umm-version (mt/version-of (:format concept))
                            accept-version (config/ingest-accept-umm-version (:concept-type concept))]
-                       (if (>= 0 (compare umm-version accept-version))
+                       ;; when the umm-version goes to v1.10 and accept-version is v1.9, we need
+                       ;; to compare the numbers after the dot.  
+                       (if (>= 0 (compare (Integer. (last (str/split umm-version #"\.")))
+                                          (Integer. (last (str/split accept-version #"\.")))))
                          (umm-spec/validate-metadata (:concept-type concept)
                                                      (:format concept)
                                                      (:metadata concept))
--- a/ingest-app/src/cmr/ingest/validation/validation.clj
+++ b/ingest-app/src/cmr/ingest/validation/validation.clj
@@ -114,8 +114,8 @@
                    (if (mt/umm-json? (:format concept))
                      (let [umm-version (mt/version-of (:format concept))
                            accept-version (config/ingest-accept-umm-version (:concept-type concept))]
-                       ;; when the umm-version goes to v1.10 and accept-version is v1.9, we need
-                       ;; to compare the numbers after the dot.  
+                       ;; when the umm-version goes to 1.10 and accept-version is 1.9, we need
+                       ;; to compare the versions with padded zeros.  
                        (if (>= 0 (compare-versions-with-padded-zeros umm-version accept-version))
                          (umm-spec/validate-metadata (:concept-type concept)
                                                      (:format concept)
--- a/ingest-app/src/cmr/ingest/validation/validation.clj
+++ b/ingest-app/src/cmr/ingest/validation/validation.clj
@@ -94,6 +94,20 @@
      :ISOTopicCategories (match-kms-keywords-validation
                           kms-index :iso-topic-categories msg/iso-topic-category-not-matches-kms-keywords)}))
 
+(defn- pad-zeros-to-version
+  "Pad 0's to umm versions. Example: 1.9.1 becomes 01.09.01, 1.10.1 becomes 01.10.01"
+  [version]
+  (let [version-splitted (str/split version #"\.")]
+    (str/join "." (map #(if (> 10 (Integer. %)) (str "0" %) %) version-splitted))))
+
+(defn- compare-versions-with-padded-zeros
+  "Compare the umm-version and accepted umm-version
+   with padded 0's."
+  [umm-version accepted-umm-version]
+  (let [umm-version-with-padded-zeros (pad-zeros-to-version umm-version)
+        accepted-umm-version-with-padded-zeros (pad-zeros-to-version accepted-umm-version)]
+    (compare umm-version-with-padded-zeros accepted-umm-version-with-padded-zeros))) 
+
 (defn validate-concept-metadata
   [concept]
   (if-errors-throw :bad-request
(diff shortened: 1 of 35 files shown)


umm-spec-lib/src/cmr/umm_spec/test/iso19115_expected_conversion.clj
```diff
@@ -51,8 +51,8 @@
  (when online-resource
   (-> online-resource
       (update :Linkage #(url/format-url % true))
-      (update :Name #(su/with-default % true))
-      (update :Description #(str (su/with-default % true) " PublicationReference:")))))
+      (update :MimeType #(when % nil))
```
daniel-zamora:
```
(update {:a 1 :b 2} :a #(when % nil))

user=>
{:b 2, :a nil}
(update {:a nil :b 2} :a #(when % nil))

user=>
{:b 2, :a nil}
(update {:b 2} :a #(when % nil))

user=>
{:b 2, :a nil}```

I tried this update, I'm still confused what its trying to do.

Comment (sxu123) [reply umm-spec-lib/src/cmr/umm_spec/test/iso19115_expected_conversion.clj]: I understand why you are confused. It took me sometime on this one. I originally just removed MimeType but it didn't  work. It seems that the generated one always have :MimeType, either nil or with value. I'm closing the PR for now because I'm almost done with the second part, will combine them together. I will take another look at this. Thanks.

File: ingest-app/src/cmr/ingest/validation/validation.clj [hunk 2]
```diff
@@ -100,7 +100,10 @@
                    (if (mt/umm-json? (:format concept))
                      (let [umm-version (mt/version-of (:format concept))
                            accept-version (config/ingest-accept-umm-version (:concept-type concept))]
-                       (if (>= 0 (compare umm-version accept-version))
+                       ;; when the umm-version goes to v1.10 and accept-version is v1.9, we need
+                       ;; to compare the numbers after the dot.  
+                       (if (>= 0 (compare (Integer. (last (str/split umm-version #"\.")))
```
mschmele: We should make sure that there is unit test coverage for this

system-int-test/test/cmr/system_int_test/ingest/ingest_lifecycle_integration_test.clj
```diff
@@ -147,6 +149,9 @@
 (deftest mmt-ingest-round-trip
   (testing "ingest and search UMM JSON metadata"
     ;; test for each UMM JSON version
+  (let [accepted-version (common-config/collection-umm-version)
```
mschmele: I've noticed that this bit of code got added to just about every one of the system-int tests. Is this something that unit tests could replace, or is it necessary for these tests to function?

sxu123: We should always test the latest version, which is ver/current-collection-version. But since ingest limits it to common-(config/collection-umm-version), we need to set it to the latest version in our tests.

ingest-app/src/cmr/ingest/validation/validation.clj
ygliuvt: This will not work for comparing `1.9` and `2.1`. We should convert the versions to float first, then compare them.

system-int-test/test/cmr/system_int_test/ingest/ingest_lifecycle_integration_test.clj
ygliuvt: How about moving this to the fixture and use `use-fixture` to set it rather than putting this in every test?

umm-spec-lib/resources/json-schemas/collection/umm/v1.10/umm-c-json-schema.json
```diff
@@ -1,6 +1,45 @@
 {
   "$schema": "http://json-schema.org/draft-04/schema#",
   "definitions": {
+    "UseConstraintsDescriptionType": {
```
ygliuvt: Why don't we make this a new version?

ingest-app/src/cmr/ingest/validation/validation.clj
sxu123: I thought about that, but we can't guarantee that it'll be a float number either.  what if we will have 1.10.1 or sth like that?

system-int-test/test/cmr/system_int_test/ingest/ingest_lifecycle_integration_test.clj
sxu123: Because there are still some tests there that want to test the config/collection-umm-version, I think.  Daniel created many of the curr-ingest-ver-example-collection-record. I didn't remove them all, I just fixed the few ones that failed for me.

umm-spec-lib/resources/json-schemas/collection/umm/v1.10/umm-c-json-schema.json
sxu123: I don't know. @eereiter?

ingest-app/src/cmr/ingest/validation/validation.clj
ygliuvt: That's a good point. We should split the number by `.` and then compare the integer in orders.

sxu123: Done.

eereiter:
I am commenting on your pull request comments. For #2, "We don't have a mechanism to really test the schema currently."  I have been using this web site to test my json changes.  With this link: https://json-schema-validator.herokuapp.com/syntax.jsp you can test the JSON schema. Then using the following site: https://json-schema-validator.herokuapp.com/ where you can add the correct schema and validate your test json documents against it.  

Jason had mentioned that the CMR would support JSON schema draft 4, so I am assuming that there is a bug in supporting oneOf and also probably anyOf and allOf.

umm-spec-lib/resources/json-schemas/collection/umm/v1.10/umm-c-json-schema.json
eereiter: Because the customer wants to know what will be in a version up front, so I have defined what will be in each version.  So the versions are now defined by capability and not sprint time boundaries.

PR #486, opened 2018-03-23
CMR-4821: Move up tag association and other system concepts cleanup ahead of provider concept table cleanup.

chris-durbin:
cleanup


chris-durbin:
This PR is to try to quickly resolve an operational performance problem, which seems to be caused by the CMR_TAG_ASSOCIATIONS table having thousands of old revisions for many concepts that have not been cleaned up since March 4th.

PR #490, opened 2018-04-02
CMR-4839: Contains the last piece of clarification for CMR-4651.

sxu123:
1. Added 1.9 to 1.10 migration for OnlineResourceType change; 
2. Modified the iso=>UMM parsing.

### DIFF
--- a/umm-spec-lib/src/cmr/umm_spec/migration/related_url_migration.clj
+++ b/umm-spec-lib/src/cmr/umm_spec/migration/related_url_migration.clj
@@ -91,10 +91,24 @@
    Need to remove MimeType, add default for Name and Description if they don't exist."
   [element]
   (if-let [ol-resource (:OnlineResource element)]
-    (-> element
-        (assoc :OnlineResource {:Linkage (:Linkage ol-resource)
-                                :Name (util/with-default (:Name ol-resource) true)
-                                :Description (util/with-default (:Description ol-resource) true)}))
+    (assoc element :OnlineResource {:Linkage (:Linkage ol-resource)
+                                    :Name (util/with-default (:Name ol-resource) true)
+                                    :Description (util/with-default (:Description ol-resource) true)})
+    element))
+
+(defn- remove-default-name-description
+  "Remove default Name and Description from online-resource."
+  [online-resource]
+  (apply dissoc online-resource (for [[k v] (select-keys online-resource [:Name :Description])
+                                      :when (= util/not-provided v)]
+                                  k))) 
+
+(defn migrate-online-resource-up
+  "Migrate online-resource from version 1.9 to 1.10.
+   Need to remove Name and Description if their values are util/not-provided."
+  [element]
+  (if-let [ol-resource (:OnlineResource element)]
+    (assoc element :OnlineResource (remove-default-name-description ol-resource))
     element))
 
 (defn migrate-related-url-to-online-resource
--- a/umm-spec-lib/src/cmr/umm_spec/migration/related_url_migration.clj
+++ b/umm-spec-lib/src/cmr/umm_spec/migration/related_url_migration.clj
@@ -105,7 +105,8 @@
 
 (defn migrate-online-resource-up
   "Migrate online-resource from version 1.9 to 1.10.
-   Need to remove Name and Description if their values are util/not-provided."
+   Need to remove Name and Description if their values are util/not-provided
+   because they are artificially added when migrating from 1.10 to 1.9."
   [element]
   (if-let [ol-resource (:OnlineResource element)]
     (assoc element :OnlineResource (remove-default-name-description ol-resource))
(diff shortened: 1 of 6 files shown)


umm-spec-lib/src/cmr/umm_spec/migration/related_url_migration.clj
```diff
@@ -91,10 +91,24 @@
...
-        (assoc :OnlineResource {:Linkage (:Linkage ol-resource)
-                                :Name (util/with-default (:Name ol-resource) true)
-                                :Description (util/with-default (:Description ol-resource) true)}))
+    (assoc element :OnlineResource {:Linkage (:Linkage ol-resource)
+                                    :Name (util/with-default (:Name ol-resource) true)
+                                    :Description (util/with-default (:Description ol-resource) true)})
+    element))
+
+(defn- remove-default-name-description
+  "Remove default Name and Description from online-resource."
```
mschmele: Try to add to this docstring something about why we want to remove defaults in this case. This is one thing that we're often inconsistent about. I think we should be better about documenting why that is, even if that's just "UMM 1.x doesn't have defaults, but UMM 1.y does"

umm-spec-lib/src/cmr/umm_spec/migration/related_url_migration.clj
```diff
@@ -91,10 +91,24 @@
...
   [element]
   (if-let [ol-resource (:OnlineResource element)]
-    (-> element
-        (assoc :OnlineResource {:Linkage (:Linkage ol-resource)
-                                :Name (util/with-default (:Name ol-resource) true)
-                                :Description (util/with-default (:Description ol-resource) true)}))
+    (assoc element :OnlineResource {:Linkage (:Linkage ol-resource)
+                                    :Name (util/with-default (:Name ol-resource) true)
+                                    :Description (util/with-default (:Description ol-resource) true)})
+    element))
```
mschmele: Does `assoc` not return what you were expecting here? It's inside the scope of the `if-let`, and it looks like you might not be getting the data that you're expecting

umm-spec-lib/src/cmr/umm_spec/migration/related_url_migration.clj
```diff
@@ -91,10 +91,24 @@
...
+(defn- remove-default-name-description
+  "Remove default Name and Description from online-resource."
+  [online-resource]
+  (apply dissoc online-resource (for [[k v] (select-keys online-resource [:Name :Description])
+                                      :when (= util/not-provided v)]
+                                  k))) 
+
+(defn migrate-online-resource-up
+  "Migrate online-resource from version 1.9 to 1.10.
+   Need to remove Name and Description if their values are util/not-provided."
```
mschmele: Again about why we're removing these values

umm-spec-lib/src/cmr/umm_spec/migration/related_url_migration.clj
sxu123: If there is no OnlineResource in element, there is no point changing the OnlineResource(removing MimeType and add default values for the Name and Description), so just return element. 
Note: I'm not really changing the code here, but to refactor it so that we don't use threading when we have less than 3 things to thread.

umm-spec-lib/src/cmr/umm_spec/migration/related_url_migration.clj
sxu123: We don't have to remove these default values for the code to function correctly in the real world. It just looks cleaner when we add default values migrating one direction and remove the default values when migrating the opposite direction. This is why it worked fine without this direction of migration in my previous PR.  I think Adding simple comment is just going to be more confusing.

umm-spec-lib/src/cmr/umm_spec/migration/related_url_migration.clj
sxu123: Again I don't like adding simple comment that doesn't tell the whole story. but if it makes you feel better, I can add a little. Done.

umm-spec-lib/src/cmr/umm_spec/migration/related_url_migration.clj
sxu123: If you use assoc only, when there is no OnlineResource, it will return invalid OnlineResource that has nil Linkage(required field) and default Name and Description, which is wrong.

umm-spec-lib/src/cmr/umm_spec/migration/related_url_migration.clj
oubiwann: Ordinarily, I'd want to see a tiny bit more in the docstring for a utility function, but this is only used once right now, and in this same file. The split was really made just to keep the code clean and simple, and it's not a public function or anything.

We will want to revisit this when another function use the utility function here, but that additional use can be used to improve the docs at that time. I'm okay with this as it is.

oubiwann:
@sxu123 Looks good!

@mschmele When you get a chance, could you re-review? @sxu123's rationales seem sound to me ...

PR #491, opened 2018-04-03
CMR-4774: Fixed a bug that caused the Sort params not logged.

sxu123:


oubiwann:
Can you write a test that breaks without this fix?

sxu123:
@oubiwann, The fix is very obvious - :sort-params key doesn't exist in the (query->execution-params query), it is replaced by :sort key.   I know usually we add tests for fixes but in this case, it only impacts the logging, so I'm not sure how to test it other than running the demo to show the actual logging, which I did.  Any suggestions?

PR #492, opened 2018-04-04
CMR-4777 adds logging for acl entry-titles/concept-ids sync function

daniel-zamora:


oubiwann:
@daniel-zamora, Without digging into your change or the ticket, I have a lazy question for you: we have a timed-function macro in the CMR (and we use it all over the place) -- did you look at that at all to see if it could suit your purpose here? Would be a shame to have some one-off here that could be avoided ...

daniel-zamora:
@oubiwann I copied this code from elsewhere in the code base.  This is how its done in at least 1 other location.

daniel-zamora:
@oubiwann looked at the macro, it does same thing but allows the logging to be toggled.  Using it would be less descriptive though, we would lose the count on synced entries OR have 2 separated log strings.

oubiwann:
@daniel-zamora Sorry, man -- got bogged down today; should have checked back in on this sooner ... approving now ...

oubiwann:
Also: Thanks for checking up on that!

PR #493, opened 2018-04-05
Cmr 4846

eereiter:
remove the required constraint for scienceKeywords in UMM-Var


common-app-lib/resources/humanizers.json
```diff
@@ -1907,6 +1907,14 @@
         "source_value": "NASA National Snow and Ice Data Center Distributed Active Archive Center",
         "type": "alias"
     },
+    {
+        "field": "organization",
```
ygliuvt: So many different ways to call NSIDC. :)

PR #495, opened 2018-04-06
CMR-4846 removing required constraint for sciencekeywords on UMM-Var

eereiter:


eereiter:
I will wait until the bamboo for my branches build is complete and is green before I merge.

oubiwann:
Btw, it's nice to see another PR from you in CMR :-D

eereiter:
Thanks,  Valerie was asking if we had any extra cycles.  Since this one was really quick I decided to go ahead and do it.

umm-spec-lib/resources/json-schemas/variable/umm/v1.1/umm-var-json-schema.json
```diff
@@ -289,7 +289,6 @@
       "Dimensions",
       "Sets",
       "Scale",
-      "Offset",
-      "ScienceKeywords"
```
sxu123: 1. Do you need to check the code to see if any default values were added before when they were required and remove them? 
2.  I think you will need migration for it just like what I did when migrating online resource from 1.9 to 1.10, I had to remove the default values?

eereiter: 1) No - there are no default values for UMM-Var - if science keywords didn't exist in the record then the record couldn't be ingested. 
2) Again there are no default values for UMM-Var so there isn't anything for me to change.

PR #496, opened 2018-04-09
CMR-4841: Log the first 255 characters of the response body.

sxu123:
The error happens when the parsing of the html response body fails. We should catch the exception and log the first 255 characters of the body when it occurs. 
Note: This fix only involves logging, so there won't be testing for it.  Have tried to manually introduce an error and ran a demo to verify the code worked.

### DIFF
--- a/transmit-lib/src/cmr/transmit/metadata_db.clj
+++ b/transmit-lib/src/cmr/transmit/metadata_db.clj
@@ -334,11 +334,11 @@
         ;; For CMR-4841 - log the first 255 characters of the response body if
         ;; the parsing of the html throws exception. 
         response-body (:body response)
-        response-body-logged (subs response-body 0 (min 255 (.length response-body)))
+        response-body-logged (subs response-body 0 (min 255 (count response-body)))
         body (try
                (json/decode response-body)
                (catch Exception e 
-                 (info "Exception occurred while parsing the response body: " response-body-logged)
+                 (warn "Exception occurred while parsing the response body: " response-body-logged)
                  (throw e)))
         {:strs [concept-id revision-id]} body]
     (case status
--- a/transmit-lib/src/cmr/transmit/metadata_db.clj
+++ b/transmit-lib/src/cmr/transmit/metadata_db.clj
@@ -331,7 +331,15 @@
                                 :throw-exceptions false
                                 :headers (ch/context->http-headers context)}))
         status (int (:status response))
-        body (json/decode (:body response))
+        ;; For CMR-4841 - log the first 255 characters of the response body if
+        ;; the parsing of the html throws exception. 
+        response-body (:body response)
+        first-255-body (subs response-body 0 (min 255 (.length response-body)))
+        body (try
+               (json/decode response-body)
+               (catch Exception e 
+                 (info "Exception occurred while parsing the response body: " first-255-body)
+                 (throw e)))
         {:strs [concept-id revision-id]} body]
     (case status
       422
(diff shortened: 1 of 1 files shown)


oubiwann:
Hrm, I just hit "approve" ... but now I' thinking about testing ... hold off on a merge; I want to check something.

transmit-lib/src/cmr/transmit/metadata_db.clj
```diff
@@ -331,7 +331,15 @@
...
         status (int (:status response))
-        body (json/decode (:body response))
+        ;; For CMR-4841 - log the first 255 characters of the response body if
+        ;; the parsing of the html throws exception. 
+        response-body (:body response)
+        response-body-logged (subs response-body 0 (min 255 (.length response-body)))
+        body (try
+               (json/decode response-body)
+               (catch Exception e 
+                 (info "Exception occurred while parsing the response body: " response-body-logged)
```
oubiwann: Let's change this to `warn` instead of `info`.

transmit-lib/src/cmr/transmit/metadata_db.clj
```diff
@@ -331,7 +331,15 @@
                                 :throw-exceptions false
                                 :headers (ch/context->http-headers context)}))
         status (int (:status response))
-        body (json/decode (:body response))
+        ;; For CMR-4841 - log the first 255 characters of the response body if
+        ;; the parsing of the html throws exception. 
+        response-body (:body response)
+        response-body-logged (subs response-body 0 (min 255 (.length response-body)))
```
oubiwann: The standard way of getting sizing info in Clojure is `count`; did you try that instead of the `.length` Java method? It may be that count won't work on a byte stream? Dunno ... would have to check. This is one of the reasons I'd like to see a test ...

oubiwann:
So, I did a `git grep transmit.meta` and there is _nothing_ in `system-int-test` :-(

If this was me, I'd probably set up an integration-only test (not a `system-int-test`) in `transmit-lib/int-test` similarly as done in some of the other CMR projects. Not going to ask you do do that, tough :-) (Should be a ticket in its own right, really ...)

I'd definitely change the `info` to `warn`, though ...

transmit-lib/src/cmr/transmit/metadata_db.clj
sxu123: response-body should be a string.  I have seen both count and .length, I just picked one, for no reason :)  I will change it to count if preferred.

transmit-lib/src/cmr/transmit/metadata_db.clj
```diff
@@ -331,7 +331,15 @@
                                 :throw-exceptions false
                                 :headers (ch/context->http-headers context)}))
         status (int (:status response))
-        body (json/decode (:body response))
+        ;; For CMR-4841 - log the first 255 characters of the response body if
+        ;; the parsing of the html throws exception. 
+        response-body (:body response)
```
ygliuvt: These should be moved to the catch block.

sxu123: Why?

transmit-lib/src/cmr/transmit/metadata_db.clj
```diff
@@ -331,7 +331,15 @@
...
+        response-body-logged (subs response-body 0 (min 255 (count response-body)))
+        body (try
+               (json/decode response-body)
+               (catch Exception e
```
ygliuvt: The exception is too general. It should be limited to just the JsonParseException that is in the stack trace.

transmit-lib/src/cmr/transmit/metadata_db.clj
ygliuvt: Didn't pay attention. So move the comment and response-body-logged to the catch block.

transmit-lib/src/cmr/transmit/metadata_db.clj
ygliuvt: Yes. Thanks!

PR #498, opened 2018-04-11
CMR-4808 adds umm s version 1.1

daniel-zamora:
UMM-S Service/Access Constraints should not be repeatable as there is only one set of access constraints per collection record. Leave as plural and should be object and not array
UMM-S Service/Access Constraints: Increase Field length to 4,000 characters
UMM-S Service/Use Constraints: Increase Field Length to 20,000 characters
UMM-S Service/Use Constraints: Field Should be set as an object and not an array (not repeatable) in the schema
UMM-S Service/ServiceQuality/Lineage: Increase Field Length to 4,000 characters
MM-S Service/RelatedURL: Field Should Be Repeatable (make multiple) in schema
UMM-S Service/RelatedURL/GetService: Sub-Field Should Not Be Required
Add "TOOL" just after "THREDDS DIRECTORY" to the list of valid RelatedURLSubTypeEnum. Depending if we use the same validation as Collections then we will also need to change cmr.umm-spec.util and add TOOL there as well.
UMM-S Service/Coverage/Type removed
UMM-S Service/Coverage/CoverageSpatialExtent/CoverageSpatialExtentType added, with enum
UMM-S Service/Coverage/CoverageSpatialExtent/CoverageSpatialExtentType added, with enum

### DIFF
--- a/umm-spec-lib/src/cmr/umm_spec/models/umm_service_models.clj
+++ b/umm-spec-lib/src/cmr/umm_spec/models/umm_service_models.clj
@@ -5,10 +5,6 @@
 
 (defrecord UMM-S
   [
-   ;; This element contains important information about the universal resource locator (URL) for the
-   ;; service.
-   RelatedURL
-
    ;; This element contains important information about the coverage for the service.
    Coverage
 
@@ -52,6 +48,10 @@
 
    ;; The name of the service, software, or tool.
    Name
+
+   ;; This element contains important information about the universal resource locator (URL) for the
+   ;; service.
+   RelatedURLs
 
    ;; A brief description of the service.
    Description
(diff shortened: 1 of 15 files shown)


oubiwann:
I've got a few questions about this, but mostly because I haven't been in the loop -- I'll do those line-by-line. I've asked @eereiter to sanity check it, as well.

Did want to give you a heads-up that the Bamboo build is red right now, though (dunno if that's just one of the intermittent network issues or not ...)

umm-spec-lib/src/cmr/umm_spec/models/umm_service_models.clj
```diff
@@ -168,9 +168,6 @@
    ;; This is the long name of the service organization.
    LongName
 
-   ;; Uuid of the service organization.
```
oubiwann: Is github going a bad job of the diffs here? Looks like some fields got dropped that I wasn't expecting? (maybe they've just shifted elsewhere in the file, and I'm not seeing where ...)

daniel-zamora: Its possible that this was dropped. I downloaded the schema v1.1 from the repo and not all changes were listed in the ticket.  But if its changed in the schema then it should be changed in this PR.

PR #502, opened 2018-04-13
CMR-4869: Fixed error in searching UMM JSON collections in native format

ygliuvt:
Fixed error in searching UMM JSON collections in native format.

### DIFF
--- a/search-app/src/cmr/search/results_handlers/metadata_results_handler.clj
+++ b/search-app/src/cmr/search/results_handlers/metadata_results_handler.clj
@@ -1,20 +1,22 @@
 (ns cmr.search.results-handlers.metadata-results-handler
   "Handles search results with metadata including ECHO10 and DIF formats."
-  (:require [cmr.common-app.services.search.elastic-results-to-query-results :as elastic-results]
-            [cmr.common-app.services.search.elastic-search-index :as elastic-search-index]
-            [cmr.common-app.services.search :as qs]
-            [cmr.search.services.query-execution :as qe]
-            [cmr.search.services.query-execution.granule-counts-results-feature :as gcrf]
-            [cmr.search.services.query-execution.facets.facets-results-feature :as frf]
-            [cmr.search.services.query-execution.tags-results-feature :as trf]
-            [cmr.common-app.services.search.results-model :as results]
-            [cmr.search.data.metadata-retrieval.metadata-cache :as metadata-cache]
-            [clojure.data.xml :as x]
-            [cmr.common.xml :as cx]
-            [clojure.string :as str]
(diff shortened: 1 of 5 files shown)


search-app/src/cmr/search/results_handlers/metadata_results_handler.clj
```diff
@@ -109,6 +111,13 @@
                                (when data
                                  (x/element :data {} (json/generate-string data))))))))]))
 
+(defn xml-escape-umm-json-metadata
+  "Returns the metadata. Xml escape the special characters in UMM JSON metadata if applicable."
+  [format metadata]
+  (if (mt/umm-json? format)
+    (string/escape metadata {\< "&lt;", \> "&gt;", \& "&amp;"})
```
oubiwann: Hrm ... this seems too hand-tooled to me (and thus too fragile). A more robust solution would be to use whatever community standard there is in Clojure for escaping HTML/XML entities ...

search-app/src/cmr/search/results_handlers/metadata_results_handler.clj
```diff
@@ -109,6 +111,13 @@
                                (when data
                                  (x/element :data {} (json/generate-string data))))))))]))
 
+(defn xml-escape-umm-json-metadata
```
mschmele: Nitpick: this name is a little confusing. Would something like `escape-special-characters-for-umm-json` make sense?

mschmele:
+1ing the comments that @oubiwann made as well

search-app/src/cmr/search/results_handlers/metadata_results_handler.clj
oubiwann: More "Hrm": I've been taking a quick look at what Clojure projects do, and it seems that most entity escaping is actually happening inside the various web-oriented projects ... which may indicate there is no standard solution for this.

If that does end up being the case, then let's put this in `cmr.common.xml/escape-entities` ... since this definitely a general-purpose function we'll need again, sooner or later.

search-app/src/cmr/search/results_handlers/metadata_results_handler.clj
ygliuvt: `escape-special-characters-for-umm-json` is not clear on what the escaping is for. Is it for url escaping or xml escaping? That's why I want the `xml` in there.

search-app/src/cmr/search/results_handlers/metadata_results_handler.clj
ygliuvt: Good point. Moved.

PR #504, opened 2018-04-18
CMR-4874 fixes bug, adds test coverage

daniel-zamora:
Publishing UMM-S in cmr-dev-system-uberjar.jar results in 500 error

### DIFF
--- a/umm-spec-lib/src/cmr/umm_spec/migration/version/service.clj
+++ b/umm-spec-lib/src/cmr/umm_spec/migration/version/service.clj
@@ -35,10 +35,8 @@
   [context s & _]
   (-> s
       (assoc :AccessConstraints (first (:AccessConstraints s)))
-      (update :AccessConstraints #(util/trunc % 1024))
-      (update :UseConstraints #(util/trunc % 1024))
-      (update-in [:ServiceQuality :Lineage] #(util/trunc % 100))
       (assoc :RelatedURLs [(:RelatedURL s)])
+      (assoc :UseConstraints (first (:UseConstraints s)))
       (update :Coverage migrate-coverage-type-up)
       (dissoc :RelatedURL)
       (util/update-in-each [:ServiceOrganizations] dissoc :Uuid)
@@ -47,7 +45,9 @@
 (defmethod interface/migrate-umm-version [:service "1.1" "1.0"]
   [context s & _]
   (-> s
-      (assoc :AccessConstraints [(:AccessConstraints s)])
+      (assoc :AccessConstraints [(util/trunc (:AccessConstraints s) 1024)])
+      (assoc :UseConstraints [(util/trunc (:UseConstraints s) 1024)])
+      (update-in [:ServiceQuality :Lineage] #(util/trunc % 100))
       (assoc :RelatedURL (first (:RelatedURLs s)))
       (update :RelatedURL migrate-related-url-subtype-down)
       (update :Coverage migrate-coverage-type-down)
--- a/umm-spec-lib/src/cmr/umm_spec/migration/version/service.clj
+++ b/umm-spec-lib/src/cmr/umm_spec/migration/version/service.clj
@@ -13,7 +13,7 @@
     related-url))
 
 (defn- migrate-types-down
-  "Migrates CoverageSpatialExtent and CoverageTemporalExtent types"
+  "Migrates CoverageSpatialExtent and CoverageTemporalExtent types from 1.1 to 1.0"
   [coverage-type]
   (if-let [type (get-in coverage-type [:CoverageSpatialExtent :CoverageSpatialExtentTypeType])]
     (-> coverage-type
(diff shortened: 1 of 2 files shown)


umm-spec-lib/src/cmr/umm_spec/migration/version/service.clj
```diff
@@ -12,21 +12,34 @@
     (assoc related-url :SubType nil)
     related-url))
 
+(defn- migrate-types-down
+  "Migrates CoverageSpatialExtent and CoverageTemporalExtent types"
```
mschmele: Nitpick: add version -> version to the docstring

umm-spec-lib/src/cmr/umm_spec/migration/version/service.clj
```diff
@@ -12,21 +12,34 @@
...
-              assoc :CoverageSpatialExtentTypeType (get coverage-type :Type))
-      (dissoc :Type)))
+  (if-let [type (or (get-in coverage-type [:CoverageSpatialExtent :Type])
+                    (get coverage-type :Type))]
+    (-> coverage-type
+        (update :CoverageSpatialExtent
+                assoc :CoverageSpatialExtentTypeType type)
+        (update :CoverageSpatialExtent dissoc :Type)
+        (dissoc :Type))
+    (dissoc coverage-type :Type)))
```
mschmele: Do you need this last `(dissoc coverate-type :Type)`? You're already doing that in the thread above

umm-spec-lib/test/cmr/umm_spec/test/migration/version/service.clj
```diff
@@ -17,12 +17,17 @@
                 :Description "OPeNDAP Service"
                 :Type "GET SERVICE"
                 :URL "https://acdisc.gesdisc.eosdis.nasa.gov/opendap/Aqua_AIRS_Level3/AIRX3STD.006/"},
-   :Coverage {:Type "SPATIAL_POINT"}
-   :AccessConstraints ["TEST"]})
+   :Coverage {:Type "SPATIAL_POINT"
+              :CoverageSpatialExtent {:Type "SPATIAL_POINT"}}
+   :AccessConstraints [(apply str (repeat 1024 "x"))]
+   :UseConstraints [(apply str (repeat 1024 "x"))]
```
oubiwann: 👍

umm-spec-lib/src/cmr/umm_spec/migration/version/service.clj
```diff
@@ -35,10 +48,8 @@
   [context s & _]
   (-> s
       (assoc :AccessConstraints (first (:AccessConstraints s)))
-      (update :AccessConstraints #(util/trunc % 1024))
-      (update :UseConstraints #(util/trunc % 1024))
-      (update-in [:ServiceQuality :Lineage] #(util/trunc % 100))
       (assoc :RelatedURLs [(:RelatedURL s)])
+      (assoc :UseConstraints (first (:UseConstraints s)))
```
ygliuvt: I don't think the UMM-S V1.1 schema is correct in making `AccessConstraints` and `UseConstraints` a single type. It should be an array type. We should talk to Simon or Erich to get it fixed.

umm-spec-lib/src/cmr/umm_spec/migration/version/service.clj
daniel-zamora: if branching

daniel-zamora:
@eereiter need your input so I can close this ticket, its blocking MMT development.

umm-spec-lib/src/cmr/umm_spec/migration/version/service.clj
eereiter: These are single types in UMM-C as well:
    "UseConstraints": {
      "description": "Designed to protect privacy and/or intellectual property by allowing the author to specify how the collection may or may not be used after access is granted. This includes any special restrictions, legal prerequisites, terms and conditions, and/or limitations on using the item. Providers may request acknowledgement of the item from users and claim no responsibility for quality and completeness. Note: Use Constraints describe how the item may be used once access has been granted; and is distinct from Access Constraints, which refers to any constraints in accessing the item.",
      "$ref": "umm-cmn-json-schema.json#/definitions/UseConstraintsType"
    },
    "AccessConstraints": {
      "description": "Allows the author to constrain access to the collection. This includes any special restrictions, legal prerequisites, limitations and/or warnings on obtaining collection data. Some words that may be used in this element's value include: Public, In-house, Limited, None. The value field is used for special ACL rules (Access Control Lists (http://en.wikipedia.org/wiki/Access_control_list)). For example it can be used to hide metadata when it isn't ready for public consumption.",
      "$ref": "umm-cmn-json-schema.json#/definitions/AccessConstraintsType"
    },

It is the same there as it is in UMM-S here, where they are not arrays.  They are plural though because of English grammar not because they should be arrays.

ygliuvt: We should change `AccessConstraints` to `AccessConstraint`, `UseConstraints` to `UseConstraint` in both UMM-C and UMM-S. I am going to approve this PR as we don't want to block EDSC, but we should be consistent on our naming convention in UMM schemas.

PR #505, opened 2018-04-18
CMR-4722

ygliuvt:
Bounding box constraint matching granules outside of bounding box.

### DIFF
--- a/orbits-lib/resources/orbits/geometry_backtracking.rb
+++ b/orbits-lib/resources/orbits/geometry_backtracking.rb
@@ -96,6 +96,11 @@ module Orbits
         return [[lat_range, LongitudeCoverage.full]]
       end
 
+      # Shift north and south off start latitude a bit if there is an exact match,
+      # so that math could work.
+      north = start_lat_rad - EPSILON if north == start_lat_rad
+      south = start_lat_rad + EPSILON if south == start_lat_rad
+
       # If the northern part of the bounding rectangle is above the orbit's starting latitude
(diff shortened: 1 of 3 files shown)


mschmele:
The theory behind this is pretty interesting and a standard practice in GIS calculations, especially where floating-point arithmetic is used (according to what bit of googling I've done). In essence, you're approximating an upper-bound for margin of error in floating-point operations. If two points are less than or equal to epsilon distance from each other, they're considered equal. It's not going to be 0, but it'll be close

If we want to generate machine epsilon to be super fancy and precise: https://en.wikipedia.org/wiki/Machine_epsilon#How_to_determine_machine_epsilon

ygliuvt:
I have fixed two issues:

1/ Coordinate at the equator are not handled correctly in math computation.

2/ Spatial bounding box search with points latitude at exactly the orbit start latitude is not handled correctly.

See attached google earth visualization. You can see the erroneous matches around the equator are eliminated with the fix.
after_fix_matches_at_the_orbit
after_fix_erroneous_matches_around_equator_are_eliminated

chris-durbin:
I'm not entirely sure how the Ruby code is packaged, but I assume we'll need to update the source Ruby repo as well, and potentially add some tests in there.

ygliuvt:
> I'm not entirely sure how the Ruby code is packaged, but I assume we'll need to update the source Ruby repo as well, and potentially add some tests in there.
All the code is in CMR. I don't know if we need to maintain the old Ruby project as well.

PR #507, opened 2018-04-23
CMR-4891: Relocate site resources for docs, etc.

oubiwann:
I've put the base templates in it's own jarfile on Clojars, so we just download them like any other dep. This works for both CMR proper and CMR OPeNDAP.

To test, simply run `lein install-with-content!`, load up a dev env, and then visit the site pages (e.g., http://localhost:3003/) to make sure they rendered properly.

Note that in creating the resources, no files were changed (neither in name nor content).


oubiwann:
Hrm. That's true -- there's no reason we can't add that code to CMR itself as a top-level project: it wouldn't impact anything.

I'll make that change.

oubiwann:
There are currently errors when building with this setup; making changes now and watching builds ...

oubiwann:
Okay, got a green build now.

PR #511, opened 2018-05-10
CMR-4850: Added CORS support for health checks.

sxu123:
health care in all apps are calling the cmr.common-app.api.health to do the job so I'm modifying this central place.

### DIFF
--- a/common-app-lib/src/cmr/common_app/api/health.clj
+++ b/common-app-lib/src/cmr/common_app/api/health.clj
@@ -55,7 +55,6 @@
   takes a request-context as a parameter to determine if the application and its dependencies are
   working as expected."
   [health-fn]
-  (OPTIONS "/" req common-routes/options-response)
   (OPTIONS "/health" req common-routes/options-response)
   (GET "/health" {:keys [request-context]}
     (get-app-health request-context health-fn)))
--- a/common-app-lib/src/cmr/common_app/api/health.clj
+++ b/common-app-lib/src/cmr/common_app/api/health.clj
@@ -2,13 +2,14 @@
   "Defines the health routes for applications."
   (:require
    [cheshire.core :as json]
+   [cmr.common-app.api.routes :as common-routes]   
    [cmr.common.cache :as cache]
    [cmr.common.cache.in-memory-cache :as mem-cache]
    [cmr.common.cache.single-thread-lookup-cache :as stl-cache]
    [cmr.common.config :refer [defconfig]]
    [cmr.common.log :refer [warn]]
    [cmr.common.mime-types :as mt]
-   [compojure.core :refer [GET]]))
+   [compojure.core :refer :all]))
 
 (def health-cache-key
   "The key used to store the health cache in the system cache map."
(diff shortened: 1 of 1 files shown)


common-app-lib/src/cmr/common_app/api/health.clj
```diff
@@ -34,7 +35,8 @@
   (when-not ok?
     (warn "Health check failed" (pr-str dependencies)))
   {:status (if ok? 200 503)
-   :headers {"Content-Type" (mt/with-utf-8 mt/json)}
+   :headers {common-routes/CONTENT_TYPE_HEADER (mt/with-utf-8 mt/json)
+             common-routes/CORS_ORIGIN_HEADER "*"}
```
sxu123: I'm not adding other CORS headers here because I think those are related to OPTIONS case. like CORS_METHODS_HEADER, CORS_MAX_AGE_HEADER.. right?

common-app-lib/src/cmr/common_app/api/health.clj
```diff
@@ -53,5 +55,7 @@
   takes a request-context as a parameter to determine if the application and its dependencies are
   working as expected."
   [health-fn]
+  (OPTIONS "/" req common-routes/options-response)
+  (OPTIONS "/health" req common-routes/options-response)
   (GET "/health" {:keys [request-context]}
```
sxu123: Do I need to add a (GET "/" ...)  case?  it won't be able to get the health check result, but it will not trigger CORS error? 
@oubiwann, I'm not sure what it means to have to take care of both "/" and "/health" cases.

common-app-lib/src/cmr/common_app/api/health.clj
```diff
@@ -53,5 +55,7 @@
   takes a request-context as a parameter to determine if the application and its dependencies are
   working as expected."
   [health-fn]
+  (OPTIONS "/" req common-routes/options-response)
```
ygliuvt: No need to add this route.

common-app-lib/src/cmr/common_app/api/health.clj
ygliuvt: yes.

common-app-lib/src/cmr/common_app/api/health.clj
ygliuvt: oh, I didn't know that. Then we need both.

oubiwann: So, the `/` paths are in each individual app (search, ingest, etc.), not in common-app-lib.

These changes are being made in common-app-lib for expediency and to support re-use ... if we just stick to supporting `/health` right now, we can do this as planned.

If we also support `/`, then we'd have to update the other apps, too. We might want to hold off on that until we get an explicit request for that ...

sxu123: In that case, I'm going to remove OPTIONS "/" in this ticket.  I prefer to separate the support for real health checks (this ticket) and the support for "/" which is treated as health checks but is not real health checks.

PR #512, opened 2018-05-11
CMR-4913 EDSC shows 0 granules for public collection when user isn't logged in

daniel-zamora:
This PR is to address bug in granule search code where entry titles with preceding white space were being properly matched against the corresponding ACLs.  The access control app having trimmed the entry titles before saving the ACL.

### DIFF
--- a/search-app/src/cmr/search/services/acls/granule_acls.clj
+++ b/search-app/src/cmr/search/services/acls/granule_acls.clj
@@ -2,6 +2,7 @@
   "Contains functions for manipulating granule acls"
   (:require
    [clojure.set :as set]
+   [clojure.string :as str]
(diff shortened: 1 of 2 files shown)


oubiwann:
We can't emulate this behaviour in a test? (I'm wondering why there's no test with this change ...)

search-app/src/cmr/search/services/acls/granule_acls.clj
```diff
@@ -2,6 +2,7 @@
   "Contains functions for manipulating granule acls"
   (:require
    [clojure.set :as set]
+   [clojure.string :as str]
```
ygliuvt: use `string` rather than `str`.

PR #515, opened 2018-05-17
CMR-4945: UMM-C default version 1.9 -> 1.10

mschmele:
NOTE: I expect some tests to fail here. I'll get to them once CI finishes

### DIFF
--- a/system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
+++ b/system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
@@ -12,7 +12,7 @@
    [cmr.mock-echo.client.echo-util :as e]
    [cmr.system-int-test.data2.core :as data2-core]
    [cmr.system-int-test.data2.umm-spec-collection :as data-umm-c]
-   [cmr.system-int-test.data2.umm-spec-common :as data-umm-cmn] 
+   [cmr.system-int-test.data2.umm-spec-common :as data-umm-cmn]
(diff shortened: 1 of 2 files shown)


system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
```diff
@@ -492,15 +492,19 @@
           ;; Check that each concept was updated
           (doseq [concept-id concept-ids
                   :let [concept (-> (search/find-concepts-umm-json :collection
-                                                                   {:concept-id concept-id})
+                                                                   {:concept-id (first concept-ids)})
```
sxu123: Why the change?  We want to test each concept-id in concept-ids, now every iteration it's testing the same first concept-id?

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
```diff
@@ -573,7 +577,7 @@
...
```
ygliuvt: This line should not have been changed.

mschmele: You are correct 😄 - I had changed it for debugging and forgot to undo that

sxu123:
The empty :Instruments is added because for version 1.10, it doesn't go through umm migration to have the it removed. Ideally, it probably should be added in the migration code when source=dest. But since MMT is depending on this, we could revisit in the future.

PR #516, opened 2018-05-17
CMR-4902: Read granule size from ISO SMAP.

sxu123:
Notes based on playing with the EDSC interface and the behaviors observed:
1. Ingest a smap granule with transferSize, retrieve it as echo10 and verify that the echo10 granule contains the SizeMBDataGranule with the same value. This is what needs to be fixed for this ticket.
2. Ingest a smap granule with transferSize, retrieve it as iso19115 is not supported. (existing behavior)
No need to fix for this ticket.
3. Ingest an echo10 granule with SizeMBDataGranule, retrieve it as iso19115, it should contain the transferSize (existing behavior). Make sure it continues to work after the fix.
4. Ingest an echo10 granule with SizeMBDataGranule, retrieve it as iso smap is not supported.(existing behavior). No need to fix for this ticket. So converting size in umm-granule to iso smap element is not needed(deleted the added code).
5. We don't support the ingest of iso mends granule. So the description in the ticket should only apply to iso smap granule, not the iso mends granule. Are we going to support it soon?

### DIFF
--- a/system-int-test/test/cmr/system_int_test/search/granule_search_format_test.clj
+++ b/system-int-test/test/cmr/system_int_test/search/granule_search_format_test.clj
@@ -47,6 +47,26 @@
           response (search/find-metadata :granule format-key params options)]
       (d/assert-metadata-results-match format-key [g1-echo] response))))
 
+(deftest search-smap-granule-with-size-in-echo10
+  (let [coll (d/ingest-concept-with-metadata-file "CMR-4902/4902_smap_iso_collection.xml"
+                                                  {:provider-id "PROV1"
+                                                   :concept-type :collection
+                                                   :native-id "iso-smap-collection"
+                                                   :format-key :iso-smap})
+        ;; The iso smap granule contains transferSize being 22.2031965255737
+        granule (d/ingest-concept-with-metadata-file "CMR-4902/4902_smap_iso_granule.xml"
+                                                     {:provider-id "PROV1"
+                                                      :concept-type :granule
+                                                      :native-id "iso-smap-granule"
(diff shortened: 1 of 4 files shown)


ygliuvt:
The granule ISO formats that we supported in ingest and search are so inconsistent. It hurts to even think about it. Hopefully, we can decide on the granule ISO format soon and make it consistent between granule ingest ISO format and granule search ISO format.

system-int-test/test/cmr/system_int_test/search/granule_search_format_test.clj
```diff
@@ -47,6 +47,29 @@
           response (search/find-metadata :granule format-key params options)]
       (d/assert-metadata-results-match format-key [g1-echo] response))))
 
+(deftest search-smap-granule-with-size-in-echo10
```
mschmele: Do we have a test case for when `size` is `nil` or not in a format that we expect?

sxu123: 1. We want to test the case when the iso map granule ingested contains transferSize, so the size is not going to be nil.
    If you really want, we could add a case when the transferSize doesn't exist, but that's like testing the current code, 
    not the code that I modified.
2. echo10 is the only retrieving format that we support right now when the original ingest format is iso smap.

mschmele: Fair enough - I'm just always in favor of being more defensive. Data isn't always perfect, so I want to make sure our code expects that to be the case, if that makes sense

PR #517, opened 2018-05-18
CMR-4939: Fixed bulk index of collections to include service associations.

ygliuvt:


system-int-test/src/cmr/system_int_test/utils/service_util.clj
```diff
@@ -224,3 +227,77 @@
...
+    (is (= [200 coll-atom]
+           [status results]))))
+
+(defn- assert-collection-json-result
+  "Verify the collection JSON response associations related fields have the correct values"
+  [coll expected-fields serv-concept-ids var-concept-ids]
+  (let [coll-with-extra-fields (merge coll
+                                      expected-fields
+                                      {:services serv-concept-ids
+                                       :variables var-concept-ids})
```
sxu123: Just trying to learn sth here...  why is that tags don't need to be here?

ygliuvt: We have issue written to move tags in here as well. We have not done that because tags are being actively used by other systems (e.g EDSC), and we have to coordinate with those systems to make the change.

PR #519, opened 2018-05-22
CMR-4879 adds header to allow old provider delete, for use in tests.

daniel-zamora:


oubiwann:
Maybe I'm misunderstanding this ... but no test?

/me looks again ...

ingest-app/src/cmr/ingest/services/provider_service.clj
```diff
@@ -10,8 +10,10 @@
 (defn verify-empty-provider
   "Throws error if provider still has collections."
   [context provider headers]
-  (when-not (= "true"
-               (get headers "force-full-provider-delete"))
+  (when-not (and (= "true"
+                    (get headers "force-full-provider-delete"))
+                 (not= "prod"
+                       (System/getenv "ENVIRONMENT")))
```
oubiwann: This should probably get a code comment ... maybe something like:

```
;; The following does not use CMR cmr. common.config due to the fact that
;; this is a non-CMR env/config value; this is evident by the lack of the CMR-
;; prefix. 
(System/getenv "ENVIRONMENT")))
```

That will help future devs when they come across this code ...

oubiwann:
So what about some tests?

PR #520, opened 2018-05-23
CMR-4819 part 1

ygliuvt:
Simplified the launchpad token validation in CMR to always require CMR Ingest NAMS workflow with launchpad token in legacy services.


access-control-app/src/cmr/access_control/api/routes.clj
```diff
@@ -303,43 +309,46 @@
 
         ;; Search for ACLs with either GET or POST
         (GET "/"
-             {ctx :request-context params :params headers :headers}
-             (search-for-acls ctx headers params))
+          {ctx :request-context params :params headers :headers}
```
oubiwann: Why are you changing these indents? It seems like this violates our style convention ...  your changes mean that we're no longer aligning the arguments ...

ygliuvt: This is done automatically by ATOM. I think that is how we indent multi-line arguments. Take a look at other lines in routes.

oubiwann: There are two ways of indenting lines in Clojure, and they can't be mixed:

1: 
```
(thing arg1
           arg2)
```

2:
```
(thing
 arg1
  arg2)
```

Note that the second form above is for general functions. Forms like `cond`, `if`, `when`, etc., use the non-aligning two-space indentation, e.g.:
```
(if thing
  a
  b)
```

Atom is doing the indentation of the routes wrong; I fixed these routes a while ago and don't want to have to fix them again; can you change these back?

oubiwann: For more context, see:
* https://stuartsierra.com/2016/clojure-how-to-ns.html
* https://github.com/bbatsov/clojure-style-guide#source-code-layout--organization

oubiwann:
Everything else looks good; also, I'm not asking you to fix the other broken indents; just the ones that Atom broke _this_ time ;-)

access-control-app/src/cmr/access_control/api/routes.clj
ygliuvt: Make sense. Done.

oubiwann:
Thanks for the fix!

PR #524, opened 2018-06-06
CMR-4948: Return correct facet count when outside of facet size range.

sxu123:
Currently, when a selected facet is outside of the facet size range, the count is assigned 0. 
As the facets_size increases, this count could change, we need to fix the inconsistency.

1. For each selected facet, say platform, instrument,   If the returned facet-result includes any 0 counts,
and the facets_size for the field is not 10000, then send the query to ES again, with facets_size increased to 10000 for the field, to insure all facets are returned for that field.

2. Update the original facet-result using the counts in the new facet-result.

### DIFF
--- a/search-app/src/cmr/search/services/query_execution.clj
+++ b/search-app/src/cmr/search/services/query_execution.clj
@@ -140,10 +140,18 @@
   [context query]
   [context (related-item-resolver/resolve-related-item-conditions query context)])
 
+(defn- update-facets
+  "Update the orig-facets-with-count using the info in the all-facets-with-count."
+  [orig-facets-with-count all-facets-with-count]
+  (for [title-val (map :title orig-facets-with-count)]
+    (some #(when (= title-val (:title %)) %) all-facets-with-count)))
+
 (defn- get-facets-with-count
   "Extract out the facet part that contains title and count, amoung other things: 
   [{:title \"t1\" :count 0} {:title \"NonExist\" :count 0} {:title \"t3\" :count 1}]
-  from the facets result."
+  from the facets result.  Note: facets-result is the result from a particular field, 
+  so there is only one child under :facets :children, which is why we use first to 
+  get to the child."
   [facets-result]
   (-> facets-result
       (get-in [:facets :children])
@@ -174,9 +182,7 @@
         orig-facets-with-count (get-facets-with-count facets-for-field)
         all-facets-with-count  (get-facets-with-count all-facets-for-field)
         ;; replace the original facets with the facets in all-facets that have the same :title.
-        updated-orig-facets-with-count
-         (for [title-val (map :title orig-facets-with-count)]
-           (some #(when (= title-val (:title %)) %) all-facets-with-count))
+        updated-orig-facets-with-count (update-facets orig-facets-with-count all-facets-with-count)
         updated-orig-first-facets-children
          (assoc orig-first-facets-children :children updated-orig-facets-with-count)]
     ;;Return the facets-for-field with the first facets children being the updated-orig-first-facets-children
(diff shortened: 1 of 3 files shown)


search-app/src/cmr/search/services/query_execution.clj
```diff
@@ -140,16 +140,69 @@
   [context query]
   [context (related-item-resolver/resolve-related-item-conditions query context)])
 
+(defn- get-facets-with-count
+  "Extract out the facet part that contains title and count, amoung other things: 
+  [{:title \"t1\" :count 0} {:title \"NonExist\" :count 0} {:title \"t3\" :count 1}]
+  from the facets result."
```
mschmele: Maybe add something in the docstring about why you're taking the first child of `facets-result`

search-app/src/cmr/search/services/query_execution.clj
```diff
@@ -140,16 +140,69 @@
...
+  [{:title \"t1\" :count 1} {:title \"NonExist\" :count 0} {:title \"t3\" :count 1} 
+   {:title \"t4\" :count 2} {:title \"t5\" :count 1}]
+  updated-orig-facets-with-count is:
+  [{:title \"t1\" :count 1} {:title \"NonExist\" :count 0} {:title \"t3\" :count 1}]"
+  [facets-for-field all-facets-for-field]
+  (let [orig-first-facets-children (first (get-in facets-for-field [:facets :children]))
+        orig-facets-with-count (get-facets-with-count facets-for-field)
+        all-facets-with-count  (get-facets-with-count all-facets-for-field)
+        ;; replace the original facets with the facets in all-facets that have the same :title.
+        updated-orig-facets-with-count
```
mschmele: This `for` block is complex enough that I think it could be its own function

search-app/src/cmr/search/services/query_execution.clj
```diff
@@ -140,16 +140,75 @@
...
                   (facet-condition-resolver/adjust-facet-query field)
                   (assoc :result-features [:facets-v2])
                   (assoc :facet-fields [field])
-                  (assoc :page-size 0))]
-    (common-qe/execute-query context query)))
+                  (assoc :page-size 0))
+        facets-size-map (:facets-size query)
+        facets-size-for-field (field facets-size-map)
+        facets-for-field (common-qe/execute-query context query)
+        query-with-all-facets-size
```
ygliuvt: Move this line down to the `let` block on line 209.

search-app/src/cmr/search/services/query_execution.clj
```diff
@@ -140,16 +140,75 @@
...
+  from the facets result.  Note: facets-result is the result from a particular field, 
+  so there is only one child under :facets :children, which is why we use first to 
+  get to the child."
+  [facets-result]
+  (-> facets-result
+      (get-in [:facets :children])
+      first
+      :children)) 
+
+(defn- get-facets-for-field-again?
```
ygliuvt: This way of checking of if one needs to get facets for field again is inefficient. Sometimes, it is correct to have count 0 for a facet field, so count 0 does not necessarily mean that the facet count is incorrect. I think the incorrect facet size 0 only happens when the facet field is in the query param does not rank in the list of facet fields returned, so we could make this check more accurate by only doing the additional get all facets in this case. 

We have existing function `create-facets-with-zero-matches` to handle the zero match case. I don't have time to look into all the details on how to do this the best. One laziest way of doing this is to make the `create-facets-with-zero-matches` function return a count of `-1` to make it apparent that we need to do a all facets search to get the count correct, and in the code here, we check for `-1` rather than `0`. 

Hope this makes sense.

ygliuvt: Looks like count 0 is not a valid value for facet field. So 0 is as good as -1. The current approach is OK.

search-app/src/cmr/search/services/query_execution.clj
sxu123: Done.

PR #526, opened 2018-06-14
CMR-4706: Fixed the issue for iso/iso19115 when include_granule_counts=true

sxu123:
This PR contains the fix for 4 parts:

Part1:  Fix the 500 error related to include_granule_counts  for iso/iso19115 cases (supported now)
Part2: Fix the 500 error related to include_granule_counts for umm_json case (supported now)
Part3: Fix the 500 error related to include_granule_counts for kml case (Throw bad request error because there is no place to put granule_count in the kms result).
Part4: Fix the 500 error related to include_granule_counts for opendata case (Throw bad request error. This has been documented as not supported so the ticket didn't complain about this one. But we should throw bad request error.)


sxu123:
Totally different, that's why I separated this part out. If there's anything that could be combined, it will be addressed later when I get to the part because I don't know how to fix them yet... :)

system-int-test/src/cmr/system_int_test/data2/umm_json.clj
```diff
@@ -80,6 +80,7 @@
   [meta]
   (util/remove-nil-keys
    (-> meta
+       (dissoc :granule_count)
```
sxu123: There are many existing tests that are making use of this without granule_count.  Since we are testing the granule_count in granule_counts_search_test.clj, we don't need to test them here.

system-int-test/src/cmr/system_int_test/data2/granule_counts.clj
```diff
@@ -77,7 +77,43 @@
   (into {} (for [{:keys [concept-id has-granules]} (:items results)]
              [concept-id has-granules])))
 
+(defmethod results->actual-has-granules :iso19115
+  [result-format results]
+  (into {} (for [{:keys [concept-id has-granules]} (:items results)]
+             [concept-id has-granules])))
+
```
sxu123: Didn't add defmethod for :umm_json case because has-granules doesn't exist in the umm_json result and it is not in the scope of this ticket - which is about granule_count.

search-app/src/cmr/search/results_handlers/umm_json_results_handler.clj
```diff
@@ -15,6 +16,24 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; Collection UMM JSON
 
+(defn- get-granule-count-for-item
+  "Get the granule count from granule-counts-map using teh concept-id in the item."
```
chris-durbin: teh typo

system-int-test/test/cmr/system_int_test/search/granule_counts_search_test.clj
```diff
@@ -349,8 +349,32 @@
...
+          :iso19115 (search/find-metadata :collection :iso19115 {:include-has-granules true :include-granule-counts true})
+
+          "granule count in umm_json format"
+          :umm_json (search/find-concepts-umm-json :collection {:include-has-granules true :include-granule-counts true})
+
+          "granule count in atom format"
+          :atom (search/find-concepts-atom :collection {:include-has-granules true :include-granule-counts true})
+
+          "granule count in json format"
+          :atom (search/find-concepts-json :collection {:include-has-granules true :include-granule-counts true}))))))
```
chris-durbin: Can you add a test that you get the appropriate error message for both opendata and kml requests?

sxu123: I was just working on adding the kml one. Didn't think of opendata, will add that too. Thanks.

search-app/src/cmr/search/results_handlers/umm_json_results_handler.clj
sxu123: Corrected.

system-int-test/test/cmr/system_int_test/search/granule_counts_search_test.clj
sxu123: opendata format does throw 500 error, even though the ticket didn't complain about this since it's not supported. Added the proper handling and the testing for it.

PR #529, opened 2018-06-26
CMR-2053: Supported granule-count in Legacy UMM JSON.

sxu123:


sxu123:
@oubiwann, Thanks for reviewing. There is no need for doc update because it is supposed to be supported. Only opendata and kml are not.

PR #530, opened 2018-06-28
CMR-4813 

daniel-zamora:
This PR adds fuzzy end-date sorting for collection search

I think I might need more test coverage, maybe testing popularity, relevance, and near-current together per the use case in the ticket.  Also, I'm not sure what to configure the near-current range to be, its currently 30 days.  Another thing is sort order for this and how to handle it.  I haven't figured out how to use :desc instead of :asc by default for near-current.

### DIFF
--- a/indexer-app/src/cmr/indexer/data/concepts/collection.clj
+++ b/indexer-app/src/cmr/indexer/data/concepts/collection.clj
@@ -59,6 +59,17 @@
   [m f]
   (reduce-kv (fn [m k v] (assoc m k (f v))) {} m))
 
+(defn- determine-ongoing-date
+  "Determines ongoing date using collection's end-date."
+  [end-date]
+  (if end-date
+    (if (t/after?
+         end-date
+         (t/minus (tk/now) (t/days (indexer-config/ongoing-days))))
+      (index-util/date->elastic (t/date-time 3000 1 1 0 0 0 0))
+      (index-util/date->elastic end-date))
+    (index-util/date->elastic (t/date-time 3000 1 1 0 0 0 0))))
+
 (defn- collection-temporal-elastic
   "Returns a map of collection temporal fields for indexing in Elasticsearch."
   [context concept-id collection]
@@ -73,15 +84,10 @@
                            nil
                            granule-end-date)
         coll-start (index-util/date->elastic start-date)
-        coll-end (index-util/date->elastic end-date)
-        ongoing (if end-date
-                       (t/after?
-                        end-date
-                        (t/minus (tk/now) (t/days (indexer-config/ongoing-days))))
-                       true)]
+        coll-end (index-util/date->elastic end-date)]
     (merge {:start-date coll-start
             :end-date coll-end
-            :ongoing ongoing}
+            :ongoing (determine-ongoing-date end-date)}
            (or (when granule-start-date
                  {:granule-start-date (index-util/date->elastic granule-start-date)
                   :granule-end-date (index-util/date->elastic granule-end-date)})
--- a/indexer-app/src/cmr/indexer/data/concepts/collection.clj
+++ b/indexer-app/src/cmr/indexer/data/concepts/collection.clj
@@ -75,10 +75,9 @@
         coll-start (index-util/date->elastic start-date)
         coll-end (index-util/date->elastic end-date)
         near-current (if end-date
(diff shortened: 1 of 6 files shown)


system-int-test/test/cmr/system_int_test/search/collection_sorting_search_test.clj
```diff
@@ -379,3 +384,55 @@
...
+               "PROV1"
+               (data-umm-c/collection 6 {:EntryTitle "Dataset6"
+                                         :TemporalExtents [(data-umm-cmn/temporal-extent
+                                                            {:beginning-date-time "2010-01-01T12:00:00Z"})]}))]
+    (index/wait-until-indexed)
+    (are [sort-key items]
+         (sort-order-correct? items sort-key)
+
+         ["-near-current" "entry-title"] [coll2 coll4 coll5 coll6 coll1 coll3]
+         ["near-current" "entry-title"] [coll1 coll3 coll2 coll4 coll5 coll6])))
```
sxu123: I don't understand the result...   So, coll1, coll2 and coll3, they should all have the same near-current value, which is false right? So, sorting of these three should only be based on entry-title -  coll1, coll2, coll3?
I think you might want to add tests that's only sorting on the near-current first to make sure it's working as expected before mixing it with other sorting parameters.   If I have c1: near-current=true, c2: near-current=true, c3: near-current-false, c4: near-current=false, what order do I expect to see when I search for them with sort key = near-current?

daniel-zamora: One question per comment is easier to respond to.
1. No, coll2 is true.  coll1 and coll3 are false.
2. The current sorting is right.
3. If entry-title is consistent: sort-key= near-current, then the order is c3, c4, c1, c2 because the default sort order for near-current is :asc so false goes before true.  I haven't figured out how to easily reverse it.

sxu123: Thanks. After Chris approves your logic, please add the doc update for this new sorting parameter.

indexer-app/src/cmr/indexer/data/concepts/collection.clj
```diff
@@ -74,7 +86,12 @@
         coll-start (index-util/date->elastic start-date)
         coll-end (index-util/date->elastic end-date)]
     (merge {:start-date coll-start
-            :end-date coll-end}
+            :end-date coll-end
+            ;; ongoing is a date-time field in elastic, and if a collection is
```
chris-durbin: Can you move this comment into the docstring for `determine-ongoing-date`.

search-app/docs/api.md
```diff
@@ -1654,7 +1654,8 @@ One or more sort keys can be specified using the `sort_key[]` parameter. The ord
   * `has_granules` - Sorts collections by whether they have granules or not. Collections with granules are sorted before collections without granules.
   * `has_granules_or_cwic` - Sorts collections by whether they have granules or they are tagged as a CWIC collection. Collections with granules or are CWIC tagged are sorted before collections without granules or a CWIC tag.
   * `usage_score` - Sorts collection by usage. The usage score comes from the EMS metrics, which are ingested into the CMR.
-
+  * `ongoing` - Sorts collection by fuzzy collection end-date in relation to ongoing-days configured. Any end-date after today, minus the configured ongoing-days (30 by default), is considered ongoing.  Any end-date before that is not near current.
```
chris-durbin: Change "not near current." to "not ongoing."

system-int-test/test/cmr/system_int_test/search/collection_sorting_search_test.clj
```diff
@@ -379,3 +384,55 @@
          "sensor" [c1 c2 c3 c4 c5]
          ;; Descending sorts by the max value of a multi value fields
          "-sensor" [c2 c5 c1 c4 c3])))
+
+(deftest collection-fuzzy-end-date-test
+  (let [today (f/unparse (f/formatters :date-time) (t/now))
+        outside-ongoing (f/unparse (f/formatters :date-time)
```
chris-durbin: Might be useful to add one more date really far in the future like the year 4000 (beyond 3000) where the second sort will put it at the end to show that 3000 is used for its ongoing date.

PR #532, opened 2018-06-29
CMR-5008: Fixed the token issue and added some tests.

sxu123:
This is the addition to the 5008 fix that was merged.

### DIFF
--- a/system-int-test/test/cmr/system_int_test/bootstrap/bulk_index/concepts_test.clj
+++ b/system-int-test/test/cmr/system_int_test/bootstrap/bulk_index/concepts_test.clj
@@ -45,7 +45,7 @@
     ;; Create and return token
     (e/login (s/context) "admin-read-update" [admin-read-update-group-concept-id])))
 
-(deftest index-system-concepts-test-valid
+(deftest index-system-concepts-test-with-update-token
   (s/only-with-real-database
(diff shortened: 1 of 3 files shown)


system-int-test/test/cmr/system_int_test/bootstrap/bulk_index/concepts_test.clj
```diff
@@ -37,6 +37,30 @@
...
+    ;; Create and return token
+    (e/login (s/context) "admin-read-update" [admin-read-update-group-concept-id])))
+
+(deftest index-system-concepts-test-with-update-token
+  (s/only-with-real-database
+   ;; Disable message publishing so items are not indexed as part of the initial save.
+   (core/disable-automatic-indexing)
+
+   ;; Remove fixture ACLs
+   (let [response (ac/search-for-acls (u/conn-context) {} {:token (tc/echo-system-token)})
```
ccuadrado: Does this test work if you don't use the echo-system-token? It should work with any token that has the System Ingest Permission.

sxu123: I tried using the token it created, it worked too. Also, looks like I don't even need to remove the fixture ACLs for this test to work.  I just copied it from the other tests.

PR #534, opened 2018-07-05
CMR-5000 CMR/ECHO does not check for token expiration during ingest

daniel-zamora:


acl-lib/src/cmr/acl/core.clj
```diff
@@ -159,21 +159,31 @@
            error-message))
         nil))))
 
+(defn- valid-token?
+  "test"
```
chris-durbin: Update docstring.

acl-lib/src/cmr/acl/core.clj
```diff
@@ -159,21 +159,31 @@
...
-      (let [acl-oit-key (echo-acls/acl-type->acl-key object-identity-type)]
-        (->> (get-permitting-acls context
-                                  object-identity-type
-                                  "INGEST_MANAGEMENT_ACL"
-                                  permission-type)
-             ;; Find acls for this provider
-             (filter #(or (nil? provider-id)
-                          (= provider-id (get-in % [acl-oit-key :provider-id]))))
-             seq))))
+  (when (valid-token? context)
```
chris-durbin: We're losing the performance optimization for the system token. I think it makes sense to always allow the system token to bypass this check.

Edit: I see you are bypassing it in the transmit library.

system-int-test/test/cmr/system_int_test/search/humanizer/community_usage_metrics_test.clj
```diff
@@ -61,7 +61,7 @@
 
   (testing "Create with unknown token"
     (is (= {:status 401
-            :errors ["Token [ABC] does not exist"]}
+            :errors ["You do not have permission to perform that action."]}
```
chris-durbin: The old error was more informative. Any chance we could make the error message `"Token [ABC] does not exist or has expired"`?

acl-lib/src/cmr/acl/core.clj
```diff
@@ -159,21 +159,31 @@
            error-message))
         nil))))
 
+(defn- valid-token?
+  "test"
+  [context]
+  (if-let [token (:token context)]
+    (try
+      (= 200 (first (echo-tokens/get-token-info context token)))
+      (catch Exception e false))
```
chris-durbin: If the return code is not 200 it probably makes sense for us to log the error message that comes back.

chris-durbin: Also if an exception is thrown we should log the exception.

acl-lib/src/cmr/acl/core.clj
```diff
@@ -159,21 +159,31 @@
...
+    true))
+
 (defn has-ingest-management-permission?
```
chris-durbin: Shouldn't we always validate that the token has not expired (not just for ingest permissions)? I assume we have to be calling legacy-services to validate a token, and I don't quite understand how it could say it's valid when it is expired. Seems like the fix should be in legacy-services to return an error on token validation.

ygliuvt:
Thanks for picking up this issue while I am out. Do you have code or test changes to verify if a token has expired?

acl-lib/src/cmr/acl/core.clj
oubiwann: Perhaps a note to that effect here would not go amiss ...

oubiwann:
I'll wait for additional changes before doing a complete review.

acl-lib/src/cmr/acl/core.clj
daniel-zamora: I couldn't find where we asked legacy-services about the token when tracing through the ingest-management check.

daniel-zamora: It looks like get-user-id is called in access control for current-sids, you might be right that we check this indirectly there.

system-int-test/test/cmr/system_int_test/search/humanizer/community_usage_metrics_test.clj
daniel-zamora: I'm thinking this is because there is some indirect token validation happening somewhere. Looking.

daniel-zamora: so it looks like its hidden in context-augmenter/add-user-id-and-sids-handler which is a part of the routes handler.  I missed this the first time around.

PR #536, opened 2018-07-10
CMR-5000

ygliuvt:
Fixed CMR does not check for token expiration during ingest.

### DIFF
--- a/mock-echo-app/src/cmr/mock_echo/data/token_db.clj
+++ b/mock-echo-app/src/cmr/mock_echo/data/token_db.clj
@@ -1,5 +1,8 @@
 (ns cmr.mock-echo.data.token-db
   (:require
+   [clj-time.core :as t]
+   [cmr.common-app.services.search.datetime-helper :as datetime-helper]
+   [cmr.common.time-keeper :as tk]
    [cmr.transmit.config :as transmit-config]))
 
 (defn initial-db-state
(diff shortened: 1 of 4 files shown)


mock-echo-app/src/cmr/mock_echo/data/token_db.clj
```diff
@@ -35,7 +38,11 @@
   :id of the token that was created"
   [context token-info]
   (let [token-db (context->token-db context)
-        token (assoc token-info :id (new-token-id token-db))]
+        token (assoc token-info :id (new-token-id token-db))
+        token (if (:expires token)
+                token
+                ;; token expires in 30 days
+                (assoc token :expires (datetime-helper/datetime->string (t/plus (tk/now) (t/days 30)))))]
```
daniel-zamora: We should have this be a defconfig

transmit-lib/src/cmr/transmit/echo/tokens.clj
```diff
@@ -47,7 +50,14 @@
                                                        "Echo-Token" (transmit-config/echo-system-token)}
                                              :form-params {:id token}})]
       (case (int status)
-        200 (get-in parsed [:token_info :user_name])
+        200 (let [expires (some-> (get-in parsed [:token_info :expires])
```
daniel-zamora: Could this be an if-let?

daniel-zamora: maybe not, disregard.

mock-echo-app/src/cmr/mock_echo/data/token_db.clj
ygliuvt: This code is only for testing. We can make this a `def`, but probably not a `defconfig`.

PR #537, opened 2018-07-17
CMR-4885: Added alternative path for ISOMends instruments.

sxu123:
1. NOAA instruments are from an alternative path;
These instruments only contain ShortName and LongName, which are combined together from short-name-xpath.
2. These instruments are not associated with any platforms. We need to add a default platform to contain these instruments.
3. It's not possible for one ISOMEND xml file to contain both instruments that are associated with platforms and instruments that are not associated with platforms.


umm-spec-lib/src/cmr/umm_spec/xml_to_umm_mappings/iso_shared/instrument.clj
```diff
@@ -82,42 +87,58 @@
...
-        :ComposedOf child-instruments
-        :OperationalModes (char-and-opsmode/parse-operationalmodes instrument-elem)}
-       {:ShortName (value-of instrument-elem iso/short-name-xpath)
-        :LongName (value-of instrument-elem iso/long-name-xpath)
-        :Characteristics (char-and-opsmode/parse-characteristics instrument-elem)
-        :Technique (without-default-value-of instrument-elem "gmi:type/gco:CharacterString")
-        :OperationalModes (char-and-opsmode/parse-operationalmodes instrument-elem)}))))
+  [doc base-xpath alt-xpath? instrument-elem]
+  (if alt-xpath?
+    ;; This is the NOAA case, where both the shortname and longname are combined
```
mschmele: Might be worth adding "See CMR-4885 for more details" to the end of this comment block

umm-spec-lib/src/cmr/umm_spec/xml_to_umm_mappings/iso_shared/instrument.clj
```diff
@@ -12,6 +13,10 @@
   (str "/gmi:MI_Metadata/gmi:acquisitionInformation/gmi:MI_AcquisitionInformation/gmi:instrument"
     "/eos:EOS_Instrument"))
 
+(def instrument-alternative-xpath
+  (str "/gmi:MI_Metadata/gmi:acquisitionInformation/gmi:MI_AcquisitionInformation/gmi:instrument"
+    "/gmi:MI_Instrument"))
```
chris-durbin: Quotes should line up directly below the quote above.

umm-spec-lib/src/cmr/umm_spec/xml_to_umm_mappings/iso_shared/instrument.clj
```diff
@@ -82,42 +87,58 @@
...
-        :Technique (without-default-value-of instrument-elem "gmi:type/gco:CharacterString")
-        :NumberOfInstruments (count child-instruments)
-        :ComposedOf child-instruments
-        :OperationalModes (char-and-opsmode/parse-operationalmodes instrument-elem)}
-       {:ShortName (value-of instrument-elem iso/short-name-xpath)
-        :LongName (value-of instrument-elem iso/long-name-xpath)
-        :Characteristics (char-and-opsmode/parse-characteristics instrument-elem)
-        :Technique (without-default-value-of instrument-elem "gmi:type/gco:CharacterString")
-        :OperationalModes (char-and-opsmode/parse-operationalmodes instrument-elem)}))))
+  [doc base-xpath alt-xpath? instrument-elem]
```
oubiwann: This is not a request for a change, but something to consider:

For refactorings like this where a new parameter is added, it is good practice to do something like the following ...

1. Keep the original arity with the same parameters (this is an easy path for backwards compatibility and helps reduce the number of surprises due to the change to nearly zero).
1. Add a new a new arity N+1 block to the function, where the new parameter is an "options" map containing the key for the new data you want to pass (you can do this in the args with destructuring). Note that this is an excellent means of future-proofing your code.
1. Add the new implementation that a) has the exact same behaviour as the original function when the options map is nil (or the expected options key isn't present, etc.), and b) has the new behaviour when the expected key _is_ present, then
1. Delete the original arity code and in that back simple call the arity N+1 function with the options map being `{}`.

Another nice benefit of this approach is that it makes it super-easy for reviewers to see what has changed, and clearly identify the backwards compatibility plan :-)

oubiwann: Btw, step 4 as outlined above is for when the logic is basically identical, modulo small variance due to the options. If the logic is completely different, you just keep the original in the first arity block, and the new logic in the secon arity block.

umm-spec-lib/src/cmr/umm_spec/xml_to_umm_mappings/iso_shared/instrument.clj
```diff
@@ -12,6 +13,10 @@
   (str "/gmi:MI_Metadata/gmi:acquisitionInformation/gmi:MI_AcquisitionInformation/gmi:instrument"
     "/eos:EOS_Instrument"))
 
+(def instrument-alternative-xpath
```
chris-durbin: Add a docstring.

oubiwann: I was wondering about that, but the other, similar `def`'s in the file didn't have them ...

umm-spec-lib/src/cmr/umm_spec/xml_to_umm_mappings/iso_shared/instrument.clj
sxu123: Done.

umm-spec-lib/src/cmr/umm_spec/xml_to_umm_mappings/iso_shared/instrument.clj
sxu123: Done.

umm-spec-lib/src/cmr/umm_spec/xml_to_umm_mappings/iso_shared/instrument.clj
sxu123: Done.

umm-spec-lib/src/cmr/umm_spec/xml_to_umm_mappings/iso_shared/instrument.clj
sxu123: Done.

sxu123:
@chris-durbin, added tests.

umm-spec-lib/test/cmr/umm_spec/test/noaa_instrument_xpath_test.clj
```diff
@@ -0,0 +1,39 @@
...
+  (let [metadata (slurp (noaa-example-file))
+        umm (js/parse-umm-c (core/parse-metadata test-context :collection :iso19115 metadata))]
+    (is (= expected-noaa-platforms (:Platforms umm)))))
+         
+(deftest test-mixed-example-file 
+  "Verify that the returned platforms doesn't include the not-provided-platform."
+  (let [metadata (slurp (mixed-example-file))
+        umm (js/parse-umm-c (core/parse-metadata test-context :collection :iso19115 metadata))]
+    (is (< 0 (count (:Platforms umm))))
+    (is (= nil (some #(= (first expected-noaa-platforms) %) (:Platforms umm))))))
```
chris-durbin: I'd change these two assertions to just verify you get back exactly what was in the metadata. Verifying the exact content will catch more potential errors.

sxu123: Chris agreed that it's not worth the effort to change it because I have to create many multiple levels of records for the expected. The existing tests should cover the verification of regular instrument parsing. This one is just to show that when both NOAA and regular instrument xpaths exist(which in reality is not possible), the regular xpath takes precedence.

PR #538, opened 2018-07-17
CMR-5031 Temporal extents not getting picked up for 2 SMAP datasets

daniel-zamora:
This PR is for changing the granule iso smap -> umm mapping such that when multiple extents are provided, it will map the first temporal it finds, instead of the only the first extent.  

Before, if there was no temporal extent in the first element of extents, the temporal mapping would be nil even if the subsequent extents included a temporal extent.

### DIFF
--- a/umm-lib/resources/data/iso_smap/granule_with_multiple_extents.xml
+++ b/umm-lib/resources/data/iso_smap/granule_with_multiple_extents.xml
@@ -318,6 +318,10 @@
                       </gmd:polygon>
                     </gmd:EX_BoundingPolygon>
                   </gmd:geographicElement>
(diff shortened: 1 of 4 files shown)


oubiwann:
Looks like something was missing from your comment  c405b33  ...? I don't see test coverage ...

daniel-zamora:
I changed the test data to cover this scenario

oubiwann:
Got a private note from you -- thanks! I don't think I know how we do testing here; I'm taking a deeper look now ...

daniel-zamora:
Going to add a new test, with a new file covering this scenario. Per Duncan's comment.

PR #539, opened 2018-07-17
CMR-4971 Adds validation on acl group-permissions group-id

daniel-zamora:
Group-ids that are invalid can cause internal errors, so we want to prevent any non CMR concept-id from populating that field in :group_permissions

### DIFF
--- a/access-control-app/int-test/cmr/access_control/int_test/acl_crud_test.clj
+++ b/access-control-app/int-test/cmr/access_control/int_test/acl_crud_test.clj
@@ -321,8 +321,8 @@
     (are3 [re acl]
           (is (thrown-with-msg? Exception re (ac/create-acl (u/conn-context) acl {:token token})))
 
-          ; Acceptance criteria: I receive an error if creating an ACL missing required fields.
-          ; Note: this tests a few fields, and is not exhaustive. The JSON schema handles this check.
+          ;; Acceptance criteria: I receive an error if creating an ACL missing required fields.
+          ;; Note: this tests a few fields, and is not exhaustive. The JSON schema handles this check.
           "Nil field value"
           #"object has missing required properties"
           (dissoc system-acl :group_permissions)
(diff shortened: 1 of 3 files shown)


access-control-app/int-test/cmr/access_control/int_test/acl_crud_test.clj
```diff
@@ -321,8 +321,8 @@
     (are3 [re acl]
           (is (thrown-with-msg? Exception re (ac/create-acl (u/conn-context) acl {:token token})))
 
-          ;; Acceptance criteria: I receive an error if creating an ACL missing required fields.
-          ;; Note: this tests a few fields, and is not exhaustive. The JSON schema handles this check.
+          ; Acceptance criteria: I receive an error if creating an ACL missing required fields.
```
chris-durbin: Not sure if this change was intentional - I think we're still standardizing on two semicolons, though don't remember the rationale anymore.

oubiwann: Should be `;;` -- part of the Clojure style guide (inherited from CL).
* 4 at top of namespace (almost never used in clj due to namespace docstrings)
* 3 for module-level code comments
* 2 for function-level code comments
* 1 for end-of-line comments

oubiwann: Note that for uses of Emacs with Lisp plugins, comment lines that start with a single semi-colon can cause issues. I think I remember @rvirding running into this in LFE ...

access-control-app/src/cmr/access_control/services/acl_validation.clj
```diff
@@ -86,6 +86,18 @@
     (:system-identity acl)          :system-identity
     (:catalog-item-identity acl)    :catalog-item-identity))
 
+(defn- group-id-validation
```
oubiwann: Thanks for splitting these functions out into named utility functions -- much cleaner!

access-control-app/int-test/cmr/access_control/int_test/acl_crud_test.clj
daniel-zamora: I used atom to comment out a bunch of stuff, then uncommented it.  I think thats why this happened.

PR #540, opened 2018-07-19
Task/cmr 4997/umm var 1.2

oubiwann:
It's been a while since I've updated a concept to a new version; not sure if I missed anything or not ...


system-int-test/test/cmr/system_int_test/ingest/granule_ingest_test.clj
```diff
@@ -414,7 +414,7 @@
         ["Collection with Entry Title [correct], Short Name [S2], Version Id [V1] referenced in granule [Gran1] provider [PROV1] does not exist."]))))
 
 (deftest ingest-granule-with-parent-umm-collection-test
-  (let [cddis-umm (-> "example_data/umm-json/1.2/CDDIS.json" io/resource slurp)
+  (let [cddis-umm (-> "example-data/umm-json/collection/v1.2/CDDIS.json" io/resource slurp)
```
ygliuvt: CMR naming convention has directory names using underscores, not dashes. Not sure why you change it.

oubiwann: Directories for Clojure namespaces are required to have underscores, but this is not source code. The dash is actually more prevalent for non-source code CMR dirs than underscores, especially for higher-level directories.

oubiwann: All dirs:
```
$ find . -type d|sort -u|egrep -v 'src|test'|grep '-'|wc -l
    1741
$ find . -type d|sort -u|egrep -v 'src|test'|grep '_'|wc -l
     596
```

Higher-level dirs:

```
$ find . -type d -maxdepth 3|sort -u|egrep -v 'src|test'|grep '-'|wc -l
     202
$ find . -type d -maxdepth 3|sort -u|egrep -v 'src|test'|grep '_'|wc -l
       9
```

umm-spec-lib/src/cmr/umm_spec/migration/version/variable.clj
```diff
@@ -3,13 +3,60 @@
   (:require
    [cmr.umm-spec.migration.version.interface :as interface]))
 
+(defn sample-11->12
```
mschmele: Why `11` and `12` rather than `1.1` and `1.2`? Maybe add a docstring or comment to explain the reasoning.

oubiwann: Clojure doesn't let you put `.` in function names; that's parsed as a namespace/Class separator. Attempting to do so will result in `java.lang.ClassNotFoundException`.

system-int-test/test/cmr/system_int_test/ingest/granule_ingest_test.clj
chris-durbin: Since we aren't consistent I'd like to change to use spaces. 😛

umm-spec-lib/src/cmr/umm_spec/migration/version/variable.clj
chris-durbin: I find `11` and `12` confusing as well. 

I _think_ I prefer something like `sample-1-1->1-2`. What do you think about that format?
Or how about `sample-1dot1->1dot2`? Alright that last one is a joke.

umm-spec-lib/src/cmr/umm_spec/migration/version/variable.clj
```diff
@@ -3,13 +3,60 @@
...
+  [dims]
+  (mapv #(dissoc % :Type) dims))
+
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Variable Migration Implementations
 
 (defmethod interface/migrate-umm-version [:variable "1.0" "1.1"]
+  ;; Migrate up to 1.1
   [context v & _]
-  (dissoc v :Services))
```
chris-durbin: It looks like this change to an old migration was intentional (and something must have been broken before). Do we have a test for the 1.0->1.1 and 1.1->1.0 migrations now?

umm-spec-lib/resources/example-data/umm-json/variable/v1.2/V1200267323-HMR_TME_v_1_2.json
```diff
@@ -0,0 +1,69 @@
+{
```
chris-durbin: There's a lot of inconsistent indentation in this example file.

umm-spec-lib/src/cmr/umm_spec/migration/version/variable.clj
chris-durbin: I found a test - ignore.

umm-spec-lib/resources/example-data/umm-json/variable/v1.2/V1200267323-HMR_TME_v_1_2.json
oubiwann: Yeah ... I just pasted 'em as I got 'em ... they're small; I can fix them up.

oubiwann:
> No changes to any indexing code for supporting search of the new parameters?

I'm actually looking into that now ...

system-int-test/test/cmr/system_int_test/ingest/granule_ingest_test.clj
ygliuvt: Man, I guess nobody paid any attention. It stands out when you actually renamed it to something else. I don't care then. :)

umm-spec-lib/src/cmr/umm_spec/migration/version/variable.clj
oubiwann: Done.

umm-spec-lib/resources/example-data/umm-json/variable/v1.2/V1200267323-HMR_TME_v_1_2.json
oubiwann: Done.

PR #544, opened 2018-07-26
CMR-5087

daniel-zamora:
This PR fixes bug in acl-fetcher where the search call to access-control is limited to 2000 results per page.  The function will now properly page through all the pages, instead of truncating results to only 2000.


acl-lib/src/cmr/acl/acl_fetcher.clj
```diff
@@ -73,18 +73,38 @@
...
   [context object-identity-types]
-  (map (comp :acl util/map-keys->kebab-case)
-       (get
-        (access-control/search-for-acls (assoc context :token (config/echo-system-token))
-                                        {:identity-type (object-identity-types->identity-strings
-                                                         object-identity-types)
-                                         :include-full-acl true
-                                         :page-size 2000})
-        :items)))
+  (let [page-size 2000
```
chris-durbin: We should add a paging test (we can do this in a separate PR). For that test you might need to make the page-size configurable and change the value for the test.

acl-lib/src/cmr/acl/acl_fetcher.clj
```diff
@@ -73,18 +73,38 @@
...
+        response (access-control/search-for-acls (assoc context :token (config/echo-system-token))
```
mschmele: It might be worth either pulling the call to `search-for-acls` into a function to clean up this one, or creating a `get-all-acls` function in the `access-control` namespace. It's nitpicky, but I feel like it'll be easier to follow

PR #549, opened 2018-08-06
Added CMR-Scroll-Id to allowed and exposed headers. Made changes to c…

CayvonH:
…ors_headers_test to test CMR-Scroll-Id header.

### DIFF
--- a/dev-system/resources/cors_headers_test.html
+++ b/dev-system/resources/cors_headers_test.html
@@ -7,23 +7,30 @@
  </head>
  <body>
    <h1 id="resp"></h1>
+   <input type="text" id="scroll-id" value=""/>
+   <button onclick="submit()" id="scroll">Submit Scroll-Id</button>
    <script>
-     var req = new XMLHttpRequest();
-     var url = 'http://localhost:3003/collections.json';
-     req.open('GET', url, true);
-     req.setRequestHeader('Client-Id', 'Sprint44Demo');
-     req.setRequestHeader('CMR-Request-Id', 'sprint44-demo');
-     req.onload = function() {
-      var hits = req.getResponseHeader("CMR-Hits");
-      var requestId = req.getResponseHeader("CMR-Request-Id");
-      var response = "CMR-Hits: " + hits + "\nCMR-Request-Id: " + requestId + "\nResult: " + req.responseText;
-       document.getElementById('resp').innerText = response;
-     };
-     req.onerror = function() {
-       document.getElementById('resp').innerText = "Fail!";
-     };
-     req.send();
+     function submit() {
+       var req = new XMLHttpRequest();
+       var url = 'http://localhost:3003/collections.json?scroll=true';
+       var scrollHeader = document.getElementById("scroll-id").value;
+       req.open('GET', url, true);
+       req.setRequestHeader('Client-Id', 'Sprint44Demo');
+       req.setRequestHeader('CMR-Request-Id', 'sprint44-demo');
+       if (scrollHeader) req.setRequestHeader('CMR-Scroll-Id', scrollHeader);
+       req.onload = function() {
+         var hits = req.getResponseHeader("CMR-Hits");
+         var requestId = req.getResponseHeader("CMR-Request-Id");
+         var scrollId = req.getResponseHeader("CMR-Scroll-Id");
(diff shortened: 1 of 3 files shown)


dev-system/resources/cors_headers_test.html
```diff
@@ -7,23 +7,30 @@
...
+       var scrollHeader = document.getElementById("scroll-id").value;
+       req.open('GET', url, true);
+       req.setRequestHeader('Client-Id', 'Sprint44Demo');
+       req.setRequestHeader('CMR-Request-Id', 'sprint44-demo');
+       if (scrollHeader) req.setRequestHeader('CMR-Scroll-Id', scrollHeader);
+       req.onload = function() {
+         var hits = req.getResponseHeader("CMR-Hits");
+         var requestId = req.getResponseHeader("CMR-Request-Id");
+         var scrollId = req.getResponseHeader("CMR-Scroll-Id");
+         var response = "CMR-Hits: " + hits + "\nCMR-Request-Id: " + requestId + "\nCMR-Scroll-Id: " + scrollId + "\nResult: " + req.responseText;
```
chris-durbin: It's not obvious to me how we're verifying that we can accept the CMR-Scroll-Id header and return the header as part of a CORS request.

Is this somehow incorporated in an automated test?

system-int-test/test/cmr/system_int_test/header_test.clj
```diff
@@ -67,3 +69,41 @@
...
+  [request-response header-to-compare expected-response]
+  (let [received-headers (get-headers-as-set
+                          request-response
+                          header-to-compare)
+        expected-headers (get-headers-as-set
+                          expected-response
+                          header-to-compare)]
+    (= expected-headers received-headers)))
+
+(deftest cors-headers
```
ygliuvt: It would be clearer if we just compare the response header value of CORS_CUSTOM_ALLOWED_HEADER with a list of Echo-Token, Client-Id, CMR-Request-Id, CMR-Scroll-Id. The same for the CORS_CUSTOM_EXPOSED_HEADER. We can remove the `compare-cors-headers` function.

system-int-test/test/cmr/system_int_test/header_test.clj
```diff
@@ -67,3 +69,16 @@
                               {:headers {:cmr-request-id cmr-request-id}
                                :throw-exceptions false})]
       (is (cmr-request-id-in-header? headers cmr-request-id)))))
+
+(deftest cors-headers
+  (testing "allowed headers in options search request"
+    (is (= (get-in (client/options (url/search-url :collection))
+                   [:headers routes/CORS_CUSTOM_ALLOWED_HEADER])
```
chris-durbin: We generally try to write our tests without referencing the source code vars so that it's clear what's being tested, and if there is a change to the source we know we did not break anything. 

For example this test would pass as written even without adding in the CMR-Scroll-Id header.

So we'd probably turn this into 4 assertions - one for each of the allowed headers. Something along the lines of:
```
  (let [allowed-headers (-> (client/options (url/search-url :collection))
                            (get-in [:headers "Access-Control-Allow-Headers"])
                            (string/split #", "))]
    (is (some #{"Echo-Token"} allowed-headers))
    (is (some #{"Client-Id"} allowed-headers))
    (is (some #{"CMR-Request-Id"} allowed-headers))
    (is (some #{"CMR-Scroll-Id"} allowed-headers)))
```
Note that you'll need to require `[clojure.string :as string]`

oubiwann:
@CayvonH Could you rebase this, ammending your first 2 commits to include the ticket number? Can pair on this, if you want.

system-int-test/test/cmr/system_int_test/header_test.clj
```diff
@@ -1,10 +1,13 @@
...
   "Tests for headers in ingest and search responses"
   (:require [clojure.test :refer :all]
+            [clj-http.client :as client]
+            [clojure.string :as string]
             [cmr.system-int-test.utils.ingest-util :as ingest]
             [cmr.system-int-test.utils.search-util :as search]
             [cmr.system-int-test.utils.index-util :as index]
             [cmr.system-int-test.data2.collection :as dc]
+            [cmr.system-int-test.utils.url-helper :as url]
             [cmr.common-app.api.routes :as routes]))
```
sxu123: Nitpick: Could you line up the require block the same way as in common-app-lib/src/cmr/common_app/api/routes.clj? All our new or newly modified require block should line up that way.

PR #550, opened 2018-08-09
CMR-5094

ygliuvt:
Added indices on variable_concept_id to variable/service associations tables to improve search performance.


ygliuvt:
This is only for performance improvement, functionally there is no change. So no, there won't be an easy way to test these through integration testing or should we even attempt performance testing in integration test suite. 

< Migrations look good - is there a way we can write tests for these?

PR #552, opened 2018-08-14
CMR-5098 Fixes index rebalancing for old deleted collections.

daniel-zamora:
When a collection has been deleted, along with its tombstone the index rebalancing for that collection would fail on the collection validation not finding a collection.  This would prevent removal of old unused indices.

### DIFF
--- a/bootstrap-app/src/cmr/bootstrap/services/bootstrap_service.clj
+++ b/bootstrap-app/src/cmr/bootstrap/services/bootstrap_service.clj
@@ -51,6 +51,15 @@
      :bad-request
      [(format "Provider: [%s] does not exist in the system" provider-id)])))
 
+(defn- validate-collection
+   "Validates to be bulk_indexed collection exists in cmr else an exception is thrown."
+   [context provider-id collection-id]
+   (let [provider (get-provider context provider-id)]
+     (when-not (bulk/get-collection context provider collection-id)
+       (errors/throw-service-errors
+        :bad-request
+        [(format "Collection [%s] does not exist." collection-id)]))))
+
 (defn index-provider
   "Bulk index all the collections and granules for a provider."
   [context dispatcher provider-id start-index]
@@ -61,15 +70,6 @@
   "Bulk index all the concepts with a revision date later than the given date-time."
   [context dispatcher date-time]
   (dispatch/index-data-later-than-date-time dispatcher context date-time))
-
-(defn- validate-collection
-  "Validates to be bulk_indexed collection exists in cmr else an exception is thrown."
-  [context provider-id collection-id]
-  (let [provider (get-provider context provider-id)]
-    (when-not (bulk/get-collection context provider collection-id)
-      (errors/throw-service-errors
-       :bad-request
-       [(format "Collection [%s] does not exist." collection-id)]))))
 
 (defn index-collection
   "Bulk index all the granules in a collection"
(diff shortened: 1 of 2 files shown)


bootstrap-app/src/cmr/bootstrap/services/bootstrap_service.clj
```diff
@@ -62,22 +62,13 @@
   [context dispatcher date-time]
   (dispatch/index-data-later-than-date-time dispatcher context date-time))
 
-(defn- validate-collection
```
chris-durbin: I think we should keep this validation when trying to move a collection to a separate index (we don't want to create a new collection index for a non-existent collection).

bootstrap-app/src/cmr/bootstrap/services/bootstrap_service.clj
```diff
@@ -62,22 +62,13 @@
...
-      (errors/throw-service-errors
-       :bad-request
-       [(format "Collection [%s] does not exist." collection-id)]))))
-
 (defn index-collection
   "Bulk index all the granules in a collection"
   ([context dispatcher provider-id collection-id]
    (index-collection context dispatcher provider-id collection-id nil))
   ([context dispatcher provider-id collection-id options]
-   (validate-collection context provider-id collection-id)
```
chris-durbin: I'd keep this validation call.

bootstrap-app/src/cmr/bootstrap/services/bootstrap_service.clj
```diff
@@ -136,7 +127,6 @@
     (info (format "Starting to rebalance granules for collection [%s] to target [%s]."
                   concept-id target))
     (rebalancing-collections/validate-target target concept-id)
-    (validate-collection context (:provider-id (concepts/parse-concept-id concept-id)) concept-id)
```
chris-durbin: Change to
```
(when (= "separate-index" target)
  (validate-collection context (:provider-id (concepts/parse-concept-id concept-id)) concept-id))
```

bootstrap-app/src/cmr/bootstrap/services/bootstrap_service.clj
```diff
@@ -164,7 +154,6 @@
 (defn finalize-rebalance-collection
   "Finalizes collection rebalancing."
   [context concept-id]
-  (validate-collection context (:provider-id (concepts/parse-concept-id concept-id)) concept-id)
```
chris-durbin: I think it makes sense to remove the validation here in finalize.

We should add a test for finalize with a collection ID that doesn't exist to be sure it doesn't need the validation. Basically the same test as the "Finalizing not started collection" test, but use an ID that doesn't exist and verify it fails with the same validation error that the concept ID is not marked as rebalancing.

system-int-test/test/cmr/system_int_test/bootstrap/rebalance_collections_test.clj
```diff
@@ -47,10 +51,6 @@
        (is (= {:status 400
                :errors ["Provider: [NON_EXIST] does not exist in the system"]}
               (bootstrap/start-rebalance-collection "C1-NON_EXIST" {:synchronous false}))))
-     (testing "Non existent collection"
```
chris-durbin: This validation should be back in place so we should put this test back in (moving to separate index which is the default target should fail validation).

bootstrap-app/src/cmr/bootstrap/services/bootstrap_service.clj
oubiwann: This is showing up in the Github diff as still deleted ...

daniel-zamora: Its moved up in the file. I can fix that if you think its important.

oubiwann: Hahaha, nah -- not important at all.

PR #553, opened 2018-08-15
CMR-4906: Added Cmr-Revision-Id header for ACL update.

sxu123:
Will need to add tests. Demo worked. Not sure where to add the tests. Looks like the current ACL related tests are just some ingest tests that test some ACL indirectly?


oubiwann:
This PR looks good, but as already stated, it does need tests. I did find some `Cmr-*` header checks in `system-int-test` ... not sure if those are good examples or not.

sxu123:
Thanks. Added tests.

PR #554, opened 2018-08-15
CMR-5107: Bounds to IndexRanges

oubiwann:
The proposed update to UMM-Var 1.2 has been made; this PR represents the changes we needed to make in order to support the update.

This update will remove a great deal of confusion that has already arisen around the metadata being saved in the given fields, and should make things much easier for future developers/project contributors.

### DIFF
--- a/umm-spec-lib/resources/example-data/umm-json/variable/v1.2/V1200241825-EDF_DEV01_v1_2.json
+++ b/umm-spec-lib/resources/example-data/umm-json/variable/v1.2/V1200241825-EDF_DEV01_v1_2.json
@@ -1,73 +1,64 @@
 {
-	"Name":"TotCH4_A",
-	"Definition":"Total Methane",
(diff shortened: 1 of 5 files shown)


oubiwann:
Looking into a failing build ...

oubiwann:
Build should be fixed now; I've emailed Simon C. with a request to change the schema (the issue was that a required field that should be an array had min items set to 0; setting it to 2 did the trick).

oubiwann:
Note that there are no changes for tests due to the fact that the code that tests this autogenerates tests based upon the schemas. As an example, the failing build was due to an issue with the schema; when that was fixed (and the changes pushed back to the Metadata Quality team), the tests passed.

mschmele:
> We got lucky that there is no additional migration needed for this change.

I thought I was going crazy when I looked at that, haha. Makes for a nice and easy review

PR #555, opened 2018-08-15
CMR-5039 Error found by MMT when updating their tests to use CMR access-control-app 

daniel-zamora:
Adds default expression to parse-collection-access-value, metadata MMT was using was causing a failure to the get-or-lazy-get call in acl-matcher because the assoc-lazy call was using parse-collection-access-value which had no condition for umm-json. Instead of throwing an error, the lazy-get will now return a nil.

### DIFF
--- a/umm-spec-lib/src/cmr/umm_spec/umm_spec_core.clj
+++ b/umm-spec-lib/src/cmr/umm_spec/umm_spec_core.clj
@@ -152,4 +152,4 @@
      mt/dif (dif-util/parse-access-constraints metadata true)
      mt/dif10 (dif-util/parse-access-constraints metadata true)
      mt/iso19115 (iso19115-2-to-umm/parse-access-constraints metadata true)
-     mt/iso-smap nil)))
+     nil)))


umm-spec-lib/src/cmr/umm_spec/umm_spec_core.clj
```diff
@@ -152,4 +152,4 @@
      mt/dif (dif-util/parse-access-constraints metadata true)
      mt/dif10 (dif-util/parse-access-constraints metadata true)
      mt/iso19115 (iso19115-2-to-umm/parse-access-constraints metadata true)
-     mt/iso-smap nil)))
+     nil)))
```
chris-durbin: The code in this namespace looks like it's missing conversions for parsing collection temporal and collection access-value when the format is UMM-JSON as well as parsing a collection access-value when the collection is in ISO SMAP format. It seems like we are missing some tests and might have some bugs.

oubiwann: @chris-durbin: Can we go over this after standup tomorrow?

daniel-zamora: I have no idea if this is even suppose to throw an exception for other formats.  It could be that its assumed to never be called in that case and there should be an error.  I'm thinking this could affect production for collections that are native umm-json.  I submitted a PR that had the smallest footprint, but I can try and add a test case for this.

chris-durbin: If this is breaking MMT I'm good with adding your fix in for now since I can't see how it would make things any worse. We should open a new ticket to deal with the formats that are missing conversions.

PR #556, opened 2018-08-15
CMR-5106 Changing Entry Title of a collection causes it to disappear from searches until the ACL is re-created

daniel-zamora:
This PR changes collection ACL matcher logic for entry-titles and concept-ids to be OR'd instead of AND'd.


daniel-zamora:
Looking into adding a test case.

PR #562, opened 2018-08-23
CMR-4955: Make collection ingest tests to test against the latest umm version

sxu123:
Some ingest tests are currently testing against cmr.common-app.config/collection-umm-version, which is the accepted collection ingest umm version.  We should make our ingest tests testing against the latest umm version, which is cmr.umm-spec.versioning/current-collection-version. 

1. Set the collection-umm-version to current-collection-version in the ingest/reset-fixture, which is used in most of the ingest tests. 
Most of the collection ingest tests are not specifying the umm version, so it makes use of the collection-umm-version. By setting it to current-collection-version in the test environment, these ingest tests will test against the current-collection-version. Note: during the test, I changed the default value of cmr.common-app.config/collection-umm-version to "1.11", which doesn't exist, to make sure no tests are using the default value. 

2. Modified an acl doc for CMR-4906.

### DIFF
--- a/system-int-test/src/cmr/system_int_test/utils/ingest_util.clj
+++ b/system-int-test/src/cmr/system_int_test/utils/ingest_util.clj
@@ -7,11 +7,11 @@
    [clojure.string :as str]
    [clojure.test :refer :all]
    [cmr.acl.core :as acl]
+   [cmr.common-app.config :as common-config]
    [cmr.common-app.test.side-api :as side]
    [cmr.common.mime-types :as mt]
    [cmr.common.util :as util]
    [cmr.common.xml :as cx]
-   [cmr.ingest.config :as icfg]
    [cmr.mock-echo.client.echo-util :as echo-util]
    [cmr.system-int-test.data2.provider-holdings :as ph]
    [cmr.system-int-test.system :as s]
@@ -21,6 +21,7 @@
    [cmr.system-int-test.utils.url-helper :as url]
    [cmr.transmit.access-control :as ac]
    [cmr.transmit.config :as transmit-config]
+   [cmr.umm-spec.versioning :as umm-versioning]
    [cmr.umm.echo10.echo10-collection :as c]
    [cmr.umm.echo10.echo10-core :as echo10]
    [cmr.umm.echo10.granule :as g])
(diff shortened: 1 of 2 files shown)


system-int-test/src/cmr/system_int_test/utils/ingest_util.clj
```diff
@@ -631,6 +632,11 @@
 
 ;;; fixture - each test to call this fixture
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+(defn set-collection-ingest-umm-version-to-current
+  "Set the collection ingest-accept-umm-version to current-collection-version."
+  []
+  (side/eval-form `(common-config/set-collection-umm-version!
+                     umm-versioning/current-collection-version)))
```
sxu123: Wonder if we have the same issue with service and variable - i.e. their ingest tests are NOT testing against the latest umm version. If that's the case, we need to fix that too.

ygliuvt: Yes, good point. Right now we do not have a way to test the latest UMM version for variable/service similar to collection before this fix. We should do similar thing here to tie the ingest version to the latest UMM version supported internally by CMR.

system-int-test/src/cmr/system_int_test/utils/ingest_util.clj
```diff
@@ -631,6 +632,11 @@
 
 ;;; fixture - each test to call this fixture
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+(defn set-collection-ingest-umm-version-to-current
```
ygliuvt: Rename this function to `set-ingest-umm-version-to-current` and include setting ingest UMM version of variables/services also to the most current version that CMR supports.

sxu123: Yeah, that's what I thought would be a good thing to do.  Done.

sxu123: Looks like all the variable/service tests passed with default variable-umm-version and service-umm-version set to 1.3 which doesn't exist.  This indicates that the tests don't use these default values for ingests.  They might have already used the latest umm versions.  Will write a ticket to investigate if it's true, if so, need to see if it's only for testing or if it's because ingest is actually using the latest umm version, rather than the accepted umm version.

PR #563, opened 2018-08-23
CMR-5080: log size of metadata to be ingested

mschmele:
I picked `metadata->concept` because it is used in all ingest functions. It can be moved easily

### DIFF
--- a/ingest-app/src/cmr/ingest/api/core.clj
+++ b/ingest-app/src/cmr/ingest/api/core.clj
@@ -1,7 +1,6 @@
 (ns cmr.ingest.api.core
   "Supports the ingest API definitions."
   (:require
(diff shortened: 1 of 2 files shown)


ingest-app/src/cmr/ingest/api/core.clj
```diff
@@ -205,6 +206,7 @@
 (defn metadata->concept
   "Create a metadata concept from the given metadata"
   [concept-type metadata content-type headers]
+  (info (format "Size of metadata to be ingested: [%s]" (memory-meter/measure metadata)))
```
ygliuvt: I don't think we care about how JVM allocate space for the metadata. We could just use JAVA's `metadata.getBytes().length` call to get the bytes of metadata.

PR #568, opened 2018-09-05
CMR-5037: Strengthen temporal validation.

CayvonH:


umm-spec-lib/src/cmr/umm_spec/xml_to_umm_mappings/dif9.clj
```diff
@@ -92,9 +92,9 @@
  "Return a list of temporal extents from the XML doc"
  [doc sanitize?]
  (if-let [temporals (select doc "/DIF/Temporal_Coverage")]
-  [{:RangeDateTimes (for [temporal temporals]
-                      {:BeginningDateTime (date/with-default (value-of temporal "Start_Date") sanitize?)
-                       :EndingDateTime (parse-dif-end-date (value-of temporal "Stop_Date"))})}]
+  [{:RangeDateTimes (doall (for [temporal temporals]
```
oubiwann: I typically like to put the expression in `(doall ...)` on its own line (if the body of the expression is more than one line, that is).

PR #569, opened 2018-09-13
CMR-5129 Add granule ISO SMAP SpatialExtent/Orbit translations

daniel-zamora:
Add to the current granule ISO SMAP translation to read in the SpatialExtent/Orbit information into the current UMM-G CMR model called spatial-coverage/orbit.
Basically read in the SMAP granule record and read the following snippet:
```
    <gmd:geographicElement>
      <gmd:EX_GeographicDescription id="Orbit">
        <gmd:geographicIdentifier>
          <gmd:MD_Identifier>
            <gmd:code>
              <gco:CharacterString>AscendingCrossing: -140.0 StartLatitude: 0.0 StartDirection: A EndLatitude: 0.0 EndDirection: A</gco:CharacterString>
            </gmd:code>
            <gmd:codeSpace>
              <gco:CharacterString>gov.nasa.esdis.umm.orbitparameters</gco:CharacterString>
            </gmd:codeSpace>
            <gmd:description>
              <gco:CharacterString>OrbitParameters</gco:CharacterString>
            </gmd:description>
          </gmd:MD_Identifier>
        </gmd:geographicIdentifier>
      </gmd:EX_GeographicDescription>
    </gmd:geographicElement>
```
so that the following fields can be populated:
spatial-coverage/orbit
spatial-coverage/orbit/ascending-crossing
spatial-coverage/orbit/start-lat
spatial-coverage/orbit/start-direction
spatial-coverage/orbit/end-lat
spatial-coverage/orbit/end-direction


common-lib/src/cmr/common/util.clj
```diff
@@ -932,3 +932,12 @@
           #(string/replace % "_" "-")
           #(string/replace % #"^CMR_" ""))
     #(string/starts-with? %1 "CMR_")))
+
+(defn get-index-or-nil
+ "Get the index of the key in the string. Return nil if the key does not
```
chris-durbin: I think of a key in a map when I see `key`. It looks like this should be called `substring` instead.

umm-lib/src/cmr/umm/iso_smap/granule/spatial.clj
```diff
@@ -1,20 +1,30 @@
...
 
 (def extent-path
   [:composedOf :DS_DataSet :has :MI_Metadata :identificationInfo :MD_DataIdentification
    :extent :EX_Extent :geographicElement])
 
+(defn- orbit?
+  "Returns true or false depending on the existence of the :ascending-crossing key existing in the
+  spatial coverage map."
+  [spatial-map]
+  (some #(= :ascending-crossing %) (keys spatial-map)))
```
chris-durbin: I think `(some #{:ascending-crossing} (keys spatial-map))` is the idiomatic way to check for the existence of a key.

umm-lib/src/cmr/umm/iso_smap/granule/spatial.clj
```diff
@@ -1,20 +1,30 @@
...
+
 (defn xml-elem->SpatialCoverage
   "Returns a UMM SpatialCoverage from a parsed XML structure"
   [xml-struct]
   (let [spatial-elems (cx/elements-at-path xml-struct extent-path)
         geometries (flatten (map spatial/decode spatial-elems))]
     (when (seq geometries)
-      (g/map->SpatialCoverage {:geometries geometries}))))
+      (g/map->SpatialCoverage (util/remove-map-keys empty?
+                                {:geometries (remove #(orbit? %) geometries)
```
chris-durbin: When a function only takes one argument you can write it as `(remove orbit? geometries)`. Similar for the `filter` on the next line.

spatial-lib/src/cmr/spatial/encoding/gmd.clj
```diff
@@ -74,3 +93,44 @@
...
+        end-lat (util/get-index-or-nil orbit-str "EndLatitude:")
+        end-direction (util/get-index-or-nil orbit-str "EndDirection:")]
+    (when (= description-type "OrbitParameters")
+      (orbit/->Orbit
+       (when ascending-crossing
+         (let [asc-c (subs orbit-str
+                           ascending-crossing
+                           (or start-lat start-direction end-lat end-direction
+                               (count orbit-str)))]
+           (Float. (str/trim (subs asc-c (inc (.indexOf asc-c ":")))))))
```
chris-durbin: I believe these calls to `Float.` will throw an exception if the value can't be represented as a float. I see validation functions on the UMM, but this parsing happens before converting to the model.

Can you add a test with invalid ISO-SMAP metadata to verify we successfully return an error to the client and not an internal server error?

umm-lib/src/cmr/umm/iso_smap/granule/orbit.clj
```diff
@@ -0,0 +1,111 @@
...
+            start-lat (convert-direction start-direction)
+            end-lat (convert-direction end-direction))))
+
+(defn- coerce-float
+  "Coerce's string to float, catches exceptions and produces error message."
+  [field value]
+  (try
+    (Float. value)
+    (catch Exception e
+      (error e (format "For orbit field [%s] the value [%s] is not a number." field value)))))
```
chris-durbin: We're going to return whatever the log message error macro returns - right now that's `nil`. I think generally we'd want to add an explicit return - so add `nil` after logging the error if that's the value we want it to return.

PR #572, opened 2018-09-20
CMR-5080: Log metadata-size-bytes with provider info, after successful ingest

sxu123:
This PR is addressing  three issues:
1. Logging other concept info together with the metadata-size-bytes, specifically provider info so that Catalino could get all the metadata size ingested for a given provider.
2. Logging the metadata size only after successful ingest.
3. Make sure we log both the attempt and the success of ingest. The difference is at the attempt, the metadata-size-bytes is not logged. Currently granule ingest is logged at the attempt, collection ingest is logged at the success, service and variable ingest are not logged.  These all need to be changed.

The log will look sth like:
[a83aa11a-281a-48c0-b2fe-f232231074f7] INFO [cmr.ingest.api.core] - Successfully ingested concept {:format "application/echo10+xml", :native-id "20040618200000-OSISAF-L3C_GHRSST-SSTsubskin-SEVIRI_SST-ssteqc_meteosat08_20040618_200000-v02.0-fv01.0.nc", :concept-type :granule, :provider-id "PODAAC", :metadata-size-bytes 2574} from client unknown

As to whether the file is always UTF-8 encoded, I'm not sure. I'm just using the same code Mark used before to calculate the size, which was approved by the team.

### DIFF
--- a/ingest-app/src/cmr/ingest/api/core.clj
+++ b/ingest-app/src/cmr/ingest/api/core.clj
@@ -236,8 +236,7 @@
   [concept request-context]
   (let [metadata-size-bytes (-> (:metadata concept) (.getBytes "UTF-8") alength)
         concept (assoc concept :metadata-size-bytes metadata-size-bytes)]
-    (info (format "Ingesting %s %s from client %s"
-                  (name (:concept-type concept)) 
+    (info (format "Successfully ingested %s from client %s"
                   (concept->loggable-string concept)
                   (:client-id request-context)))))
 
--- a/ingest-app/src/cmr/ingest/api/core.clj
+++ b/ingest-app/src/cmr/ingest/api/core.clj
@@ -205,8 +205,6 @@
 (defn metadata->concept
   "Create a metadata concept from the given metadata"
   [concept-type metadata content-type headers]
-  (info (format "Size of metadata (in bytes) to be ingested: [%s]"
-                (-> metadata (.getBytes "UTF-8") alength)))
   (-> {:metadata metadata
        :format (mt/keep-version content-type)
        :native-id (:name metadata)
@@ -232,6 +230,16 @@
   "Returns a string with information about the concept as a loggable string."
   [concept]
   (pr-str (dissoc concept :metadata)))
+
+(defn log-concept-with-metadata-size
+  "Log the concept with metadata-size-bytes added, and metadata removed."
+  [concept request-context]
+  (let [metadata-size-bytes (-> (:metadata concept) (.getBytes "UTF-8") alength)
+        concept (assoc concept :metadata-size-bytes metadata-size-bytes)]
+    (info (format "Ingesting %s %s from client %s"
+                  (name (:concept-type concept)) 
+                  (concept->loggable-string concept)
+                  (:client-id request-context)))))
 
 (defn set-default-error-format [default-response-format handler]
   "Ring middleware to add a default format to the exception-info created during exceptions. This
--- a/ingest-app/src/cmr/ingest/api/granules.clj
+++ b/ingest-app/src/cmr/ingest/api/granules.clj
@@ -65,10 +65,11 @@
     (api-core/verify-provider-exists request-context provider-id)
     (acl/verify-ingest-management-permission request-context :update :provider-object provider-id)
     (common-enabled/validate-write-enabled request-context "ingest")
-    (let [concept (api-core/body->concept! :granule provider-id native-id body content-type headers)]
-      (info (format "Ingesting granule %s from client %s"
-                    (api-core/concept->loggable-string concept) (:client-id request-context)))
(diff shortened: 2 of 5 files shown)


ingest-app/src/cmr/ingest/api/core.clj
```diff
@@ -233,6 +231,16 @@
   [concept]
   (pr-str (dissoc concept :metadata)))
 
+(defn log-concept-with-metadata-size
```
daniel-zamora: I think this should be moved out of here and into cmr.ingest.services.helper

ingest-app/src/cmr/ingest/api/core.clj
```diff
@@ -233,6 +231,16 @@
...
+  "Log the concept with metadata-size-bytes added, and metadata removed."
+  [concept request-context]
+  (let [metadata-size-bytes (-> (:metadata concept) (.getBytes "UTF-8") alength)
```
ccuadrado: Are we sure metadata is always UTF-8 encoded?

chris-durbin:
Can you add what the log line looks like in the description for the PR?

ingest-app/src/cmr/ingest/api/granules.clj
```diff
@@ -65,10 +65,11 @@
     (api-core/verify-provider-exists request-context provider-id)
     (acl/verify-ingest-management-permission request-context :update :provider-object provider-id)
     (common-enabled/validate-write-enabled request-context "ingest")
-    (let [concept (api-core/body->concept! :granule provider-id native-id body content-type headers)]
-      (info (format "Ingesting granule %s from client %s"
-                    (api-core/concept->loggable-string concept) (:client-id request-context)))
-      (api-core/generate-ingest-response headers (ingest/save-granule request-context concept)))))
+    (let [concept (api-core/body->concept! :granule provider-id native-id body content-type headers)
+          save-granule-result (ingest/save-granule request-context concept)]
+      ;;Log the size of the metadata after successful ingest.
```
daniel-zamora: Fix all of these comments, per https://wiki.earthdata.nasa.gov/display/CMR/Coding+Style+Guidelines.
They need a space between the semicolons and the text.

ingest-app/src/cmr/ingest/api/core.clj
daniel-zamora: the cmr.ingest.api.core namespace is starting to look really cluttered.

daniel-zamora: Reading through the namespace it looks like it was always a dumpster for helper functions.  I think we can probably leave this here then.

sxu123: I have no problem of moving it. But I want to understand what core is used for because other common functions were in there too, so we need to be consistent...

ingest-app/src/cmr/ingest/api/core.clj
sxu123: Please refer to the description at the beginning of the PR.  @ygliuvt,  @chris-durbin, what do you guys think?

ingest-app/src/cmr/ingest/api/granules.clj
sxu123: Thanks. Done.

ingest-app/src/cmr/ingest/api/core.clj
chris-durbin: There are probably many places in our code where we assume UTF-8. I think for the purposes of calculating bytes it is fine to assume UTF-8.

ingest-app/src/cmr/ingest/api/core.clj
```diff
@@ -233,6 +231,16 @@
...
+        concept (assoc concept :metadata-size-bytes metadata-size-bytes)]
+    (info (format "Ingesting %s %s from client %s"
+                  (name (:concept-type concept)) 
+                  (concept->loggable-string concept)
```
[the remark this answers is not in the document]
sxu123: Yes. I originally just did (format "Ingest concept %s from client %s"...   because the concept->Loggable string already contains concept type. But then I'm not sure if anything is depending on the "Ingest granule", "Ingest collection" etc, so I just made it more specific.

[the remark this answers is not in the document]
sxu123: Okay. Catalino prefers to change it to (format "Successfully ingested concept %s from client %s"...  Done.

ingest-app/src/cmr/ingest/api/granules.clj
```diff
@@ -65,10 +65,11 @@
...
+      (api-core/log-concept-with-metadata-size concept request-context)
```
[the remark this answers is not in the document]
sxu123: Yes. Done.

PR #582, opened 2018-09-27
CMR-5138: Use granule start/end date for DateDates when not provided, for opendata search.

sxu123:
This is Work In Progress PR. I'm running into issues with retrieving the granule-start-date and granule-end-date. It complains about "No implementation of method: :get-value of protocol: #'cmr.common.cache/CmrCache found for class: nil". I'm not sure what went wrong since the same method indexer-coll-gran-aggre-cache/get-coll-gran-aggregates context concept-id was used in cmr.indexer.data.concepts.collection without problem. I tried using external db, that didn't help either. Please take a look and let me know what I did wrong. Thanks.


ygliuvt:
@sxu123 The reason that method works in indexer but not in your code is because that cache only exists in the indexer-app, not the search-app. You code is in the search app.

sxu123:
@ygliuvt, Thanks Leo.

PR #583, opened 2018-09-27
CMR-5062: Added support for TemporalExtent in UMM-G.

ygliuvt:

### DIFF
--- a/umm-spec-lib/src/cmr/umm_spec/umm_g/granule.clj
+++ b/umm-spec-lib/src/cmr/umm_spec/umm_g/granule.clj
@@ -36,6 +36,19 @@
                            :short-name (:ShortName collection-ref)
                            :version-id (:Version collection-ref)})))
 
+(defn umm-g->Temporal
+  "Returns a UMM Temporal from a parsed UMM-G JSON"
+  [umm-g-json]
+  (let [temporal (:TemporalExtent umm-g-json)
+        range-date-time (when-let [range-date-time (:RangeDateTime temporal)]
+                          (umm-c/map->RangeDateTime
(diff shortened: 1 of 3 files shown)


umm-spec-lib/src/cmr/umm_spec/umm_g/granule.clj
```diff
@@ -81,5 +94,12 @@
                             {:EntryTitle entry-title}
                             {:ShortName short-name
                              :Version version-id})
+     :TemporalExtent (if-let [single-date-time (:single-date-time temporal)]
```
chris-durbin: Is there anything that prevents having both a RangeDateTime and a SingleDateTime in the schema?

ygliuvt: Yes, it is `one-of`.

umm-spec-lib/test/cmr/umm_spec/test/umm_g/granule.clj
```diff
@@ -9,6 +9,7 @@
    [cmr.common.test.test-check-ext :refer [defspec]]
    [cmr.umm-spec.umm-spec-core :as core]
    [cmr.umm.test.generators.granule :as gran-gen]
+   [cmr.umm.umm-collection :as umm-c]
```
mschmele: It might be worth moving some of this common functionality into a utility namespace in the future. Using collection-specific code in granule-specific namespaces is bound to get confusing and will probably lead to duplication

ygliuvt: This is to reference the RangeDateTime type that is defined in `umm-collection`.

PR #585, opened 2018-09-27
CMR-5205 As the CMR system, I want to support version 1.3 of UMM-Var.

daniel-zamora:


daniel-zamora:
Looking into potential tests for this.

daniel-zamora:
There is some confusion about Characteristics, in the migrations it looks like an array value, but example data it is simply a map.  The schema has it as a object, not an array.  So I think all the migration tests are wrong for this field.

oubiwann:
Yeah, we got conflicting data from the metadata team, before things were 100% ironed out ... the schema and the example didn't match up. I think I recently tried to bring them up to date, but might have missed something ...

oubiwann:
Hrm, can't seem to locate the originals; regardless, you are correct -- the data type of characteristics in the migrations should be a map from v1.2 onward.

umm-spec-lib/test/cmr/umm_spec/test/migration/version/variable.clj
```diff
@@ -35,17 +35,17 @@
...
-                      :Bounds "Text describing coord bounds"
-                      :Coordinates "Text describing coord range"
-                      :GridMapping "Text describing mapping projection"
-                      :Size 1024
-                      :SizeUnits "bytes"
-                      :ChunkSize 100
-                      :Structure "/file.hdf/MODIS_Grid_Daily_1km_LST/Data_Fields/"
-                      :MeasurementConditions "Sampled Particle Size Range: 90 - 600 nm"
-                      :ReportingConditions "STP: 1013 mb and 273 K"}]
+   :Characteristics {:StandardName "VAR_1"
```
oubiwann: Thanks for fixing that.

PR #586, opened 2018-09-28
CMR-5066: Added support for Platforms, Instruments, Characteristics,  etc. in UMM-G.

ygliuvt:


oubiwann:
I can't even begin to say how great your gif selection for this PR is ...

PR #587, opened 2018-09-28
CMR-5138: Use Granule's start/end date for issued/modified in opendata.

sxu123:
For opendata collection search, when the collection's Creation Date and Last Modified Date are not present, or when they are equal to the default 1970 date, use collection's earliest granule's start date and latest granule's end date. If the collection has no granules, we should leave Creation Date blank and Last Modified Date blank(but since it's a required field, we are adding a default currently, which is different from the 1970 date).  This is what the ticket asked.

In this PR:
1. We always use granule-start-date and granule-end-date regardless of if the collection has granules. When collection doesn't have granules, these dates contain the earliest and the latest of the collection's TemporalExtent data. @chris-durbin thinks we might just want to do it this way. If we decide we want to still do what ticket asks, I could pull has-granules and check it. But because of the following, it can't be achieved right now as it will always be false. 
2. We still need to update the index periodically to reflect the latest granule-start-date and granule-end-date which is obtained from elastic search through collection-granule-aggregation.  This is a lot more work, Chris said we might want to have a separate ticket for this :)

### DIFF
--- a/indexer-app/src/cmr/indexer/data/collection_granule_aggregation_cache.clj
+++ b/indexer-app/src/cmr/indexer/data/collection_granule_aggregation_cache.clj
@@ -174,8 +174,8 @@
           {start2 :granule-start-date-stored end2 :granule-end-date-stored} gt2]
       {:granule-start-date-stored (if (< (compare start1 start2) 0) start1 start2)
        :granule-end-date-stored (when (and end1 end2) ;; If either is nil return nil
-                           ;; else return max time
-                           (if (> (compare end1 end2) 0) end1 end2))})
+                                  ;; else return max time
+                                  (if (> (compare end1 end2) 0) end1 end2))})
     (or gt1 gt2)))
 
 (defn- merge-coll-gran-aggregates
--- a/indexer-app/src/cmr/indexer/data/collection_granule_aggregation_cache.clj
+++ b/indexer-app/src/cmr/indexer/data/collection_granule_aggregation_cache.clj
@@ -93,10 +93,10 @@
                              (get-in bucket [:max-temporal :value]))
                  some-with-no-end (> (get-in bucket [:no-end-date :doc_count]) 0)]]
        [concept-id
-        {:granule-start-date earliest-start
+        {:granule-start-date-stored earliest-start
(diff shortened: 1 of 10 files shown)


indexer-app/src/cmr/indexer/data/index_set.clj
```diff
@@ -317,8 +317,8 @@
 
           ;; Temporal range of min and max granule values or the same as collection start and end date
           ;; if the collection has not granules.
-          :granule-start-date             m/date-field-mapping
-          :granule-end-date               m/date-field-mapping
+          :granule-start-date             (m/stored m/date-field-mapping)
```
chris-durbin: We can't change Elasticsearch fields once they've been created. We have to add new fields, keep these existing fields, and change our indexing and search code to make sure they are using the new fields.

i.e.:
```
          :granule-start-date             m/date-field-mapping
          :granule-end-date               m/date-field-mapping
          :granule-start-date-stored (m/stored m/date-field-mapping)
          :granule-end-date-stored   (m/stored m/date-field-mapping)
```

sxu123: Done.

indexer-app/src/cmr/indexer/data/collection_granule_aggregation_cache.clj
```diff
@@ -93,10 +93,10 @@
                              (get-in bucket [:max-temporal :value]))
                  some-with-no-end (> (get-in bucket [:no-end-date :doc_count]) 0)]]
        [concept-id
-        {:granule-start-date earliest-start
+        {:granule-start-date-stored earliest-start
```
chris-durbin: Just an FYI.

We try not to expose the Elasticsearch field implementation beyond the `cmr.search.data.query-to-elastic` namespace. This means calling:

```(q2e/query-field->elastic-field :granule-start-date :collection)```
to pull back the value. The CMR code base is really inconsistent, and I'm seeing the field names leak out in a bunch of namespaces probably because it's much easier to understand, so I'm fine with this PR.

sxu123: Thank you!

PR #588, opened 2018-10-01
CMR-5067 As a Client User, I want to parse and generate project-refs in SWOT granule metadata format

daniel-zamora:


umm-spec-lib/src/cmr/umm_spec/umm_g/project.clj
```diff
@@ -0,0 +1,25 @@
...
+
+(defn- umm-g-project->ProjectRef
+  "Returns the umm-lib granule model ProjectRef from the given UMM-G Project."
+  [project]
+  (:Campaigns project))
+
+(defn umm-g-projects->ProjectRefs
+  "Returns the umm-lib granule model ProjectRefs from the given UMM-G Projects."
+  [projects]
+  (seq (reduce concat [] (map umm-g-project->ProjectRef projects))))
```
ygliuvt: Change this line to:
`(seq (distinct (mapcat :Campaigns project)))`, and remove the `umm-g-project->ProjectRef` function.

daniel-zamora: Why are we distincting here?

daniel-zamora: When/if this gets updated to allow for multiple ProjectTypes instead of just a list of short names,  the existing structure is more useful.

daniel-zamora: Are you thinking different campaigns can contain the same short names? So we should distinct here?

umm-spec-lib/test/cmr/umm_spec/test/umm_g/granule.clj
```diff
@@ -21,6 +23,23 @@
   "Generator for UMM-G granule in umm-lib Granule model."
   (gen/fmap #(assoc % :collection-ref (gen/generate umm-g-coll-refs)) gran-gen/granules))
 
+(defn- sanitize-operation-modes
+  "Sanitizer for operation-modes, if sequence it removes duplicates, if nil it inserts a not provided."
+  [operation-modes]
+  (cond
```
ygliuvt: Why not change this to:
```
(if (seq operation-modes)
  (distinct operation-modes)
  nil))
```

chris-durbin: Or even better:
```
(when (seq operation-modes)
  (distinct operation-modes)
```

daniel-zamora: Because I removed stuff for you!

PR #592, opened 2018-10-03
CMR-5068: Added support to RelatedUrls in UMM-G.

ygliuvt:


umm-spec-lib/src/cmr/umm_spec/umm_g/related_url.clj
```diff
@@ -0,0 +1,66 @@
+(ns cmr.umm-spec.umm-g.related-url
+  "Contains functions for parsing UMM-G JSON related-urls into umm-lib granule model
+  RelatedURLs and generating UMM-G JSON related-urls from umm-lib granule model RelatedURLs."
+  (:require
+   [cmr.umm.umm-collection :as umm-c])
+  (:import cmr.umm.umm_collection.RelatedURL))
+
+(defn- umm-g-related-url->RelatedURL
```
ygliuvt: There are rules that are still in works to map a UMM-G related url type to umm-lib related url type. Erich is still working on those mappings. Those will be done through a new issue when the mapping is defined.

PR #594, opened 2018-10-04
CMR-5069: Added support of AdditionalAttributes in UMM-G.

ygliuvt:

### DIFF
--- a/umm-spec-lib/src/cmr/umm_spec/umm_g/granule.clj
+++ b/umm-spec-lib/src/cmr/umm_spec/umm_g/granule.clj
@@ -61,21 +61,19 @@
       :data-provider-timestamps (umm-g->DataProviderTimestamps umm-g-json)
       :collection-ref coll-ref
       ; :data-granule (xml-elem->DataGranule umm-g-json)
-      ; :access-value (cx/double-at-path umm-g-json [:RestrictionFlag])
       :temporal (umm-g->Temporal umm-g-json)
       ; :orbit-calculated-spatial-domains (ocsd/xml-elem->orbit-calculated-spatial-domains umm-g-json)
       :platform-refs (platform/umm-g-platforms->PlatformRefs (:Platforms umm-g-json))
       :project-refs (project/umm-g-projects->ProjectRefs (:Projects umm-g-json))
+      ; :access-value (cx/double-at-path umm-g-json [:RestrictionFlag])
       :cloud-cover (:CloudCover umm-g-json)
       :two-d-coordinate-system (tiling-system/umm-g-tiling-identification-system->TwoDCoordinateSystem
                                  (:TilingIdentificationSystem umm-g-json))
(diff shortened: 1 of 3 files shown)


umm-spec-lib/src/cmr/umm_spec/umm_g/granule.clj
```diff
@@ -60,20 +61,19 @@
       :data-provider-timestamps (umm-g->DataProviderTimestamps umm-g-json)
       :collection-ref coll-ref
       ; :data-granule (xml-elem->DataGranule umm-g-json)
-      ; :access-value (cx/double-at-path umm-g-json [:RestrictionFlag])
       :temporal (umm-g->Temporal umm-g-json)
       ; :orbit-calculated-spatial-domains (ocsd/xml-elem->orbit-calculated-spatial-domains umm-g-json)
       :platform-refs (platform/umm-g-platforms->PlatformRefs (:Platforms umm-g-json))
       :project-refs (project/umm-g-projects->ProjectRefs (:Projects umm-g-json))
+      ; :access-value (cx/double-at-path umm-g-json [:RestrictionFlag])
```
chris-durbin: I see you and Daniel have decided to move lines around for fun :)

umm-spec-lib/test/cmr/umm_spec/test/umm_g/granule.clj
```diff
@@ -87,6 +86,26 @@
                                :end-coordinate-1 -50
                                :start-coordinate-2 50
                                :end-coordinate-2 100})
+    :product-specific-attributes [(umm-lib-g/map->ProductSpecificAttributeRef
+                                    {:name "AdditionalAttribute1 Name1"
```
chris-durbin: Additional attributes can be a bunch of different types - float, integer, booleans that get treated as strings, etc. It looks like we're only testing with strings. Shouldn't we test the other types as well?

umm-spec-lib/src/cmr/umm_spec/umm_g/granule.clj
ygliuvt: See the commit message above: "Reformatted code to minimize future merge conflict.". Then it will be like the PR giphy.

umm-spec-lib/test/cmr/umm_spec/test/umm_g/granule.clj
ygliuvt: This is Additional Attributes for granules. It is much easier than collections. Thank God!

chris-durbin: Since this is just for the translation part (validation of types should already be handled separately), I'm ok with merging this in.

I'd recommend changing the examples to:
```{:name "EVI1KM16DAYQCLASSPERCENTAGE" :values ["Value5" "Value6"]}```
rather than repeating the name within the values, but we can discuss that with Erich when he comes back.

PR #595, opened 2018-10-04
CMR-5060 As a Client User, I want to parse and generate data-granule in SWOT granule metadata format

daniel-zamora:
Some assumptions are made for this particular mapping since Erich is out on vacation including which Size should be used, and how to map ArchiveAndDistributionInformation from umm-lib DataGranule.  Haven't addressed the differing enums for DayNightFlag.


umm-spec-lib/src/cmr/umm_spec/umm_g/data_granule.clj
```diff
@@ -0,0 +1,35 @@
...
+(defn umm-g-data-granule->DataGranule
+  "Returns the umm-lib granule model DataGranule from the given UMM-G DataGranule."
+  [data-granule]
+  (when data-granule
+    (g/map->DataGranule
+      (let [{:keys [Identifiers DayNightFlag ProductionDateTime
+                    ArchiveAndDistributionInformation]} data-granule]
+        {:day-night DayNightFlag
+         :producer-gran-id (->> Identifiers
+                                (filter #(= "ProducerGranuleId" (:IdentifierType %)))
```
ygliuvt: It is probably better to rewrite this part with `some` rather than `filter` then `first`, as `some` returns right after a match is found.

umm-spec-lib/src/cmr/umm_spec/umm_g/data_granule.clj
```diff
@@ -0,0 +1,35 @@
...
+                                first
+                                :Identifier)
+         :production-date-time ProductionDateTime
+         :size (get (first ArchiveAndDistributionInformation) :Size)}))))
```
ygliuvt: Yeah, this UMM-G definition does not match umm-lib granule model at all. :(

umm-spec-lib/src/cmr/umm_spec/umm_g/data_granule.clj
```diff
@@ -0,0 +1,35 @@
...
+  (when data-granule
+    (let [{:keys [producer-gran-id day-night production-date-time size]} data-granule]
+      {:DayNightFlag day-night
+       :Identifiers (when producer-gran-id
+                      [{:Identifier producer-gran-id
+                        :IdentifierType "ProducerGranuleId"}])
+       :ProductionDateTime production-date-time
+       :ArchiveAndDistributionInformation [{:Name "Not provided"
+                                            :Size size
+                                            :SizeUnit "MB"}]})))
```
ygliuvt: Since we don't have `SizeUnit` in umm-lib, a better default value is `NA`.

umm-spec-lib/src/cmr/umm_spec/umm_g/data_granule.clj
```diff
@@ -0,0 +1,35 @@
...
+  (:import cmr.umm.umm_granule.UmmGranule))
+
+(defn umm-g-data-granule->DataGranule
+  "Returns the umm-lib granule model DataGranule from the given UMM-G DataGranule."
+  [data-granule]
+  (when data-granule
+    (g/map->DataGranule
+      (let [{:keys [Identifiers DayNightFlag ProductionDateTime
+                    ArchiveAndDistributionInformation]} data-granule]
+        {:day-night DayNightFlag
```
ygliuvt: UMM-G uses different case for DayNightFlag from umm-lib granule model which is basely what is ECHO10, this function should handle the mapping correctly.

umm-spec-lib/test/cmr/umm_spec/test/umm_g/generators.clj
```diff
@@ -26,11 +26,20 @@
                  coords-gen
                  coords-gen))))
 
+(def umm-g-data-granules
```
ygliuvt: We should use the existing generator in umm-lib. This is not creating the correct umm-lib granule model DataGranule record.

umm-spec-lib/src/cmr/umm_spec/umm_g/data_granule.clj
ygliuvt: oh. I saw you comment on the top of the PR stating that this is intentional. I guess hoping Erich would modify the umm-g schema to change the DayNightFlag Enums to match ECHO10. 

Given how easy it is to just code in the mapping, I think we should just fix it right now rather than letting loose ends hanging in the code. We can always remove the mapping code if we decide to modify the casing later.

umm-spec-lib/test/cmr/umm_spec/test/umm_g/generators.clj
ygliuvt: The effort to create this generator is pretty much the same as adding in the umm-lib to UMM-G DayNightFlag mapping. So we should tread this generator with the DayNightFlag mapping. :)

umm-spec-lib/src/cmr/umm_spec/umm_g/data_granule.clj
daniel-zamora: default in echo10 seems to be MB
```                 (when size
                   (x/element :SizeMBDataGranule {} size))
```

umm-spec-lib/test/cmr/umm_spec/test/umm_g/generators.clj
daniel-zamora: Its really a guess as to what we should do, I erred on the side of keeping the current schema.

umm-spec-lib/src/cmr/umm_spec/umm_g/data_granule.clj
daniel-zamora: The same can be said for leaving it and re visiting when Erich returns.  Now I do double to work to reverse it, only to have the chance of reversing it again.

umm-spec-lib/src/cmr/umm_spec/umm_g/data_granule.clj
```diff
@@ -0,0 +1,36 @@
...
+
+(defn umm-g-data-granule->DataGranule
+  "Returns the umm-lib granule model DataGranule from the given UMM-G DataGranule."
+  [data-granule]
+  (when data-granule
+    (g/map->DataGranule
+      (let [{:keys [Identifiers DayNightFlag ProductionDateTime
+                    ArchiveAndDistributionInformation]} data-granule]
+        {:day-night (string/upper-case DayNightFlag)
+         :producer-gran-id (->> Identifiers
```
ygliuvt: I like this better:
```
(some #(when (= "ProducerGranuleId" (:IdentifierType %)) (:Identifer %)) Identifiers)
```

umm-spec-lib/src/cmr/umm_spec/umm_g/data_granule.clj
ygliuvt: Yeah, this is hard to say. I wish we could ask Erich.

PR #596, opened 2018-10-04
CMR-5061: Added support of AccessConstraints in UMM-G.

ygliuvt:


umm-spec-lib/src/cmr/umm_spec/umm_g/granule.clj
```diff
@@ -108,6 +108,7 @@
                                              (str ending-date-time))}}))
      :Platforms (platform/PlatformRefs->umm-g-platforms platform-refs)
      :CloudCover cloud-cover
+     :AccessConstraints (when access-value {:Value access-value})
```
daniel-zamora: I think this when is redundant.

daniel-zamora: I don't think this works.

daniel-zamora: ```:access-value 42``` Not a map.

ygliuvt: This is generating UMM-G `AccessConstraints` which is a map, not umm-lib `access-value`. We don't want to make an empty map when there is no `access-value`, thus the `when`.

PR #599, opened 2018-10-11
CMR-5065 As a Client User, I want to parse and generate measured-parameters in SWOT granule metadata format

daniel-zamora:


umm-spec-lib/src/cmr/umm_spec/umm_g/granule.clj
```diff
@@ -113,7 +115,9 @@
      :Projects (project/ProjectRefs->umm-g-projects project-refs)
      :DataGranule (data-granule/DataGranule->umm-g-data-granule data-granule)
      :TilingIdentificationSystem (tiling-system/TwoDCoordinateSystem->umm-g-tiling-identification-system
-                                   two-d-coordinate-system)
+                                  two-d-coordinate-system)
```
ygliuvt: nit: the indentations are off for this line and the lines below.

umm-spec-lib/src/cmr/umm_spec/umm_g/measured_parameters.clj
```diff
@@ -0,0 +1,101 @@
...
+  [qa-flags]
+  (when qa-flags
+    (let [{:keys [automatic-quality-flag
+                  automatic-quality-flag-explanation
+                  operational-quality-flag
+                  operational-quality-flag-explanation
+                  science-quality-flag
+                  science-quality-flag-explanation]}
+          qa-flags]
+      {:AutomaticQualityFlag automatic-quality-flag
```
ygliuvt: We need to make sure the umm-lib quality flags are converted into valid UMM-G quality flags, i.e. a default ENUM value needs to be picked for umm-lib quality flags that are not in the UMM-G ENUM list; otherwise we will get invalid UMM-G records during conversion.

daniel-zamora: Which one.

daniel-zamora: I meant which value heh.

ygliuvt: Erich should be able to provide the default values.

daniel-zamora: Yes I know, I talked to him yesterday, and you as well so I'm not sure what the purpose of this comment is.

eereiter: I have added "Undetermined" as a new enumeration value to all three: AutomaticQualityFlag, OperationalQualityFlag and ScienceQualityFlag, in the https://git.earthdata.nasa.gov/projects/EMFD/repos/unified-metadata-model/browse/granule/v1.4/umm-g-json-schema.json file.  If the current UMM-G model has nil for an element then DON'T assign the new UMM-G v1.4 a value - leave it nil, they are optional.  If the current UMM-G model has a different value then what exists in the new UMM-G v 1.4 model, then use the Undetermined value.  Please also append to accompanying explanation element the following only if the current UMM-G quality flag has a value: "The original records value of {current UMM-G model quality flag value} was not recognized and the value of Undetermined was assigned." That way when analyzers of the data using UMM-G v1.4 can help the data providers clean up their records. This shouldn't happen though because only ECS DAACs use these elements and the new UMM-G v1.4 model is using the same enumeration names as the BMGT (Bulk Metadata Generation Tool).

umm-spec-lib/test/cmr/umm_spec/test/umm_g/granule.clj
```diff
@@ -3,152 +3,34 @@
...
+    [granule (gen/no-shrink generators/umm-g-granules)]
     (let [granule (sanitizer/sanitize-granule granule)
           metadata (core/generate-metadata {} granule :umm-json)]
-      (empty? (core/validate-metadata :granule :umm-json metadata)))))
+      (is (empty? (core/validate-metadata :granule :umm-json metadata))))))
 
-(defspec generate-and-parse-umm-g-granule-test 100
-  (for-all [granule generators/umm-g-granules]
+(deftest generate-and-parse-umm-g-granule-test
+  (checking "umm-g round tripping" 100
```
ygliuvt: Thanks for switching to the new CMR testing notation.

umm-spec-lib/test/cmr/umm_spec/test/umm_g/granule.clj
```diff
@@ -3,152 +3,34 @@
...
-                     :url "https://webmap.ornl.gov/sdat/pimg/957_1.png"
-                     :description "ISLSCP II EARTH RADIATION BUDGET EXPERIMENT (ERBE) MONTHLY ALBEDO, 1986-1990"
-                     :mime-type "image/png"
-                     :title "ISLSCP II EARTH RADIATION BUDGET EXPERIMENT (ERBE) MONTHLY ALBEDO, 1986-1990"
-                     :size 10})]}))
-
 (deftest parse-granule-test
   (testing "parse granule"
-    (is (= expected-granule (core/parse-metadata {} :granule :umm-json sample-umm-g-granule)))))
+    (is (= expected-util/expected-sample-granule (core/parse-metadata {} :granule :umm-json sample-umm-g-granule)))))
```
ygliuvt: It feels wrong to have the `sample-umm-g-granule` in one namespace and the expected-sample-granule` in a different namespace away from the test.

daniel-zamora: Yea, its how we do it in umm-c generate_and_parse so I figured it was ok here.

PR #601, opened 2018-10-16
CMR-5064: Added support of OrbitCalculatedSpatialDomains in UMM-G.

ygliuvt:

### DIFF
--- a/umm-spec-lib/src/cmr/umm_spec/umm_g/orbit_calculated_spatial_domain.clj
+++ b/umm-spec-lib/src/cmr/umm_spec/umm_g/orbit_calculated_spatial_domain.clj
@@ -1,0 +1,41 @@
+(ns cmr.umm-spec.umm-g.orbit-calculated-spatial-domain
+  "Contains functions for parsing UMM-G JSON OrbitCalculatedSpatialDomains into umm-lib granule model
+  OrbitCalculatedSpatialDomains and generating UMM-G JSON OrbitCalculatedSpatialDomains from umm-lib
+  granule model OrbitCalculatedSpatialDomains."
+  (:require
+   [cmr.umm.umm-granule :as g])
+  (:import cmr.umm.umm_granule.OrbitCalculatedSpatialDomain))
+
+(defn- umm-g-orbit-calculated-spatial-domain->OrbitCalculatedSpatialDomain
+  "Returns the umm-lib granule model OrbitCalculatedSpatialDomain from the given
+  UMM-G OrbitCalculatedSpatialDomain."
+  [ocsd]
+  (g/map->OrbitCalculatedSpatialDomain
+   {:orbital-model-name (:OrbitalModelName ocsd)
+    :orbit-number (:OrbitNumber ocsd)
+    :start-orbit-number (:BeginOrbitNumber ocsd)
+    :stop-orbit-number (:EndOrbitNumber ocsd)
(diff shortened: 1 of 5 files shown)


umm-spec-lib/test/cmr/umm_spec/test/umm_g/sanitizer.clj
```diff
@@ -17,4 +17,5 @@
       (update :platform-refs (fn [x] (when (seq x) (distinct x))))
       (util/update-in-each
        [:platform-refs]
-       #(util/update-in-each % [:instrument-refs] update :operation-modes sanitize-operation-modes))))
+       #(util/update-in-each % [:instrument-refs] update :operation-modes sanitize-operation-modes))
+      (util/update-in-each [:orbit-calculated-spatial-domains] #(assoc % :orbit-number nil))))
```
chris-durbin: Isn't orbit number a valid field in both the umm-lib granule model and UMM-G? I'm guessing there's a rule that it can include either an orbit-number or both a start and stop orbit number.

It seems like we're not testing both scenarios.

sxu123: That's what I'm gonna point out too.  umm-lib allows orbit-number and start, stop orbit numbers to co-exist, but UMM-G doesn't.  So, we may have the invalid data in the system. Also, UMM-G requires start/stop orbit numbers to be integer, before my merge, umm-lib allowed them to be double, so there might be invalid data in the system as well.

ygliuvt: Updated generator to generate the record according to UMM-G constraint. I don't want to mask the data problem during conversion. The bad data needs to be fixed through granule update. I would rather deal with it later when the need becomes apparent, but I did add in a simple double to int conversion to deal with the orbit number type change that Siwei made.

umm-spec-lib/src/cmr/umm_spec/umm_g/orbit_calculated_spatial_domain.clj
```diff
@@ -0,0 +1,43 @@
...
+  (when (seq orbit-calculated-spatial-domains)
+    (for [ocsd orbit-calculated-spatial-domains]
+      (let [{:keys [orbital-model-name orbit-number start-orbit-number stop-orbit-number
+                    equator-crossing-longitude equator-crossing-date-time]} ocsd]
+        {:OrbitalModelName orbital-model-name
+         :OrbitNumber orbit-number
+         ;; Added conversion to int to deal with potential invalid data from other formats,
+         ;; the start-orbit-number and stop-orbit-number were used to be double before.
+         :BeginOrbitNumber (when start-orbit-number (int start-orbit-number))
+         :EndOrbitNumber (when stop-orbit-number (int stop-orbit-number))
```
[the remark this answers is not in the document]
sxu123: Talked with Leo about this. We don't validate UMM-G after the translation from umm-lib to UMM-G, it's an issue but not directly related to this ticket.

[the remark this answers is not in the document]
ygliuvt: Erich told me to throw in the int conversion here. The schema validation code is already in place, but I can add a test for orbit numbers when I work on the granule validation issue which deals with granule business rule validation.

PR #603, opened 2018-10-18
CMR-5184-5185:  Added support for feature ids and crid ids.

sxu123:
The PR addressed:
1. Added support for grid-ids and feature-ids in old umm-g and new UMM-G v1.4 
2. Modified the generator code and the related tests so that the round trip of umm->UMM-G->umm
    tests could pass.

Note: Currently we don't support these fields in echo10 and iso_smap, so there is no need to parse them, yet. Also, the acceptance criteria says only to make the umm to UMM-G v1.4 round trip pass. So, this PR only covers that.  If we need to search for the granules using the crid-id and feature-id, I will have another PR to cover that.

### DIFF
--- a/umm-lib/test/cmr/umm/test/echo10/granule.clj
+++ b/umm-lib/test/cmr/umm/test/echo10/granule.clj
@@ -38,7 +38,9 @@
     (let [xml (echo10/umm->echo10-xml granule)
           parsed (g/parse-granule xml)
           expected-parsed (umm->expected-parsed-echo10 granule)]
-      (= parsed expected-parsed))))
+      ;; Remove crid-ids and feature-ids because they are not supported in echo10 
+      (= (update-in parsed [:data-granule] dissoc :crid-ids :feature-ids)
+         (update-in expected-parsed [:data-granule] dissoc :crid-ids :feature-ids)))))
 
 (def all-fields-granule-xml
   "<Granule>
--- a/umm-lib/test/cmr/umm/test/iso_smap/granule.clj
+++ b/umm-lib/test/cmr/umm/test/iso_smap/granule.clj
@@ -93,7 +93,9 @@
     (let [xml (iso/umm->iso-smap-xml granule)
           parsed (g/parse-granule xml)
           expected-parsed (umm->expected-parsed-smap-iso granule)]
-      (= parsed expected-parsed))))
+      ;; Remove crid-ids and feature-ids because they are not supported in iso-smap.
+      (= (update-in parsed [:data-granule] dissoc :crid-ids :feature-ids)
+         (update-in expected-parsed [:data-granule] dissoc :crid-ids :feature-ids)))))
 
 (def sample-granule-xml
   (slurp (io/file (io/resource "data/iso_smap/sample_smap_iso_granule.xml"))))
--- a/umm-spec-lib/src/cmr/umm_spec/umm_g/data_granule.clj
+++ b/umm-spec-lib/src/cmr/umm_spec/umm_g/data_granule.clj
@@ -18,6 +18,16 @@
          :producer-gran-id (->> Identifiers
                                 (some #(when (= "ProducerGranuleId" (:IdentifierType %)) %))
                                 :Identifier)
+         :crid-ids (->> Identifiers
+                        (filter #(= "CRID" (:IdentifierType %)))
+                        (map :Identifier)
+                        distinct
+                        seq) 
+         :feature-ids (->> Identifiers
+                        (filter #(= "FeatureId" (:IdentifierType %)))    
+                        (map :Identifier)
+                        distinct
+                        seq)
          :production-date-time ProductionDateTime
          :size (get (first ArchiveAndDistributionInformation) :Size)}))))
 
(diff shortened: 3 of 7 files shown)


umm-lib/test/cmr/umm/test/iso_smap/granule.clj
```diff
@@ -93,7 +93,9 @@
     (let [xml (iso/umm->iso-smap-xml granule)
           parsed (g/parse-granule xml)
           expected-parsed (umm->expected-parsed-smap-iso granule)]
-      (= parsed expected-parsed))))
+      ;; Remove crid-ids and feature-ids because they are not supported in iso-smap.
+      (= (update-in parsed [:data-granule] dissoc :crid-ids :feature-ids)
```
daniel-zamora: Can we break this out into a function.

umm-lib/test/cmr/umm/test/iso_smap/granule.clj
```diff
@@ -93,7 +93,9 @@
...
+         (update-in expected-parsed [:data-granule] dissoc :crid-ids :feature-ids)))))
```
daniel-zamora: Same.

umm-spec-lib/test/cmr/umm_spec/test/umm_g/granule.clj
```diff
@@ -25,7 +25,15 @@
           actual (core/parse-metadata {} :granule :umm-json umm-g-metadata)
           expected (expected-util/umm->expected-parsed granule)
           actual (update-in actual [:spatial-coverage :geometries] set)
-          expected (update-in expected [:spatial-coverage :geometries] set)]
+          expected (update-in expected [:spatial-coverage :geometries] set)
```
daniel-zamora: Can we pull all this extra processing into its own function we can add to, to keep the test clean and readable.

daniel-zamora: the set logic could also be included.

umm-lib/test/cmr/umm/test/echo10/granule.clj
```diff
@@ -38,7 +43,9 @@
     (let [xml (echo10/umm->echo10-xml granule)
           parsed (g/parse-granule xml)
           expected-parsed (umm->expected-parsed-echo10 granule)]
-      (= parsed expected-parsed))))
+      ;; Remove crid-ids and feature-ids because they are not supported in echo10 
+      (= (remove-crid-ids-and-feature-ids parsed)
```
chris-durbin: We should be making all modifications in `umm->expected-parsed-echo10`.

umm-lib/test/cmr/umm/test/iso_smap/granule.clj
```diff
@@ -93,7 +94,9 @@
     (let [xml (iso/umm->iso-smap-xml granule)
           parsed (g/parse-granule xml)
           expected-parsed (umm->expected-parsed-smap-iso granule)]
-      (= parsed expected-parsed))))
+      ;; Remove crid-ids and feature-ids because they are not supported in iso-smap.
```
chris-durbin: Same comment for `umm->expected-parsed-smap-iso`

umm-spec-lib/src/cmr/umm_spec/umm_g/data_granule.clj
```diff
@@ -18,18 +18,35 @@
          :producer-gran-id (->> Identifiers
                                 (some #(when (= "ProducerGranuleId" (:IdentifierType %)) %))
                                 :Identifier)
+         :crid-ids (->> Identifiers
+                        (filter #(= "CRID" (:IdentifierType %)))
+                        (map :Identifier)
+                        seq)
```
chris-durbin: I don't think we need to call `seq` here or for `:feature-ids` below.

umm-spec-lib/src/cmr/umm_spec/umm_g/data_granule.clj
```diff
@@ -18,18 +18,35 @@
...
+    (let [{:keys [producer-gran-id crid-ids feature-ids
+                  day-night production-date-time size]} data-granule]
       {:DayNightFlag (string/capitalize day-night)
-       :Identifiers (when producer-gran-id
-                      [{:Identifier producer-gran-id
-                        :IdentifierType "ProducerGranuleId"}])
+       :Identifiers (let [producer-gran-id (when producer-gran-id
+                                             [{:Identifier producer-gran-id
+                                               :IdentifierType "ProducerGranuleId"}])
+                          crid-ids (when (seq crid-ids)
```
chris-durbin: No need for the call to `when`.

umm-spec-lib/src/cmr/umm_spec/umm_g/data_granule.clj
```diff
@@ -18,18 +18,35 @@
...
+                                     (map #(assoc {:IdentifierType "CRID"} :Identifier %) 
+                                          (distinct crid-ids)))
+                          feature-ids (when (seq feature-ids)
```
chris-durbin: No need for the call to `when`

umm-spec-lib/src/cmr/umm_spec/umm_g/data_granule.clj
sxu123: I think they are needed because otherwise it's possible that crid-ids would be [] when Identifiers don't have CRIDs. and we want crid-ids to be nil in that case. Same with feature-ids.

umm-spec-lib/src/cmr/umm_spec/umm_g/data_granule.clj
sxu123: Right, since I did check on (seq ) in the end... Done.

umm-spec-lib/src/cmr/umm_spec/umm_g/data_granule.clj
chris-durbin: You are right. I was thinking that a later step called `seq`, but there is no later step - this is the step that creates the model.

umm-lib/test/cmr/umm/test/echo10/granule.clj
sxu123: Moved it to umm->expected-parsed-echo10 for expected-parsed. But for parsed, we still need to call the function because it's possible that parsed contains :crid-ids nil :feature-ids nil.

PR #610, opened 2018-10-23
CMR-5143: Add citations to opendata response.

CayvonH:
* Add collection citations fields.
* Creator, Editor, SeriesName, ReleasePlace, IssueIdentification, DataPresentationForm have their own box.
* Citation is the result of concatting the citation fields in the order needed to present full citation information.
* When creating citation field SeriesName is chosen over Title if it exists. DOI is chosen over OnlineResource/Linkage. Publisher is chosen over archive-center.
* The publisher field will be appended to either the NASA publication name or the USGS publication name.

Format of citation is as follows:
CREATOR. EDITOR. RELEASE-DATE. SERIES-NAME OR TITLE. "Version" VERSION. RELEASE-PLACE. ISSUE-IDENTIFICATION. "Archived by " (NASA OR USGS PUBLICATION NAME), PUBLISHER OR ARCHIVE-CENTER. DOI OR ONLINE-RESOURCE/LINKAGE. DATA-PRESENTATION-FORM. OTHER-CITATION-DETAILS.

Example:
collection citation


search-app/src/cmr/search/results_handlers/opendata_results_handler.clj
```diff
@@ -243,18 +255,80 @@
     time
     gran-time))
 
+(defn- title->citation-title
+  "Format citation name."
+  [title version]
+  (when title
```
chris-durbin: Some alternatives - I'm thinking the one at the bottom might be the most readable?
```
  (when title
    (let [version-string (if version (format ". Version %s" version) "")
      (format "%s%s" title version-string))
```
or
```
  (when title
    (let [version-string (if version (str ". Version " version) "")
      (str title version-string))
```

search-app/src/cmr/search/results_handlers/opendata_results_handler.clj
```diff
@@ -243,18 +255,80 @@
...
+    (str title
+         (when version
+           (str ". Version " version)))))
+
+(defn- publisher->citation-publisher
+  "Format publisher for citation."
+  [archive-center provider-id]
+  (when archive-center
+    (str "Archived by "
```
chris-durbin:
```
(let [publication-name (if (= "USGS_EROS" provider-id)
                         USGS_PUBLICATION_NAME
                         NASA_PUBLICATION_NAME)]
  (format "Archived by %s, %s" publication-name archive-center))
```

eereiter:
This looks a lot better.  Oh I see now why you only use Series or Title, because I put the title piece in the wrong place. OK my mistake, I meant for the Citation to contain both the series and the title when both exist 
If you have a Series name do the following: 
  SeriesName followed by a period followed by "Version " followed by Version followed by a period, followed by Title followed by a period.
If SeriesName does not exist then use Title followed by a period, "Version " followed by Version followed by a period

Is there a reason why we are choosing DOI over OnlineResource?  Can we have both?

search-app/src/cmr/search/results_handlers/opendata_results_handler.clj
```diff
@@ -243,18 +255,80 @@
...
+         (if (= provider-id "USGS_EROS")
```
chris-durbin: We should have a variable defined rather than using "USGS_EROS". I think there's another place in this file that checks for the same string.
We should define this as a variable (and I think there's somewhere else in the code that checks for USGS as well).
```
(def USGS_PROVIDER "The USGS provider ID." "USGS_EROS")
```

Is LPDAAC_ECS also a USGS provider rather than NASA? Are there any other providers that we shouldn't be treating as NASA? We might need to talk to Scott, Erich or Chris Lynnes.

search-app/src/cmr/search/results_handlers/opendata_results_handler.clj
```diff
@@ -243,18 +255,80 @@
...
+  (when doi
+    (doi/doi->url doi)))
+
+(defn- release-date->citation-release-date
+  "Format release-date for citation"
+  [release-date]
+  (when release-date
+    (first (string/split release-date #"T"))))
+
+(def ^:private citation-order
```
chris-durbin: Really nice.

CayvonH:
> This looks a lot better. Oh I see now why you only use Series or Title, because I put the title piece in the wrong place. OK my mistake, I meant for the Citation to contain both the series and the title when both exist
> If you have a Series name do the following:
> SeriesName followed by a period followed by "Version " followed by Version followed by a period, followed by Title followed by a period.
> If SeriesName does not exist then use Title followed by a period, "Version " followed by Version followed by a period
> 
> Is there a reason why we are choosing DOI over OnlineResource? Can we have both?

We can add both. I was thinking we could end up with two of the same URLs but now that I think about it that doesn't seem likely.

PR #611, opened 2018-10-23
CMT-5139: Add graphic to opendata response.

CayvonH:
* Get browse images from RelatedUrls.
* Add graphic-preview-file from browse image URL.
* Add graphic-preview-description from browse image description.
* Add graphic-preview-type from either the RelatedUrl/GetData/Format or RelatedUrl/GetData/MimeType.

On data.gov the Additional Metadata will now display the fields as:
graphic_meta

The image will show as such:
graphic_image

### DIFF
--- a/indexer-app/src/cmr/indexer/data/concepts/collection/opendata.clj
+++ b/indexer-app/src/cmr/indexer/data/concepts/collection/opendata.clj
@@ -32,17 +32,15 @@
   "Returns the opendata related url for the given collection related url"
   [related-url]
   (let [{:keys [Description Type Subtype URL]} related-url
-        Format (get-in related-url [:GetData :Format])
         get-data-mime-type (get-in related-url [:GetData :MimeType])
-        MimeType (get-in related-url [:GetService :MimeType])
+        get-service-mime-type (get-in related-url [:GetService :MimeType])
         size (get-in related-url [:GetData :Size])]
     {:type Type
      :sub-type Subtype
      :url URL
-     :format Format
      :get-data-mime-type get-data-mime-type
      :description Description
-     :mime-type MimeType
+     :get-service-mime-type get-service-mime-type
      :size size}))
 
 (defn publication-reference->opendata-reference
--- a/indexer-app/src/cmr/indexer/data/concepts/collection/opendata.clj
+++ b/indexer-app/src/cmr/indexer/data/concepts/collection/opendata.clj
@@ -32,11 +32,15 @@
   "Returns the opendata related url for the given collection related url"
   [related-url]
   (let [{:keys [Description Type Subtype URL]} related-url
+        Format (get-in related-url [:GetData :Format])
+        get-data-mime-type (get-in related-url [:GetData :MimeType])
         MimeType (get-in related-url [:GetService :MimeType])
         size (get-in related-url [:GetData :Size])]
     {:type Type
      :sub-type Subtype
      :url URL
+     :format Format
+     :get-data-mime-type get-data-mime-type
      :description Description
      :mime-type MimeType
      :size size}))
--- a/search-app/src/cmr/search/results_handlers/opendata_results_handler.clj
+++ b/search-app/src/cmr/search/results_handlers/opendata_results_handler.clj
@@ -331,17 +331,25 @@
                                     not-empty)]
      (str (string/join ". " citation-details) "."))))
 
-(defn- nil-if-not-provided
-  "nil if not provided."
-  [s]
-  (when-not (= umm-spec-util/not-provided s)
-    s))
+(defn- score-browse-image-related-url
+  "Score the related-url based off number of browse-image fields."
+  [related-url]
+  (->> (select-keys related-url [:get-data-mime-type :description])
+       vals
+       (remove #(= umm-spec-util/not-provided %))
+       count))
+
+(defn get-best-browse-image-related-url
+  "Get the related-url that is a browse image with the most graphic-preview fields."
+  [related-urls]
+  (->> (ru/browse-urls related-urls)
+       (sort-by score-browse-image-related-url)
+       last))
 
 (defn graphic-preview-type
-  "Use format or mime type if provided."
-  [related-url]
-  (or (nil-if-not-provided (:format related-url))
-      (nil-if-not-provided (:get-data-mime-type related-url))))
+  "Nil if browse image mime type is Not provided."
+  [browse-image-related-url]
+  (util/nil-if-value umm-spec-util/not-provided (:get-data-mime-type browse-image-related-url)))
 
 (defn- result->opendata
   "Converts a search result item to opendata."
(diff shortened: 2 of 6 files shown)


search-app/src/cmr/search/results_handlers/opendata_results_handler.clj
```diff
@@ -331,6 +331,18 @@
                                     not-empty)]
      (str (string/join ". " citation-details) "."))))
 
+(defn- nil-if-not-provided
```
mschmele: Nitpick: This is a good candidate for `cmr.common.util` (included above as `util`). It's kind of the opposite of `inflate-nil-keys`

https://github.com/nasa/Common-Metadata-Repository/blob/master/common-lib/src/cmr/common/util.clj#L279

indexer-app/src/cmr/indexer/data/concepts/collection/opendata.clj
```diff
@@ -32,11 +32,15 @@
   "Returns the opendata related url for the given collection related url"
   [related-url]
   (let [{:keys [Description Type Subtype URL]} related-url
+        Format (get-in related-url [:GetData :Format])
+        get-data-mime-type (get-in related-url [:GetData :MimeType])
         MimeType (get-in related-url [:GetService :MimeType])
```
chris-durbin: Nit: the inconsistencies in case are bugging me. I think we should use Clojure convention outside of destructuring. So the fields in line 34 are ok, but `Format` should be `get-data-format` and `MimeType` below should be `get-service-mime-type`.

indexer-app/src/cmr/indexer/data/concepts/collection/opendata.clj
```diff
@@ -32,11 +32,15 @@
...
         size (get-in related-url [:GetData :Size])]
     {:type Type
      :sub-type Subtype
      :url URL
+     :format Format
+     :get-data-mime-type get-data-mime-type
      :description Description
      :mime-type MimeType
```
chris-durbin: Since we added another mime-type to the map we should probably change this key to `:get-service-mime-type`.

search-app/src/cmr/search/results_handlers/opendata_results_handler.clj
```diff
@@ -331,6 +331,18 @@
...
+  "nil if not provided."
+  [s]
+  (when-not (= umm-spec-util/not-provided s)
+    s))
+
+(defn graphic-preview-type
+  "Use format or mime type if provided."
+  [related-url]
+  (or (nil-if-not-provided (:format related-url))
```
chris-durbin: Isn't the open data format looking for a MIME type? 

If so - Is Format already a MIME type, or do we need to do some conversion from the Format string to a MIME type?

Similarly I'm thinking that MIME type should also be our first choice if open data is looking for a MIME type here.

@eereiter - what is your take?

search-app/src/cmr/search/results_handlers/opendata_results_handler.clj
```diff
@@ -343,7 +355,8 @@
         issued-time (when-not (= issued-time (str umm-spec-date-util/parsed-default-date))
                       issued-time)
         modified-time (get-issued-modified-time update-time granule-end-date)
-        collection-citation (first collection-citations)]
+        collection-citation (first collection-citations)
+        browse-image-related-url (first (ru/browse-urls related-urls))]
```
chris-durbin: Minor:
Are we guaranteed that every browse URL will have all three of the the graphic preview fields? If not we might want to choose the URL to use based on how many fields they contain rather than just choosing the first one.

system-int-test/test/cmr/system_int_test/search/collection_search_format_test.clj
```diff
@@ -729,10 +729,16 @@
...
+      (testing "Opendata fields in response."
         (are3 [expected-result field-key opendata-test-collection]
           (is (= expected-result (field-key opendata-test-collection)))
 
+          "graphic-preview-file in response"
+          "https://docserver.gesdisc.eosdis.nasa.gov/public/project/Images/AIRX3STD_006.png" :graphic-preview-file opendata-coll-5138-1
+
+          "graphic-preview-description in response"
+          "Sample data of the \"AIRS/Aqua Level 3 daily standard physical retrieval product (Without HSB)\"." :graphic-preview-description opendata-coll-5138-1
+
```
chris-durbin: What about graphic preview type?

search-app/src/cmr/search/results_handlers/opendata_results_handler.clj
```diff
@@ -331,6 +331,18 @@
                                     not-empty)]
      (str (string/join ". " citation-details) "."))))
 
+(defn- nil-if-not-provided
+  "nil if not provided."
+  [s]
+  (when-not (= umm-spec-util/not-provided s)
+    s))
+
+(defn graphic-preview-type
```
chris-durbin: If we don't have a graphic preview type, will it still render on data.gov? e.g. Are they figuring out the type based on the extension in the URL provided if they can?

search-app/src/cmr/search/results_handlers/opendata_results_handler.clj
CayvonH: The preview type doesn't matter. It will display on data.gov based off the file extension i believe. The graphic-preview-type is just extra information.

search-app/src/cmr/search/results_handlers/opendata_results_handler.clj
CayvonH: Yes it will display regardless of what graphic-preview-type is set to.

search-app/src/cmr/search/results_handlers/opendata_results_handler.clj
eereiter: Yes , I agree.  Format is not the MimeType.  For example JPEG could be the format, but the mime type would be image/jpeg. Most of the time there is a direct mapping, but not all of the time.

CayvonH: So for the graphic-preview-type would the mime-type be the correct to display?

eereiter: yes

search-app/src/cmr/search/results_handlers/opendata_results_handler.clj
eereiter: No we are not guaranteed to have all three fields.

eereiter: ISO records won't have the mime-type.

system-int-test/test/cmr/system_int_test/search/collection_search_format_test.clj
```diff
@@ -715,24 +716,48 @@
...
 
+          "graphic-preview-file in response"
+          "http://example.com/browse-image" :graphic-preview-file opendata-coll-umm
+
+          "graphic-preview-type in response"
+          "image/png" :graphic-preview-type opendata-coll-umm
+
+          "graphic-preview-description in response"
+          "A test related url description." :graphic-preview-description opendata-coll-umm
+
```
CayvonH: I suppose I should have asked this earlier, but right now the only reason type and description and no other fields were selected to be displayed was because I modeled this off of the example in the ticket. There is no reason we couldn't add more fields if desired. Is this a good set of details about the graphic preview or is there any other information from the RelatedUrl you would like added? @eereiter @chris-durbin

PR #613, opened 2018-10-25
CMR-5186: Added support to search granules by feature-id and crid-id.

ygliuvt:

### DIFF
--- a/system-int-test/src/cmr/system_int_test/data2/granule.clj
+++ b/system-int-test/src/cmr/system_int_test/data2/granule.clj
@@ -100,13 +100,15 @@
 (defn data-granule
   "Returns a data-granule with the given attributes"
   [attribs]
-  (let [{:keys [producer-gran-id day-night size production-date-time]} attribs]
-    (when (or size producer-gran-id day-night production-date-time)
+  (let [{:keys [producer-gran-id day-night size production-date-time feature-ids crid-ids]} attribs]
+    (when (or size producer-gran-id day-night production-date-time feature-ids crid-ids)
       (g/map->DataGranule {:producer-gran-id producer-gran-id
                            :day-night day-night
                            :production-date-time (or production-date-time
                                                      (p/parse-datetime "2010-01-01T12:00:00Z"))
-                           :size size}))))
+                           :size size
+                           :feature-ids feature-ids
+                           :crid-ids crid-ids}))))
 
 (defn orbit-calculated-spatial-domain
   "Returns an orbit-calculated-spatial-domain with the given attributes"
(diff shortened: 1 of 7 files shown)


ygliuvt:
> Will we need to reindex granules? I'm assuming we won't have to because no existing metadata has feature-id or crid-id, it will only be possible once users start to ingest UMM-G as json metadata.

That's correct. No existing granules would have feature ids or crid ids.

system-int-test/test/cmr/system_int_test/search/granule_search_feature_id_crid_id_test.clj
```diff
@@ -0,0 +1,102 @@
...
+
+        ;; pattern
+        [gran1 gran2 gran4 gran5] "feature*" {:pattern true}
+        [gran1 gran2 gran4] "feature?" {:pattern true}
+
+        ;; and option
+        [gran4] ["feature1" "feature2"] {:and true}))
+
+    (testing "search by crid-id"
+      (are [items crid-ids options]
```
daniel-zamora: We shouldn't write new code using `are`, we should use `are3`.

system-int-test/src/cmr/system_int_test/data2/granule.clj
```diff
@@ -100,13 +100,15 @@
 (defn data-granule
   "Returns a data-granule with the given attributes"
   [attribs]
-  (let [{:keys [producer-gran-id day-night size production-date-time]} attribs]
-    (when (or size producer-gran-id day-night production-date-time)
+  (let [{:keys [producer-gran-id day-night size production-date-time feature-ids crid-ids]} attribs]
```
daniel-zamora: Does `when-let` work for this?

system-int-test/test/cmr/system_int_test/search/granule_search_feature_id_crid_id_test.clj
ygliuvt: Not getting that memo. :) Felt `are3` would be too verbose in this case. I can change to use `are3` if there is another downvote on the `are` here.

system-int-test/src/cmr/system_int_test/data2/granule.clj
ygliuvt: It will not. Check this out in repl:
```
(when-let [{:keys [a b]} {}] (println "don't want this to run"))
```

system-int-test/test/cmr/system_int_test/search/granule_search_feature_id_crid_id_test.clj
daniel-zamora: `are3` is more maintainable than `are` but only marginally.  During my time at CMR I've had instances where I was asked to convert tests I didn't write initially, as well as tests that I had initially written with `are`.  I don't care too much either way, I'm just passing along feedback that I've been given in the past.

ygliuvt: I agree that `are3` are better in most cases and it gives out the exact test case when the test fails which is really good. But in this case the tests are so trivial that it would make it actually more unreadable to use `are3`. No comments from other people, so I am going to merge it now.

PR #615, opened 2018-10-26
Cmr 5110

eereiter:
Fixing CMR-5110. I had to create a new UMM version 1.11. Updating the RelatedUrl URLContentType, Type, and Subtype to match the version 8.6 Keywords in the KMS for related urls.  Also fixing the translations between the supported specifications. This has been a problem for several months and the international community has started to complain because our tools are providing conflicting error during metadata curation.


umm-spec-lib/test/cmr/umm_spec/test/migration/version/collection.clj
```diff
@@ -1680,3 +1680,55 @@
...
+                                     :URLContentType "DistributionURL",
+                                     :Type "GET DATA"},
+                                    {:Subtype "GENERAL DOCUMENTATION",
+                                     :URLContentType "PublicationURL",
+                                     :Type "VIEW RELATED INFORMATION"},
+                                    {:Subtype "Earthdata Search",
+                                     :URLContentType "DistributionURL",
+                                     :Type "GET DATA"}]}
+        result (vm/migrate-umm {} :collection "1.11" "1.10" related-urls)]
+    (println result)
```
daniel-zamora: remove debug code.

eereiter: Fixed and pushed.

system-int-test/test/cmr/system_int_test/ingest/translation_test.clj
```diff
@@ -220,32 +220,3 @@
                                            :collection :dif dif9-xml :umm-json
                                            {:query-params {"skip_umm_validation" "true"}})]
       (is (= 200 status)))))
-
-(comment
```
chris-durbin: I think this is one of those comment blocks that is really useful to have around for debugging UMM issues. Check with Leo and Siwei before removing this.

ygliuvt: I have not used this comment block code personally.

umm-spec-lib/src/cmr/umm_spec/migration/version/collection.clj
```diff
@@ -238,8 +238,18 @@
...
+      (update-in-each [:CollectionCitations] related-url/migrate-online-resource-down)
       (assoc :UseConstraints (when-let [description (get-in c [:UseConstraints :Description])]
                                ;; Description in 1.10 is object/record.
                                ;; It needs to be converted to string when becoming UseConstraints in 1.9.i
                                (:Description description)))))
+
+(defmethod interface/migrate-umm-version [:collection "1.10" "1.11"]
+  [context c & _]
+  (-> c
+    (related-url/migrate-up-to-1_11)))
```
daniel-zamora: alignment is off for this, needs to be vertically aligned.

umm-spec-lib/src/cmr/umm_spec/migration/related_url_migration.clj
```diff
@@ -251,3 +336,46 @@
...
+   to GCMD Related URL Keyword updates 8.6, so we need to translate the values back to the old
+   8.5 keyword version"
+  [collection]
+  (if (nil? (:RelatedUrls collection))
+    collection
+    (assoc collection :RelatedUrls
+      (into []
+       (:RelatedUrls
+        (-> collection
+          (update-in-each [:RelatedUrls] replace-existing-related-url-keywords "1.10")))))))
```
daniel-zamora: Thread macros should be vertically aligned, I would go through all your uses and change them.  Maybe in the cases where there is only one form being used i the thread, just remove them entirely.  It also looks like in this case should thread from assoc all the way down.

umm-spec-lib/src/cmr/umm_spec/migration/version/collection.clj
daniel-zamora: Probably don't need to use a thread here anyway.

daniel-zamora: FYI, we try to use https://github.com/bbatsov/clojure-style-guide

umm-spec-lib/src/cmr/umm_spec/test/umm_record_sanitizer.clj
```diff
@@ -155,14 +243,14 @@
 
 (defn- sanitize-umm-record-urls
   "Sanitize all RelatedUrls throughout the collection by making them url strings."
-  [record]
+  [record concept]
```
daniel-zamora: Your use of `concept` here probably is not consistent with other places is cmr.  It should be concept-type.

umm-spec-lib/src/cmr/umm_spec/migration/version/collection.clj
eereiter: done

umm-spec-lib/src/cmr/umm_spec/test/umm_record_sanitizer.clj
daniel-zamora: I think this code could probably be cleaner and clearer if we used defmulti for our url-content-type mappings.  Not necessary here because it would require significant time for refactoring.  Maybe just add a comment saying we should consolidate and simplify this.  For now, changing `concept` to `concept-type` is fine.

umm-spec-lib/test/cmr/umm_spec/test/migration/version/collection.clj
```diff
@@ -1680,3 +1683,52 @@
...
+                                         :URI ["http://www.foo.com", "http://www.bar.com"]}}
+                           {:Description "Related url 2 description"
+                            :URL "http://www.foo.com"
+                            :URLContentType "VisualizationURL"
+                            :Type "GET RELATED VISUALIZATION"
+                            :Subtype "GIBS"}]))))
+
+(deftest migrate-1-10-to-1-11
+  (let [result (vm/migrate-umm {} :collection "1.10" "1.11" related-urls-UMM-1-10-example)]
+    (is (= (remove-empty-maps exp-conv/example-collection-record)
```
daniel-zamora: I'm a little worried about these remove-empty-maps calls. Maybe we should pair and work through this to see why these are needed, and if there is a better solution.

umm-spec-lib/src/cmr/umm_spec/migration/related_url_migration.clj
eereiter: done.

umm-spec-lib/src/cmr/umm_spec/test/umm_record_sanitizer.clj
eereiter: Thanks and done.

umm-spec-lib/src/cmr/umm_spec/migration/related_url_migration.clj
```diff
@@ -251,3 +336,46 @@
...
+     (assoc related-url :URLContentType (:URLContentType result)
+                        :Type (:Type result)
+                        :Subtype (:Subtype result)))))
+
+(defn migrate-up-to-1_11
+  ":RelatedUrl Type and Subtypes have changed in UMM-C version 1.11 to correspond
+   to GCMD Related URL Keyword updates 8.6."
+  [collection]
+  (def cc collection)
+  (if (nil? (:RelatedUrls collection))
```
ygliuvt: This `if` block should be able to be changed to something like below:
```
(when (:RelatedUrls collection)
  (update-in-each collection [:RelatedUrls] replace-existing-related-url-keywords "1.11"))
```

umm-spec-lib/src/cmr/umm_spec/migration/related_url_migration.clj
```diff
@@ -251,3 +336,46 @@
...
+       (:RelatedUrls
+        (-> collection
+          (update-in-each [:RelatedUrls] replace-existing-related-url-keywords "1.11")))))))
+
+(defn migrate-down-from-1_11
+  ":RelatedUrl Type and Subtypes have changed in UMM-C version 1.11 to correspond
+   to GCMD Related URL Keyword updates 8.6, so we need to translate the values back to the old
+   8.5 keyword version"
+  [collection]
+  (if (nil? (:RelatedUrls collection))
```
ygliuvt: Should be able to change to `when` similar to my comment on the migrate up function.

umm-spec-lib/src/cmr/umm_spec/test/umm_record_sanitizer.clj
```diff
@@ -35,13 +35,101 @@
...
+                                                 "RECIPE"
+                                                 "REQUIREMENTS AND DESIGN"
+                                                 "SCIENCE DATA PRODUCT SOFTWARE DOCUMENTATION"
+                                                 "SCIENCE DATA PRODUCT VALIDATION"
+                                                 "USER FEEDBACK"
+                                                 "USER'S GUIDE"]}
+   "DataCenterURL" {"HOME PAGE" []}
+   "DataContactURL" {"HOME PAGE" []}})
+
+(defn valid-types-for-url-content-type
```
ygliuvt: Use `defn-` to define private functions that are only called within the same namespace. This applies to other functions in this PR as well.

umm-spec-lib/test/cmr/umm_spec/test/migration/version/collection.clj
eereiter: Thanks for paring with me and finding a better solution.

umm-spec-lib/src/cmr/umm_spec/migration/related_url_migration.clj
```diff
@@ -251,3 +336,46 @@
...
+    (remove-nil-keys
+     (assoc related-url :URLContentType (:URLContentType result)
+                        :Type (:Type result)
+                        :Subtype (:Subtype result)))))
+
+(defn migrate-up-to-1_11
+  ":RelatedUrl Type and Subtypes have changed in UMM-C version 1.11 to correspond
+   to GCMD Related URL Keyword updates 8.6."
+  [collection]
+  (def cc collection)
```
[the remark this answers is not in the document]
eereiter: done

umm-spec-lib/src/cmr/umm_spec/migration/related_url_migration.clj
eereiter: I can't.  This function needs to return a collection.  If the when block is false nil is passed back instead of the actual collection.

umm-spec-lib/src/cmr/umm_spec/migration/related_url_migration.clj
eereiter: I can't.  This function needs to return a collection.  If the when block is false nil is passed back instead of the actual collection.

umm-spec-lib/src/cmr/umm_spec/test/umm_record_sanitizer.clj
eereiter: done

PR #616, opened 2018-10-30
CMR-5260: Added tests for UMM-G granule validation endpoint.

ygliuvt:

### DIFF
--- a/system-int-test/test/cmr/system_int_test/ingest/granule_validation_test.clj
+++ b/system-int-test/test/cmr/system_int_test/ingest/granule_validation_test.clj
@@ -328,55 +328,50 @@
             :errors errors}
            (select-keys response [:status :errors])))))
 
-;; This tests that UMM type validations are applied during collection ingest.
-;; Thorough tests of UMM validations should go in cmr.umm.test.validation.core and related
-;; namespaces.
-(deftest granule-umm-validation-test
-  (testing "Spatial validation"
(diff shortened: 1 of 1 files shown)


system-int-test/test/cmr/system_int_test/ingest/granule_validation_test.clj
```diff
@@ -431,3 +428,72 @@
...
+  ;; (e.g. application/vnd.nasa.cmr.umm+json; version=1.4)
+  ;; So we execute curl command directly from shell for UMM-G granule validation requests
+  (try
+    (spit coll-file (:metadata coll-concept))
+    (spit granule-file (:metadata granule-concept))
+    (let [{:keys [provider-id native-id]} granule-concept
+          validate-url (url/validate-url provider-id :granule native-id)
+          coll-form (format "collection=<%s;type=%s" coll-file (:format coll-concept))
+          granule-form (format "granule=<%s;type=%s" granule-file (:format granule-concept))
+          {:keys [out]} (shell/sh "curl" "-i" "-F" granule-form "-F" coll-form validate-url)]
```
chris-durbin: This adds a dependency on `curl` for our tests, but I guess that's ok. We know that our CI environment has curl installed.

ygliuvt: Yep, we are already using curl is some other tests.

PR #618, opened 2018-11-02
CMR-5166: As a Client User, I want to retrieve granules in UMM-G format.

ygliuvt:

### DIFF
--- a/system-int-test/test/cmr/system_int_test/search/granule_retrieval_test.clj
+++ b/system-int-test/test/cmr/system_int_test/search/granule_retrieval_test.clj
@@ -3,6 +3,7 @@
   (:require
     [clojure.test :refer :all]
     [cmr.common.mime-types :as mt]
+    [cmr.common.util :refer [are3]]
     [cmr.system-int-test.data2.core :as d]
     [cmr.system-int-test.data2.granule :as dg]
     [cmr.system-int-test.data2.umm-spec-collection :as data-umm-c]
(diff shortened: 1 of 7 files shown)


system-int-test/test/cmr/system_int_test/search/granule_retrieval_test.clj
```diff
@@ -35,17 +45,85 @@
...
+          (is (= (:metadata umm-g-gran) (:body response))))
+
+        "without specifying accept format"
+        nil
+
+        "specifying umm json accept format without version"
+        "application/vnd.nasa.cmr.umm+json"
+
+        "specifying umm json accept format with version"
+        "application/vnd.nasa.cmr.umm+json;version=1.4"))
```
chris-durbin: Do we already have tests that verify all potential versions can be returned successfully? It seems like that would be worthwhile.

chris-durbin:
> I think cmr.umm-spec.legacy is probably not a good namespace name because legacy means other things for CMR.
As a matter of personal principle I refuse to work on any code with 'legacy' in the name.

system-int-test/test/cmr/system_int_test/search/granule_retrieval_test.clj
ygliuvt: There is only one version for UMM-G and it is `1.4` oddly.

ygliuvt:
>> I think cmr.umm-spec.legacy is probably not a good namespace name because legacy means other things for CMR.

> As a matter of personal principle I refuse to work on any code with 'legacy' in the name.

The only way I can work on legacy code and not lose my mind is to not bother with changing their names. :) I really have no idea why this namespace was named `legacy`.

PR #619, opened 2018-11-05
CMR-5269: Fix orbit ascending crossing longitude precision issue

chris-durbin:
We were parsing orbit parameters from ISO SMAP granule metadata as a float, which resulted in a discrepancy in precision between ascending crossing longitudes and equator crossing longitudes. This precision issue resulted in polygon calculations being wildly wrong.


system-int-test/test/cmr/system_int_test/search/granule_orbit_search_test.clj
```diff
@@ -9,7 +9,7 @@
    [cmr.spatial.kml :as kml]
    [cmr.spatial.line-string :as l]
    [cmr.spatial.mbr :as m]
-   [cmr.spatial.point :as p]
+   [cmr.spatial.point :as point]
```
mschmele: 👏

ygliuvt: boo! Now `point` feels lonely because he is forced to move away from his friends `l` and `m`. :)

ygliuvt:
Wow, Great precision with the Giphy. Approved!

PR #622, opened 2018-11-06
Cmr 4575 - Using the KMS API instead of the static contents to get the KMS keywords.

eereiter:
This will allow the CMR to use the latest KMS keywords and it should help fix issues we keep having with the static KMS contents. This is also a step in the direction of deprecating the static KMS files.


transmit-lib/src/cmr/transmit/kms.clj
```diff
@@ -34,19 +34,21 @@
...
-  {:providers "providers/providers.csv"
-   :platforms "platforms/platforms.csv"
-   :instruments "instruments/instruments.csv"
-   :projects "projects/projects.csv"
-   :temporal-keywords "temporalresolutionrange/temporalresolutionrange.csv"
-   :spatial-keywords "locations/locations.csv"
-   :science-keywords "sciencekeywords/sciencekeywords.csv"
-   :concepts "concepts/concepts.xml"
-   :iso-topic-categories "concepts/concepts.xml"})
+  {:providers "providers?format=csv"
```
chris-durbin: Since they all require the format query parameter you could remove that part from this map and just add the query parameter `?format=csv` when making the call.

eereiter: I will be in here again when I really add the related-url keywords and can update it then.

PR #624, opened 2018-11-08
CMR-5141: Added test-environment-robots that disallow everything in test env.

sxu123:


search-app/src/cmr/search/routes.clj
```diff
@@ -70,10 +71,17 @@
     mt/json
     mt/xml))
 
+(defconfig test-environment? 
+   "Flag that indicates if the environment is test environment"
+   {:default false :type Boolean})
+
```
sxu123: I can add a config.clj if preferred.

search-app/src/cmr/search/routes.clj
```diff
@@ -70,10 +71,17 @@
...
 (def robots-txt-response
   "Returns the robots.txt response."
-  {:status 200
-   :body (slurp (io/resource "public/robots.txt"))})
+  (if (test-environment?)
```
chris-durbin: Since this is done in a `def` it will look at the configuration at build time rather than deploy time.

We'll need to have two different vars and perform the `(test-environment)` check when the route is called.

Note that this also will let you add a test to verify the correct contents are returned when changing the configuration.

search-app/src/cmr/search/routes.clj
```diff
@@ -70,10 +71,17 @@
     mt/json
     mt/xml))
 
+(defconfig test-environment?
```
chris-durbin: Our configurations should not have a ? in them since we use environment variables to override the value. `test-environment` will let us use the configuration parameter CMR_TEST_ENVIRONMENT.

sxu123: Removed ?

search-app/src/cmr/search/routes.clj
sxu123: Thanks. Added the test.

PR #627, opened 2018-11-12
CMR-5282: Fixed DataGranule DayNightFlag parsing and generating in UMM-G to work with ISO SMAP granules.

ygliuvt:


ygliuvt:
Intentionally did not add test as it requires changing the existing umm-lib generator and the test failures caused by it. I don't feel it is worth it.

PR #631, opened 2018-11-15
CMR-5248: Added handling of 422 error in transmit-lib; Added variable update tests.

ygliuvt:

### DIFF
--- a/system-int-test/test/cmr/system_int_test/ingest/variable_ingest_test.clj
+++ b/system-int-test/test/cmr/system_int_test/ingest/variable_ingest_test.clj
@@ -245,3 +245,47 @@
               {:keys [status concept-id revision-id]} response]
           (is (= 200 status))
           (is (= 3 revision-id)))))))
+
+(deftest variable-update-test
+  (let [{token :token} (variable-util/setup-update-acl
+                        (s/context) "PROV1" "user1" "update-group")
+        concept (variable-util/make-variable-concept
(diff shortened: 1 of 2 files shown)


system-int-test/test/cmr/system_int_test/ingest/variable_ingest_test.clj
```diff
@@ -245,3 +245,47 @@
...
+            {:keys [concept-id revision-id]} (variable-util/ingest-variable
+                                              concept
+                                              (variable-util/token-opts token))]
+        (is (mdb/concept-exists-in-mdb? concept-id revision-id))
+        (is (= 1 revision-id))))
+
+    (let [;; variable with a different variable name, but the same native-id
+           concept (variable-util/make-variable-concept
+                    {:Name "new-var-name"} {:native-id "var-to-be-updated"})]
+      (testing "update the variable with a different variable name is not allowed"
```
chris-durbin: Note that this is exactly the reason we added native-ids for collections. Providers wanted to be able to update the identifier (dataset ID), but use the same concept so they didn't need to recreate all the granules.

So in this case if someone later decides that a variable has a bad name they'll have to delete the variable, create a new one with the better name, and then redo all of the associations. I assume they are ok with this.

ygliuvt: That's what I was told.

PR #632, opened 2018-11-16
CMR-5274: Scrubed tokens from logs.

sxu123:


common-lib/src/cmr/common/util.clj
```diff
@@ -947,3 +947,13 @@
...
+
+(defn scrub-token
+  "Cut out the middle of the token, leaving 10 characters on each end.
+  For the token that contains less than 10 characters(like tokens in the test),
+  remove the last character."
+  [token]
+  (let [token-length (.length token)]
+    (if (<= 10 token-length)
+      (str (subs token 0 10) (subs token (- token-length 10) token-length))
+      (subs token 0 (- token-length 1)))))
```
sxu123: Not 100% sure if the real tokens are all more than 10 characters long. seems that they should be...  If so, I could just return the token directly here.

common-lib/src/cmr/common/util.clj
```diff
@@ -947,3 +947,13 @@
   (let [index (.indexOf s substring)]
    (when (>= index 0)
     index))))
+
+(defn scrub-token
+  "Cut out the middle of the token, leaving 10 characters on each end.
+  For the token that contains less than 10 characters(like tokens in the test),
+  remove the last character."
+  [token]
+  (let [token-length (.length token)]
```
chris-durbin: I think `count` is preferred over Java interop.

common-lib/src/cmr/common/util.clj
```diff
@@ -947,3 +947,13 @@
...
+    (if (<= 10 token-length)
+      (str (subs token 0 10) (subs token (- token-length 10) token-length))
```
chris-durbin: For something with less than 20 characters - this would contain all the characters in the token though some would be repeated.

I think you're right that we have to handle short or long tokens.

I'd propose something like this:
Take the first 5 and last 5 characters if the token has at least 15 characters. If it has less than 5 characters remove all of the characters. We wouldn't want to just remove 1 character if we were actually trying to redact the token. If it has 5-15 characters we should remove the last 5 characters.

Also we need to indicate this is not the actual token value we received. We should include some filler indicator like "..." or "XXX" to make it clear it has been partially redacted.

transmit-lib/src/cmr/transmit/echo/tokens.clj
```diff
@@ -57,7 +58,7 @@
                 (get-in parsed [:token_info :user_name])
                 (errors/throw-service-error
                  :unauthorized
-                 (format "Token [%s] has expired" token))))
+                 (format "Token [%s] has expired" (common-util/scrub-token token)))))
```
chris-durbin: I think the error message should also indicate the token is not what they passed in.

Something along the lines of:
"Token [%s] has expired. Note the token value has been partially redacted."

common-lib/src/cmr/common/util.clj
sxu123: Thanks. 
1. I'm aware of the repeating character case. I thought it was ok :)
2. I only removed the last 1 character as a place holder to indicate that we need to sth about it. 
3. I was thinking about adding sth in between too.  This is good. now I don't have to worry about the response user gets back. Thank you!

common-lib/src/cmr/common/util.clj
sxu123: Done.

transmit-lib/src/cmr/transmit/echo/tokens.clj
sxu123: Done.

common-lib/src/cmr/common/util.clj
```diff
@@ -947,3 +947,19 @@
...
     index))))
+
+(defn scrub-token
+  "Scrub token:
+  1. When at least 15 chars long keep the first and the last 5 chars. 
+  2. When at least 5 and no more than 14 chars long, remove last 5. 
+  3. When less than 5 chars long, remove all chars.  
+  4. Replace what's removed with XXX."
+  [token]
+  (let [token-length (count token)]
```
chris-durbin: This function is perfect for unit testing. Can you add a unit test and make sure to cover each of these cases?

system-int-test/test/cmr/system_int_test/ingest/provider_ingest_permissions_test.clj
```diff
@@ -240,7 +240,8 @@
                                      {:token provider-admin-update-token
                                       :allow-failure? true})]
         (is (= 401 status))
-        (is (= [(format "Token [%s] has expired" provider-admin-update-token)] errors))))))
+        (is (= [(format "Token [%s] has expired. Note the token value has been partially redacted."
```
chris-durbin: Can you change this test to display the actual returned value rather than using the source code to do the comparison?

e.g. `Token [abcdeXXXvwxyz] has expired. Note the token value has been partially redacted.`

common-lib/src/cmr/common/util.clj
sxu123: Sure. Done.

system-int-test/test/cmr/system_int_test/ingest/provider_ingest_permissions_test.clj
sxu123: Done.

common-lib/src/cmr/common/util.clj
```diff
@@ -947,3 +947,19 @@
...
+    (cond  
+      (<= 15 token-length) (str (subs token 0 5)
```
ygliuvt: This comparison syntax is odd. We should make the comparison more inline with the nature English expression that you have above: `(>= token-length 15)`. The next condition is even harder to read.

system-int-test/test/cmr/system_int_test/ingest/provider_ingest_permissions_test.clj
```diff
@@ -3,7 +3,7 @@
   (:require
    [clojure.string :as string]
    [clojure.test :refer :all]
-   [cmr.common.util :refer [are3]]
+   [cmr.common.util :as common-util :refer [are3]]
```
ygliuvt: No need to modify this line.

sxu123: Right. I needed to use it before, but not anymore. Done.

PR #636, opened 2018-11-28
CMR-5093: Validate Spatial enums for MENDS.

CayvonH:
Add check for CoordinateSystem/GranuleSpatialRepresentation

### DIFF
--- a/system-int-test/test/cmr/system_int_test/ingest/collection_umm_spec_validation_test.clj
+++ b/system-int-test/test/cmr/system_int_test/ingest/collection_umm_spec_validation_test.clj
@@ -95,8 +95,10 @@
                    options)))
 
 (defn assert-valid-spatial
-  [coord-sys hsd]
-  (assert-valid {:SpatialExtent (data-umm-c/spatial {:gsr coord-sys :hsd hsd})}))
+  ([coord-sys hsd]
+   (assert-valid-spatial coord-sys hsd nil))
+  ([coord-sys hsd options]
+   (assert-valid {:SpatialExtent (data-umm-c/spatial {:gsr coord-sys :hsd hsd})} options)))
 
 (use-fixtures :each (ingest/reset-fixture {"provguid1" "PROV1" "provguid2" "PROV2"}))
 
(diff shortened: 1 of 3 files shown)


umm-spec-lib/src/cmr/umm_spec/validation/spatial.clj
```diff
@@ -32,16 +42,29 @@
...
  (let [granule-spatial-representation (:GranuleSpatialRepresentation spatial-extent)
        geometry (util/remove-nil-keys
                  (dissoc (get-in spatial-extent [:HorizontalSpatialDomain :Geometry]) :CoordinateSystem))]
-  (when (and (not (empty? geometry))
-             (nil? granule-spatial-representation))
-   {field-path ["Granule Spatial Representation must be supplied."]})))
+  (if (and (not (empty? geometry))
+           (nil? granule-spatial-representation))
+   {field-path ["Granule Spatial Representation must be supplied."]}
+   (validate-enum field-path granule-spatial-representation valid-granule-spatial-representations))))
```
sxu123: We only want to validate if the geometry is not empty.  But your code will validate if geometry is empty too. 
(when (not (empty? geometry))
    (if (nil? granule-spatial-representation)
      {field-path ["Granule Spatial Representation must be supplied."]}
     (validate-enum field-path granule-spatial-representation valid-granule-spatial-representations)))

system-int-test/test/cmr/system_int_test/ingest/collection_umm_spec_validation_test.clj
```diff
@@ -216,6 +218,33 @@
...
+                                  ["SpatialExtent"]
+                                  ["Value (\"INVALID_GRANULE_SPATIAL_REPRESENTATION\") not found in enum (possible values: [\"CARTESIAN\",\"GEODETIC\",\"NO_SPATIAL\",\"ORBIT\"])"]
+                                  {:format :iso19115}))
+        (testing "Valid GranuleSpatialRepresentation enum"
+          (assert-valid-spatial "CARTESIAN"
+                                geometry
+                                {:format :iso19115}))
+        (testing "Invalid CoordinateSystem enum"
+          (assert-invalid-spatial "CARTESIAN"
+                                  (assoc-in geometry [:Geometry :CoordinateSystem] "INVALID_COORDINATE_SYSTEM")
```
sxu123: Why do you need to pass the valid "CARTESIAN" and associate the invalid one, rather than passing in the invalid one directly, like the "Invalid GranuleSpatialRepresentation enum" case above?  I'm sure there's a reason, just doesn't look intuitive to me.

CayvonH: I wanted to test invalid CoordinateSystems. `The assert-invalid-spatial` function accepts GranuleSpatialRepresentation as a parameter but not a CoordinateSystem (although it is labeled coord-sys as the parameter for some reason). The assoc-in was to just avoid having to create another geometry variable with an invalid CoordinateSystem but if that is easier to read I could add another one.

umm-spec-lib/src/cmr/umm_spec/validation/spatial.clj
CayvonH: Can there not exist a case where the geometry is empty but the GranuleSpatialRepresentation is provided for ingest?

sxu123: That's a question for @eereiter. I was thinking about that too. But the doc string says: "Validate that the granule spatial representation is valid if the spatial extent has geometries", so either change the doc string, or change the code.

umm-spec-lib/src/cmr/umm_spec/validation/spatial.clj
```diff
@@ -32,16 +42,29 @@
     (update-in spatial-extent path (fn [geometry] (map #(assoc % :coord-system coord-sys) geometry)))
     spatial-extent))
 
+(defn- validate-enum
+  "Validate if value exists in enums if value is not nil."
+  [field-path value enums]
+  (when value
+    (when-not (->enum value enums)
+      {field-path
+       [(format "Value (\"%s\") not found in enum (possible values: [%s])"
```
chris-durbin: Our convention has been to put user input values in brackets in our validation messages.

So `(\"%s\")` should be `[%s]`.

system-int-test/test/cmr/system_int_test/ingest/collection_umm_spec_validation_test.clj
sxu123: I think you should add one if you want to test CoordinateSystem too...  which brings another question - the ticket only says to "Add validation for GranuleSpatialRepresentation from ISO MENDS".  If you are adding the validation for CoordinateSystem too, could you update the ticket?

umm-spec-lib/src/cmr/umm_spec/validation/spatial.clj
```diff
@@ -2,20 +2,30 @@
...
+   [cmr.umm-spec.spatial-conversion :as sc]))
 
 (def valid-coord-systems
-  "Coordinate systems that our valid for umm-spec geometries."
+  "Coordinate systems that are valid for umm-spec geometries."
   #{:geodetic :cartesian})
 
+(def valid-granule-spatial-representations
+  "Granule spatial representations that are valid for umm-spec spatial extents."
+  #{:geodetic :cartesian :orbit :no-spatial})
```
chris-durbin: There's a lot of converting to and from keywords, which I think makes the code harder to understand. I'd use the values we expect:
"GEODETIC" "CARTESIAN" "ORBIT" "NO_SPATIAL" throughout. Similarly for "GEODETIC" and "CARTESIAN" above. These enumerations should really be in the UMM-C (I assume we should be updating our latest UMM-C in addition to this change).

I think this should be a UMM-C change to add the enumerations and then add a migration to set these values correctly for old versions of UMM-C.

umm-spec-lib/src/cmr/umm_spec/validation/spatial.clj
```diff
@@ -2,20 +2,30 @@
...
+
+(defn- ->enum
```
chris-durbin: I'd name this `string->enum`

umm-spec-lib/src/cmr/umm_spec/validation/spatial.clj
chris-durbin: I think validating the granule spatial representation anytime it is supplied is the correct thing to do. Cayvon explained that a nil granule spatial representation will pass the validation, so I think the code is good. Just change the docstring.

system-int-test/test/cmr/system_int_test/ingest/collection_umm_spec_validation_test.clj
CayvonH: done

umm-spec-lib/src/cmr/umm_spec/validation/spatial.clj
CayvonH: done

umm-spec-lib/src/cmr/umm_spec/validation/spatial.clj
CayvonH: done

PR #644, opened 2018-12-13
CMR-5360: Add data.json endpoint to test data.nasa.gov.

CayvonH:
Added a data.json endpoint in order to test the connector feature for data.nasa.gov. The connector feature requires an endpoint ending in /data.json to be able to harvest opendata and hopefully sync with data.nasa.gov's data.json response. 

Only testing 20 collections from production for now. If successful will change to harvest all publicly available collections from cmr.

### DIFF
--- a/search-app/src/cmr/search/api/concepts_search.clj
+++ b/search-app/src/cmr/search/api/concepts_search.clj
@@ -243,13 +243,13 @@
                :throw-exceptions false
                :query-params {:page-size 20 :pretty true}} ;; Test just 20 datasets for now
         response (client/request query)]
-    (:body response)))
+    response))
 
 (defn- find-data-json-prod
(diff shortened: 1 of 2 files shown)


search-app/src/cmr/search/api/concepts_search.clj
```diff
@@ -300,3 +322,11 @@
   (GET "/tiles"
     {params :params ctx :request-context}
     (find-tiles ctx params)))
+
+(def data-json-routes
+  "config-enabled route for data.json."
```
chris-durbin: Add a sentence here explaining why we need to do this (Socrata does not support hitting an endpoint that ends with collections.opendata - it must end with data.json). You might have to add ... seriously to the end of that sentence.

PR #646, opened 2018-12-19
CMR-5368: Update Clojure to 1.10.0

chris-durbin:
Note that we'll need to deploy an update to the spatial plugin in each environment and perform a workload run.


access-control-app/int-test/cmr/access_control/int_test/access_control_group_crud_test.clj
```diff
@@ -103,8 +103,10 @@
 
     (testing "Create group with fields at maximum length"
       (let [group (into {} (for [[field max-length] field-maxes]
-                             [field (string-of-length max-length)]))]
-        (is (= 200 (:status (u/create-group (e/login (u/conn-context) "user1") group)))))))
+                             [field (string-of-length max-length)]))
+            ;; FIXME understand why having this outside of a binding causes JVM corruption error
```
chris-durbin: This one makes me a little nervous. I'd love to understand why we see the stack corruption error if trying to make the call to create-group outside the binding.

I can't reproduce this problem outside of two tests, but it boils down to:

JVM corruption:
```
(let [group {:group-stuff "ABC"}]
  (create/group group)
```
Works fine: 🤔 
```
(let [group {:group-stuff "ABC"}]
      _ (create/group group)
```

common-lib/test/cmr/common/test/util.clj
```diff
@@ -48,15 +48,7 @@
       "The most basic case with 1"
       2 (+ 1 1)
       "A more complicated test"
-      4 (* 2 2)))
-  (testing "without comments"
```
chris-durbin: This one now throws an error at compile time, so no need for the test case. Same for `are3-test` below.

access-control-app/int-test/cmr/access_control/int_test/access_control_group_crud_test.clj
```diff
@@ -103,8 +103,10 @@
...
+            ;; in Clojure 1.10.0
+            _ (is (= 200 (:status (u/create-group (e/login (u/conn-context) "user1") group))))])))
```
oubiwann: We probably want to keep track of the issues around `is`that have been discovered in 1.10:
 * https://dev.clojure.org/jira/browse/CLJ-2458

chris-durbin: Note that without the `is` it also throws the JVM corruption issue on the call to create-group.

PR #648, opened 2018-12-20
CMR-5235: Add related url title mappings.

CayvonH:
* Add related url title mappings.
* Create related url to title conversion based on URLContentType, Type, and Subtype.
* If Subtype is nil return the default mapping based off URLContentType and Type.
* Use title mappings for opendata response.

### DIFF
--- a/system-int-test/src/cmr/system_int_test/data2/opendata.clj
+++ b/system-int-test/src/cmr/system_int_test/data2/opendata.clj
@@ -153,7 +153,11 @@
   (update-in opendata-results-map [:results :dataset]
              (fn [dataset]
                (into #{} (for [field dataset]
-                           (util/map-values #(if (sequential? %) (set %) %) field))))))
+                           ;; Dissoc title from distribtion. mimic-ingest-retrieve-metadata-conversion is returning a umm-lib
+                           ;; representation after generating and parsing metadata. This is an issue with retrieving the correct
+                           ;; URLContentType and will cause this test to fail since the url-content-type is not retrieved/in the umm-lib model.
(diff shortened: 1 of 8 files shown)


system-int-test/src/cmr/system_int_test/data2/opendata.clj
```diff
@@ -153,7 +153,11 @@
   (update-in opendata-results-map [:results :dataset]
              (fn [dataset]
                (into #{} (for [field dataset]
-                           (util/map-values #(if (sequential? %) (set %) %) field))))))
```
chris-durbin: All these lines are a little long (require scrolling). We usually set a ruler in our code editor to 100 to know when to split lines.

umm-spec-lib/src/cmr/umm_spec/related_url.clj
```diff
@@ -1,14 +1,26 @@
...
 
+(defn related-url->title
+  "Return the related url title. Title is mapped by URLContentType->Type->Subtype->Title.
+  if subtype is nil return the default title for that type."
+  ([related-url]
+   (let [{:keys [URLContentType Type Subtype]} related-url]
+     (related-url->title URLContentType Type Subtype)))
+  ([url-content-type type sub-type]
+   (let [sub-type-titles (get-in related-url-titles/related-url-titles [url-content-type type])]
+     (or (get sub-type-titles sub-type)
```
chris-durbin: I was going to suggest
```
(get sub-type-titles sub-type (get sub-type-titles "default"))
```
but I think yours is easier to read.

PR #649, opened 2018-12-21
CMR-5374: Complete data.json response.

CayvonH:
* data.json endpoint at search/socrata/data.json.
* Move def to defn.
* Return all public collections with gov.nasa.eosdis tag.
* Use query functions instead of http calls.

### DIFF
--- a/system-int-test/test/cmr/system_int_test/search/data_json_search_test.clj
+++ b/system-int-test/test/cmr/system_int_test/search/data_json_search_test.clj
@@ -70,7 +70,6 @@
         private-collections-with-tag (create-collections 5)
         private-collections-without-tag (create-collections 4)]
     (tag-util/create-tag (config/echo-system-token) tag)
-    (index-util/wait-until-indexed)
 
     ;; Make collections public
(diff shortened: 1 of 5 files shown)


system-int-test/test/cmr/system_int_test/search/data_json_search_test.clj
```diff
@@ -0,0 +1,92 @@
...
+   (map #(hash-map :concept-id (:concept-id %)) collections)))
+
+(deftest data-json-response
+  (let [tag (tag-util/make-tag {:tag-key "gov.nasa.eosdis"})
+        public-collections-with-tag (create-collections 7)
+        public-collections-without-tag (create-collections 6)
+        private-collections-with-tag (create-collections 5)
+        private-collections-without-tag (create-collections 4)]
+    (tag-util/create-tag (config/echo-system-token) tag)
+    (index-util/wait-until-indexed)
```
ygliuvt: This line can be removed as the wait done on line 79 is enough to make sure tags are available before association.

PR #650, opened 2018-12-26
CMR-5338: Add scroll id in umm json response.

CayvonH:
When scroll is true send back scroll id in headers.

### DIFF
--- a/system-int-test/test/cmr/system_int_test/search/scrolling_search_test.clj
+++ b/system-int-test/test/cmr/system_int_test/search/scrolling_search_test.clj
@@ -17,6 +17,24 @@
 
 (use-fixtures :each (ingest/reset-fixture {"provguid1" "PROV1"}))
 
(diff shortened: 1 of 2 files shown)


system-int-test/test/cmr/system_int_test/search/scrolling_search_test.clj
```diff
@@ -17,6 +17,24 @@
 
 (use-fixtures :each (ingest/reset-fixture {"provguid1" "PROV1"}))
 
+(deftest collection-scroll-id-header
```
ygliuvt: This test should check not only the `CMR-Scroll-Id` header is returned which you did, but also the returned scroll id can be used to scroll through the results. See the `granule-scrolling` test below for reference. Also UMM JSON format is also supported on granule searches. It seems you are unaware of that based on you singled out this test on collections only.

PR #652, opened 2019-01-02
CMR-5219 Updates api docs

daniel-zamora:


search-app/docs/api.md
```diff
@@ -1599,6 +1599,10 @@ Find an additional attribute with name "X\Y\Z" with value 7.
 
     curl "%CMR-ENDPOINT%/collections?attribute\[\]=float,X\Y\Z,7"
 
+Find an additional attribute with name "MISSION_NAME" with value "Big Island, HI".
+
+    curl "%CMR-ENDPOINT%/collections?attribute\[\]=string,MISSION_NAME,Big Island\, HI"
```
CayvonH: Should the space in Big Island be URL encoded?

ygliuvt: We should probably add a test for this in system-int-test if there is not one already.

daniel-zamora: They aren't encoding the commas in the docs, so I didn't encode the space.

daniel-zamora: There are already tests for escaping commas for additional attributes.

chris-durbin: Oh show me those tests - that was the first thing I looked for when Greg reported the problem.

PR #654, opened 2019-01-04
CMR-5281

ygliuvt:
As a data provider, I want ingest to reject ingest of orbit granules which specify an equator crossing date time outside of the temporal range of the granule.

### DIFF
--- a/umm-spec-lib/test/cmr/umm_spec/test/validation/umm_spec_granule_validation_tests.clj
+++ b/umm-spec-lib/test/cmr/umm_spec/test/validation/umm_spec_granule_validation_tests.clj
@@ -715,6 +715,31 @@
          {:path [:two-d-coordinate-system :start-coordinate-1]
           :errors ["The field [Start Coordinate 1] falls outside the bounds [0 ∞] defined in the collection"]}]))))
 
+(defn- make-granule-with-equator-crossing
+  "Creates a granule with the given equator-crossing-date-time.
+  This function is tied to the collection that is used in test
+  granule-ocsd-equator-crossing-date-time and is only intended to be used in that test."
+  [equator-crossing-date-time]
+  (make-granule
+   {:temporal (g/map->GranuleTemporal
+               {:range-date-time
+                {:beginning-date-time (dtp/parse-datetime "2015-01-01T00:00:00Z")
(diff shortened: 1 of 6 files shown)


umm-spec-lib/test/cmr/umm_spec/test/validation/umm_spec_granule_validation_tests.clj
```diff
@@ -714,3 +714,80 @@
...
+                               {:TemporalExtents [(cmn/map->TemporalExtentType
+                                                   {:RangeDateTimes [(helpers/range-date-time "2015-01-01T00:00:00Z" "2016-01-02T00:00:00Z")]})]
+                                :SpatialExtent
+                                {:GranuleSpatialRepresentation :orbit
+                                 :OrbitParameters {:InclinationAngle 98.2
+                                                   :Period 100.0
+                                                   :SwathWidth 2600.0
+                                                   :StartCircularLatitude 50.0
+                                                   :NumberOfOrbits 2.0}}})
+        make-granule-with-equator-crossing
```
chris-durbin: It's easier to read if we pull this function definition out of the let block.

ygliuvt: Didn't do that because this function is not generic enough, too much hard coded data just for this particular test.

ygliuvt: I knew exactly what you mean. I didn't do it because the granule has some fields that are tied to the collection that is defined in the `let` block. I don't want to move the function out of the `let` block to make it appear to be more independent.

oubiwann: I'd just put a docstring in the extracted function saying something like "This is intended only for use by ..."

Also, it's not idiomatic Clojure to define a function in a `let`; that's what `letfn` is for ...

ygliuvt: OK. Moved the function out of the let block. Thanks.

PR #655, opened 2019-01-08
CMR-4784 update cmr_metadata_preview commit

daniel-zamora:
I think this is the correct commit, it isn't the merge commit though so if its not correct let me know.


collection-renderer-lib/project.clj
```diff
@@ -14,7 +14,7 @@
    the hardcoded commit id during dev integration with cmr_metadata_preview project.
    The hardcoded commit id should be updated when MMT releases a new version of the gem."
   (or (System/getenv "CMR_METADATA_PREVIEW_COMMIT")
-      "72fe8b4a178"))
+      "991048ffbcb"))
```
chris-durbin: I would just always use the latest commit on master (the merge commit - in this case a3d370d). The code is no different, but makes it easier to confirm we are up to date.

PR #657, opened 2019-01-09
CMR-5388: Add cmr-exchange projects

oubiwann:
image

This  pulls in CMR exchange projects into the `cmr-exchange` subdir.

Note that for this PR, you won't want to use rebase, but rather simply "merge".


chris-durbin:
I think site-templates and message-queue-lib already exist in the CMR today. Did you make sure to omit those from the cmr-exchange subdir? Are there any other projects you know of that live in both locations?

oubiwann:
Yeah, I didn't include the message-queue-lib for the reason you mentioned; I'll have to check on the site-templates in CMR ... things might have diverged.

oubiwann:
Only differences in site-templates were README and project.clj, so I will remove the one brought in from cmr-exchange.

oubiwann:
Updated and ready for merge ...

oubiwann:
Just added a PR image ... seemed appropriate ;-)

chris-durbin:
So once we merge this in we'll have to update Bamboo and Travis CI to use the new location for the sub repos. We should keep versioning the individual libraries and apps the way we have been and publishing to clojars. Is there anything else that should change in our workflow as a result of moving all these projects into this single repo?

oubiwann:
Yeah, Bamboo for service-bridge should be the only thing that needs to be updated; everything else keeps operating as-is.

oubiwann:
Finished green a minute after your post :-)

PR #661, opened 2019-01-10
CMR-5375: Replace Nasa logo with CMR logo

sxu123:
There is still cmr_metadata_preview I haven't touched. Not familiar with which GUI it is and not sure if we are responsible for it.  Will work on that as second part of this ticket if needed.


common-app-lib/resources/public/site/css/eui.css
```diff
@@ -160,21 +160,25 @@ table.image-grid p {
 }
 
 .eui-masthead-logo.eui-application-logo a:link {
-  background-image: url(https://cdn.earthdata.nasa.gov/eui/latest/docs/assets/ed-logos/app-logo.png);
+  padding: 13px 0 10px 120px; 
+  background-image: url(../images/cmr-logo-white.png);
```
chris-durbin: Anyone know if the relative paths will be available in a deployed environment? Also it seems like we be using a CDN rather than serving the image from the CMR hosts, though I know very little about this topic.

PR #663, opened 2019-01-10
CMR-5231: Change references of echo.nasa.gov.

CayvonH:


search-app/src/cmr/search/results_handlers/atom_results_handler.clj
```diff
@@ -287,7 +287,7 @@
    :xmlns:dc "http://purl.org/dc/terms/"
    :xmlns:georss "http://www.georss.org/georss/10"
    :xmlns:time "http://a9.com/-/opensearch/extensions/time/1.0/"
-   :xmlns:echo "http://www.echo.nasa.gov/esip"
+   :xmlns:echo "http://www.echo.nasa.gov/esip" ;; TODO change or delete this dead link
```
ygliuvt: I think we need to make a decision now rather than kicking the can down the road.

PR #665, opened 2019-01-14
CMR-5387: As a Client user, I want to use the Ingest translation endpoint to translate a granule into ISO MENDS format.

ygliuvt:


ingest-app/src/cmr/ingest/api/translation.clj
```diff
@@ -32,14 +36,48 @@
                                  :umm-json
                                  (umm-spec/generate-metadata context umm :umm-json))))
 
+(def xsl-transformer-cache-name
+  "This is the name of the cache to use for XSLT transformer templates. Templates are thread
+  safe but transformer instances are not.
+  http://www.onjava.com/pub/a/onjava/excerpt/java_xslt_ch5/?page=9"
```
daniel-zamora: this URL redirects me to https://www.oreilly.com/ideas?page=9 which doesn't have any documentation.

ygliuvt: It is existing code that was copied from search-app. The link must have become obsolete since. I will remove it.

ingest-app/src/cmr/ingest/api/translation.clj
```diff
@@ -32,14 +36,47 @@
...
+(defn- generate-granule-iso-mends-metadata
+  "Generate granule ISO MENDS metadata from the given umm-lib granule model.
+  CMR does this by generating granule metadata in ECHO10 format and then uses xslt to convert
+  the ECHO10 metadata into ISO MENDS."
+  [context umm]
+  (let [echo10-metadata (umm-legacy/generate-metadata context umm :echo10)]
+    (xslt/transform echo10-metadata (get-template context echo10-iso19115-xslt))))
+
+(defn- generate-metadata
+  "Generate the metadatat for the given umm record and output format."
```
CayvonH: Small typo at 'metadatat'

ingest-app/docs/api.md
```diff
@@ -622,7 +622,7 @@ Example output:
 ```
 ## <a name="translate-granule"></a> Translate Granule Metadata
 
-Granule metadata can be translated between metadata standards using the translate API in Ingest. The request specifies the metadata standard being sent using the Content-Type header. Metadata is sent inside the body of the request. The output format is specified via the Accept header. The supported formats are ECHO10, ISO SMAP and UMM-G.
+Granule metadata can be translated between metadata standards using the translate API in Ingest. The request specifies the metadata standard being sent using the Content-Type header. Metadata is sent inside the body of the request. The output format is specified via the Accept header. The supported input formats are ECHO10, ISO SMAP and UMM-G. The supported output formats are ECHO10, ISO SMAP, UMM-G and ISO19115.
```
chris-durbin: Why is ISO 19115 not supported as an input format? Do we support all combinations of input and output format? For example can we use ISO SMAP as input and get back ISO-19115?

ingest-app/src/cmr/ingest/api/translation.clj
```diff
@@ -1,15 +1,18 @@
...
+   [cmr.common.xml.xslt :as xslt]
+   [cmr.umm-spec.legacy :as umm-legacy]
+   [cmr.umm-spec.test.umm-generators :as umm-generators]
+   [cmr.umm-spec.umm-spec-core :as umm-spec]
+   [cmr.umm-spec.util :as util]
+   [cmr.umm-spec.versioning :as versioning]
+   [compojure.core :refer :all]))
 
 (def concept-type->supported-formats
   "A map of concept type to the list of formats that are supported both for input and output of
```
chris-durbin: This doesn't match the documentation that says ISO-19115 is supported as an output format, but not an input format.

ingest-app/src/cmr/ingest/api/translation.clj
```diff
@@ -32,14 +36,47 @@
...
+  "This is the name of the cache to use for XSLT transformer templates. Templates are thread
+  safe but transformer instances are not."
+  :xsl-transformer-templates)
+
+(def ^:private echo10-iso19115-xslt
+  (io/resource "xslt/echo10_to_iso19115.xsl"))
+
+(defn- get-template
+  "Returns a XSLT template from the filename, using the context cache."
+  [context f]
```
chris-durbin: Replace `f` with `filename`.

ingest-app/src/cmr/ingest/api/translation.clj
```diff
@@ -32,14 +36,47 @@
...
 
+(def xsl-transformer-cache-name
+  "This is the name of the cache to use for XSLT transformer templates. Templates are thread
+  safe but transformer instances are not."
+  :xsl-transformer-templates)
+
+(def ^:private echo10-iso19115-xslt
+  (io/resource "xslt/echo10_to_iso19115.xsl"))
+
+(defn- get-template
```
chris-durbin: I think we should replace the cache with a compile-time variable. Unless I'm misunderstanding the code, all of these templates are known ahead of time and cannot change while the application is running.

e.g.
```
(def filename->template
  {"filename1" (xslt/read-template "filename1)
   "filename2" ...})
```

ingest-app/docs/api.md
ygliuvt: We don't have support in either umm-lib or umm-spec-lib to parse ISO19115 granules. That is why we can not ingest granule in ISO19115 metadata format. ISO19115 is strictly an output format for granules. We can get granule in ISO19115 from any valid granule formats including ISO SMAP.

ingest-app/src/cmr/ingest/api/translation.clj
ygliuvt: This is not my code really. It is copied from search-app. I think the reason cache is used rather than compile time variable is because the underlining JAVA code are interfaces, as a result, they can not be resolved in compile time. If we simply use variable, they are not actually executed at compile time, but rather executed every single time the code is evaluated at run time. That's why we used cache to cache the parsed template (I think).

ingest-app/src/cmr/ingest/api/translation.clj
ygliuvt: Good point. Fixed the validation to not allow ISO19115 as input format for granule translation.

ingest-app/src/cmr/ingest/api/translation.clj
chris-durbin: I'm ok with using the in-memory cache in that case.

PR #667, opened 2019-01-14
CMR-5334: Add dependency-check plugin.

CayvonH:
Primary differences between this and Catalino's PR is:
* No more log/throw options
* Remove need to do bundle-audit and instead use a properties file (only added to search-app and system-int-test as these were the only places we needed to disable bundle-audit).


README.md
```diff
@@ -183,6 +183,14 @@ To perform static analysis and linting for a project, you can run
 `cmr test lint PROJ`. As above with dependency version checking, by
 not passing a project, you can run for all projects: `cmr test lint`.
 
+#### Dependency Vulnerability Scanning
+
+You can see if your currently installed version of CMR has any reported Common Vulnerabilities and Exploits (CVEs) by running the helpful alias `lein check-sec` that you can use in each application, or at the root folder to scan all CMR apps together.
+
+You will find the vulnerability summary in `./target/dependency-check-report.html` in each application.
+
+Make sure you set up the CMR_ORACLE_JAR_REPO environment variable.
```
ygliuvt: Set `CMR_ORACLE_JAR_REPO` to what? We should provide an example.

search-app/resources/dependencycheck.properties
```diff
@@ -0,0 +1,4 @@
+analyzer.ruby.gemspec.enabled=false
```
daniel-zamora: Can we comment on whats going on with these properties, why they are needed etc.

ygliuvt: Can you add comments explaining these properties? When will we want to set them to `true`?

README.md
CayvonH: We can set this to anything. The problem right now is in oracle-lib we have:
```
(def extra-repository
  "The set of repositories to include if configured"
  (if-let [repo (get (System/getenv) oracle-jar-repo-env-var)]
    repo
    "http://example.com/no_url_specified_this_is_ignored"))
```

This creates an error:
```
Tried to use insecure HTTP repository without TLS:
 releases: http://example.com/no_url_specified_this_is_ignored
```

Which funny enough can be fixed if we change the http in:
`http://example.com/no_url_specified_this_is_ignored`
To https:
`https://example.com/no_url_specified_this_is_ignored`

But otherwise I would imagine we would set this variable to something like https://maven.earthdata.nasa.gov/

Since the jars are already on our local system we can just skip this step by setting this environment variable to 'nothing'. 

Maybe I should change the http://example.com... to something else so this step isn't even required?

chris-durbin: Just chiming in to say we should not put our internal maven repo in here.

Not knowing anything about the code - it seems like we should allow `extra-repository` to return nil and not try to do anything. So that `if-let` would become a `when-let`.

ygliuvt: So sounds like it is not needed. My vote is to remove this `Make sure` line :) and change `http://example.com...` to `https://example.com...` in `oracle-lib`.

dev-system/resources/dependencycheck.properties
```diff
@@ -0,0 +1,6 @@
+# Disable ruby analyzers, CI build image does not have ruby installed.
```
chris-durbin: Should we ask for them to be installed to the image that is used for CMR?

PR #669, opened 2019-01-15
CMR-5404: Add cmr-common-lib to clojars, update references to it

mschmele:


chris-durbin:
I think we need to come up with and document a plan for versioning in the CMR before we start moving things into clojars. When do we use snapshots vs. releases? What do we do for release/versioning at the end of a sprint? How do we handle creating tags in Github? Are nightly deployments to SIT impacted? Are we going to change our build plan(s) in Bamboo? When should someone increment the minor or the patch version for a library/application - on every change to any code within a library?

mschmele:
Hmm that's a good question. I'll write up some notes on my thoughts and we'll talk about it this week maybe?

mschmele:
We're going to put some libraries into clojars, but probably not this one

PR #671, opened 2019-01-16
CMR-5409: Move lein-nvd to lein-dependency-check.

CayvonH:
* Use lein-dependency-check instead of lein-nvd to be similar to CMR Proper.
* Move suppression files to the projects top directory. This change will be reverted and all suppressions and properties files will be placed in resources/security once lein-dependency-check allows for that.
* Rename false-positives.xml to suppression.xml.
* Change readme security-scna-badges slightly to be dependency check security instead of nvd.


CayvonH:
> Did you see any difference in the reported vulnerabilities when switching libraries?

Absolutely none. Which is why i originally left it as is but changed my mind as I couldn't think of a reason not to update it.

CayvonH:
> Did you see any difference in the reported vulnerabilities when switching libraries?

With the exception of one medium severity with low confidence (lein-dependency-check found this in some libraries when lein-nvd did not). They are identical.

PR #672, opened 2019-01-16
CMR-5378: Add the release version to the footer for the CMR core landing pages.

sxu123:
At deployment environment variable CMR_RELEASE_VERSION is set to the release version.  This is done through bamboo adding this env variable to .env, which is sourced when CMR starts up.

### DIFF
--- a/common-app-lib/src/cmr/common_app/config.clj
+++ b/common-app-lib/src/cmr/common_app/config.clj
@@ -20,4 +20,4 @@
 
 (defconfig release-version
   "Contains the release version of CMR."
-  {:default "1.1.1-r1.1.1"})
+  {:default "dev"})
--- a/common-app-lib/src/cmr/common_app/config.clj
+++ b/common-app-lib/src/cmr/common_app/config.clj
@@ -17,3 +17,7 @@
   "Flag for whether or not launchpad token is enforeced."
   {:default false
    :type Boolean})
+
+(defconfig release-version
+  "Contains the release version of CMR."
+  {:default "1.1.1-r1.1.1"})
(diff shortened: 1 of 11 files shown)


common-app-lib/src/cmr/common_app/config.clj
```diff
@@ -17,3 +17,7 @@
   "Flag for whether or not launchpad token is enforeced."
   {:default false
    :type Boolean})
+
+(defconfig release-version
+  "Contains the release version of CMR."
+  {:default "1.1.1-r1.1.1"})
```
chris-durbin: I noticed that Legacy services uses a default version of "dev". I think dev is better to indicate it's not a specific version - this numbering makes it look like a real release.

sxu123: done.

sxu123:
Tested with uberjar and the release version is dynamically changed based on the value of the environment variable CMR_RELEASE_VERSION.

search-app/test/cmr/search/test/site/routes.clj
```diff
@@ -88,13 +91,24 @@
...
+           "v dev"))
       (is (string/includes?
            (:body response)
            "site/docs/search/api"))
       (is (string/includes?
            (:body response)
            "site/docs/search/site")))))
 
+(deftest dynamic-release-version
+  (let [_ (side-api/eval-form `(common-config/set-release-version! "1.0"))
```
chris-durbin: You can move the call to `set-release-version` above the `let` block instead of doing it in a binding. 

Since there's another test that relies on the default version you should change the version back at the end of this test so that it doesn't matter what order tests run in.

e.g. add this at the end:
```
(side-api/eval-form `(common-config/set-release-version! "dev"))
```

sxu123: okay. I was just thinking of that but since the test didn't fail, I thought it was okay not to change it. Will do that. Thanks.

PR #673, opened 2019-01-18
CMR-5387: Fixed common-lib xslt translation to explicitly use Saxon lib.

ygliuvt:
This is to guarantee that xslt 2.0 processor is used for xslt translation.

### DIFF
--- a/common-lib/src/cmr/common/xml/xslt.clj
+++ b/common-lib/src/cmr/common/xml/xslt.clj
@@ -8,7 +8,8 @@
            [javax.xml.transform
             TransformerFactory
             Templates
-            URIResolver]))
+            URIResolver]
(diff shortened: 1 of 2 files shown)


common-lib/src/cmr/common/xml/xslt.clj
```diff
@@ -8,7 +8,8 @@
            [javax.xml.transform
```
chris-durbin: Fix the require and import statements in this namespace to match our standards.
```
  (:require
   [clojure.java.io :as io])
  (:import
   (java.io StringReader StringWriter)
   (javax.xml.transform.stream StreamResult StreamSource)
   (javax.xml.transform Templates TransformerFactory URIResolver)
   (net.sf.saxon TransformerFactoryImpl)))
```

PR #674, opened 2019-01-18
CMR-5410: Remove use of reflection in many places in the CMR codebase.

chris-durbin:
In order to properly setup `lein check` to run I also needed to eliminate a cyclic dependency caused by our caching namespaces in the acl-lib and common-app-lib. Moving the consistent and cubby caches to the transmit library was the simplest solution, and while not a perfect fit was a better solution that adding another library in my opinion.


mock-echo-app/src/cmr/mock_echo/api/urs.clj
```diff
@@ -109,8 +111,10 @@
       successful-login-response
       unsuccessful-login-response)))
 
+(def bees (byte-array 10))
```
mschmele: What is this? Is it used anywhere?

chris-durbin: Who doesn't need bees? Looks like some of my debugging didn't get deleted.

mschmele: image

PR #675, opened 2019-01-21
CMR-5413

CayvonH:
Output csv as well as html.
Cleanup all files relating to security (suppression.xml and dependencycheck.properties) by moving them to resources/security.


CayvonH:
> Is there any change to the way we should run the command, or will both CSV and HTML outputs be created every time?
> 
> If there's a different way to run the command please update the README.

No change in running the command. Just added csv output to the dependency-check config as I was told csv is the preferred format for when we want to parse the vulnerabilities. Although we could add any number of formats including json and xml.

PR #676, opened 2019-01-21
CMR-5423: remove data.json config option.

CayvonH:
no longer need data.json endpoint to be configurable.

### DIFF
--- a/search-app/src/cmr/search/api/concepts_search.clj
+++ b/search-app/src/cmr/search/api/concepts_search.clj
@@ -5,7 +5,6 @@
    [clj-http.client :as client]
    [cmr.common-app.api.routes :as common-routes]
    [cmr.common-app.services.search :as search]
-   [cmr.common-app.site.pages :as common-pages]
    [cmr.common.cache :as cache]
    [cmr.common.config :refer [defconfig]]
    [cmr.common.log :refer (debug info warn error)]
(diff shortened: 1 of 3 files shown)


search-app/src/cmr/search/api/concepts_search.clj
```diff
@@ -315,11 +309,10 @@
     (find-tiles ctx params)))
 
 (def data-json-routes
-  "config-enabled route for data.json. Socrata does not support harvesting from
+  "Route for data.json response. Socrata does not support harvesting from
```
ygliuvt: Now that this route is official, I think we should add it in the api doc or at least somewhere in README so that it does not get lost as time goes on.

CayvonH: hmm do you think the search-app readme would be okay or should i put it in the top level cmr readme?

ygliuvt: I am actually leaning towards in the search api doc. We document the facet v2 api in api doc even though it is not supposed to be called manually. This api is similar to that. search-app README is fine too if we don't want to expose this route in the public api doc. Definitely not in the top level CMR README.

CayvonH: I was more leaning towards the readme because I have no idea why anyone other than us would actually want to use it.

ygliuvt: Yeah, I see your point. I am fine with the search-app README.

PR #683, opened 2019-01-29
CMR-5415-CMR-5416-CMR-5417: Allow variable search by variable_alias

sxu123:


indexer-app/src/cmr/indexer/data/index_set.clj
```diff
@@ -662,6 +662,8 @@
    :provider-id.lowercase (m/doc-values m/string-field-mapping)
    :variable-name (-> m/string-field-mapping m/stored m/doc-values)
    :variable-name.lowercase (m/doc-values m/string-field-mapping)
+   :variable-alias (-> m/string-field-mapping m/stored m/doc-values)
```
chris-durbin: It doesn't have to be stored unless you return it in a search response. If you think there's a chance it will eventually be returned in either the JSON or XML references formats for variables (alias is a commonly needed variables field), then I'm fine with going ahead and storing it now rather than ending up with two fields because we have to change it later on to store it.

system-int-test/test/cmr/system_int_test/search/variable/variable_search_test.clj
```diff
@@ -136,6 +140,44 @@
...
+      [variable1]
+      {:variable-alias "Alias1"}
+
+      "By alias case sensitive, default ignore-case true"
+      [variable1]
+      {:variable-alias "alias1"}
+
+      "By alias ignore case false"
+      []
+      {:variable-alias "alias1" "options[variable-alias][ignore-case]" false}
```
chris-durbin: Our APIs all use underscores rather than dash case. It's fine if dash case also works but we should have tests verifying the underscore versions as well.

(both `variable_alias` and `options[variable_alias][...]` should be tested).

sxu123: Done.

search-app/docs/api.md
```diff
@@ -3227,6 +3227,8 @@ These parameters will match fields within a variable. They are case insensitive
   * options: pattern, ignore_case
 * native_id
   * options: pattern, ignore_case
+* variable_alias
```
ygliuvt: I think the search parameter should be `alias` rather than `variable_alias`.

indexer-app/src/cmr/indexer/data/index_set.clj
sxu123: We've decided to keep it as stored since it's an important field.

search-app/docs/api.md
chris-durbin: Good point - this is on the /search/variables route so variable is redundant.

PR #684, opened 2019-01-31
CMR-5429: Added support for UMM-Var 1.4.

sxu123:
UMM-Var version 1.4 removes :AvgCompressionRateASCII and :AvgCompressionRateNetCDF4
and replace them with the new :AverageCompressionInformation [{:Format "ASCII" :Rate <rate>} {:Format "NetCDF-4" :Rate <rate>}]

### DIFF
--- a/umm-spec-lib/src/cmr/umm_spec/migration/version/variable.clj
+++ b/umm-spec-lib/src/cmr/umm_spec/migration/version/variable.clj
@@ -115,10 +115,12 @@
   ;; Migrate down to 1.3
   (if-let [avg-comp-info (get-in v [:SizeEstimation :AverageCompressionInformation])]
     (as-> v m
-      (when-let [avg-comp-ascii (some #(when (= "ASCII" (:Format %)) (:Rate %)) avg-comp-info)]
-        (assoc-in m [:SizeEstimation :AvgCompressionRateASCII] avg-comp-ascii))
-      (when-let [avg-comp-netcdf4 (some #(when (= "NetCDF-4" (:Format %)) (:Rate %)) avg-comp-info)]
-        (assoc-in m [:SizeEstimation :AvgCompressionRateNetCDF4] avg-comp-netcdf4))
+      (if-let [avg-comp-ascii (some #(when (= "ASCII" (:Format %)) (:Rate %)) avg-comp-info)]
+        (assoc-in m [:SizeEstimation :AvgCompressionRateASCII] avg-comp-ascii)
+        m)
+      (if-let [avg-comp-netcdf4 (some #(when (= "NetCDF-4" (:Format %)) (:Rate %)) avg-comp-info)]
+        (assoc-in m [:SizeEstimation :AvgCompressionRateNetCDF4] avg-comp-netcdf4)
+        m)
       (update-in m [:SizeEstimation] dissoc :AverageCompressionInformation)
       (util/remove-nil-keys m))
     v))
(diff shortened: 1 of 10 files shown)


umm-spec-lib/src/cmr/umm_spec/migration/version/variable.clj
```diff
@@ -109,3 +109,30 @@
   ;; Migrate up to 1.3
   [context v & _]
   v)
+
+(defmethod interface/migrate-umm-version [:variable "1.4" "1.3"]
+  [context v & _]
+  ;; Migrate down to 1.3
+  (if-let [avg-comp-info (get-in v [:SizeEstimation :AverageCompressionInformation])]
+    (as-> v m
```
ygliuvt: This migration logic only works when both the ASCII and NetCDF-4 formats exist. You should add migration test cases for at least when one of the formats does not exist, and fix this migration code to make it work.

sxu123: What do you mean?  I only process v when AverageCompressionInformation exists in 1.4;  Then I only associate AvgCompressionRateASCII when "ASCII" format exists in AverageCompressionInformation;  And only associate AvgCompressionRateNetCDF4 when "NetCDF-4" format exists in AverageCompressionInformation. I've tested this function a few times. It works. I could add more test but the logic should be fine.

sxu123: I can't believe this. I tested a few times. and in the last minute I changed it to when-let. Thanks!

umm-spec-lib/test/cmr/umm_spec/test/migration/version/variable.clj
```diff
@@ -181,3 +229,35 @@
                                                             {:GroupPath "/MODIS_Grid_Daily_1km_LST/Data_Fields"
                                                              :IndexRanges {:LatRange [-45 45]
                                                                            :LonRange [90 180]}})))))
+
+(deftest migrate-13->14
```
ygliuvt: nit: I prefer the variable map be included in the test over a `def`. This way I can easily see what is being migrated. Also, we don't need to include irrelevant fields in the variable map.

umm-spec-lib/src/cmr/umm_spec/migration/version/variable.clj
```diff
@@ -109,3 +109,32 @@
...
+      (if-let [avg-comp-ascii (some #(when (= "ASCII" (:Format %)) (:Rate %)) avg-comp-info)]
+        (assoc-in m [:SizeEstimation :AvgCompressionRateASCII] avg-comp-ascii)
+        m)
+      (if-let [avg-comp-netcdf4 (some #(when (= "NetCDF-4" (:Format %)) (:Rate %)) avg-comp-info)]
+        (assoc-in m [:SizeEstimation :AvgCompressionRateNetCDF4] avg-comp-netcdf4)
+        m)
+      (update-in m [:SizeEstimation] dissoc :AverageCompressionInformation)
+      (util/remove-nil-keys m))
```
ygliuvt: Remove this line.

umm-spec-lib/test/cmr/umm_spec/test/migration/version/variable.clj
sxu123: I'm only using def because all the previous versions are using def in the test.  I don't mind using a more realistic variable for the tests.

PR #685, opened 2019-02-01
CMR-5490: Treat a platform with a blank short name in DIF9 as not provided

chris-durbin:

### DIFF
--- a/umm-spec-lib/src/cmr/umm_spec/util.clj
+++ b/umm-spec-lib/src/cmr/umm_spec/util.clj
@@ -2,7 +2,7 @@
   "This contains utilities for the UMM Spec code."
   (:require
    [clj-time.format :as f]
-   [clojure.string :as str]
+   [clojure.string :as string]
    [cmr.common.xml.parse :as p]
    [cmr.common.xml.simple-xpath :refer [select]]
    [cmr.common.util :as util]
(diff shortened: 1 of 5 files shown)


umm-spec-lib/src/cmr/umm_spec/util.clj
```diff
@@ -284,9 +284,12 @@
 (defn parse-short-name-long-name
   "Returns the list of ShortName and LongName from parsing the given doc on the given path."
   [doc path]
-  (seq (for [elem (select doc path)]
-         {:ShortName (p/value-of elem "Short_Name")
-          :LongName (p/value-of elem "Long_Name")})))
+  (seq (for [elem (select doc path)
```
ygliuvt: Am I reading the code correctly? So this fix just drops any platforms or instruments that have a blank short-name? Don't we want these kind of collections to fail validation instead?

chris-durbin: Yes. I asked Erich and Scott about rejecting these requests, but they both agreed for DIF9 they'd rather it treated the same way as the case when ShortName and LongName are missing where we use "Not provided".

ygliuvt: This is dropping these platforms/instruments. It is not the same as replace blank with "Not provided".

PR #686, opened 2019-02-01
CMR-5411: Change schema validation library

CayvonH:
Move to org.everit.json.schema.

### DIFF
--- a/common-lib/src/cmr/common/validations/json_schema.clj
+++ b/common-lib/src/cmr/common/validations/json_schema.clj
@@ -3,106 +3,68 @@
   for more details."
   (:require
    [cheshire.core :as json]
-   [cmr.common.services.errors :as errors]
+   [clojure.java.io :as io]
    [cmr.common.log :as log :refer (warn info)]
-   [clojure.string :as str])
+   [cmr.common.services.errors :as errors])
   (:import
-   (com.github.fge.jsonschema.main JsonSchema
-                                   JsonSchemaFactory)
-   (com.github.fge.jackson JsonLoader)
-   (com.github.fge.jsonschema.core.report ListProcessingReport)
-   (com.fasterxml.jackson.core JsonParseException)))
+   (org.everit.json.schema Schema ValidationException)
+   (org.everit.json.schema.loader SchemaLoader SchemaClient)
+   (org.json JSONArray JSONException JSONObject JSONTokener)))
 
-(defn- parse-error-report
-  "Parses the error-report to return a human friendly error message.
-
-  Example:
-  {:instance {:pointer \"/provider\"}
-   :message \"object instance has properties which are not allowed by the schema: [\"123\"]\"
-  ... other keys ignored}"
-  [error-report]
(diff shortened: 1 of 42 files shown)


common-lib/src/cmr/common/validations/json_schema.clj
```diff
@@ -7,102 +7,64 @@
...
   object. It's necessary to use this one if loading a JSON schema from the
   classpath that references other JSON schemas on the classpath."
   [uri]
-  (let [factory (JsonSchemaFactory/byDefault)]
-    (.getJsonSchema factory (str uri))))
+  (let [raw-schema (json-string->JSONObject (slurp (str uri)))]
+    (-> (SchemaLoader/builder)
+        (.schemaJson raw-schema)
+        (.resolutionScope (str uri))
+        .build
```
ygliuvt: `.build` twice?

CayvonH: Yeah so the first called build() returns a SchemaLoader object then the call to load() returns a Schema.Builder() object and finally the last call to build() returns a Schema Object.

So it goes:
SchemaLoader.SchemaLoaderBuilder -> SchemaLoader -> Schema.Builder -> Schema.

common-lib/src/cmr/common/validations/json_schema.clj
```diff
@@ -7,102 +7,64 @@
    [cmr.common.log :as log :refer (warn info)]
    [clojure.string :as str])
   (:import
-   (com.github.fge.jsonschema.main JsonSchema
-                                   JsonSchemaFactory)
-   (com.github.fge.jackson JsonLoader)
-   (com.github.fge.jsonschema.core.report ListProcessingReport)
-   (com.fasterxml.jackson.core JsonParseException)))
+   (org.everit.json.schema Schema
```
chris-durbin: The original code wasn't following our new guide for namespaces: https://stuartsierra.com/2016/clojure-how-to-ns.html
```
(org.everit.json.schema Schema ValidationException)
```
is our preferred style now. There's a fewer other namespaces in this PR that should also be updated.

CayvonH:
> I assume there will be a few other tests that need to be updated based on the format of the error messages changing.

Yes, there is quite a few. I'm working on that now. Just wanted to get this PR out there in case people had different ideas of schema validation so I don't change all the tests unnecessarily.

common-lib/src/cmr/common/validations/json_schema.clj
mschmele: Might be worth leaving a comment to explain why it was done this way to prevent confusion in the future

CayvonH:
> I think the new error messages look more human friendly except for `#` prepended to all of the messages. I'm wondering if we should strip that from the error messages.
> 
> Also it seems like the default is to use JSON schema 0.4. Have we tested that it really does support 0.7?

I made a dummy draft-07 schema and was able to use the new if-then-else keywords. All I had to do was change the $schema top level definition in the schema to "http://json-schema.org/draft-04/schema#".

PR #689, opened 2019-02-08
CMR-5454 1: Add support of UMM-G v1.5 in umm-spec-lib.

ygliuvt:


umm-spec-lib/src/cmr/umm_spec/test/umm_generators.clj
```diff
@@ -38,11 +39,19 @@
...
   (let [constructor-fn (if type-name
                          (record-gen/schema-type-constructor schema type-name)
                          identity)
         properties (:properties schema-type)
+        ;; The following line is a hack!!!
+        ;; Since UMM-G InstrumentType has ComposedOf field that is also an InstrumentType
+        ;; and there could be multiple levels of ComposedOf within an InstrumentType,
+        ;; We randomly (1 out of 2) drop the ComposedOf field from properties
+        ;; to not get into infinite loops of generating InstrumentTypes
+        properties (if (= 0 (rand-int 2))
```
chris-durbin: I don't quite follow why we had to use random as opposed to always picking one. Are you sure there is not a legitimate case that this code would not handle?

chris-durbin: Oh this is a generator for test cases... Ok I'm not so worried anymore.

ygliuvt: yeah, It just generates the ComposedOf part of the InstrumentType 50% of the time for testing. This way we can have a variety of test data that could go down multiple levels on the ComposedOf. It is a good hack. :)

PR #696, opened 2019-02-19
CMR-5452: Add support of spatial extent string values of cycle, pass, and tile in umm-lib granule model

ygliuvt:


umm-spec-lib/src/cmr/umm_spec/migration/version/granule.clj
```diff
@@ -54,6 +54,13 @@
     "application/xhdf5"
     mime-type))
 
+(defn- dissoc-track
+  "Migrate v1.5 Track to v1.4 by dissociating it from HorizontalSpatialDomain if applicable"
+  [g]
+  (if (get-in g [:SpatialExtent :HorizontalSpatialDomain :Track])
```
chris-durbin: Not sure if track 0 is a thing, but to be safe I'd wrap this check in `some?`.

umm-spec-lib/src/cmr/umm_spec/umm_g/track.clj
```diff
@@ -0,0 +1,38 @@
+(ns cmr.umm-spec.umm-g.track
+  "Contains functions for parsing UMM-G JSON Track into umm-lib granule model
+   Track and generating UMM-G JSON Track from umm-lib granule model Track."
+  (:require
+   [cmr.umm.umm-granule :as g])
+  (:import cmr.umm.umm_granule.UmmGranule))
```
chris-durbin: Import style is now:
```
(:import
 (cmr.umm.umm_granule UmmGranule))
```
even for a single import.

umm-spec-lib/src/cmr/umm_spec/migration/version/granule.clj
ygliuvt: Track is a map. It will not be `0` or `false`.

umm-spec-lib/src/cmr/umm_spec/umm_g/track.clj
ygliuvt: Done.

PR #698, opened 2019-02-20
CMR-5397: Update jruby version to 9.2.6.0

CayvonH:


orbits-lib/project.clj
```diff
@@ -1,7 +1,7 @@
 (def jruby-version
   "The version of JRuby to use. This is the same as used in the echo orbits java package to prevent
```
chris-durbin: Should probably update this comment to say collection-renderer-lib.

PR #700, opened 2019-02-21
CMR-5511: Implemented Tabular Ascii format.

sxu123:


cmr-exchange/service-bridge/test/cmr/opendap/tests/system/app/sizing/tabular_ascii.clj
```diff
@@ -49,4 +49,10 @@
                            granule-id
                            variable-alias)
                    options)]
-    (is (string/includes? (:body response) "[tabular_ascii] format is not implemented yet for service type: [ESI]."))))
+    (is (= 200 (:status response)))
+    (is (= "cmr-service-bridge.v2.1; format=json"
+           (get-in response [:headers :cmr-media-type])))
+    (is (= [{:bytes 26
```
chris-durbin: I was expecting a much larger size - seems like this estimate should be fairly accurate based on the formula. Why is it only 26 bytes?

sxu123: In this variable, the alias I put in is "Test Alias 6", Dimensionality is 1. 
So, the size is (inc (count "Test Alias 6")) (inc Dimensionality) = 13 * 2 = 26.

sxu123: Note: We have made three assumptions in this approach, which needs to be verified in the near future:
  1. Different variable may have different dimensionality, but tabular ascii only groups
  the variables with the same dimensionalities together, if not we will need to use the max dimensionality.
  2. Data width will never be wider than the header width.
  3. Alias put in the variable is the same as what appears in the tabular ascii.

PR #704, opened 2019-02-25
CMR-5520: Standardize collection-pre-query fmt.

CayvonH:
collection-pre-query result format needs to be standard in order to extract concept ids in cmr.search.services.query-execution.has-granules-created-at-feature.
Otherwise the call to (map :concept-id collection-results) may return (nil) in the case of say umm_json which would require (map #(get-in % [:meta :concept-id]) collection-results).

### DIFF
--- a/search-app/src/cmr/search/services/query_execution/multi_part_query_feature_common.clj
+++ b/search-app/src/cmr/search/services/query_execution/multi_part_query_feature_common.clj
@@ -43,6 +43,7 @@
   (-> query
       (dissoc :result-features)
       (assoc :page-size :unlimited
+             :result-format :xml
              :result-fields [(query-to-elastic/query-field->elastic-field
                               :concept-id :collection)])
       query-model/query
--- a/search-app/src/cmr/search/services/query_execution/multi_part_query_feature_common.clj
+++ b/search-app/src/cmr/search/services/query_execution/multi_part_query_feature_common.clj
@@ -41,8 +41,7 @@
   searched against."
   [query granule-field]
   (-> query
-      (dissoc :result-features)
-      (dissoc :result-format)
+      (dissoc :offset :page-num :page-size :result-features :result-format :scroll :sort-key)
       (assoc :page-size :unlimited
              :result-fields [(query-to-elastic/query-field->elastic-field
                               :concept-id :collection)])
--- a/search-app/src/cmr/search/services/query_execution/multi_part_query_feature_common.clj
+++ b/search-app/src/cmr/search/services/query_execution/multi_part_query_feature_common.clj
@@ -42,8 +42,8 @@
   [query granule-field]
   (-> query
       (dissoc :result-features)
+      (dissoc :result-format)
       (assoc :page-size :unlimited
-             :result-format :xml
              :result-fields [(query-to-elastic/query-field->elastic-field
                               :concept-id :collection)])
       query-model/query
--- a/system-int-test/src/cmr/system_int_test/utils/search_util.clj
+++ b/system-int-test/src/cmr/system_int_test/utils/search_util.clj
@@ -741,3 +741,34 @@
         (tk/clear-current-time!)
         ;; Resume time on CMR side
         (side/eval-form `(tk/clear-current-time!))))))
+
+(defn- parse-xml-concept-ids
+  "Parse concept-ids from xml structure. Use 2-arity to extract from attributes.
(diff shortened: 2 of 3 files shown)


system-int-test/test/cmr/system_int_test/search/collection_with_new_granule_search_test.clj
```diff
@@ -193,8 +193,12 @@
       (testing "has_granules_created_at parameter by itself"
         (util/are3
           [date-ranges expected-results]
-          (let [actual-results (search/find-refs :collection {:has-granules-created-at date-ranges})]
-            (d/assert-refs-match expected-results actual-results))
+          (let [params {:has-granules-created-at date-ranges}
+                json-search-results (search/find-concepts-json :collection params)
+                actual-results (search/find-refs :collection params)]
+            (d/assert-refs-match expected-results actual-results)
+            (is (= (set (map :concept-id expected-results))
```
chris-durbin: Since the issue seems to be related to the requested output format we should add a test for each supported output format.

Each test could be wrapped in a `(testing "has_granules_created_at in JSON format" ... )`.

search-app/src/cmr/search/services/query_execution/multi_part_query_feature_common.clj
```diff
@@ -42,6 +42,7 @@
   [query granule-field]
   (-> query
       (dissoc :result-features)
+      (dissoc :result-format)
```
chris-durbin: I see the following in `cmr.common-app.services.search.params/default-parse-query-level-params`:
`(dissoc params :offset :page-size :page-num :result-format :scroll :sort-key)`

I'm having trouble following out how this all works and of course that function doesn't have a docstring, but should we be removing all of these parameters for this query as well?

system-int-test/test/cmr/system_int_test/search/collection_with_new_granule_search_test.clj
```diff
@@ -297,7 +297,26 @@
           "Keyword"
           {:keyword "PROV1"} [coll-temporal-match coll-temporal-no-match
                               coll-spatial-match coll-w-june-2016-granule
-                              coll-w-may-2010-granule coll-platform-match])))))
+                              coll-w-may-2010-granule coll-platform-match]))
+
+      (testing "has_granules_created_at success in mutliple formats with sort"
+        (doseq [format [:atom :dif :dif10 :echo10 :iso19115 :json :opendata :xml :umm-json]]
```
chris-durbin: You can add a testing after the `doseq` as well so that when a test fails you know exactly which format failed.

```
(testing (str "format: " format)
...
```
The error report will be "has_granules_created_at success in mutliple formats with sort format :atom"

system-int-test/test/cmr/system_int_test/search/collection_with_new_granule_search_test.clj
```diff
@@ -297,7 +297,26 @@
           "Keyword"
           {:keyword "PROV1"} [coll-temporal-match coll-temporal-no-match
                               coll-spatial-match coll-w-june-2016-granule
-                              coll-w-may-2010-granule coll-platform-match])))))
+                              coll-w-may-2010-granule coll-platform-match]))
+
+      (testing "has_granules_created_at success in mutliple formats with sort"
```
chris-durbin: multiple typo

system-int-test/src/cmr/system_int_test/utils/search_util.clj
```diff
@@ -741,3 +741,34 @@
...
+  "Parse concept-ids from xml structure. Use 2-arity to extract from attributes.
+  3-arity from path."
+  ([response elements-path]
+   (map #(get-in % [:attrs :concept-id])
+        (cx/elements-at-path (fx/parse-str response) elements-path)))
+  ([response elements-path concept-id-path]
+   (map #(cx/string-at-path % concept-id-path)
+        (cx/elements-at-path (fx/parse-str response) elements-path))))
+
+(defn search-concept-ids-in-format
```
chris-durbin: Nice

system-int-test/test/cmr/system_int_test/search/collection_with_new_granule_search_test.clj
CayvonH: For some reason I didn't even realize this was a possibility, very cool.

PR #714, opened 2019-03-12
CMR-5541: Use URS4 for user-exists.

CayvonH:
* Use URS4 to retrieve user info.
* Change mock echo to work with new urs4 call.


CayvonH:
need to really clean up mock-urs. I just added the new routes to get a working implementation going.

transmit-lib/src/cmr/transmit/config.clj
```diff
@@ -36,6 +36,10 @@
   "The test user's password"
   "Password101")
 
+(def mock-urs-token
```
chris-durbin: I wouldn't put test code into the library - why is this needed? It looks like it's being checked in `assert-bearer-token`, but I don't entirely follow.

I think what's supposed to happen is transmit calls URS to get a bearer token, then in subsequent calls it passes that bearer token in.

So mock-urs should be the one that has some mock-token rather than the transmit lib.

transmit-lib/src/cmr/transmit/urs.clj
```diff
@@ -1,82 +1,63 @@
...
-(defn login-url
+(defn- login-application-url
   [conn]
-  (format "%s/login" (conn/root-url conn)))
+  (format "%s/oauth/token?grant_type=client_credentials" (conn/root-url conn)))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; Request functions
 
+(defn- get-bearer-token
```
chris-durbin: Were you able to test login and get user URL successfully against SIT URS?

CayvonH: Locally yes, I need to change some SIT configs when this goes there though.

transmit-lib/src/cmr/transmit/urs.clj
```diff
@@ -1,82 +1,63 @@
...
-                              :raw? true
-                              :http-options {:basic-auth [(config/urs-username) (config/urs-password)]
-                                             :body (login-xml username password)
-                                             :throw-exceptions false
-                                             :content-type mt/xml}})]
-     (if raw?
-       response
-       (assert-login-response-successful (:status response)(:body response))))))
-
+       (not (nil? response))))))
```
ygliuvt: I don't think this check will work with user that does not exist in URS. Can you test with retrieving a user that does not exist in URS and fix the code to handle it?

CayvonH: This check works because when we do `cmr.transmit.http-helper/request` that will process the response in `cmr.transmit.http-helper/handle-raw-fetch-or-delete-response` (when `raw`? is `true`) which returns `nil` on a 404 status response:

```
(defn- handle-raw-fetch-or-delete-response
  "Used to convert a raw response when fetching or deleting something into the actual result."
  [{:keys [status body] :as resp}]
  (cond
    (<= 200 status 299) body
    (= status 404) nil
    :else (errors/internal-error!
            (format "Received unexpected status code %s with response %s"
                    status (pr-str resp)))))
```

Here is the response locally:
`curl -i "http://localhost:3011/acls?permitted_user=fgkndfsgsdvsdfv"`
```
HTTP/1.1 400 Bad Request
Date: Wed, 13 Mar 2019 12:48:47 GMT
Content-Type: application/json
Access-Control-Allow-Origin: *
Content-Length: 65
Server: Jetty(9.4.14.v20181114)

{"errors":["The following users do not exist [fgkndfsgsdvsdfv]"]}
```

ygliuvt: It is interesting that you are calling a different endpoint from legacy services. I didn't know that URS4 actually has a `verify_uid` endpoint. My comment is based on the users endpoint used by legacy services. You can ignore it now.

ygliuvt: Ideally, we should add a test for non-existent user.

system-int-test/test/cmr/system_int_test/transmit/urs_test.clj
```diff
@@ -0,0 +1,48 @@
...
+    (transmit-config/set-urs-relative-root-url! saved-relative-root-url)))
+
+(use-fixtures :once urs-relative-root-url-fixture)
+
+(deftest urs-user-exists-test
+  (mock-urs-client/create-users (system/context)
+                                [{:username "urs-user" :password "pass"}])
+
+
+  (println "URS DEBUG INFO ------------")
```
chris-durbin: I assume you'll remove this debug.

PR #715, opened 2019-03-14
CMR-5547: Added support for UMM-Var v1.5

ygliuvt:


umm-spec-lib/src/cmr/umm_spec/models/umm_variable_models.clj
```diff
@@ -110,6 +116,28 @@
   ])
 (record-pretty-printer/enable-record-pretty-printing SamplingIdentifierType)
 
+;; Enables specification of Earth science keywords related to the collection. The Earth Science
+;; keywords are chosen from a controlled keyword hierarchy maintained in the Keyword Management
+;; System (KMS). The valid values can be found at the KMS website:
+;; https://gcmdservices.gsfc.nasa.gov/kms/concepts/concept_scheme/sciencekeywords?format=csv.
+(defrecord ScienceKeywordType
```
chris-durbin: I didn't see Science Keywords added in this PR. Looks like maybe the change was just to introduce this type. Should we add any science keywords tests as a result of this change, or will that be covered by the generated tests?

ygliuvt: ScienceKeywordType is not new. It is moved from the common schema. v1.5 moved a couple common schema types into the UMM-Var schema and deleted the common schema reference from UMM-Var.

PR #720, opened 2019-03-21
CMR-5291: As a metadata provider, I want CMR to validate variable name uniqueness.

ygliuvt:
Design approach is at: https://wiki.earthdata.nasa.gov/pages/viewpage.action?pageId=133744932

### DIFF
--- a/ingest-app/project.clj
+++ b/ingest-app/project.clj
@@ -12,6 +12,7 @@
                  [commons-codec/commons-codec "1.11"]
                  [commons-io "2.6"]
                  [compojure "1.6.1"]
+                 [digest "1.4.8"]
                  [drift "1.5.3"]
                  [gov.nasa.earthdata/cmr-site-templates "0.1.1-SNAPSHOT"]
                  [instaparse "1.4.10"]
--- a/ingest-app/src/cmr/ingest/services/ingest_service/variable.clj
+++ b/ingest-app/src/cmr/ingest/services/ingest_service/variable.clj
@@ -1,15 +1,39 @@
 (ns cmr.ingest.services.ingest-service.variable
   (:require
+   [cheshire.core :as json]
+   [clojure.string :as string]
    [cmr.common.util :refer [defn-timed]]
    [cmr.transmit.metadata-db2 :as mdb2]
-   [cmr.umm-spec.umm-spec-core :as spec]))
+   [cmr.umm-spec.umm-spec-core :as spec]
+   [digest]))
+
+(defn- normalized-string
+  "Returns the given string with leading and trailing whitespaces trimmed
+  and converted to lowercase."
+  [value]
+  (some-> value
+          string/trim
+          string/lower-case))
+
+(defn- get-variable-fingerprint
+  "Returns the fingerprint of the given variable concept."
+  [variable-concept]
+  (let [parsed (json/decode (:metadata variable-concept) true)
+        {:keys [AcquisitionSourceName Name Units Dimensions]} parsed
+        id-string (format "%s|%s|%s|%s"
+                          (normalized-string AcquisitionSourceName)
+                          (normalized-string Name)
+                          (normalized-string Units)
+                          (pr-str (mapv #(update % :Name normalized-string) Dimensions)))]
+    (digest/md5 id-string)))
 
 (defn add-extra-fields-for-variable
   "Returns collection concept with fields necessary for ingest into metadata db
   under :extra-fields."
   [context concept variable]
   (assoc concept :extra-fields {:variable-name (:Name variable)
-                                :measurement (:LongName variable)}))
+                                :measurement (:LongName variable)
+                                :fingerprint (get-variable-fingerprint concept)}))
 
 (defn-timed save-variable
   "Store a variable concept in mdb and indexer. Return name, long-name, concept-id, and
--- a/ingest-app/src/cmr/ingest/services/ingest_service/variable.clj
+++ b/ingest-app/src/cmr/ingest/services/ingest_service/variable.clj
@@ -5,7 +5,7 @@
    [cmr.common.util :refer [defn-timed]]
    [cmr.transmit.metadata-db2 :as mdb2]
    [cmr.umm-spec.umm-spec-core :as spec]
-   [digest]))
+   [digest :as digest]))
 
 (defn- normalized-string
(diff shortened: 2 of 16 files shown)


ingest-app/project.clj
```diff
@@ -12,6 +12,7 @@
                  [commons-codec/commons-codec "1.11"]
                  [commons-io "2.6"]
                  [compojure "1.6.1"]
+                 [digest "1.4.8"]
```
chris-durbin: I think we should just define the md5 function in `cmr.common.util` rather than bringing in another dependency.

```
(defn md5
  "Return the md5 for a string. See https://gist.github.com/jizhang/4325757."
  [^String s]
  (->> s
       .getBytes
       (.digest (MessageDigest/getInstance "MD5"))
       (BigInteger. 1)
       (format "%032x")))
```

ingest-app/src/cmr/ingest/services/ingest_service/variable.clj
```diff
@@ -1,15 +1,39 @@
 (ns cmr.ingest.services.ingest-service.variable
   (:require
+   [cheshire.core :as json]
+   [clojure.string :as string]
    [cmr.common.util :refer [defn-timed]]
    [cmr.transmit.metadata-db2 :as mdb2]
-   [cmr.umm-spec.umm-spec-core :as spec]))
+   [cmr.umm-spec.umm-spec-core :as spec]
+   [digest]))
```
chris-durbin: We should probably have `[digest :as digest]` even though that seems redundant just to be consistent. But anyway I think this should just go away and define the md5 function ourselves.

ingest-app/src/cmr/ingest/services/ingest_service/variable.clj
```diff
@@ -1,15 +1,39 @@
...
+(defn- get-variable-fingerprint
+  "Returns the fingerprint of the given variable concept."
+  [variable-concept]
+  (let [parsed (json/decode (:metadata variable-concept) true)
+        {:keys [AcquisitionSourceName Name Units Dimensions]} parsed
+        id-string (format "%s|%s|%s|%s"
+                          (normalized-string AcquisitionSourceName)
+                          (normalized-string Name)
+                          (normalized-string Units)
+                          (pr-str (mapv #(update % :Name normalized-string) Dimensions)))]
```
chris-durbin: Are we 100% that we should consider Dimension order significant? I guess I will follow up with Jon Pals.

metadata-db-app/src/migrations/068_add_fingerprint_to_variables_table.clj
```diff
@@ -0,0 +1,17 @@
...
+  "Adds fingerprint to variables table"
+  (:require
+   [config.mdb-migrate-helper :as h]))
+
+(defn up
+  "Migrates the database up to version 68."
+  []
+  (println "migrations.068-add-fingerprint-to-variables-table up...")
+  (h/sql "alter table cmr_variables add fingerprint VARCHAR(64)")
+  (h/sql "CREATE INDEX variables_FPI ON cmr_variables(fingerprint)"))
```
chris-durbin: I can't tell if the unique-field constraint provides the provider in the query. Can you add some debug logging and run a test against external Oracle?

The SQL seems to be generated in the end by `gen-find-concepts-in-table-sql` in `cmr.metadata-db.data.oracle.search`.

I just want to make sure the index we add matches that query.

system-int-test/test/cmr/system_int_test/ingest/variable_ingest_test.clj
```diff
@@ -9,23 +9,82 @@
...
+                                              concept
+                                              (variable-util/token-opts token))
+            var-concept-id concept-id]
         (is (mdb/concept-exists-in-mdb? concept-id revision-id))
-        (is (= 1 revision-id))))
+        (is (= 1 revision-id))
+        (is (= "7bec71103f3eada1f539acb9f1072fd7"
+               (get-in (mdb/get-concept concept-id revision-id) [:extra-fields :fingerprint])))
+
+        (testing "ingest the same concept on a different provider is OK"
```
chris-durbin: Right - so this test seems to confirm that the unique-field constraint query is using the provider.

ygliuvt: Good point. I have added provider_id to the index.

ingest-app/project.clj
ygliuvt: I don't see the benefit of putting one off functions in `cmr.common.util`. The `clj-digest` library is a cleaner way to provide that functionality. Also, I have thought about using some digest other than md5 (e.g. Sha1 or sha-256 as they suffer less collision) and that is why I have made the fingerprint column in CMR_VARIABLES a VARCHAR(64) rather than VARCHAR(32). Using the `clj-digest` library will make the switch easy if desired.

ingest-app/src/cmr/ingest/services/ingest_service/variable.clj
ygliuvt: Done.

ingest-app/src/cmr/ingest/services/ingest_service/variable.clj
ygliuvt: I went with Jon's suggestion because: 1) Jon said that the order is significant. 2) Based on Chris L.'s comments, the variables will almost always be auto-generated which makes the change of the Dimension objects orders more unlikely. I have sent Jon an email for confirmation again, but he is off for the next couple of days. We might have to add more aliases to Dimension Names anyways based on Jon's comments. I expect that we have to change the fingerprint algorithm in the future even though that is something that we don't want to do.

ingest-app/project.clj
chris-durbin: Alright if no vulnerabilities are found I won't fight you (`lein check-sec` runs successfully).

ygliuvt: Thanks! if anything, I think `clj-digest` would be more secure than rolling our own, and it is such a simple small wrapper lib, we are not pulling in much.

metadata-db-app/src/migrations/068_add_fingerprint_to_variables_table.clj
ygliuvt: Yes, you are right. The query does include provider_id. Thanks!

PR #724, opened 2019-03-25
CMR-5588: Store size as double.

CayvonH:


CayvonH:
Wasn't sure if a test case was really necessary for this.

chris-durbin:
Make sure to update the release notes page to indicate reindexing all the granules is required for this release.

ygliuvt:
Three points:
1) I think a test is necessary and should be easy to add. Just change the size value in one of the search in ATOM format tests, we should be able to reproduce the problem and verify the fix works.
2) I think we could get both the elasticsearch index change and code fix deployed in the one sprint. 
We can retrieve both `size-double-doc-values` and `size-doc-values` in ATOM result handler, and use `size-double-doc-values` if it is present, otherwise use `size-doc-values`. After the granule reindexing is complete, we can make another code change to drop `size-doc-values` from both the elasticsearch index and code.
3) We should probably change the size field in elasticsearch to a `double` as well to improve search accuracy. We have to do that in two sprints. First add `size-double` in this sprint. Once the granule reindexing is complete, make the code change to use `size-double` in the next sprint

chris-durbin:
Agree with Leo's point 1.

I thought about point 2 as well, but ruled that out too quickly as being complicated. Giving it some more thought, I think Leo's right, and this is how we should handle all of these granule field migrations in the future when the field is just something we return in the results (we can't do this for a field that we search against). One more added benefit is that it allows us to test the search change when we add the field, otherwise we might miss something in the sprint we add the indexing change.

I don't follow point 3.
> We should probably change the size field in elasticsearch to a double as well to improve search accuracy. 

The original `size` field is obsolete and should not be used anywhere. `size-doc-values` is the second iteration of the field and the one we are replacing now. `size-double-doc-values` is the third iteration of the field. Am I misunderstanding?

ygliuvt:
Ignore my third point. We don't support search granule by granule size, so there is no search accuracy to improve. :) I didn't even know that the `size` field is obsolete. I have thought that it was used for searching. We should remove these obsolete elasticsearch fields from our codebase.

PR #725, opened 2019-03-26
CMR-5537

ygliuvt:
As the UVG, I need to be able to find UMM-VAR by instruments.

### DIFF
--- a/indexer-app/test/cmr/indexer/test/data/concepts/keyword_util.clj
+++ b/indexer-app/test/cmr/indexer/test/data/concepts/keyword_util.clj
@@ -319,87 +319,87 @@
                                              :MiddleName "D."
                                              :LastName "Eve"}]}]})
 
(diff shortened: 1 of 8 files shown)


indexer-app/test/cmr/indexer/test/data/concepts/keyword_util.clj
```diff
@@ -319,87 +319,87 @@
                                              :MiddleName "D."
                                              :LastName "Eve"}]}]})
 
-(deftest fields->fn-mapper-single-valued
+(deftest extract-field-single-valued
   (is (= "An abstract summary"
```
indiejames: It seems like it might be a good idea to replace all these `is` calls with `are2`.

ygliuvt: Yeah, that is a good idea.

PR #728, opened 2019-03-28
CMR-5526: Changed the page_size to 2000.

sxu123:
This is to account for:
1. Fix the group node wildcard case, where (count [/a/b/* /a/b]) is obviously not right for the page_size.
2. For normal variable concept-id case, where page_size = (count variable-info) is correct, but it can't be more than 2000. 

Tried to implement scrolling in the event when the returning number of variables > 2000, but the response in the code doesn't return any header info where CMR-Scroll-Id can be used for subsequent calls. I think some changes need to be made in the http code but it doesn't look intuitive to me.  Since the existing code before this ticket used (count variable-info) for page_size, where it didn't take into account of 2000+ case either. So, it's an existing bug. I think perhaps I could just file a new ticket to fix it using scrolling, although it's unlikely that we will have more than 2000 variables per collection for subsetting.

### DIFF
--- a/cmr-exchange/metadata-proxy/project.clj
+++ b/cmr-exchange/metadata-proxy/project.clj
@@ -15,7 +15,7 @@
        ns
        "\u001B[35m]\u001B[33m λ\u001B[m=> "))
 
-(defproject gov.nasa.earthdata/cmr-metadata-proxy "0.2.5-SNAPSHOT"
+(defproject gov.nasa.earthdata/cmr-metadata-proxy "0.2.6-SNAPSHOT"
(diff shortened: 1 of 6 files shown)


sxu123:
> Can you add a test to verify this fixes the issue that Cayvon found? Also please write a ticket to fix the issue you found that is preventing scrolling from working for variables.

Sure. planning to. Just takes time to fabricate the Aliases so that it returns exactly the number of variables. Thanks.

sxu123:
Filed CMR-5655 for the issue. Searches for granules/services/variables in car-exchange all have the same issue.

PR #729, opened 2019-03-28
CMR-5308,CMR-5313,CMR-5324: Fix vulnerabilities.

CayvonH:
Vulnerability fixes for: bootstrap, dev-system, search-app:

* Suppressed everything in dev-system.
* suppressed guava vulnerability. Only used for validation of ip addresses in a subdependency of swagger.

### DIFF
--- a/bootstrap-app/resources/security/suppression.xml
+++ b/bootstrap-app/resources/security/suppression.xml
@@ -1,0 +1,106 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--For non-false positives suppress with <suppress until="YYYY-MM-DD">...-->
+<suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.2.xsd">
+  <suppress>
+    <notes><![CDATA[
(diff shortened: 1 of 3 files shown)


bootstrap-app/resources/security/suppression.xml
```diff
@@ -0,0 +1,106 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
indiejames: I don't suppose there is a JSON option for configuring the suppression?

CayvonH: Sadly I don't think so. The xml schema is so small it really isn't that bad once you get used to it.

https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.2.xsd
https://jeremylong.github.io/DependencyCheck/general/suppression.html

chris-durbin: https://jeremylong.github.io/DependencyCheck/general/suppression.html

There is at least a button you can click to generate the XML for you from the report.

PR #733, opened 2019-04-03
CMR-5534: Create about page.

CayvonH:
Overview copied from:
https://developer.earthdata.nasa.gov/cmr/overview

Taking suggestions for additions / design...


chris-durbin:
Can you paste a screenshot of the about page for one of the apps?

CayvonH:
Screen Shot 2019-04-03 at 5 04 11 PM
I'm not the greatest at design but I could _try_ to make it prettier / add any details you think we might want. 

For reference here are some pages about CMR:
https://earthdata.nasa.gov/about/science-system-description/eosdis-components/common-metadata-repository
https://developer.earthdata.nasa.gov/cmr/overview

chris-durbin:
We could probably add links to Earthdata Search and Worldview where they are mentioned at the end of the text. Remove the word Client from Earthdata Search. We talk about collections and granules as the two main concepts, but now we also have variables and services. It might be work mentioning them. Change UMMC to UMM-C. I'm fine with the way the page is displayed.

PR #735, opened 2019-04-04
CMR-4904 Remove DistributionType/FileSizeType

daniel-zamora:


ygliuvt:
Hmm...UMM-C v1.13 has been enabled for ingest for 3 weeks. There probably already are UMM-C v1.13 records with DistributionType exist in CMR. They will become invalid with this change. So question:
1) Should we make this a new UMM-C version?
2) If we don't make a new version and make this part of UMM-C v1.13, what is the plan to clean up the existing bad records?

PR #738, opened 2019-04-08
Mask tokens in error messages

abarciauskas-bgse:
Proposed changes:
* Mask token in error messages in transmit echo rest
* Mask token in common API errors module

TODOs:
- [ ] A few more unit tests
- [ ] Passing unit tests

### DIFF
--- a/common-lib/test/cmr/common/test/api/errors.clj
+++ b/common-lib/test/cmr/common/test/api/errors.clj
@@ -1,9 +1,6 @@
-(ns cmr.common.test.api.web-server
-  "This tests capabilities of the web server component."
+(ns cmr.common.test.api.errors
+  "This tests capabilities of the API error utilities."
   (:require
-   [clj-http.client :as h]
-   [clojure.java.io :as io]
-   [clojure.string :as str]
    [clojure.test :refer :all]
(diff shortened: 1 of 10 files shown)


common-lib/test/cmr/common/test/api/errors.clj
```diff
@@ -0,0 +1,12 @@
...
+  (:require
+   [clj-http.client :as h]
+   [clojure.java.io :as io]
+   [clojure.string :as str]
+   [clojure.test :refer :all]
+   [cmr.common.api.errors :as errors]))
+
+(deftest test-error-masked
+  (let [error-message "Token 123 does not exist"]
+  (is (= "Token does not exist" (errors/mask-token-error error-message)))))
```
chris-durbin: Nit: indent so the `is` is within the `let`:
```
(let [error-message "Token 123 does not exist"]
  (is (= "Token does not exist" (errors/mask-token-error error-message)))))
```

abarciauskas-bgse: good call - I should have noticed that! https://github.com/nasa/Common-Metadata-Repository/pull/738/commits/2df88609ba6592a005da00251391fcb3e2364ff6

PR #746, opened 2019-04-16
CMR-5691: Allow EDSC search with JSON requests through CORS.

ygliuvt:


CayvonH:
nit: maybe add a test to `dev-system/resources/cors_headers_test.html`

PR #747, opened 2019-04-16
CMR-5665: Remove the use of AverageSizeOfGranulesSampled in SES.

sxu123:
Marty's study shows that a simple (total-bytes * avg-compression-rate) is more accurate than Bob's approach of using (total-bytes * arg-compression-rate * ((total-bytes/gran-count)/AverageSizeOfGrnaulesSamples))).

### DIFF
--- a/cmr-exchange/ses-plugin/src/cmr/sizing/formats.clj
+++ b/cmr-exchange/ses-plugin/src/cmr/sizing/formats.clj
@@ -90,13 +90,6 @@
     (* total-dimensionality
        (max 2 (util/data-type->bytes data-type)))))
 
-(defn- get-avg-gran-size
-  "Gets SizeEstimation value for AverageSizeOfGranulesSampled and parses it to a number."
-  [variable]
-  (-> variable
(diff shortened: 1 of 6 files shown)


cmr-exchange/ses-plugin/src/cmr/sizing/formats.clj
```diff
@@ -90,13 +90,6 @@
     (* total-dimensionality
        (max 2 (util/data-type->bytes data-type)))))
 
-(defn- get-avg-gran-size
-  "Gets SizeEstimation value for AverageSizeOfGranulesSampled and parses it to a number."
-  [variable]
-  (-> variable
-      (get-in [:umm :SizeEstimation :AverageSizeOfGranulesSampled])
```
eereiter: Can I get rid of AverageSizeOfGranulesSampled from the UMM-Var schema since it is no longer being used?

sxu123: Yes.

PR #752, opened 2019-05-03
CMR-5695: Recreate SNS and SQS clients when they encounter an error

chris-durbin:
This PR makes changes our SNS and SQS client connections to be atoms so that if they encounter an unrecoverable error, we will recreate the client and all future requests will use the new client.

I made some changes to our dev-system setup so that testing with :aws as the :messaging component will work. I reproduced the same problem we saw in operations and verified that the connection was recreated successfully and all subsequent ingests worked correctly (messages made it to the SNS exchange and were picked up and indexed successfully).

The way I reproduced the problem was running the following in the REPL:
```(-> user/system :apps :metadata-db :queue-broker :sns-client-atom deref .shutdown)```

I'm also going to run a workload run before merging this PR in.

### DIFF
--- a/message-queue-lib/src/cmr/message_queue/queue/sqs.clj
+++ b/message-queue-lib/src/cmr/message_queue/queue/sqs.clj
@@ -167,6 +167,66 @@
                      (into {}))]
    (get q-attrs queue-arn-attribute)))
 
+
+(defn set-aws-client-attr!
+  "Conditionally configure the AWS client instance with by calling the given
+  method with the given args."
+  [client-obj method args]
+  (when-not (nil? (first args))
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: message-queue-lib/src/cmr/message_queue/queue/sqs.clj [hunk 1]
```diff
@@ -167,44 +167,109 @@
                      (into {}))]
    (get q-attrs queue-arn-attribute)))
 
+
+(defn set-aws-client-attr!
+  "Conditionally configure the AWS client instance with by calling the given
```
ygliuvt: nit: remove the first `with`.

PR #754, opened 2019-05-08
CMR-5705: Added collection temporal subsetting indication

sxu123:


search-app/docs/api.md
```diff
@@ -921,6 +923,7 @@ __Example__
       "has-formats" : false,
       "has-transforms" : false,
       "has-spatial-subsetting": false
```
ygliuvt: Add a `,` as the end of the line.

search-app/docs/api.md
```diff
@@ -805,6 +805,7 @@ __Example__
       "has_formats" : false,
```
chris-durbin: We should document this field in the ATOM response section.
Look for this in the documentation to see where it goes:
echo:hasSpatialSubsetting (collections only) | true if any of its associated services support spatial subsetting

PR #759, opened 2019-06-04
Adding debug to see which collecion/revision are being reindexed.

sxu123:
1. Check to see if the collection/revisions are fetched from metadata_db correctly.
2. If so, and the bulk_index in the indexer doesn't return error, then that means the bulk reindexing is not working correctly.

### DIFF
--- a/indexer-app/src/cmr/indexer/services/index_service.clj
+++ b/indexer-app/src/cmr/indexer/services/index_service.clj
@@ -123,11 +123,11 @@
   ([context concept-batches]
    (bulk-index context concept-batches nil))
   ([context concept-batches options]
+   (doseq [batch concept-batches]
+     (doseq [concept batch]
+       (let [cptinfo (str (:concept-id concept) "/" (:revision-id concept))]
+              (info "Reindexing concept-id/revision-id: " cptinfo))))
    (reduce (fn [num-indexed batch]
-             (map (fn [concept] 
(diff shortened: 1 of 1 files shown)


indexer-app/src/cmr/indexer/services/index_service.clj
```diff
@@ -123,6 +123,10 @@
   ([context concept-batches]
    (bulk-index context concept-batches nil))
   ([context concept-batches options]
+   (doseq [batch concept-batches]
+     (doseq [concept batch]
+       (let [cptinfo (str (:concept-id concept) "/" (:revision-id concept))]
+              (info "Reindexing concept-id/revision-id: " cptinfo))))
```
sxu123: Tested locally with 4 collections and each with 2 revisions.
curl -i -XPOST -H "Echo-Token: mock-echo-system-token" http://localhost:3002/jobs/reindex-all-collections
All the collection revisions appear in the concept-batches correctly.

sxu123: Also tested with bootstrap:
curl -i -XPOST -H "Content-Type: application/json" -H "Echo-Token: mock-echo-system-token" -d '{"provider_id": "PROV1", "collection_id":"C1200000005-PROV1"}' "http://localhost:3006/bulk_index/collections"
But it only shows "Adding collection C1200000005-PROV1 to collection index channel". Not sure how long to wait before it's processed and if it's going through the same code here.

PR #764, opened 2019-06-13
CMR-5747: Fixed ignore_conflict parameter.

sxu123:
This has been tested in SIT. Basically what gets selected in index-svc/index-concept-by-concept-id-revision-id is :ignore-conflict.

### DIFF
--- a/indexer-app/src/cmr/indexer/api/routes.clj
+++ b/indexer-app/src/cmr/indexer/api/routes.clj
@@ -47,7 +47,7 @@
       ;; Index a concept
       (POST "/" {body :body context :request-context params :params headers :headers}
         (let [{:keys [concept-id revision-id]} (walk/keywordize-keys body)
-              options {:ignore_conflict? (ignore-conflict? params)}]
+              options {:ignore-conflict? (ignore-conflict? params)}]
           ;; indexing all revisions index, does nothing for concept types that do not support all revisions index
           (index-svc/index-concept-by-concept-id-revision-id
             context concept-id revision-id (assoc options :all-revisions-index? true))
--- a/system-int-test/src/cmr/system_int_test/utils/index_util.clj
+++ b/system-int-test/src/cmr/system_int_test/utils/index_util.clj
@@ -70,6 +70,41 @@
                     :throw-exceptions false})]
     (is (= 200 (:status response)) (:body response))))
 
+(defn reindex-concept-ignore-conflict-default
+  "Re-index concept"
+  [concept-id revision-id]
+  (let [response (client/post (url/indexer-url)
+                   {:connection-manager (s/conn-mgr)
+                    :headers {transmit-config/token-header (transmit-config/echo-system-token)
+                             "content-type" "application/json"}
+                    :throw-exceptions false
+                    :body (str "{\"concept-id\": " "\"" concept-id "\", " "\"revision-id\": " "\"" revision-id "\"" "}")})]
+    (is (= 201 (:status response)) (:body response))))
+
+(defn reindex-concept-ignore-conflict-true
+  "Re-index concept"
+  [concept-id revision-id]
+  (let [response (client/post (url/indexer-url)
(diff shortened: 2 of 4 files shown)


indexer-app/src/cmr/indexer/api/routes.clj
```diff
@@ -47,7 +47,7 @@
       ;; Index a concept
       (POST "/" {body :body context :request-context params :params headers :headers}
         (let [{:keys [concept-id revision-id]} (walk/keywordize-keys body)
-              options {:ignore_conflict? (ignore-conflict? params)}]
+              options {:ignore-conflict? (ignore-conflict? params)}]
```
chris-durbin: We should be able to test this with an integration test.

1.) Ingest a collection and then update that collection so that you have two revisions.
2.) Call this endpoint to index revision 1 with ignore_conflict false and see that it returns a 409.
3.) Call this endpoint to index revision 1 with ignore_conflict true and see that it returns a 2xx status code.

sxu123: That's what I did in SIT.

system-int-test/src/cmr/system_int_test/utils/index_util.clj
```diff
@@ -70,6 +70,41 @@
                     :throw-exceptions false})]
     (is (= 200 (:status response)) (:body response))))
 
+(defn reindex-concept-ignore-conflict-default
+  "Re-index concept without passing in ignore_conflict param should default to conflict being ignored."
+  [concept-id revision-id]
```
chris-durbin: I'd just have one function that can take optional query parameters in the util namespace. It should just call the function to reindex the concept with whatever parameters you want to use and return the response object. It wouldn't call `is`.

Then in your tests you would call `is` with what you are expecting to happen for each test to test that the status code matches what you expected.

I like that you are passing in the body of the response to `is` as the last parameter so that if there's a failure it prints out some extra info - we don't do that enough.

ygliuvt: +1

system-int-test/test/cmr/system_int_test/ingest/collection_ingest_test.clj
```diff
@@ -55,6 +55,27 @@
...
+          revision-id2 (:revision-id ingested-concept2)]
+      (index/wait-until-indexed)
+      (is (mdb/concept-exists-in-mdb? concept-id1 revision-id1))
+      (is (mdb/concept-exists-in-mdb? concept-id2 revision-id2))
+      (is (= 1 revision-id1))
+      (is (= 2 revision-id2))
+      (is (= concept-id1 concept-id2))
+      (index/reindex-concept-ignore-conflict-default concept-id1 revision-id1)
+      (index/reindex-concept-ignore-conflict-true concept-id1 revision-id1)
+      (index/reindex-concept-ignore-conflict-false concept-id1 revision-id1)
```
ygliuvt: As Chris has pointed out, the reindex status check should happen here. The `util` namespace would only have a reindex function that takes the concept-id, revision-id and ignore-conflict params.

sxu123: I was thinking of the same thing actually. it's much more readable. Thanks.

system-int-test/src/cmr/system_int_test/utils/index_util.clj
```diff
@@ -70,6 +70,41 @@
...
+  (let [response (client/post (url/indexer-url)
+                   {:connection-manager (s/conn-mgr)
+                    :headers {transmit-config/token-header (transmit-config/echo-system-token)
+                             "content-type" "application/json"}
+                    :throw-exceptions false
+                    :body (str "{\"concept-id\": " "\"" concept-id "\", " "\"revision-id\": " "\"" revision-id "\"" "}")})]
```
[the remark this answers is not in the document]
sxu123: Thanks!

PR #765, opened 2019-06-14
CMR-5628 Dockerfile and project.clj updates for NGAP 2.0 deployment

daniel-zamora:
Will also fix builds that have been failing due to outdated versions.

### DIFF
--- a/cmr-exchange/service-bridge/Dockerfile
+++ b/cmr-exchange/service-bridge/Dockerfile
@@ -1,2 +1,9 @@
-FROM openjdk:8u162-jre-slim
-MAINTAINER <http://cmr.earthdata.nasa.gov/>
+FROM clojure
+
+COPY target/cmr-service-bridge-*-SNAPSHOT-standalone.jar /app/cmr-standalone.jar
+COPY resources/ /app/resources/
+EXPOSE 3013
+
+WORKDIR /app
+
+CMD "java" "-jar" "cmr-standalone.jar"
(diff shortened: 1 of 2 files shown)


cmr-exchange/service-bridge/Dockerfile
```diff
@@ -1,2 +1,9 @@
-FROM openjdk:8u162-jre-slim
-MAINTAINER <http://cmr.earthdata.nasa.gov/>
+FROM clojure
```
chris-durbin: We've switched to using:
FROM clojure:openjdk-8-lein

so that we don't have unexpected versions of Java.

cmr-exchange/service-bridge/Dockerfile
```diff
@@ -1,2 +1,9 @@
...
+
+COPY target/cmr-service-bridge-*-SNAPSHOT-standalone.jar /app/cmr-standalone.jar
```
chris-durbin: I would make sure to have service-bridge in the name of the jar.

cmr-exchange/service-bridge/Dockerfile
```diff
@@ -1,2 +1,9 @@
...
+COPY resources/ /app/resources/
+EXPOSE 3013
+
+WORKDIR /app
+
+CMD "java" "-jar" "cmr-standalone.jar"
```
chris-durbin: I think we've left out the command in our other Dockerfiles, not sure if we want to take this out for consistency with the other apps.

PR #782, opened 2019-07-17
CMR-5743: Fixed the relative path for css files.

sxu123:
On Chrome it complained about http(instead of https) access to the css files because it's using relative paths. Changed them to absolute paths.

### DIFF
--- a/collection-renderer-lib/resources/collection_preview/collection_preview.erb
+++ b/collection-renderer-lib/resources/collection_preview/collection_preview.erb
@@ -21,8 +21,8 @@ metadata =  JSON.parse(umm_json)
   <%= csrf_meta_tag %>
   <link href="https://cdn.earthdata.nasa.gov/eui/1.1.3/stylesheets/application.css" rel="stylesheet" />
   <link href="https://fonts.googleapis.com/css?family=Open+Sans:400,700,300" rel="stylesheet" type="text/css">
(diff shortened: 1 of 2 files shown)


collection-renderer-lib/resources/collection_preview/collection_preview.erb
```diff
@@ -21,8 +21,8 @@ metadata =  JSON.parse(umm_json)
   <%= csrf_meta_tag %>
   <link href="https://cdn.earthdata.nasa.gov/eui/1.1.3/stylesheets/application.css" rel="stylesheet" />
   <link href="https://fonts.googleapis.com/css?family=Open+Sans:400,700,300" rel="stylesheet" type="text/css">
-  <link href="../../../site/css/eui.css" rel="stylesheet" /> 
-  <link href="../../../site/css/cmr.css" rel="stylesheet" /> 
+  <link href="https://cmr.earthdata.nasa.gov:443/search/site/css/eui.css" rel="stylesheet" />
```
chris-durbin: This is always pointing to production rather than the environment where the application is running (so we wouldn't be able to promote stylesheet changes one environment at a time).

I think there's a way to still do what you want with an absolute path since this is an ERB file. You'll just need to use a configuration value for public CMR search root to substitute here.

PR #787, opened 2019-07-18
CMR-5781: fix image timeout issue

mschmele:


browse-scaler/src/index.js
```diff
@@ -43,9 +44,9 @@ const resizeImageFromConceptId = async (conceptType, conceptId, height, width) =
   // If given an image url, fetch the image and resize. If no image
   // exists, return the not found response
   const imageUrl = await getImageUrlFromConcept(conceptId, conceptType);
-  const imageBuffer = await slurpImageIntoBuffer(imageUrl);
+  const imageBuffer = await withTimeout(1500, slurpImageIntoBuffer(imageUrl));
```
chris-durbin: I'd recommend making it a Lambda configuration rather than a hardcoded number.

1.5 seconds is pretty short, not sure if you'll get false timeouts because of this, but we can test it in SIT.

PR #798, opened 2019-08-28
CMR-5851: add Echo-Token header to cmr requests

mschmele:


browse-scaler/src/cmr.js
```diff
@@ -1,9 +1,10 @@
 const fetch = require('node-fetch');
 
 /* CMR ENVIRONMENT VARIABLES */
-const cmrRootUrl = `http://${process.env.CMR_ROOT}`;
+const cmrRootUrl = `https://${process.env.CMR_ROOT}`;
```
CayvonH: `CMR_ROOT` points to the services load balancer? I don't think the lb can do https.

mschmele: Ah it does - I may want to rethink how I'm doing this then

PR #802, opened 2019-09-23
Cmr 5873 browse scaler throws 502

sxu123:
The issue was we set request timeout to be 3 seconds. When the image was not found, the time it took to get to that point was already very close to 3 seconds(through withTimeout function). so it didn't have enough time to complete notFound function before the request timed out, which threw 502.

1. Shortened the default timeoutInterval to be 1000, instead of 2000ms. 
2. Added withTimeout to getImageUrlFromConcept, just in case it takes too long.
3. Modified withTimeout function to handle the case when the promise is rejected.

### DIFF
--- a/browse-scaler/src/index.js
+++ b/browse-scaler/src/index.js
@@ -68,7 +68,7 @@ const resizeImageFromConceptId = async (conceptType, conceptId, height, width) =
   // If given an image url, fetch the image and resize. If no image
   // exists, return the not found response
   const imageUrl = await getImageUrlFromConcept(conceptId, conceptType);
-  console.log (`BEFORE SLURPING`)
+  console.log (`BEFORE SLURPING`);
   const imageBuffer = await withTimeout(timeoutInterval, slurpImageIntoBuffer(imageUrl));
   console.log (`AFTER SLURPING ${imageBuffer}`);
 
--- a/browse-scaler/src/index.js
+++ b/browse-scaler/src/index.js
@@ -71,6 +71,7 @@ const resizeImageFromConceptId = async (conceptType, conceptId, height, width) =
   const imageBuffer = await withTimeout(timeoutInterval, slurpImageIntoBuffer(imageUrl));
 
   if (imageUrl === null || imageBuffer === null || imageBuffer === undefined) {
+    console.log (`IMAGENOTRETURNED`);
     const imgNotFound = await notFound();
     return buildResponse(imgNotFound);
   }
--- a/browse-scaler/src/index.js
+++ b/browse-scaler/src/index.js
@@ -3,8 +3,8 @@ const { getCollectionLevelBrowseImage, getGranuleLevelBrowseImage } = require('.
 const { cacheImage, getImageFromCache } = require('./cache');
 const { withTimeout, slurpImageIntoBuffer } = require('./util');
 
-const timeoutInterval = 1000;
-
+const timeoutInterval = process.env.EXTERNAL_REQUEST_TIMEOUT || 1000;
+console.log(`TIMEOUT INTERVAL IS: ${timeoutInterval}`);
 /**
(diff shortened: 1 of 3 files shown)


browse-scaler/src/index.js
```diff
@@ -67,10 +67,16 @@ const resizeImageFromConceptId = async (conceptType, conceptId, height, width) =
 
   // If given an image url, fetch the image and resize. If no image
   // exists, return the not found response
-  const imageUrl = await getImageUrlFromConcept(conceptId, conceptType);
-  const imageBuffer = await withTimeout(timeoutInterval, slurpImageIntoBuffer(imageUrl));
+  const imageUrl = await withTimeout(timeoutInterval, getImageUrlFromConcept(conceptId, conceptType));
```
mschmele: This is a call to the CMR, which shouldn't ever time out unless we're experiencing degraded service. If that is the case, this may end up caching that an image is not found for a concept that does contain a browse image. I think this line should just be:
```const imageUrl = await getImageUrlFromConcept(conceptId, conceptType);```

Definitely keep the ```if (imageUrl === null) { ... }``` check though. That's good thinking

File: browse-scaler/src/index.js [hunk 2]
```diff
@@ -67,10 +67,16 @@ const resizeImageFromConceptId = async (conceptType, conceptId, height, width) =
...
+  if (imageUrl === null) {
+    console.log("IMAGE URL IS NOT RETURNED.");
```
mschmele: I think we should either remove this log statement or add the concept id to make this log message more useful. We can just use string templating like so: ```console.log(`No image url found for concept: ${conceptId]`);```

browse-scaler/src/index.js
```diff
@@ -67,10 +67,16 @@ const resizeImageFromConceptId = async (conceptType, conceptId, height, width) =
...
+    const imgNotFound = await notFound();
+    return buildResponse(imgNotFound);
+  }
 
-  if (imageUrl === null || imageBuffer === null || imageBuffer === undefined) {
+  const imageBuffer = await withTimeout(timeoutInterval, slurpImageIntoBuffer(imageUrl));
+  if (imageBuffer === null || imageBuffer === undefined) {
+    console.log("IMAGE IS NOT RETURNED.");
```
mschmele: Same as above log message

browse-scaler/src/index.js
```diff
@@ -3,8 +3,8 @@ const { getCollectionLevelBrowseImage, getGranuleLevelBrowseImage } = require('.
 const { cacheImage, getImageFromCache } = require('./cache');
 const { withTimeout, slurpImageIntoBuffer } = require('./util');
 
-const timeoutInterval = process.env.EXTERNAL_REQUEST_TIMEOUT || 2000;
-
+const timeoutInterval = process.env.EXTERNAL_REQUEST_TIMEOUT || 1000;
+console.log(`TIMEOUT INTERVAL IS: ${timeoutInterval}`);
```
mschmele: I don't think we need this log message

browse-scaler/src/index.js
mschmele: On second thought, it should be okay to leave this - not necessary, as I doubt the CMR will have queries that will ever take this long given a concept-id under normal conditions, but it shouldn't be a _problem_

browse-scaler/src/index.js
sxu123: I think it really helps the debugging - especially after we have the environment variable. if somewhere it times out again, we will need to know why.

browse-scaler/src/index.js
sxu123: This is added not because of concerning getImageUrlFromConcept will timeout, it's because we don't want any function to take too long. If we don't want to see 502, we have to make sure the request won't time out before all the functions are returned with either the image, or the no image found.

PR #806, opened 2019-10-08
CMR-5912: Add redis utils lib. Embedded redis.

CayvonH:
Added a redis-utils-lib which includes the ability to use an embedded redis. Added install script to retrieve the version of redis we want to use.

This PR does not change any functionality (currently) within CMR proper. 
This PR does not conclude the work for CMR-5912 but intended to split up changes rather than include everything in one massive PR.


redis-utils-lib/project.clj
```diff
@@ -0,0 +1,69 @@
+(def redis-distribution
+  "The version of redis to use and hash value of executable. See:
+  https://github.com/antirez/redis-hashes/blob/master/README"
+  {:version "3.2.10"
+   :hash "411c604a716104f7f5a326abfad32de9cea10f15f987bec45cf86f315e9e63a0"})
+
+(def redis-install-path
+  "The directory within this library where redis is to be installed"
+  "redis")
+
```
indiejames: Should these be `defconfig`s instead?

redis-utils-lib/src/cmr/redis_utils/embedded_redis_server.clj
```diff
@@ -0,0 +1,57 @@
...
+  (:require
+   [cmr.common.lifecycle :as lifecycle]
+   [cmr.common.log :refer [debug info]]
+   [cmr.redis-utils.config :as redis-config])
+  (:import
+   (redis.embedded RedisExecProvider
+                   RedisServer)
+   (redis.embedded.util OS)))
+
+(def ^:private redis-server-executable-path
```
indiejames: Should this use `redis-install-path` from `project.clj` instead of assuming `redis` as the install directory?

redis-utils-lib/dev/user.clj
```diff
@@ -0,0 +1,17 @@
+(ns user
```
indiejames: Why do we need a `user` namespace here?

redis-utils-lib/src/cmr/redis_utils/embedded_redis_server.clj
```diff
@@ -0,0 +1,57 @@
...
+      this))
+
+  (stop
+    [this system]
+    (when-let [^RedisServer redis-server (:redis-server this)]
+      (debug "Stopping redis server")
+      (.stop redis-server))
+    (assoc this :redis-server nil)))
+
+(defn create-redis-server
```
indiejames: Will you be adding this to `/dev/system` as well? Or did you need SOMEONE ELSE to issue PR for that? :)

redis-utils-lib/project.clj
CayvonH: defconfig is a macro in cmr common, I wouldn't want to use that in a `project.clj`

redis-utils-lib/src/cmr/redis_utils/embedded_redis_server.clj
CayvonH: Ok so my clojure is a bit rusty but I believe that the src namespace's have no knowledge of the `project.clj` defined variables.

redis-utils-lib/dev/user.clj
CayvonH: The repl loads the user namespace so we can load some convenience namespace's that we only want for dev.

redis-utils-lib/src/cmr/redis_utils/embedded_redis_server.clj
CayvonH: haha I actually have some stuff for dev-system already written but once that is done I do need some PRs for replacing the cubby routes / connection pooling.

redis-utils-lib/README.md
```diff
@@ -0,0 +1,7 @@
+# redis-utils-lib
+
+A Clojure library containing utilities for dealing with Redis.
+
+## Usage
+
+FIXME
```
ygliuvt: We need to put something in here to explain what this library is and how do we plan to use it.

CayvonH: Yeah I was going to add that when we get a little further along into the cubby replacement. I'm not entirely sure the full extent of uses yet.

PR #811, opened 2019-10-15
CMR-5495 Fix issue with translation endpoint causing 500 error.

daniel-zamora:
This was causing 500 errors in ingest when a user passed an accept header in the format "Accept: "

No test was added because I could not figure out how to simulate this with our internal http library and I didn't want to use (shell curl ...). Any suggestions appreciated.


daniel-zamora:
@ygliuvt The only other place in the code that curls in a test specifically states not to do this unless absolutely necessary.  Also, simply removing the header from the post call does not work, as stated in the description, you cannot reproduce this issue via our http library.  As to what the response is for this request, it will look like all the other accept header errors.

PR #818, opened 2019-10-21
CxMR-6024: fix and cleanup code to help improve new concept creation(Part1)

sxu123:


umm-spec-lib/src/cmr/umm_spec/migration/version/core.clj
```diff
@@ -15,11 +15,7 @@
    [cmr.umm-spec.migration.organization-personnel-migration :as op]
    [cmr.umm-spec.migration.related-url-migration :as related-url]
    [cmr.umm-spec.migration.spatial-extent-migration :as spatial-extent]
-   [cmr.umm-spec.migration.version.collection]
```
ygliuvt: I believe these imports are needed for loading in the concrete `defmethod` implementations.

sxu123: You're right! Added back.

PR #820, opened 2019-10-23
CMR-6016: Add timeout to Elasticsearch queries

indiejames:
This adds a timeout to the Elasticsearch queries to avoid 504 errors from CloudFront. This does not cause an error if the query times out - it simply returns whatever results are available when the timeout occurs.

Note that there are no additional (automated) tests for this as it is not possible to force a query timeout in Elasticsearch.

### DIFF
--- a/search-app/docs/api.md
+++ b/search-app/docs/api.md
@@ -14,6 +14,7 @@ Join the [CMR Client Developer Forum](https://wiki.earthdata.nasa.gov/display/CM
     * [Collection Result Feature Parameters](#collection-result-features)
     * [Headers](#headers)
     * [Extensions](#extensions)
+    * [Request Timeouts](#request-timeouts)
   * [Supported Result Formats](#supported-result-formats)
     * [ATOM](#atom)
     * [CSV](#csv)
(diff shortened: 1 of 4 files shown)


search-app/docs/api.md
```diff
@@ -290,6 +291,12 @@ Here is a list of supported extensions and their corresponding MimeTypes:
   * `umm_json_vX_Y` "application/vnd.nasa.cmr.umm_results+json; version=X.Y"
     * X and Y should be replaced with a major and minor number of the UMM version requested.
 
+#### <a name="request-timeouts"></a> Request Timeouts
+
+The CMR operating environment imposes a hard limit of 180 seconds on any request, after which a 504 error is
+returned. To avoid this, the CMR has an internal query timeout of 170 seconds - any query taking longer will time
+out and a subset of the total hit results will be returned instead of an error.
```
sxu123: Elastic search doesn't give any indication that it timed out?  How do we know that the result might only be the subsetl?

sxu123: Regarding not being able to force elastic search to timeout, did you mean that you can't set the timeout small enough for any normal queries to timeout?

chris-durbin: This is a good point - I think we need some way to communicate to a client that the results may potentially have been truncated. ECHO catalog-rest used to have an additional header that said whether results were exact/estimated (or something like that). I'm thinking a header may be a good way to indicate these are partial results due to a timeout.

indiejames: Regarding testing, the minimum timeout setting is 1 second, and it is a "best effort" timeout, so it might actually run longer. It is impractical to execute a query that takes longer than that in a test.

PR #821, opened 2019-10-25
CMR-5958: Fixed exception when ingest virtual granule for a deleted collection.

sxu123:
Will need to figure out how to test this in another PR. Unit test shows the re-matches worked 
for the exception error message in the ticket:

user=> (let [granule-ur "SC:AST_FRBT.003:2347928096"
  #_=>       pattern (re-pattern (str ".*Collection.*referenced in granule.*" granule-ur ".*does not exist.*"))
  #_=>       str "{:errors [\"Collection with Short Name [AST_FRBT], Version Id [003] referenced in granule [SC:AST_FRBT.003:2347928096] provider [LPDAAC_ECS] does not exist.\"]}"]
  #_=> (re-matches pattern str))
"{:errors [\"Collection with Short Name [AST_FRBT], Version Id [003] referenced in granule [SC:AST_FRBT.003:2347928096] provider [LPDAAC_ECS] does not exist.\"]}"

### DIFF
--- a/virtual-product-app/src/cmr/virtual_product/services/virtual_product_service.clj
+++ b/virtual-product-app/src/cmr/virtual_product/services/virtual_product_service.clj
@@ -76,7 +76,8 @@
   fall between 200 and 299 or not equal to 409 will cause an exception which in turn causes the
   corresponding queue event to be put back in the queue to be retried."
   [response granule-ur]
-  (let [{:keys [status body]} response]
+  (let [{:keys [status body]} response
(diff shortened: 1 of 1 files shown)


mschmele:
> I think the current CMR behavior is correct. No code change is needed. All we need to do is to ingest the virtual collection. It is a setup issue, not a coding issue.
> 
> This code change will cause CMR to simply ignore the virtual product configuration error when the virtual collection does not exist. This is not the desired behavior.

Code-wise I think this looks good. If the behavior is not what we want then I'm with Leo

PR #823, opened 2019-10-29
CMR-5920: validate granule data format with KMS values

mschmele:


common-app-lib/src/cmr/common_app/services/kms_fetcher.clj
```diff
@@ -74,6 +75,7 @@
   "Calls GCMD KMS endpoints to retrieve the keywords. Response is a map structured in the same way
   as used in the KMS cache."
   [context]
+  (def keyword-scheme :granule-data-format)
```
ygliuvt: Remove this line. Not sure why it is here.

ingest-app/src/cmr/ingest/services/messages.clj
```diff
@@ -49,6 +49,15 @@
    (format "Data center short name [%s] was not a valid keyword."
           (:ShortName data-center)))
 
+(defn data-format-not-matches-kms-keywords
+  "Error message when FileDistributionInformation -> Format is
+  not in the KMS."
+  [archive-and-distribution-information]
+  (format "Format [%s] within ArchiveAndDistributionInformation [%s] was not a valid keyword."
```
ygliuvt: I don't think it is a good error message. The ArchiveAndDistributionInformation might be too much for an error message.

ygliuvt: Can we add a test for validating granule data format with KMS values? Then we can see what this error message really looks like.

common-app-lib/src/cmr/common_app/services/kms_fetcher.clj
mschmele: It was for debugging purposes and I just missed it. Fixed

ingest-app/src/cmr/ingest/services/messages.clj
mschmele: Fixed and added tests

PR #824, opened 2019-11-01
CMR-6040

ygliuvt:
has_formats should only be true if the UMM-S record includes Output formats.

### DIFF
--- a/indexer-app/src/cmr/indexer/data/concepts/service.clj
+++ b/indexer-app/src/cmr/indexer/data/concepts/service.clj
@@ -72,11 +72,9 @@
   (let [service (concept-parser/parse-concept context service-concept)
         input-formats (distinct (get-in service [:ServiceOptions :SupportedInputFormats]))
         output-formats (distinct (get-in service [:ServiceOptions :SupportedOutputFormats]))]
-    (if (or (= (count output-formats) 0)
-            (and (= (count input-formats) (count output-formats) 1)
-                 (= input-formats output-formats)))
-      false
-      true)))
+    (not (or (= (count output-formats) 0)
+             (and (= (count input-formats) (count output-formats) 1)
+                  (= input-formats output-formats))))))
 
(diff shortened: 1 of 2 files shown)


indexer-app/src/cmr/indexer/data/concepts/service.clj
```diff
@@ -70,10 +70,13 @@
...
-                                     (get-in service [:ServiceOptions :SupportedInputFormats])
-                                     (get-in service [:ServiceOptions :SupportedOutputFormats])))]
-    (> (count supported-formats) 1)))
+        input-formats (distinct (get-in service [:ServiceOptions :SupportedInputFormats]))
+        output-formats (distinct (get-in service [:ServiceOptions :SupportedOutputFormats]))]
+    (if (or (= (count output-formats) 0)
+            (and (= (count input-formats) (count output-formats) 1)
+                 (= input-formats output-formats)))
+      false
+      true)))
```
indiejames: I don't follow the logic here. It seems like it's saying "if there are no output formats or there is one output format and one input format and those formats are the same then return false, otherwise return true." Is that correct?

indiejames: Also, you don't need to use an `if` here, you could just put it in a `not` or change the logic a bit.

indiejames: What should the response be if `(count input-formats)` equals 0?

ygliuvt: This is the directive I got from Erich:
```
All they need for now is for the has_format metadata element to be false if the there is no output format in the umm-s or the output format has one value and the input format also has one value and it is the same value as the output format. For anything else we will work with separate issues
```
I have asked many questions on the rule, but this is the direction I got for the current implementation. For details, see the comments of CMR-6040.

PR #826, opened 2019-11-07
Cmr 6032

indiejames:
Fixes 502 errors caused by bad image urls, adds a workaround for bad Redis cached values, fixes linter issues, and adds resizing for the "Not found" image.


mschmele:
Looks great

PR #832, opened 2019-11-13
CMR-6037

ygliuvt:
As a CMR Operator, I want access to CloudFront logs in Splunk -- Transfer logs from S3 to CloudWatch.


CayvonH:
I don't think we should bake this into the CMR. We should put this in the cloud-deployments-repo.

ygliuvt:
> I don't think we should bake this into the CMR. We should put this in the cloud-deployments-repo.
I agree. I am going to close this PR.

PR #837, opened 2019-11-21
CMR-6059: Fix intermittent test failure.

CayvonH:
Remove redis query with KEYS * in favor of SCAN.


system-int-test/test/cmr/system_int_test/search/acls/collection.clj
```diff
@@ -425,6 +425,7 @@
         user1-token (e/login (s/context) "user1")
         user2-token (e/login (s/context) "user2")]
 
+    (ingest/reindex-collection-permitted-groups (tc/echo-system-token))
```
CayvonH: I have no idea why this wasn't an issue before.

PR #853, opened 2019-12-10
Cmr 5952

indiejames:
Adds system token to browse-scaler calls to CMR.

### DIFF
--- a/browse-scaler/src/cmr.js
+++ b/browse-scaler/src/cmr.js
@@ -14,10 +14,14 @@ const cmrGranuleUrl = `${cmrRootUrl}/search/granules.json?concept_id=`;
  */
 const getEchoToken = async () => {
   console.log(process.env.CMR_ENVIRONMENT);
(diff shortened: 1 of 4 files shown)


browse-scaler/src/cmr.js
```diff
@@ -1,10 +1,32 @@
...
+const getEchoToken = async () => {
+  console.log(process.env.CMR_ENVIRONMENT);
+  const response = await getSecureParam(
+    `/${process.env.CMR_ENVIRONMENT}/browse-scaler/CMR_ECHO_SYSTEM_TOKEN`
+  );
+
+  if (response === undefined) {
+    throw new Error('ECHO Token not found. Please update config!');
+  } else {
+    console.log('Retrieved ECHO TOKEN');
```
CayvonH: Not sure if logging this every time the lambda is called is very useful. I think the error on failure would suffice.

indiejames: I have this here because I want to see the token retrieval happening.

PR #856, opened 2019-12-12
CMR-6087: Retire index-set

CayvonH:
Move the index-set api into indexer.


transmit-lib/src/cmr/transmit/indexer.clj
```diff
@@ -1,10 +1,77 @@
...
+(defn- start-rebalance-collection-url
+  [conn index-set-id concept-id]
+  (str (rebalance-collection-url conn index-set-id concept-id) "/start"))
+
+(defn- finalize-rebalance-collection-url
+  [conn index-set-id concept-id]
+  (str (rebalance-collection-url conn index-set-id concept-id) "/finalize"))
+
+(defn get-index-set
+  "Submit a request to index-set app to fetch an index-set assoc with an id"
```
indiejames: There is no index-set app anymore :)

CayvonH: was a lot easier to global replace cubby to redis :(. Fixed.

PR #858, opened 2019-12-13
CXMR-5764: Fixed the invalid version response.

sxu123:


ingest-app/src/cmr/ingest/services/ingest_service/util.clj
```diff
@@ -1,6 +1,7 @@
 (ns cmr.ingest.services.ingest-service.util
   (:require
    [cheshire.core :as json]
+   [clojure.string :as str]
```
ygliuvt: change the alias `str` to `string`

sxu123:
@ygliuvt, I'm not sure I follow your comment. The ticket said it needs better error msg, or to strip off the quotes.  So I strip off the quotes and for any other invalid versions, like abc, for example, the exception will be caught and 400 error will be returned, rather than 500.

ygliuvt:
> @ygliuvt, I'm not sure I follow your comment. The ticket said it needs better error msg, or to strip off the quotes. So I strip off the quotes and for any other invalid versions, like abc, for example, the exception will be caught and 400 error will be returned, rather than 500.

I don't think we need to strip out the quotes in version, we should just fail the request if the version has quotes in it, and when we do the validation of version, we should do it against the `cmr.umm-spec.versioning.versions`, rather than try and catch exception.

PR #861, opened 2019-12-18
CMR-6098: Added UMM-G to umm record mappings.

sxu123:


umm-spec-lib/src/cmr/umm_spec/umm_g/data_granule.clj
```diff
@@ -59,4 +64,8 @@
        :ProductionDateTime production-date-time
        :ArchiveAndDistributionInformation [{:Name util/not-provided
                                             :Size size
+                                            :SizeInBytes size-in-bytes
+                                            :Checksum (when checksum
+                                                        {:Value (:value checksum)
+                                                         :Algorithm (:algorithm checksum)})
```
mschmele: Is it okay for this value to be `nil`, or do we want to remove the empty ones?

sxu123: I think nil value is automatically removed.

PR #862, opened 2019-12-19
CMR-4920 adds new ECHO10 schema to umm-lib.

daniel-zamora:


sxu123:
Have you tested this?  I see you added new xml file, but how is it used?

PR #863, opened 2019-12-19
CMR-6098: Addressed the issue of having the same size in umm record that's mapped to different Sizes (in different unit).

sxu123:


umm-lib/src/cmr/umm/echo10/granule.clj
```diff
@@ -35,6 +35,17 @@
                  (x/element :Campaign {}
                             (x/element :ShortName {} pref))))))
 
+(defn- get-size-in-mb
+  "Get size in megabytes based on the size-unit."
+  [size size-unit]
+  (condp = size-unit
+    "KB" (/ size 1000)
```
ygliuvt: shouldn't it be `1024` rather than `1000`?

PR #867, opened 2020-01-02
CMR-6076: Catch redis exception in browse scaler.

CayvonH:
Browse scaler to not fail on redis issues.

### DIFF
--- a/browse-scaler/src/cache.js
+++ b/browse-scaler/src/cache.js
@@ -5,15 +5,34 @@ const redisPort = process.env.REDIS_PORT || 6379;
 const redisKeyExpireSeconds = process.env.REDIS_KEY_EXPIRE_SECONDS || 84000;
 
 
-const client = redis.createClient({
-  return_buffers: true,
-  host: redisHost,
-  port: redisPort
-});
+var redisClient = null;
 
 const { promisify } = require('util');
 
-const getAsync = promisify(client.get).bind(client);
+/**
+* getClient: Retrieve the Redis client. Start new client if client is null or not connected.
+* @returns {RedisClient} RedisClient object containing the connection to Redis.
+*/
+function getClient() {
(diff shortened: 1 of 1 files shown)


browse-scaler/src/cache.js
```diff
@@ -34,19 +53,19 @@ exports.cacheImage = (key, image) => {
...
+      console.error("Unable to retrieve image from Redis.");
+      return null;
+    })
+    .then(image => {
+      // workaround for bad cache entries that have the string "null" for the value
+      if (image && image !== 'null') {
+        console.log('got image from cache');
+        return image;
+      }
+      console.log('image is not in cache');
```
mschmele: This might help with troubleshooting
```suggestion
      console.log(`image is not in cache for concept: ${key}`);
```

CayvonH: done

browse-scaler/src/cache.js
```diff
@@ -34,19 +53,19 @@ exports.cacheImage = (key, image) => {
...
+  let client = getClient();
+  return promisify(client.get).bind(client)(key)
+    .catch(err => {
+      console.error(`Unable to retrieve image ${key} from Redis.`);
+      return null;
+    })
+    .then(image => {
+      // workaround for bad cache entries that have the string "null" for the value
+      if (image && image !== 'null') {
+        console.log(`got image ${key} from cache`);
```
ygliuvt: Do we want to log this all the time?

CayvonH: Eh, I don't think it hurts to know what key is being added to the cache.

ygliuvt: While it is good to know what key is being retrieved from the cache, we don't necessarily want to log it every single time we retrieve something from the cache. This seems excessive to me and it could potentially cause performance and space issues.

PR #868, opened 2020-01-03
CMR-5494: Fixed a bug for umm-json format.

sxu123:


system-int-test/test/cmr/system_int_test/search/provider_holdings_test.clj
```diff
@@ -137,14 +137,22 @@
...
                     :body "{\"errors\":[\"Unsupported format: opendata on the provider holdings endpoint.\"]}"}
                    (select-keys (ex-data e) [:status :body]))))))
 
+      (testing "Retrieve all provider holdings in unsupported format umm-json"
+        (try
+          (search/provider-holdings-in-format :umm-json {:token user-token})
+          (catch clojure.lang.ExceptionInfo e
+            (is (= {:status 400
+                    :body "{\"errors\":[\"Unsupported format: umm-json-results on the provider holdings endpoint.\"]}"}
+                   (select-keys (ex-data e) [:status :body]))))))
```
mschmele: Nitpick: is there a reason you aren't using the test utility we have the expects errors?

sxu123: Could you elaborate on which one? There are other tests that are using the try/catch so I'm doing the same here.

mschmele: We have a search api utility that expects errors, but I don't think we have one for provider holdings, now that I think about it more

PR #869, opened 2020-01-07
CMR-6097: Provided support for both X-Request-Id and CMR-Request-Id headers/response headers

sxu123:


ingest-app/docs/api.md
```diff
@@ -107,10 +115,18 @@ The user id header allows specifying the user-id to use when saving or deleting
...
 
     828ef0b8-a876-4579-85db-3cc9d1b5f6e5
 
+#### <a name="X-Request-Id-header"></a> x-request-id
+
+This header returns the value passed in through CMR-Request-Id or X-Request-Id request header or a unique id generated for the client request when no value is passed in, This can be used to help debug client errors. The generated value is a long string of the form
+
+    828ef0b8-a876-4579-85db-3cc9d1b5f6e5
+
+Note: X-Request-Id response header always contains the same value as the CMR-Request-Id response header.
```
ccuadrado: If I pass in X-Request-Id and Cmr-Request-Id that are different, do I get a 400 Bad Request?

sxu123: No. CMR-Request-Id takes precedence.  There is no need for the client to pass in both headers, but there is no harm if they want to. So I don't want to add extra code to fail this case, as long as we clearly document it.

system-int-test/test/cmr/system_int_test/header_test.clj
```diff
@@ -26,10 +27,13 @@
         cmr-hits (search-headers "CMR-Hits")
         cmr-took (search-headers "CMR-Took")
         search-request-id (search-headers "CMR-Request-Id")
+        search-x-request-id (search-headers "X-Request-Id")
         req-id-regex #"\w{8}-\w{4}-\w{4}-\w{4}-\w{12}"]
     (is (re-matches #"application\/echo10\+xml.*" content-type))
     (is (= aca-origin "*"))
     (is (= cmr-hits "1"))
+    (= ingest-request-id ingest-x-request-id)
```
CayvonH: You need to add `is` for these assertions to be ran (unless i'm misreading this)

sxu123: Good catch! Thanks!

PR #872, opened 2020-01-16
CMR-6123 

ygliuvt:
Add Launchpad Token Service sample JAVA client code into CMR repo.


token-service-client/cmr/client/SslClient.java
```diff
@@ -0,0 +1,144 @@
...
+import java.security.SecureRandom;
+import java.util.List;
+
+public class SslClient {
+
+	private static final String LINE_BREAKER = System.getProperty("line.separator");
+	// Launchpad Token Service root url, this is set to the Sandbox
+	private static final String LAUNCHPAD_TOKEN_SERVICE_URL_ROOT = "https://api.launchpad-sbx.nasa.gov/";
+	// full path of cacerts which has the PKI certificate for connecting to token service
+	private static final String CACERTS_PATH = "/Users/yliu10/cacerts";
```
daniel-zamora: do we need to change this?

ygliuvt: This is just an example. Anybody actually running the test would have to update it and the keystore password.

mschmele: Nitpick suggestion
```suggestion
	private static final String CACERTS_PATH = "~/cacerts";
```

ygliuvt: I don't like relative path. I can change it to `/path-to-cacerts/cacerts`.

PR #879, opened 2020-02-04
Cmr 5969 cmr 5970 cmr 5971 variable association restrictions

sxu123:
Implemented:
CMR-5969: Multiple variables with the same name can't be associated with one collection.
CMR-5970: Each variable can only be associated with one collection.
CMR-5971: variable and collection can only be associated if they belong to the same provider.

### DIFF
--- a/metadata-db-app/src/cmr/metadata_db/data/memory_db.clj
+++ b/metadata-db-app/src/cmr/metadata_db/data/memory_db.clj
@@ -152,8 +152,8 @@
     (when (and variable-concept-id associated-concept-id)
       (let [;;get variable name for the variable-concept-id from cmr_variables.
             variable (->> @(:concepts-atom db)
-                            (filter #(= variable-concept-id (:concept-id %)))
-                            first)
+                          (filter #(= variable-concept-id (:concept-id %)))
+                          first)
             variable-name (get-in variable [:extra-fields :variable-name])
 
             ;;get all the variable associations with variable concept id being different from variable-concept-id
@@ -419,9 +419,9 @@
 
   (if-let [error (or (validate-concept-id-native-id-not-changing db provider concept)
                      (when (variable-association-concept? concept)
-                          (or (validate-same-provider-variable-association concept)
-                              (validate-collection-not-associated-with-another-variable-with-same-name db concept)
-                              (validate-variable-not-associated-with-another-collection db concept))))]
+                       (or (validate-same-provider-variable-association concept)
+                           (validate-collection-not-associated-with-another-variable-with-same-name db concept)
+                           (validate-variable-not-associated-with-another-collection db concept))))]
    ;; There was a concept id, native id mismatch with earlier concepts
    error
    ;; Concept id native id pair was valid
--- a/metadata-db-app/src/cmr/metadata_db/data/memory_db.clj
+++ b/metadata-db-app/src/cmr/metadata_db/data/memory_db.clj
@@ -161,7 +161,7 @@
             v-associations (->> @(:concepts-atom db)
                                 (filter #(not= variable-concept-id (get-in % [:extra-fields :variable-concept-id])))
                                 (filter #(= associated-concept-id (get-in % [:extra-fields :associated-concept-id]))))
-            v-concept-ids (map #(get-in % [:extra-fields :variable-concept-id]) v-associations)
+            v-concept-ids-in-aassociations (map #(get-in % [:extra-fields :variable-concept-id]) v-associations)
 
             ;;get all the deleted variable associations with variable concept id being different from variable-concept-id
             ;;and associated with the associated-concept-id, from cmr_variable_association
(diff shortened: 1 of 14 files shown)


metadata-db-app/src/cmr/metadata_db/data/memory_db.clj
```diff
@@ -131,6 +132,111 @@
        :existing-concept-id existing-concept-id
        :existing-native-id existing-native-id})))
 
+(defn- concept-id-not-in-list?
```
CayvonH: nit: I'm not sure having two functions for `concept-id-in-list` is really necessary when `not` works well but in any case the logic of `concept-id-in-list?` should be contained in one function. So `concept-id-not-in-list?` should just call `(not (concept-id-in-list? list concept-id))`

sxu123: I thought about the same thing. but the logic is so simple that I thought it doesn't really matter.

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
```diff
@@ -339,12 +413,23 @@
                         :transaction-id (long (:transaction_id result))})
           (su/query conn stmt)))))
 
+(defn- variable-association-concept?
```
ygliuvt: This check should be just checking the concept-type is `:variable-association`.

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
```diff
@@ -224,6 +224,80 @@
...
+                                       "because the variable is already associated with another collection [%s].")
+                                  variable-concept-id associated-concept-id associated_concept_id)})))))
+
+(defn validate-same-provider-variable-association
+  "Validates that variable and the collection in the concept being saved are from the same provider.
+  Returns nil if valid and an error response if invalid."
+  [concept]
+  (let [variable-concept-id (get-in concept [:extra-fields :variable-concept-id])
+        associated-concept-id (get-in concept [:extra-fields :associated-concept-id])]
+    (when (and variable-concept-id associated-concept-id)
```
ygliuvt: There is no need for this check.

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
```diff
@@ -224,6 +224,80 @@
...
+  [db concept]
+  (let [variable-concept-id (get-in concept [:extra-fields :variable-concept-id])
+        associated-concept-id (get-in concept [:extra-fields :associated-concept-id])]
+    (when (and variable-concept-id associated-concept-id)
+      ;;Get all the same variables that are not deleted and are associated with other collections.
+      (let [stmt [(format "select va.associated_concept_id
+                           from   cmr_variable_associations va,
+                                  (select concept_id, max(revision_id) maxrev
+                                   from cmr_variable_associations group by concept_id) latestva
+                           where  va.revision_id = latestva.maxrev
```
ygliuvt: This logic is wrong. the variable association revision-ids do not need to match.

ygliuvt: You should add a test to cover this case: where a variable is associated with a collection twice, and you then want to associate the variable to a different collection. The test should fail. With the current implementation, the test will pass.

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
```diff
@@ -224,6 +224,80 @@
...
+        associated-concept-id (get-in concept [:extra-fields :associated-concept-id])]
+    (when (and variable-concept-id associated-concept-id))
+      ;; Get all the other variables sharing the same name; not deleted
+      ;; and are associated with the same collection.
+      (let [stmt [(format "select va.variable_concept_id
+                           from   cmr_variable_associations va, 
+                                  (select concept_id, max(revision_id) maxrev
+                                   from cmr_variable_associations group by concept_id) latestva,
+                                  cmr_variables v
+                           where  va.revision_id = latestva.maxrev
```
ygliuvt: same logic error here with the variable association revisions. Also I think you could get the variable_name from the concept without querying the cmr_variables table. Also this query will not work for the case when there is another variable with the same name in a different provider (BTW, you should have a test covering this case).

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
sxu123: I tested this numerous times with demo.  It worked fine. I needed to have the revision-id to have the latest so that I can check if the variable association is deleted. because only when the latest revision is deleted, the association is deleted.

ygliuvt: ok. test it this way:
1) associate variable A with collection A
2) dissociate variable A with collection A
3) associate variable A with collection A
4) associate variable A with collection B

I am not certain, but I think that will create a variable association with revision id greater than 1.

ygliuvt: Never mind. I think you are right. I was misreading the query. Don't really have time to review this in detail. Too much going on right now.

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
sxu123: I'm not sure an association concept will contain the variable_name. I don't  think it does because the cmr_variable_associations table doesn't contain the variable_name. even if it does, I still need to query the cmr_variables table to get "other" variables with the same name.
Regarding the provider, I checked the provider constraint before checking the name constraint, so it's guaranteed that the variable and the collection are from the same provider. and the existing associations between the collection and other variables should be from the same provider too for the associations to be valid.

metadata-db-app/src/cmr/metadata_db/data/memory_db.clj
sxu123: Removed one function. I thought it's more readable this way, but it's readable the other way too.

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
sxu123: Done.

PR #880, opened 2020-02-05
CMR-6182: add support for searching by humanized granule data format

mschmele:


indexer-app/src/cmr/indexer/data/concepts/collection.clj
```diff
@@ -168,6 +168,16 @@
...
     (merge variable-docs service-docs {:has-variables has-variables})))
 
+(defn- get-granule-data-format
+  "Returns the Format field from
+  ArchiveAndDistributionInformation -> FileDistributionInformation"
+  [file-distribution-information]
+  (->> file-distribution-information
+       (map #(:Format %))
+       distinct
+       (remove #(nil? %))
```
CayvonH: ultra-nit: `(remove #(nil? %))` can just be `(remove nil?)`. Same with `(map :Format)`

indexer-app/src/cmr/indexer/data/concepts/collection.clj
```diff
@@ -168,6 +168,16 @@
...
+       concat))
```
CayvonH: What is concat doing here?

indexer-app/src/cmr/indexer/data/concepts/collection/humanizer.clj
```diff
@@ -34,15 +34,32 @@
                                 (add-humanized-lowercase value-with-priorities))]
     {field value-with-lowercases}))
 
+(defn humanized-field->elastic-doc
+  "Extracts humanized fields from the science keyword and places them into an elastic doc with
+  the same shape/keys as science-keyword->elastic-doc."
+  [field]
+  (let [humanized-fields (filter #(-> % key namespace (= "cmr.humanized")) field)
```
CayvonH: Could you explain a bit what is going on here? I'm having trouble following.

indexer-app/src/cmr/indexer/data/concepts/collection.clj
mschmele: Ohhhh yeah that's right! I couldn't remember the nicer way to do this so thanks

indexer-app/src/cmr/indexer/data/concepts/collection.clj
mschmele: I saw a lot of the other indexed values using `mapcat` so I followed suit. I can also get rid of it very easily

indexer-app/src/cmr/indexer/data/concepts/collection/humanizer.clj
mschmele: It's essentially just converting the humanized values to an elastic doc and associng the `.lowercase` values

indexer-app/src/cmr/indexer/data/concepts/collection.clj
```diff
@@ -297,6 +310,11 @@
                                    (some?
                                     (some #(= (common-config/cwic-tag) %)
                                           (map :tag-key.lowercase tags))))
+            :granule-data-format granule-data-format
+            :granule-data-format.humanized (-> humanized-values
```
ygliuvt: `granule-data-format` could have multiple formats, why do we only pick the first humanized value rather than all unique ones?

indexer-app/src/cmr/indexer/data/index_set.clj
```diff
@@ -232,6 +232,12 @@
    :quantity m/string-field-mapping
    :quantity.lowercase m/string-field-mapping})
 
+(defnestedmapping granule-data-format-mapping
```
ygliuvt: Why is this defined as a nested map? It should just be string field.

indexer-app/src/cmr/indexer/data/concepts/collection.clj
ygliuvt: Actually this line can be removed. It is overwritten by `humanized-field->elastic-doc` call anyways.

indexer-app/src/cmr/indexer/data/index_set.clj
mschmele: I was getting elasticsearch errors while it was a string field

indexer-app/src/cmr/indexer/data/index_set.clj
```diff
@@ -378,6 +378,9 @@
           :science-keywords science-keywords-field-mapping
           :science-keywords.humanized science-keywords-field-mapping
 
+          :granule-data-format                  (m/stored m/string-field-mapping)
```
ygliuvt: Shouldn't we have a `:granule-data-format.lower-case` field to support case insensitive search?

indexer-app/src/cmr/indexer/data/index_set.clj
```diff
@@ -378,6 +378,9 @@
...
+          :granule-data-format.humanized        (m/stored m/string-field-mapping)
```
ygliuvt: We don't need to store humanized field. Change this to `m/string-field-mapping`.

indexer-app/src/cmr/indexer/data/index_set.clj
ygliuvt: This issue is only for humanized granule-data-format, we could not index `:granule-data-format` at all, but I guess we would want to search/retrieve it eventually.

search-app/docs/api.md
```diff
@@ -1387,6 +1388,10 @@ Supports ignore_case and the following aliases for "NEAR\_REAL\_TIME": "near\_re
 
      curl "%CMR-ENDPOINT%/collections?collection_data_type\[\]=NEAR_REAL_TIME&collection_data_type\[\]=OTHER"
 
+#### <a name="c-granule-data-format"></a> Find collections by format of data in granules
```
ygliuvt: You didn't actually implement the full support for this. See my comment on missing the `granule-data-format.lowercase` field in collection index mapping. I think you should take this out.

search-app/src/cmr/search/services/parameters/conversion.clj
```diff
@@ -33,6 +33,8 @@
    :entry-id :string
    :entry-title :string
    :exclude :exclude
+   :granule-data-format :string
+   :granule-data-format-h :string
```
ygliuvt: I think this should be `:granule-data-format-h :humanizer`. I am worried about your test now. How did they pass?

indexer-app/src/cmr/indexer/data/index_set.clj
mschmele: Fair point. I added the `.lowercase` field anyway

PR #891, opened 2020-02-21
CMR-6222: As a CMR Developer, I Want to Run Dev-system on Top of ES 7.5.2

ygliuvt:
This PR runs dev-system against the external Elasticsearch 7.5.2 that is running locally via docker. The ES can be started by running `docker-compose up` in the `resources/docker` directory under `elastic-utils-lib`. The dev-system can be started successfully and most of the search integration tests can also run successfully with the code changes in this PR. 

I tried `https://github.com/spandex-project/spandex`, but found it too low level and is too big a departure from our existing code base. On the other end, the Elastisch 5.0.0-beta1 release fixed most of upgrade issues for us and aligns well with our existing code base. So I decided to continue using the Elastisch library and add in fixes in the helper namespaces in `elastic-utils-lib` when necessary.

This PR will not run in CI because CI still depends on embedded ES which is covered in CMR-6133, so don't be alarmed if you see build failure on this branch.

### DIFF
--- a/elastic-utils-lib/project.clj
+++ b/elastic-utils-lib/project.clj
@@ -8,7 +8,7 @@
                [potemkin]]
   :dependencies [[cheshire "5.8.1"]
                  [clj-http "2.3.0"]
-                 [clojurewerkz/elastisch "2.2.2"]
+                 [clojurewerkz/elastisch "5.0.0-beta1"]
                  [commons-codec/commons-codec "1.11"]
                  [commons-io "2.6"]
                  [log4j/log4j "1.2.17"]
--- a/elastic-utils-lib/resources/docker/docker-compose.yml
+++ b/elastic-utils-lib/resources/docker/docker-compose.yml
@@ -1,0 +1,33 @@
+version: '3'
+services:
+
+  elasticsearch:
+    image: docker.elastic.co/elasticsearch/elasticsearch:7.5.2
+    container_name: cmr-es-elastic
+    environment:
+      - http.host=0.0.0.0
+      - transport.host=127.0.0.1
+      - ELASTIC_PASSWORD=kibana
+      - bootstrap.memory_lock=true
+      - "ES_JAVA_OPTS=-Xms1024m -Xmx1024m"
+      - xpack.security.enabled=false
+      - xpack.monitoring.enabled=false
+    ulimits:
+      memlock:
+        soft: -1
+        hard: -1
+    volumes:
+      - ${PWD}/data/elastic:/usr/share/elasticsearch/data
+    ports:
+      - 9209:9200
+    networks:
+      - cmr-es-net
+
+volumes:
+  data:
+    driver: local
+  logs:
+    driver: local
+
+networks:
+  cmr-es-net:
--- a/elastic-utils-lib/src/cmr/elastic_utils/es_helper.clj
+++ b/elastic-utils-lib/src/cmr/elastic_utils/es_helper.clj
@@ -1,0 +1,19 @@
+(ns cmr.elastic-utils.es-helper
+  "Defines helper functions for invoking ES"
+  (:require
+   [clojurewerkz.elastisch.rest.document :as doc]))
+
+(defn search
+  "Performs a search query across one or more indexes and one or more mapping types"
+  [conn index mapping-type opts]
+  (doc/search conn index mapping-type opts))
+
+(defn count-query
+  "Performs a count query over one or more indexes and types"
+  [conn index mapping-type query]
+  (doc/count conn index mapping-type query))
+
+(defn scroll
+  "Performs a count query over one or more indexes and types"
+  [conn scroll-id opts]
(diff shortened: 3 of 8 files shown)


elastic-utils-lib/resources/docker/docker-compose.yml
```diff
@@ -0,0 +1,33 @@
+version: '3'
+services:
+
+  elasticsearch:
+    image: docker.elastic.co/elasticsearch/elasticsearch:7.5.2
+    container_name: cmr-es-elastic
+    environment:
+      - http.host=0.0.0.0
+      - transport.host=127.0.0.1
+      - ELASTIC_PASSWORD=kibana
```
daniel-zamora: kibana? weird password.

elastic-utils-lib/project.clj
```diff
@@ -8,7 +8,7 @@
                [potemkin]]
   :dependencies [[cheshire "5.8.1"]
                  [clj-http "2.3.0"]
-                 [clojurewerkz/elastisch "2.2.2"]
+                 [clojurewerkz/elastisch "5.0.0-beta1"]
```
daniel-zamora: I'm a little wary of using an unstable library that has a history of extremely slow updates for something this important.  We are potentially trading reliability for expedience, but I wasn't a huge fan of using spandex either.  If this works then great, it just makes me nervous.

ygliuvt: We always have the option to roll our own. This makes it more convenient now that we have ruled out Spandex.

sxu123:
@ygliuvt, What does "most of the search integration tests can also run successfully with the code changes in this PR." imply?  So once this is merged, we are expecting some tests failures locally?

elastic-utils-lib/src/cmr/elastic_utils/es_helper.clj
```diff
@@ -0,0 +1,19 @@
...
+  [conn index mapping-type opts]
+  (doc/search conn index mapping-type opts))
+
+(defn count-query
+  "Performs a count query over one or more indexes and types"
+  [conn index mapping-type query]
+  (doc/count conn index mapping-type query))
+
+(defn scroll
+  "Performs a count query over one or more indexes and types"
```
sxu123: doc string needs to be changed.

elastic-utils-lib/resources/docker/docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+    ulimits:
+      memlock:
+        soft: -1
+        hard: -1
+    volumes:
+      - ${PWD}/data/elastic:/usr/share/elasticsearch/data
+    ports:
+      - 9209:9200
+    networks:
+      - cmr-es-net
```
sxu123: where is this network defined?

elastic-utils-lib/src/cmr/elastic_utils/es_helper.clj
sxu123: Where is this es_helper.clj used?  I can't seem to find it.

elastic-utils-lib/resources/docker/docker-compose.yml
ygliuvt: last line of the file

elastic-utils-lib/src/cmr/elastic_utils/es_helper.clj
ygliuvt: Done.

elastic-utils-lib/resources/docker/docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+      - xpack.security.enabled=false
+      - xpack.monitoring.enabled=false
+    ulimits:
+      memlock:
+        soft: -1
+        hard: -1
+    volumes:
+      - ${PWD}/data/elastic:/usr/share/elasticsearch/data
+    ports:
+      - 9209:9200
```
CayvonH: Any reason for 9209?

ygliuvt: It is the port that is configured for default external elasticsearch port in dev-system. I use this port number to make it easy to run with dev-system where we only need to switch the elasticsearch to :external mode.

ygliuvt:
@sxu123 This PR is the starting point of the update ES feature epic. As I have explained in the description of the PR, the build will fail because 1) we don't have ES 7.5.2 running in embedded mode which is used by CI, and 2) we only added 7.5.2 support to the basic index creation/deletion apis of ES 7.5.2. The support for the other apis that CMR uses will be added as we work through the rest of the tickets in the epic. Only until the end of the epic will we see all the tests passing in CI.

elastic-utils-lib/resources/docker/docker-compose.yml
ygliuvt: This is an internal password that nobody cares. :)

PR #897, opened 2020-02-25
CMR-6236: Updated Collection Preview Gems.

sxu123:
Verified with Chris that the html file for the collection ingested looks good.


jceaser:
I would approve this if I could

PR #899, opened 2020-02-25
CMR-6134: More changes to insert and searching via ES 7.5.2 API.

ygliuvt:
With this PR, dev-system starts up fully without error.

### DIFF
--- a/elastic-utils-lib/src/cmr/elastic_utils/es_helper.clj
+++ b/elastic-utils-lib/src/cmr/elastic_utils/es_helper.clj
@@ -46,23 +46,17 @@
   ([conn index mapping-type id document]
    (put conn index mapping-type id document nil))
   ([conn index mapping-type id document opts]
-   (if (empty? opts)
-     (rest/put conn (rest/record-url conn index "_doc" id)
-               {:content-type :json
-                :body document})
-     (rest/put conn (rest/record-url conn index "_doc" id)
-               {:content-type :json
-                :body document
-                :query-params opts}))))
+   (rest/put conn (rest/record-url conn index "_doc" id)
+             {:content-type :json
+              :body document
+              :query-params opts})))
 
 (defn delete
   "Deletes document from the index."
   ([conn index mapping-type id]
    (delete conn index mapping-type id nil))
   ([conn index mapping-type id opts]
-   (if (empty? opts)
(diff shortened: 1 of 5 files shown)


elastic-utils-lib/src/cmr/elastic_utils/es_helper.clj
```diff
@@ -26,3 +28,62 @@
...
+     (if (not-found? result)
+       nil
+       result))))
+
+(defn put
+  "Creates or updates a document in the search index, using the provided document id"
+  ([conn index mapping-type id document]
+   (put conn index mapping-type id document nil))
+  ([conn index mapping-type id document opts]
+   (if (empty? opts)
```
CayvonH: i dont really understand why we need this check and can't just pass the nil opts to `query-params` but if it works i won't complain

indexer-app/src/cmr/indexer/data/index_set_elasticsearch.clj
```diff
@@ -63,40 +65,47 @@
...
 (defn get-index-sets
   "Fetch all index-sets in elastic."
   [{:keys [conn]} index-name idx-mapping-type]
-  (when (esi/exists? conn index-name)
-    (let [result (doc/search conn index-name idx-mapping-type {"fields" "index-set-request"})]
-      (map (comp #(decode-field (first % )) :index-set-request :fields) (get-in result [:hits :hits])))))
+  (when (esi-helper/exists? conn index-name)
+    (let [result (es-helper/search
+                  conn index-name idx-mapping-type {"_source" "index-set-request"})]
+      (map (comp #(decode-field (first % )) :index-set-request :_source) (get-in result [:hits :hits])))))
```
CayvonH: nit: extra space in `(first % )`

elastic-utils-lib/src/cmr/elastic_utils/es_helper.clj
ygliuvt: I added this to solve a NullPointerException, but I don't think it is the cause of that. I will remove this check.

indexer-app/src/cmr/indexer/data/index_set_elasticsearch.clj
ygliuvt: Done.

PR #900, opened 2020-02-26
CMR-6235 Updated autocomplete parameters and response to follow CMR standards

jaybarra:
Autocomplete parameters have been updated to follow CMR standard parameter format.
Response has been modified to follow standard CMR responses.

<h2>Changes</h2>

- _types_ has been removed and no longer accepts CSV values
- _types_ has been replaced by an optional list of _type[]_
  * e.g.  `&type[]=instrument&type[]=platform`
- CMR-Hits value now included in response header denoting total number of hits found
  * may be different than total returned in response if using paging options
- response format now follows standard CMR response structure
```
HTTP/1.1 200 OK
Content-Type: application/json; charset=UTF-8
CMR-Hits: 1

{
  "feed": {
    "entry": [
      {
        "score": 88.00000,
        "type": "instrument",
        "value": "flux capacitor"
      }
    ]
  }
}
```

### DIFF
--- a/system-int-test/src/cmr/system_int_test/utils/url_helper.clj
+++ b/system-int-test/src/cmr/system_int_test/utils/url_helper.clj
@@ -1,7 +1,6 @@
 (ns cmr.system-int-test.utils.url-helper
   "helper to provide the urls to various service endpoints"
   (:require
-   [clojure.string :as string]
    [cmr.common.config :as config]
    [cmr.elastic-utils.config :as es-config]
    [cmr.transmit.config :as transmit-config]
--- a/system-int-test/src/cmr/system_int_test/utils/url_helper.clj
+++ b/system-int-test/src/cmr/system_int_test/utils/url_helper.clj
@@ -1,7 +1,7 @@
 (ns cmr.system-int-test.utils.url-helper
   "helper to provide the urls to various service endpoints"
   (:require
-   [clojure.string :as str]
+   [clojure.string :as string]
    [cmr.common.config :as config]
    [cmr.elastic-utils.config :as es-config]
    [cmr.transmit.config :as transmit-config]
@@ -241,13 +241,16 @@
   (format "http://localhost:%s/%ss" (transmit-config/search-port) (name type)))
 
 (defn autocomplete-url
-  "Autocomplete URL with query term and optional types"
+  "Autocomplete URL with query term and optional types collection"
   ([term]
-   (if term
-    (format "http://localhost:%s/autocomplete?q=%s" (transmit-config/search-port) term)
-    (format "http://localhost:%s/autocomplete" (transmit-config/search-port))))
+   (if-not term
+    (format "http://localhost:%s/autocomplete" (transmit-config/search-port)))
+    (format "http://localhost:%s/autocomplete?q=%s" (transmit-config/search-port) term))
   ([term types]
-   (format "http://localhost:%s/autocomplete?q=%s&types=%s" (transmit-config/search-port) term types)))
+   (format "http://localhost:%s/autocomplete?q=%s&%s"
+           (transmit-config/search-port)
+           term
+           (string/join "&" (map #(str "type[]=" %) types)))))
 
 (defn enable-search-writes-url
   []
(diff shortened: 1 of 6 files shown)


system-int-test/src/cmr/system_int_test/utils/url_helper.clj
```diff
@@ -241,13 +240,13 @@
...
-   (if term
-    (format "http://localhost:%s/autocomplete?q=%s" (transmit-config/search-port) term)
-    (format "http://localhost:%s/autocomplete" (transmit-config/search-port))))
-  ([term types]
-   (format "http://localhost:%s/autocomplete?q=%s&types=%s" (transmit-config/search-port) term types)))
+  "Autocomplete URL with query term and optional types collection"
+  ([]
+   (autocomplete-url nil))
+  ([query]
+   (if-not query
```
indiejames: nitpick: use more canonical `if` here and reverse the expressions.

system-int-test/test/cmr/system_int_test/search/autocomplete_collection_facet_search_test.clj
```diff
@@ -27,64 +27,103 @@
...
+  (testing "returns CMR-Hits header"
+     (let [response (search/get-autocomplete-suggestions "q=foo")
+           headers (:headers response)]
+       (is (:CMR-Hits headers))
+       (is (= "1" (:CMR-Hits headers))))))
+
+(deftest autocomplete-functionality-test
   (testing "partial value match search"
-   (let [response (search/get-autocomplete-suggestions "f")
+   (let [response (search/get-autocomplete-suggestions "q=f")
```
indiejames: You are basically running the same test over and over with different input and expected responses. Take a look at the `are3` macro used in other tests for a way to do this more concisely.

system-int-test/test/cmr/system_int_test/search/autocomplete_collection_facet_search_test.clj
```diff
@@ -27,64 +27,103 @@
...
+  (testing "page_size 2 test"
+   (let [response (search/get-autocomplete-suggestions "q=*&page_size=2")
+         headers (:headers response)
+         body (json/parse-string (:body response) true)
+         entry (:entry (:feed body))]
+     (is (= "3" (:CMR-Hits headers)))
+     (is (= 2 (count entry)))))
+
+  (testing "page_num should yield different 'first' results test"
+   (let [page-one-response (search/get-autocomplete-suggestions "q=b&page_size=1")
```
indiejames: This is a good place to use the thread first macro `->` instead of creating a bunch of bindings that are only used to create the next binding.

mschmele: I think this one is a little too complicated for just thread-first and would need something like `as->`

system-int-test/test/cmr/system_int_test/search/autocomplete_collection_facet_search_test.clj
```diff
@@ -27,64 +27,103 @@
...
+         page-one-headers  (:headers page-one-response)
+         page-one-body     (json/parse-string (:body page-one-response) true)
+         page-one-entry    (first (:entry (:feed page-one-body)))
+
+         page-two-response (search/get-autocomplete-suggestions "q=b&page_size=1&page_num=1")
+         page-two-headers  (:headers page-two-response)
+         page-two-body     (json/parse-string (:body page-two-response) true)
+         page-two-entry    (first (:entry (:feed page-two-body)))]
```
mschmele:
```suggestion
   (let [page-one-entry (as-> (search/get-autocomplete-suggestions "q=b&page_size=1") response
                              (:headers response)
                              (json/parse-string (:body response) true)
                              (first (:entry (:feed response))))
         page-two-entry (as-> (search/get-autocomplete-suggestions "q=b&page_size=1&page_num=1") response
                              (:headers response)
                              (json/parse-string (:body response) true)
                              (first (:entry (:feed response))))]
```

PR #902, opened 2020-02-26
CMR-6224 Adding the capability to parse multiple data formats that ex…

eereiter:
…ist in one ECHO 10 DataFormat element.


umm-spec-lib/src/cmr/umm_spec/test/dif10_expected_conversion.clj
```diff
@@ -337,22 +336,10 @@
...
-    (for [[index value] (map-indexed vector parsed-format)]
-      (if (= index 0)
-        (assoc file-dist-info :Format value)
-        (conj (assoc file-dist-info :Format value))))))
-
 (defn- expected-file-dist-info
   "Created expected FileDistributionInformation for ArchiveAndDistributionInformation map."
   [file-dist-infos]
-  (when-let [file-dist-infos (flatten (map expand-file-dist-infos file-dist-infos))]
+  (when-let [file-dist-infos (flatten (map conversion-util/expand-file-dist-infos file-dist-infos))]
```
indiejames: You should be able to replace `(flatten (map` with `(mapcat`.

umm-spec-lib/src/cmr/umm_spec/test/expected_conversion_util.clj
```diff
@@ -222,3 +223,15 @@
                  (update pr :OnlineResource dif-online-resource)
                  pr))
       check-nil-pub-ref))
+
+(defn expand-file-dist-infos
+  "Parse each format to see if the format contains multiple values. If it does
+   then replace the distribution info that exists and add new ones for each
+   multple format that exists."
+  [file-dist-info]
+  (let [formats (get (select-keys file-dist-info [:Format]) :Format)
```
indiejames: You don't need to do the `select-keys` here. Just do this
`(let [formats (:Format file-dist-info)]`

umm-spec-lib/src/cmr/umm_spec/test/expected_conversion_util.clj
```diff
@@ -222,3 +223,15 @@
...
+        parsed-format (format-util/parse-distribution-formats formats)]
```
indiejames: I'm not sure what you are trying to do here, but `for` is a list comprehension, so you are going to get a list back here. It seems like you are trying to mutate `file-dist-info` using a `for loop`, which is not what this does. If you _are_ trying to get a list, I don't think the `conj` on line 237 is doing anything since you are only giving it one argument.

indiejames: Also, if you are trying to create a list here, you don't need a `for` _and_ `map-indexed`. Just call `map-indexed` with a function that takes the index and value and do whatever you want with them in there.

umm-spec-lib/src/cmr/umm_spec/xml_to_umm_mappings/echo10.clj
```diff
@@ -188,11 +205,7 @@
       [{:Fees price
         :Format u/not-provided
         :FormatType "Native"}]
-      (for [format formats]
-        (util/remove-nil-keys
-          {:Fees price
-           :Format (value-of format ".")
-           :FormatType "Native"})))))
+      (flatten (map #(add-data-formats price %) (select doc "Collection/DataFormat"))))))
```
indiejames: `mapcat` instead of `(flatten (map`

umm-spec-lib/src/cmr/umm_spec/test/expected_conversion_util.clj
eereiter: You are correct, I don't need conj and therefore I don't need to know the index anymore.  I was able to reduce this function to the following:

(let [formats (:Format (assoc file-dist-info :Format "try, this"))
      parsed-format (format-util/parse-distribution-formats formats)]
  (map #(assoc file-dist-info :Format %) parsed-format))

umm-spec-lib/src/cmr/umm_spec/test/dif10_expected_conversion.clj
eereiter: I was. thanks

umm-spec-lib/src/cmr/umm_spec/test/expected_conversion_util.clj
eereiter: Thanks. I fixed the code.

umm-spec-lib/src/cmr/umm_spec/xml_to_umm_mappings/echo10.clj
eereiter: fixed the code.

umm-spec-lib/src/cmr/umm_spec/test/echo10_expected_conversion.clj
```diff
@@ -42,7 +42,7 @@
    In ECHO 10 there is only 1 price (fees) but there can be many data formats. The first fee
    applies to all data formats."
   [file-dist-infos]
-  (when file-dist-infos
+  (when-let [file-dist-infos (flatten (map conversion-util/expand-file-dist-infos file-dist-infos))]
```
indiejames: You can use `mapcat` here as well.

eereiter: fixed

PR #906, opened 2020-02-28
Cmr 4376 - Use major.minor.micro to distinguish UMM Versions, and added enum to UMM-C v 1.15.1

eereiter:
CMR-4376, CMR-6174 CMR-6176 - Added the capability to use 3 or more numbers to separate out the UMM version numbers. This version will be UMM-C version 1.15.1. For 1.15.1 Added the enumeration of DEPRECATED to the UMM-C CollectionProgress element and add translations to DIF 9/10, ECHO 10 and ISO MENDS/SMAP.


daniel-zamora:
In the future it would be nice if the three tickets were their own pull request, so its easier to differentiate what code change goes with what.

umm-spec-lib/src/cmr/umm_spec/versioning.clj
```diff
@@ -10,7 +10,7 @@
   "A map of concept type to a sequence of valid UMM Schema versions, with the newest one last.
   The sequence must be updated when new schema versions are added for the concept type."
   {:collection ["1.0" "1.1" "1.2" "1.3" "1.4" "1.5" "1.6" "1.7" "1.8" "1.9" "1.10" "1.11" "1.12"
-                "1.13" "1.14" "1.15"]
+                "1.13" "1.14" "1.15" "1.15.1"]
```
daniel-zamora: We might want a more elegant solution, I don't now how many minor version updates we are going to do but this vector might get out of hand.

umm-spec-lib/src/cmr/umm_spec/migration/version/core.clj
```diff
@@ -27,24 +27,36 @@
...
-        part1-compare (- begin-part1 end-part1)
-        part2-compare (- begin-part2 end-part2)]
-    (if (zero? part1-compare)
-      part2-compare
-      part1-compare)))
+  (let [begin-tmp (map #(Integer/parseInt %) (clojure.string/split begin #"\."))
+        end-tmp (map #(Integer/parseInt %) (clojure.string/split end #"\."))]
+    (loop [begin begin-tmp
+           end end-tmp]
+      (println begin end)
```
daniel-zamora: Remove println or change it to use the logging framework.

umm-spec-lib/src/cmr/umm_spec/versioning.clj
daniel-zamora: It doesn't need to be in this PR though

umm-spec-lib/src/cmr/umm_spec/migration/version/core.clj
daniel-zamora: I'm still parsing through this, will update this comment when I figure out a better way to do this.

eereiter: Thanks

eereiter: Fixed

umm-spec-lib/src/cmr/umm_spec/versioning.clj
eereiter: I created issue https://bugs.earthdata.nasa.gov/browse/CMR-6249

PR #910, opened 2020-03-04
CMR-6134: Fix tests, change bulk API.

CayvonH:


elastic-utils-lib/src/cmr/elastic_utils/es_helper.clj
```diff
@@ -74,3 +75,22 @@
                                           (conj doc/optional-delete-query-parameters
                                                 :ignore_unavailable))
                :body {:query query}})))
+
+(defn ^:private bulk-with-url
+  ([conn url operations] (bulk-with-url conn url operations nil))
```
daniel-zamora: I think this needs to be on a seperate line, also doc string

CayvonH: done

PR #911, opened 2020-03-04
CMR-6221 Autocomplete relevancy scoring on return values

jaybarra:
Updated autocomplete to use scoring feature from Elasticsearch to provide better relevancy scoring of returned values


common-app-lib/src/cmr/common_app/services/search/query_model.clj
```diff
@@ -228,6 +228,12 @@
 (defrecord MatchNoneCondition
   [])
 
+(defrecord MatchCondition
+  [field value])
+
+(defrecord MatchFilterCondition
+  [field value filter])
+
```
CayvonH: Could you add a comment for each argument in the defrecord that describes what it is. Similar to the other Conditions in that namespace.

jaybarra: updated

jaybarra:
re-creating PR

PR #913, opened 2020-03-10
CMR-6134: Configured embedded es log level.

CayvonH:
be able to configure embedded es log level with dev-system persistent settings.

### DIFF
--- a/dev-system/src/cmr/dev_system/system.clj
+++ b/dev-system/src/cmr/dev_system/system.clj
@@ -97,6 +97,9 @@
   [system]
   (update-in system [:apps] set-web-server-options))
 
+(def in-memory-elastic-log-level-atom
+  (atom :info))
+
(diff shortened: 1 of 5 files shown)


dev-system/src/cmr/dev_system/system.clj
```diff
@@ -97,6 +97,9 @@
   [system]
   (update-in system [:apps] set-web-server-options))
 
+(def in-memory-elastic-log-level-atom
```
daniel-zamora: Can we have a TODO or something to remind us to turn this off eventually

CayvonH: I'm not sure i understand what you mean?

daniel-zamora: We don't want to leave INFO on? pretty straight forward.

PR #919, opened 2020-03-16
CMR-6246 Adding more constraints when parsing data format.

eereiter:


umm-spec-lib/src/cmr/umm_spec/xml_to_umm_mappings/distributed_format_util.clj
```diff
@@ -24,26 +24,32 @@
...
+  (if (string/includes? formats ", or ")
     (string/replace formats #", or " ", ")
     formats))
 
+(defn remove-text-various
+  "Remove the text 'Various: ' from the passed in string. The input is a
+   string and the output is either the replaced string or the input string if no
+   changes occured."
+  [formats]
+  (if (string/includes? formats "Various: ")
```
mschmele: Do we need the `if` here? I think the `string/replace` should be good enough. Same goes for the other functions

daniel-zamora: Looks like they are all that way.

eereiter: I though that an if statement that does a comparison was faster than the small number of times that a replacement call was made in addition than to always try to replace a string.

eereiter: If you all don't agree then I will remove the if calls.

eereiter: Did a timed test.  with 10,000 calls to the function as is it took 4.492631 msecs, calling the same function without the if took 5.117906 msecs
1,000,000 times with if took Elapsed time: 249.397825 msecs
1,000,000 times without if took Elapsed time: 2047.724538 msecs  almost 2 seconds longer. 
I created a string where nothing is replaced and it took 17.208532 msecs with the if statement and 354.613901 msecs without the if.  This is the case that will be used the most.  
Since the performance is a lot better with the if, I am going to leave the code as is. Mark agreed on Slack that this is OK.

PR #920, opened 2020-03-16
CMR-6239

daniel-zamora:
Investigate and Implement a Solution For 'Time To Live (TTL)' For Elasticsearch 7.5.2 Upgrade

Added some messages to indexer when cleanup expired granules job is run.


daniel-zamora:
I would avoid reviewing this until I can figure out a way to clean it up.

PR #921, opened 2020-03-16
CMR-6239 Investigate and Implement a Solution For 'Time To Live (TTL)' For Elasticsearch 7.5.2 Upgrade

daniel-zamora:
Collections seems to work already with how mdb saves concepts, but for granules we needed a TTL solution.  I opted for adding a message when the cleanup job is run that tells indexer to un-index the expired granule.

### DIFF
--- a/indexer-app/src/cmr/indexer/data/elasticsearch.clj
+++ b/indexer-app/src/cmr/indexer/data/elasticsearch.clj
@@ -183,7 +183,7 @@
   es-doc is the elasticsearch document to be passed on to elasticsearch
   elastic-id is the _id of the document in the index
   revision-id is the version of the document in elasticsearch"
-  [f conn es-index es-type es-doc elastic-id revision-id ttl]
+  [f conn es-index es-type es-doc elastic-id revision-id]
   (let [options {:version revision-id :version_type "external_gte"}]
     (try
       (f conn es-index es-type elastic-id es-doc options)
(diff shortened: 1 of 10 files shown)


indexer-app/src/cmr/indexer/data/elasticsearch.clj
```diff
@@ -209,30 +209,13 @@
 
 (defn- non-tombstone-concept->bulk-elastic-doc
   "Takes a non-tombstoned concept map (a normal revision) and returns an elastic document suitable
-   with ttl fields for bulk indexing."
+   for bulk indexing."
   [context concept]
   (let [parsed-concept (cp/parse-concept context concept)
         delete-time (get-in parsed-concept
```
ygliuvt: This line can be deleted now.

indexer-app/src/cmr/indexer/data/elasticsearch.clj
```diff
@@ -209,30 +209,13 @@
...
-      elastic-doc
-      (info
-       (str
-        "Skipping expired concept ["
-        (:concept-id concept)
-        "] with delete-time ["
-        (f/unparse (f/formatters :date-time) delete-time)
-        "]")))))
+        elastic-doc (parsed-concept->elastic-doc context concept parsed-concept)]
+    elastic-doc))
```
ygliuvt: nit: no need to define `elastic-doc` in the let block, just return the value here.

PR #932, opened 2020-03-31
CMR-6278 Updates variable search documentation.

daniel-zamora:


search-app/docs/api.md
```diff
@@ -3675,7 +3677,7 @@ Content-Length: 168
 ]
 ```
 
-On occassions when variable association cannot be processed at all due to invalid input, variable association request will return a failure status code with the appropriate error message.
+On occasions when variable association cannot be processed at all due to invalid input, variable association request will return a failure status code with the appropriate error message.
```
CayvonH: ultra nit: `On occasions when variable association` -> `On occasions where a variable association`.
`variable association request` -> `the variable association request`.

search-app/docs/api.md
```diff
@@ -3418,6 +3418,8 @@ These parameters will match fields within a variable. They are case insensitive
   * options: pattern, ignore_case
 * instrument
   * options: pattern, ignore_case
+* measurement
```
ygliuvt: Do we have requirement to provide search by measurement? There is no tests for searching by measurements in integration tests. Maybe you want to double checked with system engineering team to confirm that variable measurement is indeed what we have indexed and ideally you should add in the missing tests.

daniel-zamora: I'm not adding tests for a documentation ticket.

daniel-zamora: Measurement is currently supported via search, if we don't want that we should write a ticket to remove it.

PR #934, opened 2020-04-01
CMR-6217 Enforcing group name uniqueness rule when updating groups

jaybarra:
Preventing non-unique named groups during updates. This enforces the same behavior that is already implemented on creation.

### DIFF
--- a/access-control-app/src/cmr/access_control/services/group_service.clj
+++ b/access-control-app/src/cmr/access_control/services/group_service.clj
@@ -261,17 +261,22 @@
     (auth/verify-can-update-group context (assoc existing-group :concept-id concept-id))
     ;; Verify the name does not conflict with an existing group
     (when (not= (:name updated-group) (:name existing-group))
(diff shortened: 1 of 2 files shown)


access-control-app/src/cmr/access_control/services/group_service.clj
```diff
@@ -259,11 +259,27 @@
...
     (auth/verify-can-update-group context (assoc existing-group :concept-id concept-id))
-      ;; Avoid clobbering :members by merging the updated-group into existing-group. If updated-group
-      ;; specifies :members then it will overwrite the existing value.
+    ;; Verify the name does not conflict with an existing group
+    (when (not= (:name updated-group) (:name existing-group))
+      (debug (format "Group name update [%s] -> [%s] checking for conflicts"
+                       (:name existing-group)
+                       (:name updated-group)))
+      (when-let [conflicting-concept-id
+                 (->> (search-for-groups context
```
ygliuvt, line 268: This only searches for system groups. What if there are conflict on provider owned groups?

PR #935, opened 2020-04-01
CMR-6171: Added search related functionalities for subscription concept.

sxu123:


indexer-app/src/cmr/indexer/data/elasticsearch.clj
```diff
@@ -59,6 +59,10 @@
         (:transaction-id concept)
         (map :transaction-id (:service-associations concept))))
 
+(defmethod get-elastic-version :subscription
```
ygliuvt, line 62: why does this have to be `transaction-id`?

search-app/docs/api.md
```diff
@@ -4077,6 +4084,263 @@ Content-Length: 168
...
+
+<?xml version="1.0" encoding="UTF-8"?>
+<results>
+    <hits>1</hits>
+    <took>5</took>
+    <references>
+        <reference>
+            <name>someSub1</name>
+            <id>SUB1200000005-PROV1</id>
+            <location>http://localhost:3003/concepts/SUB1200000005-PROV1/1</location>
```
jaybarra, line 4156: nitpick: replace localhost:3003 with %CMR-ENDPOINT%

indexer-app/src/cmr/indexer/services/index_service.clj
```diff
@@ -412,6 +417,10 @@
   (let [service-associations (meta-db/get-associations-for-service context concept)]
     (get-elastic-version-with-associations context concept {:service-associations service-associations})))
 
+(defmethod get-elastic-version :subscription
```
ygliuvt, line 420: This function is already defined in `elasticsearch` namespace above. why is it defined here again?

search-app/docs/api.md
```diff
@@ -4077,6 +4084,263 @@ Content-Length: 168
 
 On occasions when service dissociation cannot be processed at all due to invalid input, service dissociation request will return a failure status code with the appropriate error message.
 
+### <a name="subscription"></a> Subscription
+
+A subscription enables data to be accessed via a universal resource locator, Subscription metadata is in JSON format and conforms to [UMM-Sub Schema](https://git.earthdata.nasa.gov/projects/EMFD/repos/unified-metadata-model/browse/subscription).
```
ygliuvt, line 4089: Can we add more text here to indicate what is a subscription and how it is supposed to be used?

indexer-app/src/cmr/indexer/services/index_service.clj
sxu123: Good question! This really threw me off because I had the one defined in elasticsearch namespace first, and it didn't work because that one is not the one being used, when you don't have the associations.  These are the two defmulti defined in different name spaces but sharing the same name!  I could remove the other one, but I thought I would just leave it there because in the future, if somehow we need to association subscription with collection, it will be needed.

indexer-app/src/cmr/indexer/data/elasticsearch.clj
sxu123: Because if you look at the service and variable, they pick the max of transaction-id and the transaction-ids of the associations. so it makes sense to pick the transaction-id when you don't have associations.

search-app/docs/api.md
sxu123: Good point. Done.

search-app/docs/api.md
sxu123: Yes we can.

system-int-test/test/cmr/system_int_test/search/subscription/concept_retrieval_test.clj
```diff
@@ -0,0 +1,124 @@
...
+   (search/retrieve-concept concept-id revision-id {:accept accept-format
+                                                    :throw-exceptions true})))
+
+(defn- assert-retrieved-concept-error
+  "Verify the expected error code and error message are returned when retrieving the given concept."
+  [concept-id revision-id accept-format err-code err-message]
+  (let [{:keys [status errors]} (handle-retrieve-concept-error
+                                 concept-id revision-id accept-format)]
+    (is (= err-code status))
+    (is (= [err-message] errors))))
```
ygliuvt, line 59: Rather than copy paste these code from `cmr.system-int-test.search.service.concept-retrieval-test`, can you refactor these into an existing helper namespace and just call them from there?

indexer-app/src/cmr/indexer/data/elasticsearch.clj
ygliuvt: Subscription concept is not associated with any concepts. It makes more sense to just delete this function and let the default implementation (which is for any concepts without associations) to handle the elastic version.

system-int-test/test/cmr/system_int_test/search/subscription/concept_retrieval_test.clj
sxu123: Maybe a separate ticket to refactor all the service and variable and subscription ones.

indexer-app/src/cmr/indexer/data/elasticsearch.clj
sxu123: This is not the one being used right now(see my other comment about the same name defmethod ) The default one over there is using revision-id, which I was using before but it didn't work. Remember I was able to search for deleted concept at one point?  That's because when the concept is deleted, the delete-document didn't work in elastic search. the revision-id is lower than the current id in ES, which was using transaction.

ygliuvt: I don't follow what you are saying.

indexer-app/src/cmr/indexer/data/concepts/subscription.clj
```diff
@@ -0,0 +1,44 @@
...
+   [clojure.string :as string]
+   [cmr.common.mime-types :as mt]
+   [cmr.indexer.data.elasticsearch :as es]))
+
+(defmethod es/parsed-concept->elastic-doc :subscription
+  [context concept parsed-concept]
+  (let [{:keys [concept-id revision-id deleted provider-id native-id user-id
+                revision-date format extra-fields]} concept
+        {:keys [subscription-name subscriber-id collection-concept-id]} extra-fields]
+    (if deleted
```
CayvonH, line 13: nit: These two maps appear the same except for `:metadata-format`. Define one map and and do a 
```
(if-not deleted 
   (merge sub-map 
                {:metadata-format (name (mt/format-key format))}) 
   sub-map)
```

search-app/src/cmr/search/api/concepts_lookup.clj
```diff
@@ -42,6 +42,9 @@
    :service #{mt/any
               mt/xml
               mt/umm-json}
+   :subscription #{mt/any
+              mt/xml
+              mt/umm-json}
```
CayvonH, line 47: nit: fix indentation

indexer-app/src/cmr/indexer/data/concepts/subscription.clj
sxu123: Actually I don't need to have so many fields when deleted. Will remove a few.

search-app/src/cmr/search/api/concepts_lookup.clj
sxu123: Done.

indexer-app/src/cmr/indexer/services/index_service.clj
sxu123: Removed the one in elasticsearch.clj.

sxu123: Can't remove it. it's being used too, and the default one in elasticsearch.clj doesn't work either.

PR #939, opened 2020-04-07
CMR-6287: update response, add tests

mschmele:


indexer-app/src/cmr/indexer/services/index_service.clj
```diff
@@ -230,7 +230,7 @@
              {:type type
               :_id id
               :value v
-              :fields (camel-snake-kebab/->snake_case_keyword (name (key value)))
+              :fields v
```
jaybarra, line 233: Why are fields and value the same value?

system-int-test/test/cmr/system_int_test/search/autocomplete/suggestion_reindex_test.clj
```diff
@@ -76,8 +75,26 @@
...
+                                                          :concept-type :collection
+                                                          :format-key :echo10})
         _ (index/wait-until-indexed)]
 
     (index/reindex-suggestions)
     (index/wait-until-indexed)
-    (search/get-autocomplete-suggestions "q=level2")))
+    (testing "Ensure that response is in proper format and results are correct"
+      (is (= (get-in (search/get-autocomplete-json "q=level2") [:feed :entry])
+            [{:score 0.25604635 :type "organization" :value "Langley DAAC User Services" :fields "Langley DAAC User Services"}
```
jaybarra, line 88: I am guilty of hard-coding these as well, but do we have a "near" operator for float comparison?

indexer-app/src/cmr/indexer/services/index_service.clj
mschmele: It's really just to populate this value. `fields` is only really important for science keywords, and the bit of code it replaced wasn't actually pulling anything

system-int-test/test/cmr/system_int_test/search/autocomplete/suggestion_reindex_test.clj
mschmele: Not in the way that you're thinking of as far as I know. We have a way to compare collections such that the order of the items doesn't matter, but that's not really useful here. I assume that if the data never changes then none of the results should either

indexer-app/src/cmr/indexer/services/index_service.clj
CayvonH: i don't really follow, you're indexing the same thing twice?

system-int-test/test/cmr/system_int_test/search/autocomplete/suggestion_reindex_test.clj
CayvonH: Does the actual score value matter here? Can you just do two assertions, one with the fields you expect and one making sure the scores are in descending order. This seems like it could easily change the next time you add tests to it.

indexer-app/src/cmr/indexer/services/index_service.clj
abbottry: But.. even for science keywords, would this mean that fields would be ultimately be assigned the leaf node of the science keyword?

system-int-test/test/cmr/system_int_test/search/autocomplete/suggestion_reindex_test.clj
abbottry: Seems like you could just test for the existence of the `score`.

indexer-app/src/cmr/indexer/services/index_service.clj
mschmele: The science keyword parsing code is in a different function

Yes, just because i needed to pick something to fill the value. I guess I could go with `not-provided`?

abbottry: Ah i gotchu, this is fine from my perspective.

mschmele:
> Does this mean we’ll see “not provided” in the response now?

Yes, but only for things that aren't science keywords

indexer-app/src/cmr/indexer/services/index_service.clj
CayvonH: I think my confusion here is i don't understand your use case. It seems you're creating some sort of hierarchy by joining a path on `:` where `value` is your leaf. What is the purpose of this and why not just have one indexed value and have it be a nested map?

system-int-test/test/cmr/system_int_test/search/autocomplete/suggestion_reindex_test.clj
```diff
@@ -76,8 +90,29 @@
...
+                                                         {:provider-id "PROV1"
+                                                          :concept-type :collection
+                                                          :format-key :echo10})
         _ (index/wait-until-indexed)]
 
     (index/reindex-suggestions)
     (index/wait-until-indexed)
-    (search/get-autocomplete-suggestions "q=level2")))
+    (testing "Ensure that response is in proper format and results are correct"
+      compare-autocomplete-results
```
jaybarra, line 102: I don't think this has the expected and results being passed into it. Can you confirm? I had to add '()' aroudn it and the following arguments to get it run

mschmele: Oh yeah looks like parinfer bit me. Fixed

PR #943, opened 2020-04-10
CMR-6295 catching parsing exceptions during autocomplete suggestion reindexing

jaybarra:
Updated autocomplete indexing to be more resilient to parsing errors for badly formatted data. This will prevent the autocomplete from failing midway through a re-index operation.

The root issue is badly formatted data that was attempted to be parsed.

### DIFF
--- a/system-int-test/test/cmr/system_int_test/search/autocomplete/suggestion_reindex_test.clj
+++ b/system-int-test/test/cmr/system_int_test/search/autocomplete/suggestion_reindex_test.clj
@@ -88,6 +88,7 @@
                             (fu/processing-level-id "PL1")
                             {:DataCenters [(data-umm-spec/data-center {:Roles ["ARCHIVER"] :ShortName "DOI/USGS/CMG/WHSC"})]
                              :ScienceKeywords [(:ScienceKeywords (fu/science-keywords sk1 sk2))]})
+
         coll3 (d/ingest-concept-with-metadata-file "CMR-6287/C1000000029-EDF_OPS.xml"
                                                    {:provider-id "PROV1"
                                                     :concept-type :collection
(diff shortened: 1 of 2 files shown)


system-int-test/test/cmr/system_int_test/search/autocomplete/suggestion_reindex_test.clj
```diff
@@ -14,25 +15,21 @@
...
-   (is (= expected actual))))
+  "Compare expected to actual response for the following:
+  - Items are ordered by score in descending order
+  - Ensure that the other fields match"
+  [actual expected]
+  (let [scores (map :score actual)
+        actual-scores-descending? (->> actual
+                                       (map :score)
+                                       (apply >=))
+                                        ; we don't need to actually compare scores
```
CayvonH, line 28: nit: Change indentation of this comment to be over what it is referring to. Use two `;;` for comments on their own line.

system-int-test/test/cmr/system_int_test/search/autocomplete/suggestion_reindex_test.clj
```diff
@@ -90,29 +88,46 @@
...
+     "shorter match"
+     "q=solar"
+     [{:score 3.4995544, :type "science_keywords", :value "Solar Irradiance", :fields "Sun-Earth Interactions:Solar Activity:Solar Irradiance"}
+      {:score 1.9641972, :type "science_keywords", :value "Solar Irradiance", :fields "Atmosphere:Atmospheric Radiation:Solar Irradiance"}
+      {:score 0.20964509, :type "organization", :value "ACRIM SCF", :fields "ACRIM SCF"}
+      {:score 0.16771607, :type "organization", :value "Langley DAAC User Services", :fields "Langley DAAC User Services"}]
+
+     "more complete match"
+     "q=solar irradiation"
+     [{:score 5.0869484, :type "science_keywords", :value "Solar Irradiance", :fields "Sun-Earth Interactions:Solar Activity:Solar Irradiance"}
```
CayvonH, line 129: I know this isn't part of this PR but can you explain why fields is constructed like it is? It looks like a colon separated path with `value` as the tail but why is this needed / why not a vector or a nested map?

jaybarra: @CayvonH this was to provide EDSC values in a format they could use a single function to parse instead of checking for hierarchical vs flat suggestions. There is still some discussion over science keywords and the "fields" field

PR #950, opened 2020-04-22
Cmr 6203 allow one collection passed to associations

sxu123:


search-app/docs/api.md
```diff
@@ -3654,12 +3654,10 @@ A variable identified by its concept id can be associated with collections throu
...
 Note: The following new features are added to support UVG:
 1. Each variable can only be associated with one collection.
-2. Each collection can only be associated with variables that don't share the same name.
+2. Each collection can only be associated with multiple variables that don't share the same name.
 3. variable and collection can only be associated if they are from the same provider.
 
-Future work:
-1. Currently we are not addressing any existing data that doesn't satisfy the above requirements.
-2. We still require a list of collection concept revisions to be passed in, even though only one collection revision is allowed in the list. A ticket is filed to address these issues in the future if necessary.
+As a consequence, we now allow one collection, as well as a list of collections, to be passed in to all association endpoints.
```
ygliuvt, line 3660: A variable can only be associated with only one collection. So I do not get this line: "As a consequence, we now allow on collection, as well as a list of collections, to be passed in to all association endpoints.". We should not have allowed a list of collections based on the rule. 
Also I don't think this fix is necessary if all we changed is the input body can be either a map or a list of a map. 
The change we should make, if any, should be to change this endpoint to something like:
```
%CMR-ENDPOINT%/variables/V1200000008-PROV1/collections/C1200000005-PROV1/associations
```

ygliuvt: If we fix it properly, it will be a API breaking change and need to be communicated to clients.

sxu123: Now it allows a map to be passed in, and allow a list of one collection to be passed in. So no api change even though I have communicated to the clients that I might change the api.  The ticket was written because of one comment made at the time, I wrote it as a "potential" issue - if people complain, I will fix it, otherwise leave it as is: i.e. allow a list of ONE collection to be passed in.  And it was pulled to the sprint to be fixed.  I was going to change the schema but since all the associations share one tagging association schema,  I don't want to make it a special case for variable and I don't think it hurts anything to change it either way.

sxu123: Now, for all associations, you can either pass a list of collections, a list of one collection, or a collection.   In variable case, when you pass a list of collections, it will say you can only pass one, which is the baseline behavior.  I don't see any problem with that.  Like you said, this fix is not "necessary",  but it will satisfy the need for those who do care and doesn't break the api.

ygliuvt: I really do NOT see much value in allowing a single map being passed in as the request body. Any value that is added will be offset by the added confusion. If we don't change the route of the API endpoint, just require a list of maps as request body is totally fine (even though there can only be one map for variable association) as it is consistent with all the other type of associations.

sxu123: MMT still prefers to have this ticket fixed, while keeping the old legacy api.  I see the value in a general sense:  when you pass in one single object, you don't have to pass it in through a list, which applies to all associations.

CayvonH: Why though? It seems weird that we are making an API change that makes absolutely no sense to satisfy mmt not wrapping their json map in an array. This route should either remain unchanged or more correctly changed to a path based route `/variables/:variable-concept-id/assocations/:collection-concept-id`.

sxu123: The many to many relationship between collection and variable is changed to one to many by a UVG feature.  One argument is it's not unlikely that it may change back to many to many in the future. They want to keep that flexibility.  Another is you should be able to either pass one collection directly, or pass it in a list.

PR #954, opened 2020-04-29
CMR-6285 Variable dissociate endpoint only allows one at a time

jaybarra:


search-app/docs/api.md
```diff
@@ -3686,15 +3686,16 @@ On occasions when variable association cannot be processed at all due to invalid
 
 #### <a name="variable-dissociation"></a> Variable Dissociation
 
-A variable identified by its concept id can be dissociated from collections through a list of collection concept revisions similar to variable association requests.
+A variable identified by its concept id can be dissociated from collections similar to variable association requests.
+Future work:
```
ygliuvt, line 3690: It becomes apparent that we need to change the variable dissociation route once you put this in the API doc. We should just change the dissociation route to something like:
```
DELETE %CMR-ENDPOINT%/associate/variables/V1200000008-PROV1/collections/C1200000002-PROV1
```
rather than changing this awkward dissociation endpoint that is for multiple concepts which has been eliminated from our use cases. This endpoint will be in-sync with the future change to make variable association route to the same route with POST. So I think we should add the above dissociation endpoint and mark this existing endpoint as deprecated. We don't need to change the existing route to only accept a single collection concept.
Sorry that I didn't come to this realization immediately when I first reviewed the PR.

search-app/docs/api.md
```diff
@@ -3672,11 +3695,11 @@ Content-Length: 168
 [
   {
     "variable_association":{
-      "concept_id":"VA1200000009-CMR",
+      "concept_id":"V1200000008-PROV1
```
ygliuvt, line 3698: The original documentation is correct. We should not have changed this line.

search-app/docs/api.md
```diff
@@ -3672,11 +3695,11 @@ Content-Length: 168
...
       "revision_id":1
     },
     "associated_item":{
-      "concept_id":"C1200000005-PROV1"
+      "concept_id":"C1200000005-PROV1
```
ygliuvt, line 3702: Missed a quotation mark.

search-app/src/cmr/search/api/association.clj
```diff
@@ -57,6 +57,18 @@
     400
     200))
 
+(defmulti valid-association-args-count?
```
ygliuvt, line 60: For things this simple, we should just use a single function with `if` statement.

search-app/src/cmr/search/api/association.clj
```diff
@@ -84,6 +96,30 @@
   (validate-association-content-type headers)
   (info (format "Dissociating %s [%s] from collections: %s by client: %s."
                 (name concept-type) concept-id body (:client-id context)))
-  (let [results (assoc-service/dissociate-from-collections context concept-type concept-id body)
-        status-code (association-results->status-code concept-type results)]
-    (api-response status-code results)))
+  (if-not (valid-association-args-count? concept-type
```
ygliuvt, line 99: nit: use `if` rather than `if-not`.

ygliuvt: of course flip the statements within the block after you switch. It is just more readable to me that way.

transmit-lib/src/cmr/transmit/association.clj
```diff
@@ -28,6 +28,19 @@
   [context concept-id]
   (associations-by-concept-ids-url context :variable concept-id))
 
+(defmulti single-association-url
```
ygliuvt, line 31: I would do the generalization when we have a use case later, but you can keep it since you already did it.

search-app/src/cmr/search/api/association.clj
```diff
@@ -84,6 +96,30 @@
...
+                ;; TODO get rid of redundant json->string conversion
+                (associate-concept-to-collections
+                  request-context headers (json/generate-string [{:concept-id collection-concept-id}]) :variable variable-concept-id))
+          
+          ;; Dissociate a variable from a collection
+          (DELETE "/:collection-concept-id" [collection-concept-id
+                                             req :as {:keys [request-context headers]}]
+                  ;; TODO get rid of redundant json->string conversion
+                  (dissociate-concept-from-collections
+                    request-context headers (json/generate-string [{:concept-id collection-concept-id}]) :variable variable-concept-id)))))))
```
sxu123, line 125: @ygliuvt , you guys don't want to do end point like `%CMR-ENDPOINT%/associate/variables/V1200000008-PROV1/collections/C1200000005-PROV1`

sxu123: Personally, I would like to see all the association routes in one place, rather than some here and some in the cmr.search.api.variables.

search-app/src/cmr/search/api/routes.clj
```diff
@@ -56,6 +57,9 @@
         ;; Add routes for tagging
         tags-api/tag-api-routes
 
+        ;; Add routes for assoication
+        association-api/association-api-routes
+
```
sxu123, line 62: Like here, it's confusing to add routes for association, then followed by "Add routes for variable association".

search-app/src/cmr/search/api/association.clj
jaybarra: @sxu123 I feel like consolidating them would make sense. But right now only variable has the limit on association numbers. The others could be moved here eventually and deprecate the others routes. That can be discussed further.

sxu123: You don't need to consolidate others, just put all the variable associations together in the variable.clj.

sxu123: It's no big deal. If you want to do it this way, then mark which one is preferred way and which will be deprecated.

PR #956, opened 2020-05-06
CMR-6370: Add metadata support for UMM-T.

sxu123:


metadata-db-app/int-test/cmr/metadata_db/int_test/concepts/utils/tool.clj
```diff
@@ -0,0 +1,33 @@
+(ns cmr.metadata-db.int-test.concepts.utils.tool
+  "Defines implementations for all of the multi-methods for tools in the metadata-db
+  integration tests."
+  (:require
+   [cheshire.core :as json]
+   [cmr.metadata-db.int-test.concepts.utils.interface :as concepts]))
+
+(def tool-json
```
ygliuvt, line 8: nit: this should only define a map and `json/generate-string` should be in the `get-sample-metadata` function on line 15.

PR #958, opened 2020-05-12
CMR-5906

ygliuvt:
Add point and radius search functionality in CMR.

### DIFF
--- a/search-app/docs/api.md
+++ b/search-app/docs/api.md
@@ -74,6 +74,7 @@ Join the [CMR Client Developer Forum](https://wiki.earthdata.nasa.gov/display/CM
         * [Bounding Box](#c-bounding-box)
         * [Point](#c-point)
         * [Line](#c-line)
+        * [Circle](#c-circle)
     * [Shapefile](#c-shapefile)
     * [Additional Attribute](#c-additional-attribute)
     * [Author](#c-author)
@@ -95,6 +96,7 @@ Join the [CMR Client Developer Forum](https://wiki.earthdata.nasa.gov/display/CM
         * [Bounding Box](#g-bounding-box)
         * [Point](#g-point)
         * [Line](#g-line)
+        * [Circle](#g-circle)
     * [Shapefile](#g-shapefile)
     * [Orbit number](#g-orbit-number)
     * [Orbit equator crossing longitude](#g-orbit-equator-crossing-longitude)
@@ -1695,6 +1697,12 @@ Lines are provided as a list of comma separated values representing coordinates 
 
     curl "%CMR-ENDPOINT%/collections?line=-0.37,-14.07,4.75,1.27,25.13,-15.51"
 
+##### <a name="c-circle"></a> Circle
+
+Circle defines a circle area on the earth with a center point and a radius. The center parameters must be 3 comma-separated numbers: longitude of the center point, latitude of the center point, radius of the circle in meters. The circle center cannot be on North or South pole. The radius of the circle must be between 10 and 1,000,000.
+
+    curl "%CMR-ENDPOINT%/collections?circle=-87.629717,41.878112,1000"
+
 Note: A query could consist of multiple spatial parameters of different types, two bounding boxes and a polygon for example. If multiple spatial parameters are present, all the parameters irrespective of their type are AND'd in a query. So, if a query contains two bounding boxes and a polygon for example, it will return only those collections which intersect both the bounding boxes and the polygon.
 
 #### <a name="c-shapefile"></a> Find collections by shapefile
@@ -1949,6 +1957,10 @@ The parameters used for searching granules by spatial are the same as the spatia
 ##### <a name="g-line"></a> Line
 
     curl "%CMR-ENDPOINT%/granules?collection_concept_id=%CMR-EXAMPLE-COLLECTION-ID%&line=-0.37,-14.07,4.75,1.27,25.13,-15.51"
+
+##### <a name="g-circle"></a> Circle
+
+    curl "%CMR-ENDPOINT%/granules?collection_concept_id=%CMR-EXAMPLE-COLLECTION-ID%&circle=-87.629717,41.878112,1000"
 
 #### <a name="g-shapefile"></a> Find granules by shapefile
 
--- a/search-app/src/cmr/search/services/parameters/converters/spatial.clj
+++ b/search-app/src/cmr/search/services/parameters/converters/spatial.clj
@@ -15,7 +15,7 @@
   {:default 32 :type Long})
 
 (defn- shape->search-shape
-  "Returns the search shape of the given shape. For cirle, we convert it into a poygon to search."
+  "Returns the search shape of the given shape. For cirle, we convert it into a polygon to search."
   [shape]
   (if (= cmr.spatial.circle.Circle (type shape))
     (spatial-circle/circle->polygon shape (points-on-circle))
--- a/search-app/src/cmr/search/services/parameters/converters/spatial.clj
+++ b/search-app/src/cmr/search/services/parameters/converters/spatial.clj
@@ -1,11 +1,25 @@
 (ns cmr.search.services.parameters.converters.spatial
   "Contains parameter converters for spatial parameters"
-  (:require [cmr.common-app.services.search.params :as p]
-            [cmr.search.models.query :as qm]
-            [cmr.common-app.services.search.group-query-conditions :as gc]
-            [cmr.spatial.codec :as spatial-codec]
-            [cmr.common.services.errors :as errors]
-            [clojure.string :as str]))
+  (:require
+   [clojure.string :as string]
+   [cmr.common-app.services.search.group-query-conditions :as gc]
+   [cmr.common-app.services.search.params :as p]
+   [cmr.common.config :as cfg :refer [defconfig]]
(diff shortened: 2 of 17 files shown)


spatial-lib/src/cmr/spatial/circle.clj
```diff
@@ -0,0 +1,106 @@
...
+(defrecord Circle
+  [
+   ^Point center
+   ^double radius
+   ])
+
+(record-pretty-printer/enable-record-pretty-printing Circle)
+
+(defn circle
+  "Creates a new minimum bounding rectangle"
```
CayvonH, line 28: docstring should say cricle instead of bounding rectangle

system-int-test/test/cmr/system_int_test/search/granule_spatial_search_test.clj
```diff
@@ -405,6 +405,32 @@
         [-5.95,-23.41,12.75,-23.69,11.11,-10.38,-6.62,-10.89,-5.95,-23.41]
         [whole-world polygon-with-holes-cart wide-south-cart normal-poly-cart]))
 
+    (testing "valid circle searches"
```
CayvonH, line 408: Can you add tests that cross the antemeridian?

spatial-lib/src/cmr/spatial/circle.clj
```diff
@@ -0,0 +1,106 @@
...
+
+(defn circle->polygon
+  "Returns the polygon approximation of the circle with the given number of points.
+   Reference: https://github.com/gabzim/circle-to-polygon."
+  [^Circle cir ^long n]
+  (let [{:keys [^Point center ^double radius]} cir
+        lon1 (.lon_rad center)
+        lat1 (.lat_rad center)
+        r-rad (/ radius EARTH_RADIUS_APPROX)
+        points (persistent!
```
mschmele, line 73: This should be its own function with a docstring and whatnot - there is a lot of logic here

spatial-lib/src/cmr/spatial/lr_binary_search.clj
```diff
@@ -267,3 +267,15 @@
            (warn "Use mbr from one of the points in the polygon because lr is not found "
                  (pr-str polygon))
            (m/point->mbr (-> polygon :rings first :points first))))))))
+
+(comment
```
mschmele, line 271: Not sure if you want this comment block here still

search-app/src/cmr/search/services/parameters/converters/spatial.clj
```diff
@@ -1,26 +1,41 @@
...
+   [cmr.common-app.services.search.params :as p]
+   [cmr.common.config :as cfg :refer [defconfig]]
+   [cmr.common.services.errors :as errors]
+   [cmr.search.models.query :as qm]
+   [cmr.spatial.circle :as spatial-circle]
+   [cmr.spatial.codec :as spatial-codec]))
+
+(defconfig points-on-circle
+  "The number of vertices of the polygon used to approximate a circle"
+  {:default 32 :type Long})
```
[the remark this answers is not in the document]
ygliuvt: It is configurable. From the tests I have run, 32 is a magic number that provides a good balance between precision and performance. I thought about making it a search parameter, but that opens up to abuse, so I didn't do it.

spatial-lib/src/cmr/spatial/lr_binary_search.clj
ygliuvt: It is very useful for debugging. I intentionally kept it.

search-app/docs/api.md
```diff
@@ -1695,6 +1697,12 @@ Lines are provided as a list of comma separated values representing coordinates
 
     curl "%CMR-ENDPOINT%/collections?line=-0.37,-14.07,4.75,1.27,25.13,-15.51"
 
+##### <a name="c-circle"></a> Circle
```
CayvonH, line 1700: Wonder if `radius` would be more clear than `circle`. thats super nit-picky though

spatial-lib/src/cmr/spatial/circle.clj
ygliuvt: The function is small already, most of the code is just calling this piece of code that does the math. It will be more unreadable to have it in its own function. The docstring has listed the reference already.

search-app/docs/api.md
ygliuvt: If we use `radius`, where is the `point` info? I am not sure I follow.

spatial-lib/src/cmr/spatial/circle.clj
```diff
@@ -0,0 +1,106 @@
...
+
+(def MAX_RADIUS
+  "Maximum radius in meters"
+  1000000)
+
+(def ^:const ^double EARTH_RADIUS_APPROX
+  "Radius of the earth in meters for polygon approximation of the circle"
+  6378137.0)
+
+(defn- validate-center
```
CayvonH, line 46: What would happen if a vertex point landed on a pole? Is this a possibility?

system-int-test/test/cmr/system_int_test/search/granule_spatial_search_test.clj
ygliuvt: Good point. Added tests.

spatial-lib/src/cmr/spatial/circle.clj
ygliuvt: I don't think it is possible for math to get there which is infinity.

search-app/docs/api.md
CayvonH: Fair enough

spatial-lib/src/cmr/spatial/circle.clj
mschmele: Right, but this function has its own `let` - we usually want to try to keep such complex logic out of `let` blocks for readability/maintainability. I'll still approve, but I disagree with you here. The logic looks good, though :)

ygliuvt:
> Right, but this function has its own let - we usually want to try to keep such complex logic out of let blocks for readability/maintainability. I'll still approve, but I disagree with you here. The logic looks good, though :)

I know the code is dense here. I still found this more readable as the 5 variables needed are right there in the function. It would be messy to define the 5 parameters again in a function call. It just don't feel natural. In the referenced javascript code, the center longitude and latitude are dereferenced and radian of the radius is re-computed multiple times in a loop. That's why I didn't do it that way even though it would've looked sort of cleaner. Sorry, I didn't make the change you wanted. I will make it up to you next time. :)

search-app/src/cmr/search/services/parameters/converters/spatial.clj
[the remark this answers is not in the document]
indiejames: This will be somewhat coarse for large circles, and for small circles, you might run into the problem of points being too close together (the spatial library doesn't like that).

spatial-lib/src/cmr/spatial/circle.clj
indiejames: Why can't the center be on a pole? And is it ok for the center to be near a pole?

search-app/src/cmr/search/services/parameters/converters/spatial.clj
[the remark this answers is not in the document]
indiejames: It might be better if this were a function of the radius - bigger radius, more points.

spatial-lib/src/cmr/spatial/circle.clj
```diff
@@ -0,0 +1,117 @@
...
+(record-pretty-printer/enable-record-pretty-printing Circle)
+
+(defn circle
+  "Creates a circle"
+  ([^Point center ^double radius]
+   (->Circle center radius))
+  ([^double lon ^double lat ^double radius]
+   (->Circle (p/point lon lat) radius)))
+
+(def MIN_RADIUS
```
indiejames, line 34: where did MIN_RADIUS and MAX_RADIUS come from?

ygliuvt: We defined them to be reasonable in size. The idea is that circle search is not intended for large areas. This address your suggestion to make the points proportional to radius. I didn't want to make it so fluid at the beginning as I have not done enough tests and have data to show that it would make sense. It is something we can easily add in if we have enough data to see what we should do in the future. Based on my tests in PROD, 16 points is often times slower than 32 points. More than 32 points, the delay will add up. I would like to collect some data before making any optimizations.

spatial-lib/src/cmr/spatial/circle.clj
ygliuvt: Near the pole is OK. On the pole causes the math calculation to be exactly zero and we can't draw the circle at all.

search-app/src/cmr/search/services/parameters/converters/spatial.clj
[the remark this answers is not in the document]
ygliuvt: See my comments on the definition of minimum and maximum radius.

spatial-lib/src/cmr/spatial/circle.clj
ygliuvt: I will merge this now. I will talk this over with the team again and make necessary changes once we decide on how to pick the number of points of the polygon.

PR #959, opened 2020-05-12
Cmr 6205 reject complex shapefiles

indiejames:
This PR adds limits to the size, number of features, and number of points for a shapefile.


search-app/src/cmr/search/middleware/shapefile.clj
```diff
@@ -0,0 +1,61 @@
...
+   (org.opengis.feature.type Name)))
+
+(defconfig max-shapefile-size
+  "The maximum size in bytes a shapefile can be"
+  {:default 1000000 :type Long})
+
+(defn- progress
+  "Progress function for `wrap-multipart-params`. This function simply throws an error if
+  the uploaded file exceeds a given limit."
+  [_reqest _bytes-read content-length _item-count]
```
CayvonH, line 44: reqest -> request. _Maybe_ useful to also log the `content-length` a user attempted to upload on failure.

search-app/src/cmr/search/services/parameters/converters/shapefile.clj
```diff
@@ -40,6 +40,15 @@
...
   {:default false :type Boolean})
 
+
+(defconfig max-shapefile-features
+  "The maximum number of feature a shapefile can have"
+  {:default 100000 :type Long})
+  
+(defconfig max-shapefile-points
+  "The maximum number of points a shapefile can have"
+  {:default 5000000 :type Long})
```
CayvonH, line 50: I believe this should be less than the max shapefile size. Even if i assume 1 byte per point thats 5mb where `max-shapefile-size` is 1mb. Or am I not accounting for some sort of compression?

search-app/src/cmr/search/services/parameters/converters/shapefile.clj
```diff
@@ -74,6 +83,15 @@
           :let [sub-geometry (.getGeometryN geometry index)]]
       (geo/geometry->condition sub-geometry options))))
 
+(defn geometry-point-count
+  "Get the number of points in the given Geometry"
+  [geometry]
```
CayvonH, line 88: nit: Type hint geometry for readability. Unrelated but i really like this library and wish we would have used it for spatial-lib.

system-int-test/test/cmr/system_int_test/search/granule_shapefile_search_test.clj
```diff
@@ -32,6 +33,99 @@
   {"ESRI" {:extension "zip" :mime-type mt/shapefile}
    "GeoJSON" {:extension "geojson" :mime-type mt/geojson}
    "KML" {:extension "kml" :mime-type mt/kml}})
+
+(deftest granule-shapefile-failure-cases
+  (side/eval-form `(shapefile/set-enable-shapefile-parameter-flag! true))
+  (let [saved-shapefile-max-size (shapefile-middleware/max-shapefile-size)
+        _ (side/eval-form `(shapefile-middleware/set-max-shapefile-size! 50000))
```
CayvonH, line 40: nit: move these to a fixture. I can't remember if your revert to default code below will get ran on an exception.

search-app/src/cmr/search/middleware/shapefile.clj
indiejames: done

search-app/src/cmr/search/services/parameters/converters/shapefile.clj
indiejames: Yeah - these were wrong. The max number of points should be 100,000 and the max number of features 5000.  These were arrived at through some simple testing with shapefiles of various size/complexity. There are additionally performance tests scheduled for down the road. These values are intended to be intentionally low for now until performance tests can establish higher safe numbers.

search-app/src/cmr/search/services/parameters/converters/shapefile.clj
indiejames: Done.

system-int-test/test/cmr/system_int_test/search/granule_shapefile_search_test.clj
indiejames: I only want these to run for the first set of tests (the failure cases), whereas I believe `use-fixtures` applies to the whole file. Also, exceptions in the tests do not prevent the revert to default.

system-int-test/test/cmr/system_int_test/search/granule_shapefile_search_test.clj
```diff
@@ -32,6 +33,99 @@
...
+                                            params)))))))
+                                  
+        "All granules query"
+        "polygon_with_hole.geojson" {} #"The CMR does not allow querying across granules in all collections with a spatial condition"
+
+        "Failed to parse GeoJSON file"
+        "invalid_json.geojson" {:name "provider" :content "PROV1"} #"Failed to parse GeoJSON file"
+
+        "Shapefile has too many features"
+        "too_many_features.geojson" {:name "provider" :content "PROV1"} #"GeoJSON feature count \[3\] exceeds the 2 feature limit"
```
ygliuvt, line 95: Just a comment: I find the file name and all the variable names confusing. I wish this file was named "three_features.geojson" and so did everything related to `too_many_features`. It is only too many when you set the maximum features to `2`.

ygliuvt:
Don't you want to document these limits in the API doc?

system-int-test/test/cmr/system_int_test/search/granule_shapefile_search_test.clj
indiejames: I think it's more important that the names identify what they are used for. 'three_features.geojson' doesn't indicate what it is for. The fact that it has three features is not important.

PR #961, opened 2020-05-13
CMR-6365 collections search accepts usage_score sort parameter

jaybarra:
`usage_score` should now correctly sort collections

### DIFF
--- a/system-int-test/test/cmr/system_int_test/search/collection_sorting_search_test.clj
+++ b/system-int-test/test/cmr/system_int_test/search/collection_sorting_search_test.clj
@@ -12,13 +12,34 @@
    [cmr.system-int-test.data2.core :as d]
    [cmr.system-int-test.data2.umm-spec-collection :as data-umm-c]
    [cmr.system-int-test.data2.umm-spec-common :as data-umm-cmn]
+   [cmr.system-int-test.utils.humanizer-util :as hu]
    [cmr.system-int-test.utils.dev-system-util :as dev-sys-util]
    [cmr.system-int-test.utils.index-util :as index]
    [cmr.system-int-test.utils.ingest-util :as ingest]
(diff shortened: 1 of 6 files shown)


system-int-test/test/cmr/system_int_test/search/collection_sorting_search_test.clj
```diff
@@ -444,3 +465,37 @@
...
+(deftest collection-usage-score-sorting-test
+  (let [make-named-collection (fn [cname]
+                                (d/ingest
+                                  "PROV1"
+                                  (dc/collection
+                                    {:entry-title cname
+                                     :short-name cname
+                                     :long-name (str cname "-long")
+                                     :version-id "1"})))
+        c1 (make-named-collection "a")
```
indiejames, line 478: You might want to add a collection for which there is no usage data.

jaybarra: It is assigned min/max int depending on the query so unused actually have the highest usage when sorting in an ascending fashion and worst usage when sorting in descending. This results in collections without an explicit usage always being last in the list.

system-int-test/test/cmr/system_int_test/search/collection_sorting_search_test.clj
```diff
@@ -12,13 +12,34 @@
    [cmr.system-int-test.data2.core :as d]
    [cmr.system-int-test.data2.umm-spec-collection :as data-umm-c]
    [cmr.system-int-test.data2.umm-spec-common :as data-umm-cmn]
+   [cmr.system-int-test.utils.humanizer-util :as hu]
```
mschmele, line 15: Nitpick: we're getting away from abbreviating requires like this - something like `humanizer-util` would be preferred

PR #962, opened 2020-05-13
CMR-5906

ygliuvt:
Updated rule to pick the number of points of the polygon to approximate a circle.
Increased the allowable radius to 6000km. Added configuration to make configuring number of points more flexible. Current rule is:
if radius < 5km, 32 points; otherwise 100 points.

### DIFF
--- a/search-app/src/cmr/search/services/parameters/converters/spatial.clj
+++ b/search-app/src/cmr/search/services/parameters/converters/spatial.clj
@@ -10,15 +10,31 @@
    [cmr.spatial.circle :as spatial-circle]
    [cmr.spatial.codec :as spatial-codec]))
 
-(defconfig points-on-circle
-  "The number of vertices of the polygon used to approximate a circle"
+(defconfig min-points-on-circle
+  "The minimum number of vertices of the polygon used to approximate a circle"
   {:default 32 :type Long})
+
+(defconfig max-points-on-circle
+  "The maximum number of vertices of the polygon used to approximate a circle"
+  {:default 100 :type Long})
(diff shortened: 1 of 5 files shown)


search-app/src/cmr/search/services/parameters/converters/spatial.clj
```diff
@@ -10,15 +10,31 @@
...
+  "The minimum radius in meters to trigger the use of maximum number of vertices of the polygon
+  to approximate the circle"
+  {:default 5000 :type Long})
+
+(defn- points-on-circle
+  "Returns the number of vertices of the polygon to approximate the circle"
+  [circle]
+  (if (< (:radius circle) (min-radius-for-max-points))
+    (min-points-on-circle)
+    (max-points-on-circle)))
```
sxu123, line 31: So 32 points for radius between 10 and 5000; 100 points for radius between 5000+ and 6,000,000.  Would it be better to make this incremental between 5000+ and 6,000,000 since it's a much bigger range? I know you don't want to make the max-points-on-circle a parameter but you can increase it proportionally based on the radius of the circle - maybe you can change it to max-starting-points-on-circle?

sxu123: unless 100 is really the absolute limit due to performance concern

ygliuvt: Based on our tests, we don't see any difference between 64 points and 100 points. And it appears to be not worth it to make the rule even murkier.

PR #965, opened 2020-05-15
CMR-6410: Add full-path field to variables index and related search p…

indiejames:
Add full-path field to variables index and related search parameter

### DIFF
--- a/indexer-app/src/cmr/indexer/data/concepts/variable.clj
+++ b/indexer-app/src/cmr/indexer/data/concepts/variable.clj
@@ -31,12 +31,21 @@
       (map #(merge base-doc (measurement-quantity->elastic-doc %)) quantities)
       [base-doc])))
 
+(defn- get-full-path
+  "Returns the full path to the variable in a granule"
+  [group-path name]
+  (if (seq group-path)
+    (str group-path "/" name)
+    name))
+  
+
 (defmethod es/parsed-concept->elastic-doc :variable
   [context concept parsed-concept]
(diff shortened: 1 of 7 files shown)


indexer-app/src/cmr/indexer/data/concepts/variable.clj
```diff
@@ -31,12 +31,21 @@
       (map #(merge base-doc (measurement-quantity->elastic-doc %)) quantities)
       [base-doc])))
 
+(defn- get-full-path
+  "Returns the full path to the variable in a granule"
+  [group-path name]
+  (if (seq group-path)
```
CayvonH, line 37: why `seq` here on a string?

indiejames: `seq` returns `nil` for an empty string. If `group-path` is empty, I just want to use `name` for the full-path.

ygliuvt: wouldn't `clojure.string/blank? ` be better for your purpose? It will also filter out any full-path that are all whitespaces.

indiejames: Good point. I'll make that change on monday and add a test for blank group-path.

indiejames: Done

indiejames:
Added docs and addressed PR comments.

PR #966, opened 2020-05-18
CMR-6431

ygliuvt:
Fixed orbit code to not throw exception when crossing range is nil.

### DIFF
--- a/system-int-test/test/cmr/system_int_test/search/granule_orbit_search_test.clj
+++ b/system-int-test/test/cmr/system_int_test/search/granule_orbit_search_test.clj
@@ -89,7 +89,7 @@
                                                    :format-key :dif10})]
     (index/wait-until-indexed)
 
-    (testing "Orbit search will not throw exception."
+    (testing "Orbit search will not throw exception when crossing range is nil"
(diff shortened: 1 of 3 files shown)


system-int-test/test/cmr/system_int_test/search/granule_orbit_search_test.clj
```diff
@@ -81,6 +81,22 @@
...
 
+(deftest orbit-bug-CMR-6431
+  (let [coll (d/ingest-concept-with-metadata-file "CMR-6431/sample-orbit-collection.xml"
+                                                  {:provider-id "PROV1"
+                                                   :concept-type :collection
+                                                   :native-id "orbit-collection"
+                                                   :format-key :dif10})]
+    (index/wait-until-indexed)
+
+    (testing "Orbit search will not throw exception."
```
indiejames, line 92: can you add "when crossing range is nil" to this to explain what is being tested a little better?

ygliuvt: Done.

PR #971, opened 2020-05-20
CMR-6330 updated autocomplete indexing to filter anti-values

jaybarra:


indexer-app/src/cmr/indexer/services/index_service.clj
```diff
@@ -253,6 +253,22 @@
                                    (remove nil?))]]
     suggestion-docs))
 
+(defn- anti-value?
+  "Returns whether if the term is an anti-value. e.g. \"not applicable\" or \"not provided\".
+  This is case-insensitive"
+  [term]
+  {:pre [(some? term)]}
```
daniel-zamora, line 260: Good use of pre function check

indexer-app/src/cmr/indexer/services/index_service.clj
```diff
@@ -268,8 +284,10 @@
                                        nil)))
                              (remove nil?))
         humanized-fields (map #(humanizer/collection-humanizers-elastic context %) parsed-concepts)
-        suggestion-docs (map get-suggestion-docs humanized-fields)]
-    (flatten suggestion-docs)))
+        suggestion-docs (->> (map get-suggestion-docs humanized-fields)
+                             flatten
```
daniel-zamora, line 288: Its hard to tell, but do we need flatten here, or can mapcat be used instead?

system-int-test/test/cmr/system_int_test/search/autocomplete/suggestion_reindex_test.clj
```diff
@@ -61,6 +66,27 @@
 (def sk6 (umm-spec-common/science-keyword {:Category "EARTH SCIENCE"
                                            :Topic "Popular"
                                            :Term "Omega"}))
+
+(def sk7 (umm-spec-common/science-keyword {:Category "missing"
```
daniel-zamora, line 70: I'm guessing nil values aren't possible here or do not need to be checked? only these anit-value string values?

system-int-test/test/cmr/system_int_test/search/autocomplete/suggestion_reindex_test.clj
```diff
@@ -15,19 +15,24 @@
...
   - Items are ordered by score in descending order
   - Ensure that the other fields match"
   [actual expected]
-  (let [actual-scores-descending? (->> actual
-                                       (map :score)
-                                       (apply >=))
-        ;; we don't need to actually compare scores
-        expected (map #(dissoc % :score) expected)
+  ;; check score returns
+  (when (>= 1 (count actual))
```
daniel-zamora, line 30: this is kinda weird, does (seq actual) work?

jaybarra: Hadn't thought to try seq, I'll see if I can switch that over.

system-int-test/test/cmr/system_int_test/search/autocomplete/suggestion_reindex_test.clj
jaybarra: Not sure if nil is possible here, or in general. I'll try it and see what happens and adjust if necessary.

indexer-app/src/cmr/indexer/services/index_service.clj
jaybarra: If the result is unaffected I can simplify

mschmele: I think I tried to use mapcat a while ago but couldn't get it to work correctly so I went with flatten

indexer-app/src/cmr/indexer/services/index_service.clj
```diff
@@ -268,8 +284,10 @@
...
+                             (remove anti-value-suggestion?))]
```
mschmele, line 289: Nitpick: I always find it easier to read threaded statements when you thread a bound value rather than an expression
```suggestion
        suggestion-docs (->> humanized-fields
                             (map get-suggestion-docs)
                             flatten
                             (remove anti-value-suggestion?))]
```

indexer-app/src/cmr/indexer/services/index_service.clj
```diff
@@ -243,7 +243,7 @@
         :let [key (key humanized-field)
               key-name (-> key
                            name
-                           (s/replace #"(\.humanized2|\.humanized|-sn|-id)" ""))
+                           (s/replace #"(\.humanized(_?2)?|-sn|-id)" ""))
```
mschmele, line 246: I'm pretty sure I tried to do something like this originally but couldn't because I am bad with regexes - nice

system-int-test/test/cmr/system_int_test/search/autocomplete/suggestion_reindex_test.clj
jaybarra: nil cannot be used here. It throws an exception when ingesting so I think it's safe to ignore that case.

PR #975, opened 2020-05-28
Cmr 6448 added DMRPP data format and mime type to the new UMM-G 1.6.1

eereiter:


umm-spec-lib/src/cmr/umm_spec/migration/version/granule.clj
```diff
@@ -96,6 +96,30 @@
...
+   this function again to iterate over its maps.
+   An updated map is returned."
+  [archive-and-dist-info-file]
+  (-> archive-and-dist-info-file
+      (update :Format #(if (= "DMRPP" %)
+                          "Not provided"
+                          %))
+      (update :MimeType #(if (= "application/vnd.opendap.dap4.dmrpp+xml" %)
+                           "Not provided"
+                           %))
```
mschmele, line 111: You can probably use `string/replace` here instead to simplify this a little bit

https://clojuredocs.org/clojure.string/replace

eereiter: The problem with replace is that it can't take a nil string. An if check is required to make sure that a string exists and then do the replace. Once that is done then the code isn't any simpler.  I will leave it as it is.

PR #979, opened 2020-06-05
Cmr 6415 enforce email subscription management acl

sxu123:
Part1 of 2(This PR): Enforce EMAIL_SUBSCRIPTION_MANAGEMENT ACL on Subscription ingest.
Part2 of 2(To be continued): Enforce EMAIL_SUBSCRIPTION_MANAGEMENT ACL on Subscription search.

### DIFF
--- a/acl-lib/src/cmr/acl/core.clj
+++ b/acl-lib/src/cmr/acl/core.clj
@@ -219,7 +219,7 @@
   ([context]
    (verify-subscription-management-permission context :update :system-object nil))
   ([context permission-type]
-   (verify-subscription-management-permission context permission-type :system-object nil))
+   (verify-subscription-management-permission context permission-type :provider-object nil))
   ([context permission-type object-identity-type provider-id]
    (verify-management-permission
      context
--- a/acl-lib/src/cmr/acl/core.clj
+++ b/acl-lib/src/cmr/acl/core.clj
@@ -216,18 +216,14 @@
 (defn verify-subscription-management-permission
   "Verifies the current user has been granted EMAIL_SUBSCRIPTION_MANAGEMENT
   permission in ECHO ACLs"
-  ([context]
-   (verify-subscription-management-permission context :update :system-object nil))
-  ([context permission-type]
-   (verify-subscription-management-permission context permission-type :provider-object nil))
-  ([context permission-type object-identity-type provider-id]
-   (verify-management-permission
-     context
-     permission-type
-     object-identity-type
-     provider-id
-     token-smp-cache-key
-     has-subscription-management-permission?)))
+  [context permission-type object-identity-type provider-id]
+  (verify-management-permission
+    context
+    permission-type
+    object-identity-type
+    provider-id
+    token-smp-cache-key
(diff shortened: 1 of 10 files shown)


acl-lib/src/cmr/acl/core.clj
```diff
@@ -157,38 +166,81 @@
...
+    (when-not has-permission?
+      (errors/throw-service-error
+        :unauthorized
+        "You do not have permission to perform that action."))))
+
+(defn verify-subscription-management-permission
+  "Verifies the current user has been granted EMAIL_SUBSCRIPTION_MANAGEMENT
+  permission in ECHO ACLs"
+  ([context]
+   (verify-subscription-management-permission context :update :system-object nil))
```
ygliuvt, line 220: The newly added `EMAIL_SUBSCRIPTION_MANAGEMENT_ACL` is on `provider-object`. It is unnecessary to create these functions on `system-object`.

sxu123: Sorry, cut and paste error. I'll change it to :provider-object

ygliuvt:
I checked Legacy services logs, and I don't see any exceptions similar to what we have in CMR-5364 for undefined ACE targets in Legacy Services, so I don't know if this issue has become OBE or MMT's use case of these new targets does not follow the same workflow as the `DASHBOARD_DAAC_CURATOR` target. I will approve it for now. We should keep an eye out for validation exceptions in Legacy Services and fix those exceptions when they pop up.

acl-lib/src/cmr/acl/core.clj
```diff
@@ -121,6 +121,10 @@
   "The cache key for the token to ingest management permission cache."
   :token-imp)
 
+(def token-smp-cache-key
+  "The cache key for the token to subscription management permission cache."
+  :token-smp)
+
 (def TOKEN_IMP_CACHE_TIME
   "The number of milliseconds token information will be cached."
   (* 5 60 1000))
```
jceaser, line 130: is 5 min standard for CMR for this kind of cache or is there a logic to the number?

sxu123: I guess through experience, 5 min can maximize the performance and  still provides that value that's not out of date.

acl-lib/src/cmr/acl/core.clj
```diff
@@ -157,38 +166,77 @@
...
       (let [acl-oit-key (access-control/acl-type->acl-key object-identity-type)]
         (->> (get-permitting-acls context
                                   object-identity-type
-                                  "INGEST_MANAGEMENT_ACL"
+                                  target
                                   permission-type)
              ;; Find acls for this provider
              (filter #(or (nil? provider-id)
-                          (= provider-id (get-in % [acl-oit-key :provider-id]))))
+                           (= provider-id (get-in % [acl-oit-key :provider-id]))))
```
jaybarra, line 182: nitpick: extra space isn't necessary

sxu123: will adjust it in part2 of this ticket.

PR #982, opened 2020-06-15
CMR-6450 Port Autocomplete changes to ES update branch

jaybarra:
Logic for autocomplete ported to ES 7 style. Reverting to use original edge_ngram analyzer instead of search_as_you_type due to scoring changes. 

Scoring has still changed somewhat but it closer to original than the initial implementation.

Removed `na` from anti-value list as sodium may be shortened to its elemental symbol.

### DIFF
--- a/indexer-app/src/cmr/indexer/services/index_service.clj
+++ b/indexer-app/src/cmr/indexer/services/index_service.clj
@@ -247,7 +247,7 @@
         :let [key (key humanized-field)
               key-name (-> key
                            name
-                           (s/replace #"(\.humanized(_?2)?|-sn|-id)" ""))
+                           (s/replace #"((_\.)?humanized(_?2)?|-sn|-id)" ""))
               value-map (as-> humanized-field h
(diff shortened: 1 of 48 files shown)


indexer-app/src/cmr/indexer/services/index_service.clj
```diff
@@ -247,7 +247,7 @@
         :let [key (key humanized-field)
               key-name (-> key
                            name
-                           (s/replace #"(\.humanized(_?2)?|-sn|-id)" ""))
+                           (s/replace #"((_\.)?humanized(_?2)?|-sn|-id)" ""))
```
mschmele, line 250: Are the changes to this file something we'd also want to add to master for 1.6 as well or do we only want these for the 7.x upgrades?

ygliuvt:
I don't see you updating the feature branch with master changes first. How would you avoid merge conflicts when you have to do that later? I think we need to talk this over.

PR #985, opened 2020-06-18
CMR-6481: Changed Subscription ACL.

sxu123:
1. Changed the name from EMAIL_SUBSCRIPTION_MANAGEMENT to SUBSCRIPTION_MANAGEMENT.

2. Changed the granting permissions from ["create" "read" "update" "detete"] to ["update" "read"] only, so that it would be consistent with the INGEST_MANAGEMENT_ACL behavior.


mock-echo-app/src/cmr/mock_echo/client/echo_util.clj
```diff
@@ -98,10 +98,10 @@
   "An ACL for managing access to ingest management functions."
   "INGEST_MANAGEMENT_ACL")
 
-;; Email subscription management acl
-(def email-subscription-management
+;; subscription management acl
+(def subscription-management
   "An ACL for managing access to email subscription functions."
```
charleshuang80, line 103: I think email should be removed in this description as well

search-app/docs/api.md
```diff
@@ -4670,7 +4670,7 @@ __Sample response__
 ```
 #### <a name="subscription-access-control"></a> Subscription Access Control
 
-Access to subscription is granted through the provider via the INGEST_MANAGEMENT_ACL and EMAIL_SUBSCRIPTION_MANAGEMENT
+Access to subscription is granted through the provider via the INGEST_MANAGEMENT_ACL and SUBSCRIPTION_MANAGEMENT
```
charleshuang80, line 4673: So a user needs to have INGEST_MANAGEMENT_ACL in addition to SUBSCRIPTION_MANAGEMENT to CRUD Subscriptions?

sxu123: Yes. INGEST_MANAGEMENT ACL was there before.

mock-echo-app/src/cmr/mock_echo/client/echo_util.clj
sxu123: I think it doesn't hurt to leave it there because it is for email subscription. It's just that in ACL name, for some reason, they don't like EMAIL to be there.

search-app/docs/api.md
sxu123: I mean we want to control who can ingest with INGEST_MANAGEMENT ACL for all concepts.

charleshuang80: I thought we could have Subscriptions just be managed by SUBSCRIPTION_MANAGEMENT. But if it needs the other one, I think that is ok, we just need to be very clear about it so everyone knows. That it is here in the documentation is helpful.

mock-echo-app/src/cmr/mock_echo/client/echo_util.clj
charleshuang80: Dana's comment in the first demo was actually that we should remove all reference to 'email' for subscriptions because there can be other types of notifications that a user will receive and we don't want to be boxed into this one notification type. I forget some of the other ideas, but I think RSS was one. It's not a huge deal, but might be easier to remove 'email' here now, instead of having to remember to remove it in the future when we add more notification types.

sxu123: Okay. Removed.

search-app/docs/api.md
ygliuvt, line 4673: Can you add description as to how the permission is granted through the two ACLs? I don't think anyone reading this document can follow without more explanation.

system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
```diff
@@ -22,12 +22,12 @@
                                                                   [:read :update])]))
 
 (deftest subscription-ingest-on-prov2-test
```
ygliuvt, line 24: We are missing tests on specific ACL permission for update/delete. e.g. 1) with Ingest Management ACL update permission, but no Subscription Management ACL update permission trying to update a subscription concept. 2) with  Subscription Management ACL update permission, but no  Ingest Management ACL update permission trying to update a subscription concept, and.delete a subscription concept etc. Will these tests be added in a later PR?

search-app/docs/api.md
ygliuvt: I couldn't tell exactly what kind of permissions are needed to update/delete a subscription concept as I can't find API doc or tests describing the expected behavior.

system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
sxu123: Ingest Management ACL permissions were tested before when the subscription was first added. Now we want to add the Subscription Management ACL on top of Ingest Management ACL, so the tests have Ingest Management permissions granted already, just test the Subscription ACL permissions. I don't think it's necessary to test all the permutations, You test one, when it's done, you can just grant that permission, then test the other.

sxu123: provider_ingest_permissions_test.clj contains the tests for Ingest Management ACL.  For Subscription Management ACL, it's covered by subscription_ingest_tests.  Whenever you add something new, you test on top of what's already there, rather than testing all the permutations of all the existing ACLs, validations.

search-app/docs/api.md
sxu123: As far as I can tell, update/delete subscription requires "update" permission granted through both INGEST_MANAGEMENT_ACL and SUBSCRIPTION_MANAGEMENT ACL.  So, I moved this part to the Ingest api doc. Not sure why INGEST_MANAGEMENT_ACL for other concepts are in the search api doc, but absent in the ingest doc.   For the search api doc here, I don't see INGEST_MANAGEMENT_ACL plays a role, so I just added the SUBSCRIPTION_MANAGEMENT here.

ygliuvt: The permissions should be listed in Ingest API doc. The reason Ingest Management ACL is mentioned in search API doc is because we don't ingest of certain concepts in the search API, e.g. tags and we need to mention it for that. We should constantly improve the API docs.

system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
sxu123: Done.

PR #988, opened 2020-06-19
Cmr 6206 shapefile simplification

indiejames:
Adds a new parameter, simplify-shapefile, which, when set to true, will attempt to reduce the number of points in an uploaded shapefile until it is under the 5000 point limit.


search-app/docs/api.md
```diff
@@ -1735,10 +1743,21 @@ Examples:
 
   curl -XPOST "%CMR-ENDPOINT%/collections" -F "shapefile=@box.kml;type=application/vnd.google-earth.kml+xml" -F "provider=PROV1"
 
-Internally a WGS 84 Coordinate Reference System (CRS) is used. The system will attempt to tranform shapefile geometry that uses a differnt CRS, but this is not guaranteed to work and the request will be rejected if a suitable tranformation is not found.
+Internally a WGS 84 Coordinate Reference System (CRS) is used. The system will attempt to transform shapefile geometry that uses a different CRS, but this is not guaranteed to work and the request will be rejected if a suitable transformation is not found.
 
 **NOTE:** This is an experimental feature and may not be enabled in all environments.
 
+#### <a name="c-shapefile-simplification"></a> Simplifying shapefiles during collection search
+Shapefiles are limited to 5000 points by deafult. A user can request that the CMR try to simplify (reduce the number of points) a shapefile by setting the `simplify-shapefile` parameter to `true` in the POST used to pass in the shapefile.
```
ygliuvt, line 1751: Add some language such as "User using a shapefile larger than the CMR supported shapefile limits can use the `simplify-shapefile` parameter to ...." to make it clear why a user would want to use the parameter.

search-app/docs/api.md
```diff
@@ -1735,10 +1743,21 @@ Examples:
...
+
+Example:
+
+  curl -XPOST "%CMR-ENDPOINT%/collections" -F "simplify-shapefile=true"  -F "shapefile=@africa.zip;type=application/shapefile+zip" -F "provider=PROV1"
+
+Note that the simplification process attempts to preserve topology, i.e., the relationship between polygon outer boundaries and holes. The process uses the [Douglas-Peucker algorithm](https://en.wikipedia.org/wiki/Ramer%E2%80%93Douglas%E2%80%93Peucker_algorithm) and as such may result in geometries with less coverage than the original shapefile and potentially a loss of matching results.
+
+The simplification may fail if the process cannot reduce the number of points in the file to below the limit. Also the simplification only reduces the number of points in the file, so a shapefile will still fail if the file size  is too large or there are too many features.
```
ygliuvt, line 1759: remove the extra space in "file size  is too large" before the `is`.

search-app/docs/api.md
indiejames: Done.

search-app/docs/api.md
indiejames: Done.

search-app/src/cmr/search/middleware/shapefile_simplification.clj
```diff
@@ -0,0 +1,289 @@
...
+
+(defn- simplify-feature
+  "Simplify a feature. Returns simplfied feature and information about the simplificiton"
+  [feature tolerance mime-type]
+  (let [crs (when (.getDefaultGeometryProperty feature)
+              (-> feature .getDefaultGeometryProperty .getDescriptor .getCoordinateReferenceSystem))
+        feature-type (.getFeatureType feature)
+        feature-builder (SimpleFeatureBuilder. feature-type)
+        properties (.getProperties feature)
+        _ (doseq [p properties
```
chris-durbin, line 72: Move outside of the let binding.

system-int-test/test/cmr/system_int_test/search/collection_shapefile_simplification_search_test.clj
```diff
@@ -0,0 +1,180 @@
...
+(use-fixtures :each (ingest/reset-fixture {"provguid1" "PROV1"}))
+
+(defn- polygon
+  "Creates a single ring polygon with the given ordinates. Points must be in counter clockwise
+  order."
+  [& ords]
+  (poly/polygon [(apply umm-s/ords->ring ords)]))
+
+(def formats
+  "Shapfile formats to be tested"
```
chris-durbin, line 34: Shapefile typo

system-int-test/test/cmr/system_int_test/search/collection_shapefile_simplification_search_test.clj
```diff
@@ -0,0 +1,180 @@
...
+              (dc/collection
+               {:entry-title et
+                :spatial-coverage (dc/spatial {:gsr coord-sys
+                                               :sr coord-sys
+                                               :geometries shapes})}))))
+
+(deftest collection-shapefile-simplification-search-test
+  (let [_ (side/eval-form `(shapefile/set-enable-shapefile-parameter-flag! true))
+        saved-shapefile-max-size (shapefile-middleware/max-shapefile-size)
+        _ (side/eval-form `(shapefile-middleware/set-max-shapefile-size! 2000000))
```
chris-durbin, line 52: This number doesn't match the max size in the documentation. Not sure if that's intentional.

system-int-test/test/cmr/system_int_test/search/collection_shapefile_simplification_search_test.clj
```diff
@@ -0,0 +1,180 @@
...
+              "south_america_with_hole"
+              [49881 3492]
+              [wide-south-cart wide-south whole-world]
+
+              "Many Point Polygons Africa"
+              "africa"
+              [12861 2257]
+              [normal-line very-wide-cart wide-south-cart wide-south polygon-with-holes whole-world normal-brs normal-line-cart])))
+
+    ;; ESRI must be tested separately fromn other formats, because it is simplified down to slightly
```
chris-durbin, line 141: fromn and binry typos

chris-durbin, line 141: I assume you've verified that the shape looks correct, but just has slightly fewer points.

system-int-test/test/cmr/system_int_test/search/granule_shapefile_simplification_search_test.clj
```diff
@@ -0,0 +1,216 @@
...
+(def formats
+  "Shapfile formats to be tested"
+  ;; ESRI is handled separately
+  {"GeoJSON" {:extension "geojson" :mime-type mt/geojson}
+   "KML" {:extension "kml" :mime-type mt/kml}})
+
+(deftest granule-shapefile-simlpification-failure-cases
+  (side/eval-form `(shapefile/set-enable-shapefile-parameter-flag! true))
+  (let [saved-shapefile-max-size (shapefile-middleware/max-shapefile-size)
+        _ (side/eval-form `(shapefile-middleware/set-max-shapefile-size! 2500000))]
```
chris-durbin, line 41: Again this looks like a different max size from the documentation and from the collection test.

system-int-test/test/cmr/system_int_test/search/granule_shapefile_simplification_search_test.clj
```diff
@@ -0,0 +1,216 @@
...
+              "south_america_with_hole"
+              [49881 3492]
+              [wide-south-cart wide-south whole-world]
+
+              "Many Point Polygons Africa"
+              "africa"
+              [12861 2257]
+              [normal-line very-wide-cart wide-south-cart wide-south polygon-with-holes whole-world normal-brs normal-line-cart])))
+
+    ;; ESRI must be tested separately fromn other formats, because it is simplified down to slightly
```
chris-durbin, line 177: same fromn and binry typos

search-app/src/cmr/search/middleware/shapefile_simplification.clj
indiejames: Done.

system-int-test/test/cmr/system_int_test/search/collection_shapefile_simplification_search_test.clj
indiejames: Fixed.

system-int-test/test/cmr/system_int_test/search/collection_shapefile_simplification_search_test.clj
indiejames: Yes, it's intentional. The sample files I have are over the 1MB limit (which is enforced before simplification can happen).

system-int-test/test/cmr/system_int_test/search/collection_shapefile_simplification_search_test.clj
indiejames: Fixed.

indiejames: Correct.

system-int-test/test/cmr/system_int_test/search/granule_shapefile_simplification_search_test.clj
indiejames: This one is even bigger than the one used for collections (or below for the other granule tests) because it uses a really complicated test file that is > 2MB. Again, I'm not testing the max limit code here, so this gets around it so I can test the max features code with a realistic test file.

system-int-test/test/cmr/system_int_test/search/granule_shapefile_simplification_search_test.clj
indiejames: At least I'm consistent. Fixed.

abbottry:
I see a few mentions about counts (points, original points, etc) being returned. Can you summarize  the values returned? Are they in headers, or in the response? We're discussing API consistency right now, it'd make sense to ensure we're not creating more inconsistencies here.

indiejames:
> I see a few mentions about counts (points, original points, etc) being returned. Can you summarize the values returned? Are they in headers, or in the response? We're discussing API consistency right now, it'd make sense to ensure we're not creating more inconsistencies here.

The original point count and new point count values are returned in the `CMR-Shapfile-Simplification` header (only present when simplification is requested). For example:
```
CMR-Shapfile-Simplification: {"original-point-count":12861,"new-point-count":2252}
```
This header is listed in the API docs in the `Headers` section.

abbottry:
It looks like `shapefile` is misspelled through the PR. I was looking for the header but its `CMR-Shapfile-Simplification` instead of `CMR-Shapefile-Simplification`.

abbottry:
From a consumer perspective, I've never used an API that uses a JSON value as a response header. While the HTTP spec doesn't call it out as being illegal, it will result in a frustrating amount of steps to get that value. A consumer will need to include this custom header in `access-control-expose-headers`, call `JSON.parse` (assuming its a JSON string? please excuse my Clojure illiteracy) on the value, then do the standard checks on the value. Custom headers are a pain but if you're completely against returning these values in the actual response can the header be split into multiple, specific headers instead of an object?

abbottry:
Again, my Clojure knowledge isn't up to par with someone who can comment on functionality, but there is a noticeable difference in the validation methods for each type of shapefile, ESRI and GeoJSON are consistent, but KML has some different log and an additional error message starting [here](https://github.com/nasa/Common-Metadata-Repository/blob/a68c8280ad5dada99e339787a7d30ff789544619/search-app/src/cmr/search/services/parameters/converters/shapefile.clj#L257) -- Obviously different shapefiles have different contents, this seems like something outside of that though.

indiejames:
> From a consumer perspective, I've never used an API that uses a JSON value as a response header. While the HTTP spec doesn't call it out as being illegal, it will result in a frustrating amount of steps to get that value. A consumer will need to include this custom header in `access-control-expose-headers` (which in NGAP land requires a NASD ticket to allow) when making the request, include the response headers in methods that parse the response and then once its there, call `JSON.parse` (assuming its a JSON string? please excuse my Clojure illiteracy) on the value, then do the standard checks on the value. Custom headers are a pain but if you're completely against returning these values in the actual response can the header be split into multiple, specific headers instead of an object?

I will break these up into two separate headers if that is easier for you. The reason we are using headers is A) the simplification stats are just meta-info about the search process (like CMR-Took) and not really part of response data and  B) we would have to modify all the response formats (possibly breaking a lot of clients). With respect to `access-control-expose-headers`, is that not something we set on our side? I'm not very familiar with vagaries of CORS, so it's possbile there is something about it I don't understand.

abbottry:
Two headers would be preferable over a JSON header, for sure. So, NGAP blocks non standard headers with Lambda@Edge (this is clearly custom, so we have to file a NASD ticket to get that included in the permitted custom header list). As for CORS on your end, yes you need to update the permitted response headers (I was waiting on the decision about the headers to mention it) -- You'll need to update your configuration [here](https://github.com/nasa/Common-Metadata-Repository/blob/master/common-app-lib/src/cmr/common_app/api/routes.clj#L58) (I think that is the only place). But when a client makes that request we need to ensure that our clients expose those because they are non-standard.

Regarding meta information, there is conversation about that meta information and its inclusion in **some** response bodies, but not all. Collections and Granules return those headers; Services, Variables and Tools return hits as part of the response body _and_ in the headers. No real action being requested here but just pointing it out for consideration.

PR #991, opened 2020-06-22
CMR-6489: Add provider-id to ACL cache so that the permission given to the user can be provider dependent in the cache.

sxu123:
Currently, the cache key is [token permission] in the cache. So for a given user, if it has permission on one provider, this user could get hits in the cache for other providers that the user is not granted permission. Also, this can potentially make the user wait for a long time for the correct permission to take effect - as long as there are hits(has permission) in the cache, regardless which provider(s) they are for, the user can't be taken away the permission using the ACL for the right provider.

### DIFF
--- a/system-int-test/test/cmr/system_int_test/ingest/provider_ingest_permissions_test.clj
+++ b/system-int-test/test/cmr/system_int_test/ingest/provider_ingest_permissions_test.clj
@@ -24,8 +24,8 @@
                                                                   [:read :update]
                                                                   [:read :update])
                                     (subscription-util/grant-all-subscription-fixture {"provguid1" "PROV2"}
-                                                                  [:read :update]
-                                                                  [:read :update])
+                                                                  [:read]
+                                                                  [:read])
                                     (dev-system/freeze-resume-time-fixture)]))
 
 (defn- assert-ingest-succeeded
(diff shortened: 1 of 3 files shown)


system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
```diff
@@ -13,34 +13,36 @@
...
-                (subscription-util/grant-all-subscription-fixture {"provguid1" "PROV2"}
+               [(ingest/reset-fixture
+                  {"provguid1" "PROV1" "provguid2" "PROV2" "provguid3" "PROV3"})
+                (subscription-util/grant-all-subscription-fixture
+                  {"provguid1" "PROV1" "provguid2" "PROV2"}
+                  [:read :update]
+                  [:read :update])
+                (subscription-util/grant-all-subscription-fixture {"provguid1" "PROV3"}
                                                                   [:read]
                                                                   [:read :update])]))
```
sxu123, line 24: Need to test special cases on a provider that doesn't impact the other tests. There is one other test that's using PROV2 before, but it succeeded because it was making use of the permission in the cache for PROV1.

system-int-test/test/cmr/system_int_test/ingest/provider_ingest_permissions_test.clj
```diff
@@ -46,21 +46,12 @@
         prov-admin-update-group-concept-id (echo-util/get-or-create-group
                                             (s/context)
                                             "prov-admin-update-group")
-        prov-admin-update-group-concept-id-np (echo-util/get-or-create-group
```
sxu123, line 49: These -np were added so that we can have different user tokens. Now we the permission for a user is provider dependent in the cache, we don't need to distinguish the permissions through different user tokens.

jceaser:
I'm very green here, but I can actually read this a bit. Looks fine to me.

PR #996, opened 2020-06-30
CMR-6515: Fixed size 0 problem in granule counts search.

ygliuvt:


search-app/src/cmr/search/services/query_execution/granule_counts_results_feature.clj
```diff
@@ -114,8 +114,10 @@
 ;; This find granule counts per collection.
 (defmethod query-execution/post-process-query-result-feature :granule-counts
   [context query elastic-results query-results feature]
-  (->> query-results
-       (extract-granule-count-query query)
-       (query-execution/execute-query context)
-       search-results->granule-counts
-       (assoc query-results :granule-counts-map)))
+  (if (= 0 (:hits query-results))
```
jaybarra, line 117: prefer to use `zero?` instead of `(if (= 0 `

ygliuvt: `hits` is a whole number. Here we are comparing it with integer `0`, not with numeric value `0.0` and we also don't care about type check, so `(= 0 hits)` is more appropriate than `(zero? hits)`. That's why we have been using the `=` check in CMR when checking `hits` or `count`.

PR #998, opened 2020-07-01
CMR-6451: Autocomplete auth filtering

mschmele:
Update autocomplete indexing and search to not show private suggestions to users not logged in

### DIFF
--- a/indexer-app/src/cmr/indexer/services/index_service.clj
+++ b/indexer-app/src/cmr/indexer/services/index_service.clj
@@ -197,7 +197,7 @@
 
 (defn- science-keywords->elastic-docs
   "Convert hierarchical science-keywords to colon-separated elastic docs for indexing"
-  [science-keywords]
+  [science-keywords public-collection?]
   (let [keyword-hierarchy [:topic :term :variable-level-1
                            :variable-level-2 :variable-level-3 :detailed-variable]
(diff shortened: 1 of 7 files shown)


eereiter:
Does this also cover the case when users are logged in, but they don't have permission to view a record?  Or does this case already work?

mschmele:
> Does this also cover the case when users are logged in, but they don't have permission to view a record? Or does this case already work?

It doesn't, but this work is going to need to be reimplemented once the elasticsearch version upgrade is deployed. This is the easiest way to filter suggestions without affecting performance, which was a big concern

indexer-app/src/cmr/indexer/services/index_service.clj
```diff
@@ -288,7 +300,7 @@
                                        (debug %)
                                        nil)))
                              (remove nil?))
-        humanized-fields (map #(humanizer/collection-humanizers-elastic context %) parsed-concepts)
+        humanized-fields  (map #(humanizer/collection-humanizers-for-suggestion-docs context %) parsed-concepts)
```
jaybarra, line 303: nitpick: extra space added

mreese84:
Just looked at this with Mark. I agree with Leo. This feature should look at ACLs and not AccessConstraints. @mschmele, I recommend closing this PR and opening a new one when you get it reworked.

PR #999, opened 2020-07-02
CMR-6346

ygliuvt:
This PR enables source in ES index mappings and use `_source` to retrieve field values. All the stored fields in old index mappings are dropped except the two fields used by spatial plugin.


access-control-app/src/cmr/access_control/data/access_control_index.clj
```diff
@@ -25,27 +25,27 @@
 
 (defmapping ^:private group-mappings group-type-name
   "Defines the field mappings and type options for indexing groups in elasticsearch."
-  {:concept-id (m/stored m/string-field-mapping)
```
mschmele, line 28: Just curious: why get rid of `m/stored`?

access-control-app/src/cmr/access_control/data/acl_json_results_handler.clj
```diff
@@ -64,9 +64,7 @@
 
 (defmethod elastic-results/elastic-result->query-result-item [:acl :json]
   [context query elastic-result]
-  (let [item (->> elastic-result
-                  :fields
-                  (util/map-values first))
+  (let [item (:_source elastic-result)
```
mschmele, line 67: Nitpick: name this instance of `item` something like `source` or `field-values` like you have in the file below

ygliuvt: My changes were not related to naming of the variables either here or below. I will rename it to `source`.

access-control-app/src/cmr/access_control/data/access_control_index.clj
ygliuvt: By enabling `source`, all fields are stored in `_source` by default and we don't need to explicitly store any fields anymore.

ygliuvt:
> Code change looks good to me, just curious about the reasoning behind the decision to remove stored values, and the ticket doesn't have much information

When we first set up the index mappings in CMR, `_source` was not enabled by default and we have very few fields that need to be stored, so we explicitly make those fields that we need to retrieve values as `stored`. But as features are added, we now have lots of fields are marked as stored. This is not following the best practice of Elasticsearch index mappings. So here we are enabling `_source` in index mappings to have all fields stored in `_source` and drop the explicit stored index mappings.

PR #1000, opened 2020-07-02
CMR-6365 Update community-usage-metrics to also accept nil as a version value

jaybarra:
Fixed issue where 'nil' versions were not being captured correctly in usage metrics calculations.


indexer-app/src/cmr/indexer/data/concepts/collection/community_usage_metrics.clj
```diff
@@ -5,9 +5,16 @@
     [cmr.indexer.data.metrics-fetcher :as metrics-fetcher]))
 
 (def not-provided-version
-  "In EMS community usage CSV, the version value when the version is unknown"
+  "DEPRECATED
+   In EMS community usage CSV, the version value when the version is unknown"
   "N/A")
 
+(defn- valid-version?
```
ygliuvt, line 12: Rename this function and add a doc string. It is not checking if a version is valid or not. It is checking if the collection's version matches the community usage metrics' version.

indexer-app/src/cmr/indexer/data/concepts/collection/community_usage_metrics.clj
```diff
@@ -5,9 +5,16 @@
...
+  [parsed-version version]
+  (or (= (util/parse-version-id version) parsed-version)
```
ygliuvt, line 14: nit: switch the position of `parsed-version` and `(util/parse-version-id version)`.

system-int-test/test/cmr/system_int_test/search/collection_sorting_search_test.clj
```diff
@@ -492,6 +495,29 @@
 
     (index/wait-until-indexed)
 
+    (testing "reversing usage_score reverses sort-order"
```
ygliuvt, line 498: Why don't you use the existing functions to search and compare sorting orders? I like the old way where it lists the collections and the order explicitly. I might be missing something, but I don't see any benefit in doing this test in a different way from before.

indexer-app/src/cmr/indexer/data/concepts/collection/community_usage_metrics.clj
```diff
@@ -5,9 +5,18 @@
...
+  "DEPRECATED
+   In EMS community usage CSV, the version value when the version is unknown"
   "N/A")
 
+(defn- coll-version-matches-metrics-version?
+  "Returns whether the collection version matches the community usage metrics version.
+   This also returns true if the metrics version is nil or N/A"
+  [parsed-version version]
+  (or (= parsed-version (util/parse-version-id version))
+      (= not-provided-version version)
```
jceaser, line 17: what does it mean to be deprecated if you are still using it in new code? What is the project convention here with relation to line 8?

jaybarra: Marked as deprecated since it's not in the documentation for updating community usage metrics that that is what the value is supposed to be. But it's still in use by tests and possibly used by other parties. It may be removed entirely eventually.

PR #1003, opened 2020-07-14
CMR-6543 - updating documentation

jceaser:
Updated documentation for the transformation API to reflect current format version numbers and output.


ingest-app/docs/api.md
```diff
@@ -679,7 +679,11 @@ To disable validation of the parsed UMM metadata against the UMM spec, pass `ski
 Example: Translate ECHO10 metadata to UMM JSON
 
 ```
-curl -i -XPOST -H "Content-Type: application/echo10+xml" -H "Accept:  application/vnd.nasa.cmr.umm+json;version=1.2" %CMR-ENDPOINT%/translate/collection?skip_umm_validation=true -d \
+curl -i -XPOST \
+  -H "Content-Type: application/echo10+xml" \
+  -H "Accept: application/vnd.nasa.cmr.umm+json;version=1.14" \
```
jceaser, line 684: I question why we even specify a version number here, I can pull it out.

ingest-app/docs/api.md
```diff
@@ -768,7 +1026,11 @@ Granule metadata can be translated between metadata standards using the translat
 Example: Translate ECHO10 metadata to UMM-G
 
 ```
-curl -i -XPOST -H "Content-Type: application/echo10+xml" -H "Accept: application/vnd.nasa.cmr.umm+json;version=1.4" %CMR-ENDPOINT%/translate/granule -d \
+curl -i -XPOST \
+  -H "Content-Type: application/echo10+xml" \
+  -H "Accept: application/vnd.nasa.cmr.umm+json;version=1.5" \
```
jceaser, line 1031: I think I'm wrong, 1.6 is the latest, correct?

ingest-app/docs/api.md
ygliuvt: This demonstrates that user can specify a specific UMM version as the target output format. We should make that clearer in the api doc.

ingest-app/docs/api.md
ygliuvt: It can be any version that we support.

ingest-app/docs/api.md
```diff
@@ -60,7 +60,7 @@ Content-Type is a standard HTTP header that specifies the content type of the bo
 Note: UMM JSON accepts an additional version parameter for both Content-Type: and Accept: headers. Like charset, it is appended with a semicolon (;). If no version is appended, the latest version is assumed.
 
 ```
-application/vnd.nasa.cmr.umm+json;version=1.1
+application/vnd.nasa.cmr.umm+json;version=1.14
```
ygliuvt, line 63: This should say something like `For example: application/vnd.nasa.cmr.umm+json;version=1.14 means UMM JSON format and version 1.14.`

ingest-app/docs/api.md
```diff
@@ -60,7 +60,7 @@ Content-Type is a standard HTTP header that specifies the content type of the bo
 Note: UMM JSON accepts an additional version parameter for both Content-Type: and Accept: headers. Like charset, it is appended with a semicolon (;). If no version is appended, the latest version is assumed.
```
ygliuvt, line 60: Remove the extra `:`s from the sentence.

ingest-app/docs/api.md
```diff
@@ -698,18 +702,59 @@ Example output:
 
 ```
 {
```
ygliuvt, line 704: I don't feel strong about this. Just my 2 cents: It is probably not necessary to provide the exact input and output here. It is too excessive and overwhelms the real useful info in the API doc. We may want to just drop the full input and output from the API doc.

ingest-app/docs/api.md
```diff
@@ -738,28 +782,242 @@ Example output:
     xmlns:gss="http://www.isotc211.org/2005/gss"
     xmlns:gts="http://www.isotc211.org/2005/gts"
     xmlns:srv="http://www.isotc211.org/2005/srv"
-    xmlns:swe="http://schemas.opengis.net/sweCommon/2.0/"
     xmlns:xlink="http://www.w3.org/1999/xlink"
-    xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+    xmlns:xs="http://www.w3.org/2001/XMLSchema"
```
ygliuvt, line 786: Same comment as above. Is it really necessary to put the full ISO metadata in here?

jceaser: I don't think so, I would be fine with cutting out much of this and replacing it with '…'

ingest-app/docs/api.md
jceaser: I'm going to move it to 1.6, the idea is that we don't want to encourage users to run older versions unless they need to.

ingest-app/docs/api.md
jceaser: for consistency I think it is needed, however the ISO example is a bit long. What if we did another format that is not as verbose? The other idea I had was to add vertical scroll bars to the rendering CSS

PR #1004, opened 2020-07-14
CMR-6441 Prevent autocomplete indexing unless the correct index exists in Elasticsearch

jaybarra:
Prevents autocomplete-reindexing from occurring unless the index has been created in Elasticsearch. If indexing occurs without the correct mapping autocomplete-queries will fail.


elastic-utils-lib/src/cmr/elastic_utils/index_util.clj
```diff
@@ -213,3 +213,10 @@
   "Creates the alias for the index."
   [conn index alias]
   (esi/update-aliases conn [{:add {:index index :alias alias}}]))
+
+(defn index-exists?
+  "Returns whether an index with a given name exists in Elasticsearch."
+  [conn index-name]
+  (let [stats (esi/stats conn index-name)]
+    (not= 404 (:status stats))))
```
jceaser, line 221: won't 500 pass this test?

jaybarra: Updated

indexer-app/src/cmr/indexer/data/elasticsearch.clj
```diff
@@ -324,11 +323,13 @@
 (defn bulk-index-autocomplete-suggestions
   "Save a batch of suggestion documents in Elasticsearch."
   [context docs]
-  (doseq [docs-batch (partition-all MAX_BULK_OPERATIONS_PER_REQUEST docs)]
-    (let [bulk-operations (cmr-bulk/create-bulk-index-operations docs-batch)
-          conn (context->conn context)
-          response (bulk/bulk conn bulk-operations)]
-      (handle-bulk-index-response response))))
+  (if (esi/index-exists? (context->conn context) "1_autocomplete")
```
ygliuvt, line 326: I don't think this check is necessary. The `1_autocomplete` index should be created as part of the `db-migration` like all the other indices used by CMR. If `1_autocomplete` index is not automatically initialized for any reason, we should fix that in db-migration. It is wrong to check if the index exists in every single bulk index request.

elastic-utils-lib/src/cmr/elastic_utils/index_util.clj
```diff
@@ -213,3 +214,16 @@
...
+             :opt-un [::_shards ::_all ::error ::status]))
+
+(defn index-exists?
+  "Returns whether an index with a given name exists in Elasticsearch."
+  [conn index-name]
+  (let [stats (esi/stats conn index-name)]
+    (and (spec/valid? ::elasticsearch-index-stats stats)
+         ;; status 404 is only given when not found, status is not normally returned
+         (not= 404 (:status stats)))))
+
```
sxu123, line 229: Just curious what should happen when the index exists but it contains error?  In which case, do you or do you not want to do autocomplete indexing?

jaybarra: If the index exists but there is an error, I would rather Elasticsearch itself fail during an ingest and have that error be available in the logs.

jaybarra:
Closing as it cannot be reproduced and is not an issue in any currently deployed environment.

PR #1006, opened 2020-07-14
CMR-5370 - updating jar files to resolve a CVE

jceaser:
Tests were run locally, both applications were built without errors, not sure what else to check to ensure it is working.


jaybarra:
I do question why only those two services are using that specific library and not the rest.

PR #1007, opened 2020-07-14
CMR-6332: Added index removal of subscription and tool at provider deletion

sxu123:


indexer-app/src/cmr/indexer/services/index_service.clj
```diff
@@ -761,7 +761,23 @@
        context
        index
        (concept-mapping-types :service)
-       {:term {(query-field->elastic-field :provider-id :service) provider-id}}))))
+       {:term {(query-field->elastic-field :provider-id :service) provider-id}}))
+
+    ;; delete the tools
```
ygliuvt, line 766: nit: The calls for `:variable`, `:service`, `:tool` and `:subscription` concept types can be put into a `doseq` of itself. This way, we don't have to repeat the code for doing the `delete-by-query` calls four times.

jceaser: I'm not on the review, but I agree (which means I can actually read the code enough to have an opinion).

sxu123: Done.

PR #1010, opened 2020-07-17
CMR-6561: Fixed CMR build issue caused by Opendata.

ygliuvt:


ygliuvt:
> Do we have the right schema already in the project or do we just not need it?

We have opendata schemas locally in CMR. There is no need to retrieve it from opendata site.

PR #1023, opened 2020-08-12
CMR-6434: Add OR option for spatial parameters

mschmele:


system-int-test/test/cmr/system_int_test/search/collection_spatial_search_test.clj
```diff
@@ -464,7 +491,37 @@
         :point [17.73 2.21] [whole-world normal-brs]
         :line [-0.37,-14.07,4.75,1.27,25.13,-15.51]
         [whole-world polygon-with-holes polygon-with-holes-cart normal-line-cart normal-line
-         normal-poly-cart]))))
+         normal-poly-cart])
+
+     (testing "ORed spatial search"
```
indiejames, line 496: I'm confused by this test. It looks like you are only sending one geometry per request. Shouldn't you send more than one with an 'OR'?

mschmele: Yeah those changes didn't actually make it into the commit 🤦 
I'm adding documentation now and will push the fix up with that

ygliuvt:
I will review when the real changes are in.

system-int-test/test/cmr/system_int_test/search/collection_spatial_search_test.clj
```diff
@@ -441,6 +441,33 @@
...
+            (println "Actual:" (->> found :refs (map :name) sort pr-str)))
+          matches?)
+
+        ["0,89,100" "0,89,1000000"] [along-am-line
+                                     north-pole
+                                     on-np
+                                     touches-np
+                                     very-tall-cart
+                                     whole-world]
+        ["0,0,1000" "0,89,1000" "0,89,1000000"] [along-am-line
```
ygliuvt, line 463: The matched granules should have included `north-pole`.

system-int-test/test/cmr/system_int_test/search/collection_spatial_search_test.clj
```diff
@@ -441,6 +441,33 @@
...
+                                                 north-pole on-np
```
ygliuvt, line 464: `north-pole` and `on-np` should not be on the same line.

search-app/docs/api.md
```diff
@@ -1686,33 +1686,46 @@ Find collections matching service concept id.
 ##### <a name="c-polygon"></a> Polygon
 
 Polygon points are provided in counter-clockwise order. The last point should match the first point to close the polygon. The values are listed comma separated in longitude latitude order, i.e. lon1, lat1, lon2, lat2, lon3, lat3, and so on.
+This parameter supports the and/or option as shown below.
```
ygliuvt, line 1689: This is not what I had in mind when I filed CMR-6434. From the description of CMR-6434:
```
For now, we will just add a single option `options[spatial][or]=true` to apply the OR option across all spatial shapes. The current idea is to only apply this option on explicit shapes, such as point, bounding_box, polygon, circle, etc, but not to shape file or geojsons.
```
The idea is that any spatial condition within search will be ORed together regardless of the specific spatial types. I have thought that would be more useful and the behavior more consistent. 

with this implementation, only specific spatial types can be ORed, but between spatial types, it is still ANDed together and can't be changed.

search-app/docs/api.md
```diff
@@ -1686,33 +1686,46 @@ Find collections matching service concept id.
...
 
     curl "%CMR-ENDPOINT%/collections?polygon=10,10,30,10,30,20,10,20,10,10"
 
+    curl "%CMR-ENDPOINT%/collections?polygon=10,10,30,10,30,20,10,20,10,10&polygon=11,11,31,11,31,21,11,21,11,11&options[polygon][or]=true"
```
macrouch, line 1693: In all of these examples I think you need `[]` after the spatial param. 

curl "%CMR-ENDPOINT%/collections?polygon[]=10,10,30,10,30,20,10,20,10,10&polygon[]=11,11,31,11,31,21,11,21,11,11&options[polygon][or]=true"

common-app-lib/src/cmr/common_app/services/search/params.clj
```diff
@@ -215,6 +216,55 @@
...
+                           :and)
+        user-spatial-params (when-not (empty? params)
+                              (select-keys params valid-spatial-params))
+        user-non-spatial-params (-> params
+                                    (data/diff user-spatial-params)
+                                    first)
+        conditions (generate-conditions context concept-type params options)]
+    (if (and (= :or spatial-option)
+             (not (empty? user-spatial-params))
+             (= concept-type :collection))
```
ygliuvt, line 258: Why only collections?

common-app-lib/src/cmr/common_app/services/search/params.clj
```diff
@@ -215,6 +216,55 @@
...
+  (map (fn [[param value]]
+           (parameter->condition context concept-type param value options))
+        params))
+
+(defn- create-combined-conditions
+  "When valid spatial parameters are given and the spatial OR option is passed,
+   create proper search conditions"
+  [context concept-type options spatial-params non-spatial-params]
+  (let [spatial-conditions (when-not (empty? spatial-params)
+                       (generate-conditions
```
jaybarra, line 231: nitpick: spacing

common-app-lib/src/cmr/common_app/services/search/params.clj
```diff
@@ -215,6 +216,55 @@
...
+        non-spatial-conditions (when-not (empty? non-spatial-params)
+                                 (generate-conditions
+                                  context concept-type non-spatial-params options))
+        combined-non-spatial-conditions (when non-spatial-conditions
+                                          (gc/and-conds non-spatial-conditions))]
+    (if (and combined-spatial-conditions combined-non-spatial-conditions)
+        (gc/and-conds (conj combined-spatial-conditions combined-non-spatial-conditions))
+        (conj combined-spatial-conditions combined-non-spatial-conditions))))
+
+(defn- combine-query-conditions
```
ygliuvt, line 243: This function needs a better name. It is not combining query conditions, but rather converting params into  query conditions.

common-app-lib/src/cmr/common_app/services/search/params.clj
jaybarra, line 258: Swap keyword and concept-type positions

common-app-lib/src/cmr/common_app/services/search/params.clj
```diff
@@ -215,6 +216,55 @@
...
+  (let [valid-spatial-params [:bounding-box :circle :point :line :polygon :bounding-box]
+        spatial-option (if (= (get-in options [:spatial :or]) "true")
+                           :or
+                           :and)
+        user-spatial-params (when-not (empty? params)
+                              (select-keys params valid-spatial-params))
+        user-non-spatial-params (-> params
+                                    (data/diff user-spatial-params)
+                                    first)
+        conditions (generate-conditions context concept-type params options)]
```
ygliuvt, line 255: It is a waste to build the conditions here as when spatial options is `OR`, we don't use the conditions that is build here.

system-int-test/test/cmr/system_int_test/search/collection_spatial_search_test.clj
```diff
@@ -441,6 +433,23 @@
...
 
+    (testing "valid circle searches with ORed results"
+      (are [lon-lat-radius items]
+        (let [found (search/find-refs
+                     :collection
+                     {:circle lon-lat-radius
+                      :page-size 50
+                      "options[spatial][or]" "true"})
+              matches? (d/refs-match? items found)]
+          (when-not matches?
```
jaybarra, line 444: Are the `println` invoked if the assertion fails?

common-app-lib/src/cmr/common_app/services/search/params.clj
```diff
@@ -215,6 +216,55 @@
...
+  [context concept-type params options]
+  (map (fn [[param value]]
+           (parameter->condition context concept-type param value options))
+        params))
+
+(defn- create-combined-conditions
+  "When valid spatial parameters are given and the spatial OR option is passed,
+   create proper search conditions"
+  [context concept-type options spatial-params non-spatial-params]
+  (let [spatial-conditions (when-not (empty? spatial-params)
```
ygliuvt, line 230: use `seq` rather than not empty for the check.

common-app-lib/src/cmr/common_app/services/search/params.clj
```diff
@@ -215,6 +216,55 @@
   [concept-type params]
   (default-parse-query-level-params concept-type params))
 
+(defn- generate-conditions
+  "Converts given parameters into elasticsearch conditions"
```
ygliuvt, line 220: Change `elasticsearch` to `query`.

common-app-lib/src/cmr/common_app/services/search/params.clj
```diff
@@ -215,6 +216,55 @@
...
+                           :or
+                           :and)
+        user-spatial-params (when-not (empty? params)
+                              (select-keys params valid-spatial-params))
+        user-non-spatial-params (-> params
+                                    (data/diff user-spatial-params)
+                                    first)
+        conditions (generate-conditions context concept-type params options)]
+    (if (and (= :or spatial-option)
+             (not (empty? user-spatial-params))
```
ygliuvt, line 257: use seq rather than not empty for the check.

search-app/docs/api.md
```diff
@@ -1686,33 +1687,46 @@ Find collections matching service concept id.
...
+
 ##### <a name="c-bounding-box"></a> Bounding Box
 
 Bounding boxes define an area on the earth aligned with longitude and latitude. The Bounding box parameters must be 4 comma-separated numbers: lower left longitude, lower left latitude, upper right longitude, upper right latitude.
+This parameter supports the and/or option as shown below.
 
     curl "%CMR-ENDPOINT%/collections?bounding_box=-10,-5,10,5
 
+    curl "%CMR-ENDPOINT%/collections?bounding_box=-10,-5,10,5&bounding_box=-11,-6,11,6&options[bounding_box][or]=true
```
ygliuvt, line 1703: With the addition of spatial options, should we drop the individual shape options now? I don't feel strongly against this, but concerned that we are offering too many options that are probably too confusing.

mschmele: I figured I would just leave them in if for some reason a user wanted to fine-tune their search more, and the work is already done

common-app-lib/src/cmr/common_app/services/search/params.clj
mschmele: Because I forgot to include granules - fixed

system-int-test/test/cmr/system_int_test/search/collection_spatial_search_test.clj
mschmele: Yes, it's just the test output that gives the expected vs actual

common-app-lib/src/cmr/common_app/services/search/params.clj
```diff
@@ -215,6 +216,58 @@
...
+        (conj combined-spatial-conditions combined-non-spatial-conditions))))
+
+(defn- params->query-conditions
+  "Appropriately combine query operators based on given options"
+  [context concept-type options params query-attribs]
+  (let [valid-spatial-params [:bounding-box :circle :point :line :polygon :bounding-box]
+        spatial-option (if (= (get-in options [:spatial :or]) "true")
+                           :or
+                           :and)
+        spatial-params (when (seq params)
```
ygliuvt, line 252: no need of the `when`.

ygliuvt: just do `(select-keys params valid-spatial-params))` should be OK.

common-app-lib/src/cmr/common_app/services/search/params.clj
```diff
@@ -215,6 +216,58 @@
...
+                         (select-keys params valid-spatial-params))]
+    (if (and (= :or spatial-option)
+             (seq spatial-params)
+             (or (= :collection concept-type)
```
ygliuvt, line 256: Why do we limit to just collections and granules? We can remove this check and support it for all concept types. Even though only collections and granules have spatial parameters, we don't have to have this check.

system-int-test/test/cmr/system_int_test/search/collection_spatial_search_test.clj
```diff
@@ -464,7 +473,31 @@
         :point [17.73 2.21] [whole-world normal-brs]
         :line [-0.37,-14.07,4.75,1.27,25.13,-15.51]
         [whole-world polygon-with-holes polygon-with-holes-cart normal-line-cart normal-line
-         normal-poly-cart]))))
+         normal-poly-cart]))
+
+    (testing "ORed spatial search"
```
ygliuvt, line 478: For completeness, we should add test cases with `options[spatial][or]=false`.

system-int-test/test/cmr/system_int_test/search/collection_spatial_search_test.clj
```diff
@@ -464,7 +473,31 @@
...
+                                               "23.59,-15.47,25.56,-4"]
+                                              "options[spatial][or]" "true"})
+             bbox-items [across-am-poly very-wide-cart am-point along-am-line normal-line-cart whole-world across-am-br]
+             combined-refs (search/find-refs
+                            :collection
+                            {:circle "179.8,41,100000"
+                             :bounding-box "166.11,-19.14,-166.52,53.04"
+                             "options[spatial][or]" "true"})
+             combined-items [across-am-poly along-am-line whole-world across-am-br am-point very-wide-cart]]
+        (is (d/refs-match? combined-items combined-refs))
```
ygliuvt, line 498: nit: it's better to have the expected items explicitly listed in the verification statements rather than have it far away in the `let` block.

common-app-lib/src/cmr/common_app/services/search/params.clj
mschmele: I was getting errors and added this to fix the issue, but some of the other code I added seems to have covered that case as well so I'll remove it

ygliuvt:
We should add the granule search tests with spatial OR parameter.

PR #1024, opened 2020-08-16
CMR-6597 Added feature to ingest association at variable ingest.

sxu123:
This PR is Part1 of several parts of the feature: Variable can not exist without association.

It covers the part that ingest the association at variable ingest time.  No update of the association, and no cascading delete of variables when associations are deleted.  

1. Copied/modified variable association related functions from search-app to metadata-db-app.
This is because we will retire the existing variable association endpoint.

2. Added :search-index component in metadata-db-app to allow execute-query when doing association validation.

3. Added perform-post-commit-association in try-to-save function. so that if the association fails, it'll rollback
the variable insert.

4. Modified the variable response to include variable-association and association-item, on top of concept-id and revision-id.

5. Added cascading collection deletion to associated variables.

6. Removed fingerprint check which is obsolete. The new uniqueness (variable Name + collection) is guaranteed by the code that won't allow multiple variables with the same Name to be associated with the same collection, at association creation time, which is variable's ingest time.

Remaining work in separate tickets:
1. Disallow variable creation using current variable ingest endpoint, can only use it to update.
2. Disable association and dissociation endpoint. Will be done after the invalid data are fixed.
3. Association update. This PR doesn't provide a way to update an association through the new endpoints. Once we finalize the approach, will implement in a separate ticket.

### DIFF
--- a/ingest-app/src/cmr/ingest/api/core.clj
+++ b/ingest-app/src/cmr/ingest/api/core.clj
@@ -45,7 +45,7 @@
    mt/json :json})
 
 (defn- create-nested-elements
-  "Create xml element for the nested maps."
+  "Create nested elements for the nested maps."
   [m]
   (reduce-kv (fn [memo k v]
                  (conj memo (xml/element (keyword k) {} (if (map? v) 
--- a/ingest-app/src/cmr/ingest/api/core.clj
+++ b/ingest-app/src/cmr/ingest/api/core.clj
@@ -44,6 +44,16 @@
    mt/xml :xml
    mt/json :json})
 
+(defn- create-nested-elements
+  "Create xml element for the nested maps."
+  [m]
+  (reduce-kv (fn [memo k v]
+                 (conj memo (xml/element (keyword k) {} (if (map? v) 
+                                                          (create-nested-elements v)
+                                                          v))))
+             []
+             m))
+
 (defn- result-map->xml
   "Converts all keys in a map to tags with values given by the map values to form a trivial
   xml document.
@@ -60,10 +70,7 @@
    (xml/element
     :result
     {}
-    (reduce-kv (fn [memo k v]
-                 (conj memo (xml/element (keyword k) {} (if (map? v) (str v) v))))
-               []
-               m))))
+    (create-nested-elements m))))
 
 (defn get-ingest-result-format
   "Returns the requested ingest result format parsed from the Accept header or :xml
--- a/ingest-app/src/cmr/ingest/api/core.clj
+++ b/ingest-app/src/cmr/ingest/api/core.clj
@@ -61,7 +61,7 @@
     :result
     {}
     (reduce-kv (fn [memo k v]
-                 (conj memo (xml/element (keyword k) {} v)))
+                 (conj memo (xml/element (keyword k) {} (if (map? v) (str v) v))))
                []
                m))))
 
--- a/ingest-app/src/cmr/ingest/api/variables.clj
+++ b/ingest-app/src/cmr/ingest/api/variables.clj
@@ -27,29 +27,41 @@
 
 (defn ingest-variable
   "Processes a request to create or update a variable."
-  [provider-id native-id request]
-  (let [{:keys [body content-type headers request-context]} request
-        concept (api-core/body->concept!
-                 :variable provider-id native-id body content-type headers)]
-    (lt-validation/validate-launchpad-token request-context)
-    (api-core/verify-provider-exists request-context provider-id)
-    (acl/verify-ingest-management-permission
-      request-context :update :provider-object provider-id)
-    (common-enabled/validate-write-enabled request-context "ingest")
-    (let [concept (validate-and-prepare-variable-concept concept)
-          {concept-format :format metadata :metadata} concept
-          variable (spec/parse-metadata request-context :variable concept-format metadata)
-          _ (v/umm-spec-validate-variable
-             variable request-context (not (ingest-config/validate-umm-var-keywords)))
-          concept-with-user-id (api-core/set-user-id concept request-context headers)
-          ;; Log the ingest attempt
-          _ (info (format "Ingesting variable %s from client %s"
-                          (api-core/concept->loggable-string concept-with-user-id)
-                          (:client-id request-context)))
-          save-variable-result (ingest/save-variable request-context concept-with-user-id)]
-      ;; Log the successful ingest, with the metadata size in bytes.
-      (api-core/log-concept-with-metadata-size concept-with-user-id request-context)
-      (api-core/generate-ingest-response headers save-variable-result))))
+  ([provider-id native-id request]
+   (ingest-variable provider-id native-id request nil nil))
+  ([provider-id native-id request coll-concept-id]
+   (ingest-variable provider-id native-id request coll-concept-id nil))
+  ([provider-id native-id request coll-concept-id coll-revision-id]
+   (let [provider-id (if coll-concept-id
+                       (last (string/split coll-concept-id #"C[0-9]+-"))
+                       provider-id)
+         {:keys [body content-type headers request-context]} request
+         concept (api-core/body->concept!
+                  :variable provider-id native-id body content-type headers)]
+
+     (lt-validation/validate-launchpad-token request-context)
+     (api-core/verify-provider-exists request-context provider-id)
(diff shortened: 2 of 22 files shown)


ingest-app/src/cmr/ingest/api/core.clj
```diff
@@ -61,7 +61,7 @@
     :result
     {}
     (reduce-kv (fn [memo k v]
-                 (conj memo (xml/element (keyword k) {} v)))
+                 (conj memo (xml/element (keyword k) {} (if (map? v) (str v) v))))
```
jceaser, line 64: I want to ask why this is a needed change? Will this result in json like text in the XML? This is a private function, so was this needed to fix a bug?

Maybe a better example would help?

metadata-db-app/src/cmr/metadata_db/services/concept_constraints.clj
```diff
@@ -160,6 +160,7 @@
    :service [(partial pfn-constraint :service-name)]
    :tool [(partial pfn-constraint :tool-name)]
    :subscription [(partial pfn-constraint :subscription-name)]
+   ;; will remove this fingerprint check when updating tests later.
```
jceaser, line 163: is there a standard or practice to use "TODO" or something like that to make it easy to find these latter with grep?

ingest-app/src/cmr/ingest/api/core.clj
sxu123: This change is needed because now variable ingest returns not only concept-id and revision-id, it also returns variable association and associated-item, both of them are map. There is no requirement on how they need to be returned, so I'm just returning them as strings.
<?xml version="1.0" encoding="UTF-8"?><result><concept-id>V1200000007-PROV1</concept-id><revision-id>1</revision-id><variable-association>{:concept-id "VA1200000008-CMR", :revision-id 1}</variable-association><associated-item>{:concept-id "C1200000005-PROV1"}</associated-item></result>

metadata-db-app/src/cmr/metadata_db/services/concept_constraints.clj
sxu123: No. just decided not to do it for this ticket since it might break some tests.

ingest-app/src/cmr/ingest/api/core.clj
william-valencia: Looks like what is returned is a mix of XML and JSON. If I request JSON to be returned, will I get all properly formatted JSON returned?

sxu123: {"concept-id":"V1200000012-PROV1","revision-id":1,"variable-association":{"concept-id":"VA1200000007-CMR","revision-id":1},"associated-item":{"concept-id":"C1200000005-PROV1","revision-id":1}}

jceaser: I think the mixed content is a problem, don't we have a method to convert it to some generic xml, like <key>value</key>?

sxu123: Sure. Now that I got the main functionality working, I can go back to take care of that. Thanks.

eereiter: Or you can flatten it out as son: 
{"concept-id":"V1200000012-PROV1","revision-id":1,"variable-association-concept-id":"VA1200000007-CMR","variable-association-revision-id":1,"associated-item-concept-id":"C1200000005-PROV1","associated-itemrevision-id":1}
as XML:
<?xml version="1.0" encoding="UTF-8"?><result><concept-id>V1200000007-PROV1</concept-id><revision-id>1</revision-id><variable-association-concept-id>VA1200000008-CMR</<variable-association-concept-id><variable-association-revision-id>1</variable-association-revision-id><associated-item-concept-id>C1200000005-PROV1</associated-item-concept-id></result>

sxu123: <?xml version="1.0" encoding="UTF-8"?><result>
    <concept-id>V1200000006-PROV1</concept-id>
    <revision-id>2</revision-id>
    <variable-association>
        <concept-id>VA1200000007-CMR</concept-id>
        <revision-id>2</revision-id>
    </variable-association>
    <associated-item>
        <concept-id>C1200000005-PROV1</concept-id>
        <revision-id>1</revision-id>
    </associated-item>
</result>

metadata-db-app/src/cmr/metadata_db/services/concept_service.clj
```diff
@@ -334,7 +516,20 @@
           provider
           concept
           rollback-fn)
-        concept)
+
+        ;; Perform post commit variable association
+        (if (and (= :variable (:concept-type concept))
+                 (:coll-concept-id concept)
```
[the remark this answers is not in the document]
ygliuvt: Regardless how you get here, when the variable concept is to be inserted in Oracle, the concept must have the collection concept id, otherwise it would be invalid.

system-int-test/test/cmr/system_int_test/search/variable/variable_association_test.clj
```diff
@@ -85,18 +85,17 @@
             _ (index/wait-until-indexed)
             response (association-util/associate-by-concept-ids
                       token concept-id [{:concept-id c1-p1}])]
-        (vu/assert-variable-association-bad-request
-         {[c1-p1] {:errors [(format "Collection [%s] does not exist or is not visible." c1-p1)]}}
-         response)))
-
+        ;; when the collection was deleted, the variable got deleted too
```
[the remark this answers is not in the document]
ygliuvt: Please update variable_ingest_with_association_test.clj to make the collection delete test stands out with a `testing` clause rather than a comment (that is why I missed it). Then delete the testing here as it is a duplication.

system-int-test/test/cmr/system_int_test/search/variable/variable_association_test.clj
```diff
@@ -85,18 +85,17 @@
...
+ 
     (testing "ACLs are applied to collections found"
       ;; None of PROV3's collections are visible
       (let [response (association-util/associate-by-concept-ids token
                                                                 concept-id
                                                                 [{:concept-id c4-p3}])]
-        (vu/assert-variable-association-bad-request
-         {[c4-p3] {:errors [(format "Collection [%s] does not exist or is not visible." c4-p3)]}}
-         response)))))
+        (is (= 404 (:status response)))
```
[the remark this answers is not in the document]
ygliuvt: This is getting too much for me to follow as it is about future work that is not in this PR. I will just believe you that you are going to clean it up. :)

metadata-db-app/src/cmr/metadata_db/services/concept_service.clj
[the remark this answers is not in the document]
ygliuvt: OK. got it. You are saying right now there is a need for this check because we have not removed the old variable update endpoint. Once we remove the old variable update endpoint, we would be able to remove this check.

system-int-test/test/cmr/system_int_test/search/variable/variable_association_test.clj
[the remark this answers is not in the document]
sxu123: Done

metadata-db-app/src/cmr/metadata_db/services/concept_service.clj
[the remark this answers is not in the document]
sxu123: Old variable endpoint will not go away, as a mean of updating the variable alone, without updating the association. so this check will stay.

ingest-app/src/cmr/ingest/api/variables.clj
```diff
@@ -27,29 +26,47 @@
...
+  ([provider-id native-id request coll-concept-id coll-revision-id]
+   (let [provider-id (if coll-concept-id
+                       (last (string/split coll-concept-id #"C[0-9]+-"))
+                       provider-id)
+         {:keys [body content-type headers request-context]} request
+         concept (api-core/body->concept!
+                  :variable provider-id native-id body content-type headers)]
+
+     ;; Ensures the associated collection is visible and not deleted.
+     (v/validate-variable-associated-collection
```
ygliuvt, line 40: Move this validation down to line 50.

PR #1027, opened 2020-08-19
Feature cmr-6612 - adding debug line

jceaser:
To debug a formatting issue with the subscription email, I need to add the subscription object to logs.


jceaser:
we solved this another way

PR #1030, opened 2020-08-21
Cmr 6619 - UMMS-v1.3.4 

eereiter:
Remove ServiceOptions/SupportedInput/OutputFormats, 
Add Harmony as a new Service Type Enum, 
Added a new Supported Format - Shapefile+zip
changed SubsetTypes from enumerations to a set of objects that provide more information called Subset.


umm-spec-lib/src/cmr/umm_spec/migration/service_service_options.clj
```diff
@@ -2,7 +2,8 @@
   "Contains functions for migrating between versions of the UMM Service schema
    for service options."
   (:require
-   [clojure.set :as set]))
+   [clojure.set :as set]
+   [cmr.common.util :refer [remove-nil-keys remove-empty-maps]]))
```
mschmele, line 6: Nitpick: is there a reason you're using `refer` here?
```suggestion
   [cmr.common.util :as util]))
```

umm-spec-lib/src/cmr/umm_spec/migration/service_service_options.clj
```diff
@@ -228,3 +227,122 @@
...
+  [input-format coll]
+  (first
+    (keep-indexed #(when (= (:SupportedInputFormat %2) input-format) %1) coll)))
+
+(defn move-supported-formats-to-reformattings-for-1_3_4
+  "Take the passed in input-formats and output-formats and convert the data
+   to the passed in supported-reformattings. Any duplicate data in the
+   output-formats sub element is omitted."
+  [supported-reformattings input-formats output-formats]
+  (let [join-formats (fn [sup-ref input-format]
```
mschmele, line 296: This should be broken into its own function

umm-spec-lib/src/cmr/umm_spec/migration/service_service_options.clj
```diff
@@ -228,3 +227,122 @@
...
+            (assoc (count supported-format-vector) "Shapefile")
+            (distinct))))
+    supported-format-vector))
+
+(defn- remove-reformattings-when-input-not-valid-1_3_4-to-1_3_3
+  "Get the replaced vector values and place them in the proper place in the
+   supported reformatting map."
+  [reformatting]
+  (let [input (replace-non-valid-formats-1_3_4-to-1_3_3 (vector (:SupportedInputFormat reformatting)))]
+    (when input
```
mschmele, line 340: Could you use `when-let` here?

umm-spec-lib/src/cmr/umm_spec/migration/service_service_options.clj
eereiter: Since these are the only 2 functions that are being used, I didn't think that I needed to load or get the entire library. But I can just use as.

umm-spec-lib/src/cmr/umm_spec/migration/service_service_options.clj
eereiter: So I don't know that I can.  Reduce only allows for 2 functions to be passed in, but I need all three variables that are used in the function.

umm-spec-lib/src/cmr/umm_spec/migration/service_service_options.clj
eereiter: yes, I have updated the code.

umm-spec-lib/src/cmr/umm_spec/migration/service_service_options.clj
eereiter: Based on our slack conversation, I will leave this as is.

PR #1031, opened 2020-08-25
CMR-6629 As a client application, I want to pass an Earthdata Login token without an app id because Earthdata Login may not need it and I may not have it

daniel-zamora:
Aswell as Echo-Token=<token>:<app-id>, Authorization: Bearer <access_token>, which EDL supports.

Better, at the discretion of CMR, would be implementing CMR-6630. Precise details are in work in #analysis-authn-authz in EOSDIS Slack, but the gist is that you could use a different, more standard header for tokens without an app ID to be consistent with the rest of the enterprise and possibly ease implementation / compatibility. It seems likely that token would be "Authorization=bearer <token>"

Example of how apps will pass tokens 

Both giver and receiver applications will be in the same ECHO EDL group.

HTTP request with header:

### DIFF
--- a/search-app/docs/api.md
+++ b/search-app/docs/api.md
@@ -212,7 +212,7 @@ The Maximum URL Length supported by CMR is indirectly controlled by the Request 
 
 #### <a name="cors-header-support"></a> CORS Header support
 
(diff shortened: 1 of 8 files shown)


search-app/docs/api.md
```diff
@@ -278,6 +278,8 @@ These are query parameters that control what extra data is included with collect
   * Accept - specifies the MimeType to return search results in. Default is "application/xml".
     * `curl -H "Accept: application/xml" -i "%CMR-ENDPOINT%/collections"`
   * `Echo-Token` - specifies an ECHO token to use to authenticate yourself.
+  * `Authorization: Bearer` - specifies an URS Authorization Bearer token to use to authenticate yourself.
```
ygliuvt, line 281: Change `URS` to `EDL`.

PR #1034, opened 2020-08-27
CMR-6631 - Adding New Table for Subscription Notification Time

jceaser:
Adding a new column to the subscription table for storing last notification date-time. This column should auto populate with the creation time so that subscriptions start from the second a subscription is created. This value will then change over time as the user has been sent notifications.

This column is purposely mutable.


ygliuvt:
The new discussion on future subscription functionality got me thinking:
Should we add a new notification table for the `last_notification_at`?

The difficulty in having a new table is that:
1) We have to populate the new notification table when subscription is created
2) We have to update the notification table when subscription is updated or deleted

but we might want to trigger notification on different intervals for different subscriptions; and we might want to suspend and resume notification on subscriptions in the future. 

The benefit of having a new notification table is that:
1) we keep the immutability of subscription concept and we are consistent; 
2) we have separation between subscription and notification. 
3) we only have one revision of the subscription and it is more performant and less complicated in updating the notification.
4) we are future proof. 
I would like to know what you guys think.

mschmele:
@ygliuvt those are all good points, and I assumed those decisions had already been made given that this PR exists. I think separating subscriptions and notifications is a good idea, just given how many features we will likely have to support around it. Should we pivot?

ygliuvt:
> @ygliuvt those are all good points, and I assumed those decisions had already been made given that this PR exists. I think separating subscriptions and notifications is a good idea, just given how many features we will likely have to support around it. Should we pivot?

Yes, I told @jceaser to go with just the subscriptions table because at the time, I assumed this is going to be just one filed and I really don't want to add the extra complexity with transactions and table joins, so I opted for something simple but hacky. But as we know more and I think about this more. I think we should move to a separate notifications table.

jceaser:
Changes made to create a new table instead of using the existing Subscription table. In this way I hope to future proof the database as more notification details are envisioned in future requests. This will also help isolate mutable data away from immutable data such as concept tables.

mschmele (commented):
Will the metadata-db-lib need to be modified at all so that existing functionality still works?

metadata-db-app/src/cmr/metadata_db/migrations/072_add_cmr_sub_notifications_table.clj
```diff
@@ -0,0 +1,25 @@
+(ns cmr.metadata-db.migrations.072-add-cmr-sub-notifications-table
+  "Add a table for time of last notification to join on subscription table"
+  (:require
+   [config.mdb-migrate-helper :as h]))
+
+(def ^:private sub-notifications-column-sql
+ "id NUMBER,
+ subscription_concept_id VARCHAR(255) NOT NULL,
+ notification_date TIMESTAMP WITH TIME ZONE DEFAULT SYSTIMESTAMP NOT NULL")
```
ygliuvt, line 9: change `notification_date` to `last_notified_at`?

ygliuvt (approved):

PR #1045, opened 2020-09-08
CMR-6495 Fixes issue with 1.7 to 1.6 migration

daniel-zamora:
This PR addresses issue with migrating variable version 1.7 down to 1.6. Before, StandardName wasn't being removed.


jaybarra (approved):

mschmele (approved):

mschmele (approved):
If tests weren't failing before this PR was merged, we should add some that would fail without this change

daniel-zamora:
> If tests weren't failing before this PR was merged, we should add some that would fail without this change

all-migrations-produce-valid-umm-spec should cover this, still trying to figure out why it didn't

jaybarra (approved):

jaybarra (approved):

PR #1050, opened 2020-09-10
CMR-6649 Granule responses now have optional include_polygons parameter

jaybarra:
Due to the size of orbital granule polygons in the responses, polygons have been stripped from the default response unless explicitly requested. A new parameter of `include_polygons` has been added to granule searches. This does not affect collection responses.

This PR includes XML validation updates due to a potential XML parsing library version inconsistency affecting local builds. Strict string comparisons have been replaced with regex comparisons.


daniel-zamora (approved):

daniel-zamora:
I think it would be preferable to avoid changing the xml validation tests if possible and resolve the issue with the library locally.

search-app/src/cmr/search/results_handlers/atom_json_results_handler.clj
```diff
@@ -126,34 +126,45 @@
...
     (frf/facets->echo-json facets)))
 
 (defmethod results->json false
-  [context echo-compatible? include-facets? concept-type results]
+  [context opts concept-type results]
   (let [{:keys [items facets]} results
-        items (if (= :granule concept-type)
-                (atom/append-collection-links context items)
-                items)]
+        result-items (if (= :granule concept-type)
```
jaybarra, line 141: To eliminate a nested if, update multi-method to handle granules separately

mschmele (approved):

search-app/docs/api.md
```diff
@@ -2287,8 +2288,18 @@ Exclude granule by concept id
 Exclude granule by parent concept id
 
     curl "%CMR-ENDPOINT%/granules?provider=PROV1&provider=PROV2&echo_granule_id\[\]=G1000000002-CMR_PROV1&echo_granule_id\[\]=G1000000003-CMR_PROV1&echo_granule_id\[\]=G1000000006-CMR_PROV2&exclude\[concept_id\]\[\]=C1000000001-CMR_PROV2"
+    
+#### <a name="g-include-polygons"></a> Include generated orbital polygons.
 
-#### <a name="sorting-granule-results"></a> Sorting Granule Results
+Include generated polygons in orbital granules. False by default.
```
ygliuvt, line 2294: We should make it clear that this flag only apples to ATOM and JSON response formats, not to other formats.

ygliuvt (commented):
It just occurred to me that we are saving the ATOM result in Elasticsearch. It is probably a bad idea to save the huge ATOM result in ES when we had no use of it. :(

ygliuvt (approved):

PR #1051, opened 2020-09-10
CMR-6660: Allow variable association update when updating the variables.

sxu123:
When using the new variable ingest endpoint to update a variable, the association will be updated too. There can be one and only one association for each variable, with or without collection revision info. This decision is based on the feedback from NSIDC that there is no need for a variable to be associated with multiple revisions of a collection. When a new association info is passed in, the old one will be replaced, when the exact same association info is passed in, a new revision of the old association is created.


daniel-zamora (approved):

metadata-db-app/src/cmr/metadata_db/data/memory_db.clj
```diff
@@ -408,8 +408,7 @@
   (if-let [error (or (validate-concept-id-native-id-not-changing db provider concept)
                      (when (= :variable-association (:concept-type concept))
                        (or (validate-same-provider-variable-association concept)
-                           (validate-collection-not-associated-with-another-variable-with-same-name db concept)
-                           (validate-variable-not-associated-with-another-collection db concept))))]
```
ygliuvt, line 412: why do we remove this validation?

metadata-db-app/src/cmr/metadata_db/services/concept_service.clj
```diff
@@ -378,10 +377,10 @@
...
 
-(defn- delete-association
-  "Delete the association with the given native-id using the given embedded metadata-db context."
-  [mdb-context source-concept-type native-id]
-  (let [source-concept-type-str (name source-concept-type)
+(defn- delete-variable-association
+  "Delete variable association with the given native-id using the given embedded metadata-db context."
+  [mdb-context native-id]
+  (let [source-concept-type-str "variable"
         assoc-concept-type (keyword (str source-concept-type-str "-association"))
```
ygliuvt, line 384: there is no need to define `source-concept-type-str` and you can delete this line as well. Just replace `assoc-concept-type on line 386 with `:variable-association`.

metadata-db-app/src/cmr/metadata_db/services/concept_service.clj
```diff
@@ -560,12 +575,16 @@
...
           ;; {:variable_association {:concept_id VA1-CMR :revision_id 1},
           ;;  :associated_item {:concept_id C5-PROV1 :revision_id 2}}
-          (let [assoc-info (perform-post-commit-association
-                             context
-                             concept
-                             rollback-fn)]
+          (let [assoc-info (perform-post-commit-variable-association context concept rollback-fn)]
              (merge concept assoc-info))
-           concept))
+           (if (and (= :variable-association (:concept-type concept))
```
ygliuvt, line 580: indentation is off.

system-int-test/test/cmr/system_int_test/ingest/variable_ingest_with_association_test.clj
```diff
@@ -214,3 +214,78 @@
...
+            (is (mdb/concept-exists-in-mdb? concept-id revision-id))
+            (is (= 2 revision-id))
+            (is (mdb/concept-exists-in-mdb? va-concept-id-2 va-revision-id-2))
+            (is (= 1 va-revision-id-2))
+            (is (not= va-concept-id-2 va-concept-id-1))
+
+            ;; Verify va-concept-1 is deleted.
+            (is (= true (:deleted (mdb/get-concept va-concept-id-1))))
+
+            (testing  "update the variable above with different collection on PROV1 is OK"
```
ygliuvt, line 270: I don't like this update feature. Having user specify a `native-id` which can silently update any variables on the same provider with the same native-id is very risky. That is a lot of trust to put on the client user to manage the native-id correctly. I wouldn't trust myself to keep track of the native-ids that I have used.

ygliuvt (changes requested):
Need to change the variable update rule regarding associations (in my opinion).

ygliuvt (approved):
Talked with Siwei. We allow variable update with native-id before. It is just with the variable uniqueness change, the variable uniqueness is no longer defined by variable metadata anymore. So even though this update feature destroy the existing variable associations, it is an extension of the new variable uniqueness rule (i.e. one variable can only be associated with one collection). This requires client to be more careful when updating variables (to not use an existing native-id accidentally), but it is an acceptable approach. Delete then insert maybe too cumbersome. Approved assuming the unused functions will be cleaned up.

metadata-db-app/src/cmr/metadata_db/services/concept_service.clj
sxu123: Done.

metadata-db-app/src/cmr/metadata_db/services/concept_service.clj
sxu123: Done.

metadata-db-app/src/cmr/metadata_db/data/memory_db.clj
sxu123: Done.

PR #1054, opened 2020-09-11
CMR-6689

ygliuvt:
As a CMR client user, I would like to explicitly clear scroll context.

### DIFF
--- a/search-app/src/cmr/search/api/concepts_search.clj
+++ b/search-app/src/cmr/search/api/concepts_search.clj
@@ -1,9 +1,10 @@
 (ns cmr.search.api.concepts-search
   "Defines the API for search-by-concept in the CMR."
   (:require
+   [cheshire.core :as json]
+   [clj-http.client :as client]
    [clojure.string :as string]
    [clojure.walk :as walk]
-   [clj-http.client :as client]
    [cmr.common-app.api.routes :as common-routes]
    [cmr.common-app.services.search :as search]
    [cmr.common.cache :as cache]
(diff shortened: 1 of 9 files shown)


daniel-zamora (approved):

search-app/src/cmr/search/api/concepts_search.clj
```diff
@@ -252,6 +252,32 @@
...
+  (validate-content-type headers mt/json)
+  (if-let [short-scroll-id (-> body
+                               (json/parse-string true)
+                               :scroll_id)]
+    (if-let [scroll-id (->> short-scroll-id
+                            (core-api/get-scroll-id-and-search-params-from-cache context)
+                            :scroll-id)]
+      (query-svc/clear-scroll context scroll-id)
+      (svc-errors/throw-service-error
+       :invalid-data (format "scroll_id [%s] not found." short-scroll-id)))
```
mschmele, line 276: Having a nested `if-let` makes this hard to read. Is there a way for you to break this into a separate function or make `scroll-id` and `short-scroll-id` a part of the same `let` block?

mschmele (approved):

search-app/src/cmr/search/api/concepts_search.clj
ygliuvt: This function is effectively two `if-let` blocks. I don't know how to make it easier to read. :) Breaking it into more functions would be harder to follow in my opinion.

ygliuvt: Added comments. Hopefully, this makes the code easier to follow.

PR #1055, opened 2020-09-14
CMR-6451: fix autocomplete ACL group id filtering

mschmele:
Suggestions are not being found when an autocomplete suggestion is not public but belongs to multiple groups. Updated the test to put the expected items into multiple groups and update the field to support patterned searches

### DIFF
--- a/search-app/src/cmr/search/services/autocomplete_service.clj
+++ b/search-app/src/cmr/search/services/autocomplete_service.clj
@@ -12,18 +12,18 @@
 (defn- user-groups-condition
   "Create condition to filter suggestions based on a user's permissions.
   Only create the :permitted-group-id condition if a user belongs to groups."
-  [user-acls]
-  (if-not (empty? user-acls)
+  [group-ids]
+  (if-not (empty? group-ids)
     (gc/or-conds [public-collections-condition
                   (gc/or-conds
                    ; case insensitive pattern condition to allow searching through
                    ; concatenated group names
-                   (map (fn [acl-id]
+                   (map (fn [group-id]
                           (qm/string-condition :permitted-group-ids
-                                               (format "*%s*" acl-id)
+                                               (format "*%s*" group-id)
                                                :false
                                                :true))
-                        (into user-acls ["registered"])))])
+                        group-ids))])
     public-collections-condition))
 
 (defn- empty-token-with-type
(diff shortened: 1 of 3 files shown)


eereiter (approved):

eereiter:
Approve assuming the tests all pass.

system-int-test/test/cmr/system_int_test/search/autocomplete/suggestion_reindex_test.clj
```diff
@@ -135,7 +135,9 @@
                           (dc/collection {:entry-title "c1-echo" :access-value 1})
                           {:format :echo10})
         group1-concept-id (e/get-or-create-group (s/context) "group1")
-        group-acl (e/grant-group (s/context) group1-concept-id (e/coll-catalog-item-id "PROV2" (e/coll-id ["Secret Collection"])))]
+        group2-concept-id (e/get-or-create-group (s/context) "group2")
+        group-acl (e/grant-group (s/context) group1-concept-id (e/coll-catalog-item-id "PROV2" (e/coll-id ["Secret Collection"])))
```
indiejames, line 139: Most of the bindings in this `let` are unused and should therefore be replaced with `_` or better yet moved outside the `let` block.

search-app/src/cmr/search/services/autocomplete_service.clj
```diff
@@ -16,7 +16,14 @@
   (if-not (empty? user-acls)
     (gc/or-conds [public-collections-condition
                   (gc/or-conds
-                   (map #(qm/text-condition :permitted-group-ids %) user-acls))])
+                   ; case insensitive pattern condition to allow searching through
```
indiejames, line 19: There appears to be no explicit test of the change you are introducing - in fact I see no explicit tests of ACL functionality at all here. I assume ACLs are being tested implicitly with the other search functionality, but it is unclear to me. It would be better to have explicit tests related to ACLs. If this is happening in a different file then maybe add a comment here.

sxu123 (approved):

indiejames (changes requested):

indiejames (approved):

search-app/src/cmr/search/services/autocomplete_service.clj
```diff
@@ -16,7 +16,14 @@
   (if-not (empty? user-acls)
```
daniel-zamora, line 16: these aren't "user-acls" they are group ids, so its a little misleading from a readability standpoint.

search-app/src/cmr/search/services/autocomplete_service.clj
```diff
@@ -16,7 +16,14 @@
...
                   (gc/or-conds
-                   (map #(qm/text-condition :permitted-group-ids %) user-acls))])
+                   ; case insensitive pattern condition to allow searching through
+                   ; concatenated group names
+                   (map (fn [acl-id]
+                          (qm/string-condition :permitted-group-ids
+                                               (format "*%s*" acl-id)
+                                               :false
+                                               :true))
+                        (into user-acls ["registered"])))])
```
daniel-zamora, line 26: are you assuming if any "user-acls" are passed, that all registered is automatically allowed?  If so this is incorrect.

daniel-zamora (changes requested):

search-app/src/cmr/search/services/autocomplete_service.clj
mschmele: I was - I'll fix this

search-app/src/cmr/search/services/autocomplete_service.clj
mschmele: Good call - fixed

daniel-zamora (approved):

PR #1062, opened 2020-09-23
Cmr 5934 - Add type as a search parameter to  the service concept.  Also add type to the keyword service search.  Add service_type parameter to the collection search.

eereiter:


indexer-app/src/cmr/indexer/data/index_set.clj
```diff
@@ -499,6 +499,7 @@
           ;; associated services
           :service-names (m/doc-values m/string-field-mapping)
           :service-names-lowercase (m/doc-values m/string-field-mapping)
+          :service-types-lowercase (m/doc-values m/string-field-mapping)
```
ygliuvt, line 502: `doc-value` is on by default, no need to explicitly set it. Too bad we didn't clean up the other fields in services mapping while we did the ES upgrade.

system-int-test/test/cmr/system_int_test/search/service/service_search_test.clj
```diff
@@ -219,7 +220,16 @@
       ;; Combination of params
       "Combination of params"
       [service3]
-      {:native-id "svc*" :provider "PROV2" "options[native-id][pattern]" true})))
+      {:native-id "svc*" :provider "PROV2" "options[native-id][pattern]" true}
+
+      "By service type"
```
ygliuvt, line 225: Need to add tests for `ignore_case` and `pattern` searches.

search-app/docs/api.md
```diff
@@ -3732,6 +3738,7 @@ These parameters will match fields within a service. They are case insensitive b
 
 * name
   * options: pattern, ignore_case
+* type
```
ygliuvt, line 3741: Add the options description for `type` parameter.

search-app/docs/api.md
```diff
@@ -1672,13 +1672,18 @@ Collections can be found by searching for associated services. The following ser
 
 * service_name
   * supports `pattern`, `ignore_case` and option `and`
+* service_type
```
ygliuvt, line 1675: Add options description for `service_type`

search-app/src/cmr/search/services/parameters/parameter_validation.clj
```diff
@@ -176,6 +176,7 @@
 
    ;; service related parameters
    :service-name cpv/string-plus-and-options
+   :service-type cpv/string-param-options
```
ygliuvt, line 179: Do we want to support the `AND` option similar to `service-name`?

ygliuvt (changes requested):
Good job! Looks really clean. Just need to add some tests and update API doc to be more precise.

indexer-app/src/cmr/indexer/data/concepts/service.clj
```diff
@@ -46,6 +48,7 @@
        :deleted deleted
        :service-name service-name
        :service-name-lowercase (string/lower-case service-name)
+       :service-type-lowercase (string/lower-case service-type)
```
sxu123, line 51: Why is it that you have only lowercase for the service-type, while service-name have both the lower-case and the normal service-name?

search-app/src/cmr/search/services/parameters/parameter_validation.clj
eereiter: We support the following curl https://...?service_type=Harmony&service_type=OPeNDAP.  Since each service can only have 1 type it doesn't make sense to logically "and" them, you will never get any results.  We allow multiples which "or"s them together.  Is "AND"ing a logical "OR"?

indexer-app/src/cmr/indexer/data/concepts/service.clj
eereiter: service-type is controlled vocabulary and we will never have an instance where Harmony and harmony are distinct different values. I think that it takes up more resources to have both and it is not necessary.

indexer-app/src/cmr/indexer/data/index_set.clj
eereiter: I am not sure if you are asking me to change it or are just informing me that the definition is not needed?  I didn't know and I was just following how the code is written to be consistent.  Should I go through this class and clean it up?  Although I don't know exactly what can be removed, but I can look it up.

ygliuvt: I think you can just leave it the way it is to be consistent with the other existing fields. But when adding new fields in collection or granule mappings in the future, you don't need to add the `doc-values`.

search-app/src/cmr/search/services/parameters/parameter_validation.clj
ygliuvt: cool. Sounds like we don't have a use case for the `AND` option. Thanks.

search-app/docs/api.md
eereiter: done

search-app/docs/api.md
eereiter: done

system-int-test/test/cmr/system_int_test/search/service/service_search_test.clj
eereiter: Added tests

sxu123 (approved):

system-int-test/test/cmr/system_int_test/search/service/service_search_test.clj
```diff
@@ -219,7 +220,31 @@
...
+      [service1 service2 service3]
+      {:type "OPeN*" "options[type][pattern]" true}
+
+      "By service type pattern false"
+      []
+      {:type "OPeN*" "options[type][pattern]" false}
+
+      "By service type ignore case true"
+      [service1 service2 service3]
+      {:type "OPeNDAp" "options[type][ignore-case]" true})))
```
ygliuvt, line 247: The tests should show searching with something like "opendap" found service1, service2 and service3 when ignore_case is `true` or ignore_case is missing (the default case) and found nothing when ignore_case is `false`.

ygliuvt:
There is a merge conflict in api doc that needs to be resolved.

ygliuvt (approved):
There is one more comment on the test that has not been addressed. 
Approved assuming that has been addressed.

ygliuvt:
It appears that there is still merge conflict.

system-int-test/test/cmr/system_int_test/search/service/service_search_test.clj
eereiter: I have added tests using all lower case opendap, The last test case above I used OPeNDAp where the last character is lower case, but the actual value has an upper case P, so it is really covering the same test, but it is harder to see.  I added testing with ignore-case false when opendap is lower case and mixed case.

PR #1064, opened 2020-09-24
CMR-6731, CMR-6732

ygliuvt:
Fixed bulk indexing of deleted granules that are also expired.
Fixed bulk indexing of granules/tags to not try to index into all revisions index.


daniel-zamora (changes requested):
I think we are missing tests for these two scenarios.

ygliuvt:
I have thought about testing. These are cases that not worth testing: 
1) not indexed deleted granules that are also expired, it takes too much set up to test, not worth it.
2) not index into the non-existent all revisions index for granules and tags, basically need to check there is no exception in logs. there is no good way to test it.
Basically, both cases are not worth setting up via integration testing. We can verify them via manual testing.

daniel-zamora (approved):

eereiter (approved):

PR #1067, opened 2020-10-01
Cmr 6706 UMM-G 1.6.1 is not preserving ArchiveAndDistributionInformation Name field of UMM-G 1.6 record

eereiter:
Added archive-distribution-file-name field to the old CMR granule model so that the new UMM-G ArchiveAndDistributionInformation Name element can be translated between UMM-G versions.

### DIFF
--- a/common-lib/src/cmr/common/util.clj
+++ b/common-lib/src/cmr/common/util.clj
@@ -302,14 +302,13 @@
    and empty sequences from the passed in form."
   [x]
   (cond
-    (nil? x) nil
     (map? x) (let [clean-map (remove-nil-keys
                                (zipmap (keys x) (map remove-nils-empty-maps-seqs (vals x))))]
                (when (seq clean-map)
                  clean-map))
     (vector? x) (when (not-empty x)
                   (into [] (keep remove-nils-empty-maps-seqs x)))
-    (sequential? x) (when (not-empty x)
+    (sequential? x) (when (seq x)
                       (keep remove-nils-empty-maps-seqs x))
     :else x))
 
--- a/common-lib/src/cmr/common/util.clj
+++ b/common-lib/src/cmr/common/util.clj
@@ -297,6 +297,22 @@
     :else
       x))
 
+(defn remove-nils-empty-maps-seqs
+  "Recursively removes nils, maps with nil values, empty maps, empty vectors,
+   and empty sequences from the passed in form."
+  [x]
+  (cond
+    (nil? x) nil
+    (map? x) (let [clean-map (remove-nil-keys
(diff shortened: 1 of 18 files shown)


daniel-zamora (approved):

ygliuvt (changes requested):
ArchiveAndDistributionInformation in UMM-G is an array, i.e. more than one ArchiveAndDistributionInformationType can be present in a granule. This change will not be able to handle it. 

More importantly, this issue happens when retrieving a UMM-G granule in UMM_JSON format, i.e. the granule is converted from one UMM-G version to another. In this case, rather than going through the umm-lib Granule structure then to the latest UMM-G version, CMR should have gone through the UMM-G migration directly instead. If we make the change this way, we don't have to add a new field in umm-lib Granule, the conversion will be faster and the final result will also be lossless.

Recommend to change the transformation strategy to fix this issue.

system-int-test/test/cmr/system_int_test/ingest/granule_translation_test.clj
```diff
@@ -89,6 +90,15 @@
           ;; UMM fields will be added when ISO SMAP granule support is added.
           (is (= (:granule-ur expected) (:granule-ur actual-parsed)))
 
+          (= :echo10 output-format)
```
jceaser, line 93: what does this line do?

jceaser (approved):

common-lib/src/cmr/common/util.clj
```diff
@@ -297,6 +297,22 @@
     :else
       x))
 
+(defn remove-nils-empty-maps-seqs
+  "Recursively removes nils, maps with nil values, empty maps, empty vectors,
+   and empty sequences from the passed in form."
+  [x]
+  (cond
+    (nil? x) nil
```
ygliuvt, line 305: this line is unnecessary.

common-lib/src/cmr/common/util.clj
```diff
@@ -297,6 +297,22 @@
...
+    (map? x) (let [clean-map (remove-nil-keys
+                               (zipmap (keys x) (map remove-nils-empty-maps-seqs (vals x))))]
+               (when (seq clean-map)
+                 clean-map))
+    (vector? x) (when (not-empty x)
```
ygliuvt, line 310: replace `not-empty` with `seq`

ygliuvt (approved):
Looks good now. Just a couple minor comments.

common-lib/src/cmr/common/util.clj
eereiter: removed it

common-lib/src/cmr/common/util.clj
eereiter: replaced it.

PR #1074, opened 2020-10-08
CMR-6747 Fixes product ids csv lines in usage metrics

daniel-zamora:
this PR attempts to address some product ids that weren't being processed in the csv file properly.  Some lines have values that take the form entry-title:version-id, so we split on : and take the entry-title for searching.

### DIFF
--- a/search-app/src/cmr/search/services/community_usage_metrics/metrics_service.clj
+++ b/search-app/src/cmr/search/services/community_usage_metrics/metrics_service.clj
@@ -37,7 +37,8 @@
   "Query elastic for a collection with a given entry-title, returns short-name"
   [context entry-title]
   (when (seq entry-title)
-    (let [condition (qm/string-condition :entry-title (str entry-title "*") false true)
+    (let [entry-title (first (str/split entry-title #":"))
+          condition (qm/string-condition :entry-title entry-title false false)
           query (qm/query {:concept-type :collection
                            :condition condition
                            :page-size 1
--- a/search-app/src/cmr/search/services/community_usage_metrics/metrics_service.clj
+++ b/search-app/src/cmr/search/services/community_usage_metrics/metrics_service.clj
@@ -6,6 +6,7 @@
    [clojure.data.csv :as csv]
    [clojure.string :as str]
    [cmr.common-app.services.search.parameter-validation :as cpv]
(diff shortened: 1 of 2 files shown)


jaybarra (approved):

ygliuvt (changes requested):
I don't know why you make this code change. The entry listed in CMR-6747 is `CIESIN:410`. There is no collections in PROD that has entry-title of `CIESIN`. So how would this code change fix the entry documented in the issue? 
Note: we don't have to match every entry in the community usage to a collection, if it truly does not make sense, we need to skip the entry and get back to the provider. I feel this entry is this case and the current implementation is doing the right thing.

search-app/src/cmr/search/services/community_usage_metrics/metrics_service.clj
```diff
@@ -33,51 +34,40 @@
...
+                             (qm/string-condition :short-name parsed-product-id false false)
+                             (qm/string-condition :short-name product-id false false))
+            query (qm/query {:concept-type :collection
+                             :condition condition
+                             :page-size 1
+                             :all-revisions? true
+                             :result-format :query-specified
+                             :result-fields [:short-name]})
+            results (qe/execute-query context query)
+            short-name (first (map :short-name (:items results)))]
```
ygliuvt, line 56: every item found is guaranteed to have a `short-name`,  so you could just use `(:short-name (first (:items results)))` and put this on line 57.

search-app/src/cmr/search/services/community_usage_metrics/metrics_service.clj
```diff
@@ -33,51 +34,40 @@
...
+    (when-let [parsed-product-id (as-> (re-find #"(^.+):|(^.+)" product-id) matches
+                                       (remove nil? matches)
+                                       (last matches))]
+      (let [condition (gc/or (qm/string-condition :entry-title parsed-product-id false false)
+                             (qm/string-condition :short-name parsed-product-id false false)
+                             (qm/string-condition :short-name product-id false false))
+            query (qm/query {:concept-type :collection
+                             :condition condition
+                             :page-size 1
+                             :all-revisions? true
```
ygliuvt, line 52: Why do we care about older revisions? I don't think we should.

ygliuvt (approved):

PR #1078, opened 2020-10-14
CMR-6828

ygliuvt:
Update CMR jdbc driver to work with Oracle 12c.


daniel-zamora (approved):

jceaser (approved):
The only thing I can think of is that we may want to tell Joe that a suppression was removed in case he wants to report that somewhere as good news.

PR #1086, opened 2020-10-26
CMR-6812 Code Coverage Reports During CI Builds

jaybarra:
Introduced kaocha and cloverage to generate code coverage reports during automated builds. 
Kaocha is a test library that uses the existing clojure.test engine to run tests but provides plugins for junit output and cloverage reports.

Some tests were touched to fix missing `is` statements and tagged with metadata to denote if they require an external database connection or in-memory-db.

There will be additional artifacts produced during the builds to capture each modules code coverage report.


README.md
```diff
@@ -263,6 +263,22 @@ Once these are in place, you will be able to run the following:
...
+
+Unit Tests:
+```
+lein modules utest
+```
+
+Integration Tests against a running CMR system:
+```
+# If running CMR with the in-memory database (default)
+lein modules itest --skip-meta :oracle
```
mschmele, line 276: Do you have these commands swapped?

es-spatial-plugin/test/cmr/elasticsearch/plugins/spatial/plugin_test.clj
```diff
@@ -0,0 +1,5 @@
+(ns cmr.elasticsearch.plugins.spatial.plugin-test
+  (:require
+   [clojure.test :refer :all]))
+
+(deftest ^:kaocha/pending spatial-plugin-test)
```
mschmele, line 5: Maybe add a comment about why this file is here. It's hard to infer based on the contents of the file

oracle-lib/test/cmr/oracle/sql_utils_test.clj
```diff
@@ -0,0 +1,5 @@
+(ns cmr.oracle.sql-utils-test
+  (:require
+   [clojure.test :refer :all]))
+
+(deftest ^:kaocha/pending utils-placeholder-test)
```
mschmele, line 5: Similar to the comment on the spatial test - maybe have a comment about why kaocha requires placeholder tests

redis-utils-lib/tests.edn
```diff
@@ -0,0 +1,24 @@
...
+
+ :color? #profile {:default true
+                   :ci false}
+
+ :reporter #profile {:default kaocha.report.progress/report
+                     :ci kaocha.report/documentation}
+
+ :kaocha.plugin.junit-xml/target-file "target/junit.xml"
+
+ :kaocha.plugin.profiling/count 3
```
mschmele, line 23: I've seen this in all of the `tests.edn` files and it's kind of a magic number. Maybe a comment about why 3 is the right number here would be useful

system-int-test/test/cmr/system_int_test/search/collection_temporal_search_test.clj
```diff
@@ -26,70 +26,131 @@
 ;; values of the min and max granule of the collection or the collection's temporal values if the
 ;; collection has no granules
 (deftest search-by-temporal-limit-to-granules
-  (let [coll1 (d/ingest-umm-spec-collection "PROV1" (data-umm-c/collection 1 {:EntryTitle "coll1"
```
mschmele, line 29: This looks much nicer now :)

mschmele (approved):
A few minor comments mostly just looking for added documentation

README.md
jaybarra: skip-meta will ignore anything with the tags provided. The command will skip tests requiring a live connection.

jceaser: where is the full list of tags?

jaybarra: Those are the only two tags available that are specific to CMR. Kaocha has :pending and :skip available but those aren't important to a user running the tests.

jceaser (approved):

README.md
jceaser: I think the tests you can run should be listed, new developers will not know what is possible.

PR #1087, opened 2020-10-26
CMR-6257 - NGDA Tags

jceaser:
Changed the rules on how NGDA tags are added to the opendata output. This change adds instrument data to the metadata used by the open data results handler. These values are then used to drive the inclusion of the NGDA tags. Instruments are not included in the opendata output.

### DIFF
--- a/search-app/src/cmr/search/results_handlers/opendata_results_handler.clj
+++ b/search-app/src/cmr/search/results_handlers/opendata_results_handler.clj
@@ -91,6 +91,7 @@
                          "granule-end-date"
                          "granule-start-date"
                          "insert-time"
+                         "instruments"
                          "ords"
                          "ords-info"
                          "personnel"
@@ -147,6 +148,7 @@
           entry-title :entry-title
           ords-info :ords-info
           ords :ords
+          instruments :instruments
           doi :doi
           personnel :personnel
           start-date :start-date
@@ -173,6 +175,7 @@
             :collection-citations collection-citations
             :project-sn project-sn
             :shapes (srl/ords-info->shapes ords-info ords)
+            :instruments instruments
             :doi doi
             :personnel personnel
             :start-date start-date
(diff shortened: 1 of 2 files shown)


search-app/src/cmr/search/results_handlers/opendata_results_handler.clj
```diff
@@ -242,11 +245,24 @@
     {:name (or archive-center umm-spec-util/not-provided)
      :subOrganizationOf hierarchy}))
 
+(defn check_for_modis_aster
```
jaybarra, line 248: Is this going to be such a unique case that we won't need to generalize this to a multi-method?

jaybarra: Also please rename to use dashes to be inline with code conventions. Suggestion **modis-aster?**

mitchstartzel (approved):

search-app/src/cmr/search/results_handlers/opendata_results_handler.clj
jceaser: I think the long term solution is to move this logic to cmr tags and not drive it through business logic. At least that is what Dana has suggested. Until then I think it is unique.

search-app/src/cmr/search/results_handlers/opendata_results_handler.clj
```diff
@@ -242,11 +245,24 @@
...
+  "Logic to test for collections that use the MODIS or ASTER instruments as
+  these are the only ones that should be included in the data.gov output as
+  defined by The Data Curation for Discovery (DCD) team at Marshall Space Flight
+  Center (MSFC)"
+  [x]
+  (let [short-name (:short-name x)]
```
jaybarra, line 254: Prefer destructured syntax and more verbose input value than x
'let [{:keys [short-name]} collection]`

search-app/src/cmr/search/results_handlers/opendata_results_handler.clj
```diff
@@ -242,11 +245,24 @@
     {:name (or archive-center umm-spec-util/not-provided)
      :subOrganizationOf hierarchy}))
 
+(defn check_for_modis_aster
+  "Logic to test for collections that use the MODIS or ASTER instruments as
+  these are the only ones that should be included in the data.gov output as
+  defined by The Data Curation for Discovery (DCD) team at Marshall Space Flight
+  Center (MSFC)"
+  [x]
```
mschmele, line 253: `x` -> `collection` or something more descriptive

jaybarra (changes requested):

search-app/src/cmr/search/results_handlers/opendata_results_handler.clj
jceaser: @jaybarra - do you have issue with this still?

jaybarra (approved):

mschmele (approved):

PR #1089, opened 2020-10-28
CMR-6745: Fix granule counts with ORed spatial parameters

mschmele:


search-app/src/cmr/search/services/query_execution/granule_counts_results_feature.clj
```diff
@@ -77,22 +81,60 @@
         (assoc :limit-to-granules false))
     condition))
 
+(defn- extract-spatial-operator
+  "Determine whether spatial conditions should be ORed or ANDed together"
+  [query]
+  (let [conditions (get-in query [:condition :conditions])
+        condition-is-shape? (->> conditions
```
indiejames, line 88: nit: you don't really need these bindings - you can just use the value you have for `condition-is-shape?` as the predicate of a `when` and use the value of your `parent-group-operator` as the body of the `when`.

jaybarra (approved):

indiejames (approved):

search-app/src/cmr/search/services/query_execution/granule_counts_results_feature.clj
mschmele: Fair point. I ultimately did that for readability, but if people feel strongly I can change it

PR #1094, opened 2020-11-04
CMR-6684 Catching previously unhandled unique constraint violation

jaybarra:
Update the regex to catch a concept-id  / revision-id unique constraint that was previously causing a 500. A 409 should now be returned.


jaybarra:
OBE

PR #1096, opened 2020-11-05
CMR-6757 Remove log4j

jaybarra:
Log4j is not being used and is redundant.


mschmele (approved):
"Are we even using log4j" should have been my first question. Good call

ygliuvt (approved):

PR #1100, opened 2020-11-06
Cmr 6783 removing test file temporarily.

eereiter:


mschmele (commented):
Why not just delete this test and the revert the commit when you want to add it back in?

abbottry:
> Why not just delete this test and the revert the commit when you want to add it back in?

Why not fix the test?

PR #1107, opened 2020-11-12
CMR-6623 - bypassing ACLs

jceaser:
Bypass ACLs if the requesting user is the same as the one the subscription is for

### DIFF
--- a/ingest-app/src/cmr/ingest/api/subscriptions.clj
+++ b/ingest-app/src/cmr/ingest/api/subscriptions.clj
@@ -19,51 +19,59 @@
     (v/validate-concept-metadata concept)
     concept))
 
-(defn subscription-user-matches-token?
-  "Check to see if the user for the supplied token matches the user in the subscription request"
+(defn- perform-subscription-ingest
+  "This function assumes all checks have already taken place and that a
+  subscription is ready to be saved"
   [concept request-context headers]
-  (let [token-user (api-core/get-user-id request-context headers)
-        subscriber (get (json/decode (:metadata concept)) "SubscriberId")]
-        (= subscriber token-user)))
-
-(defn ingest-subscription
-  "Processes a request to create or update a subscription."
-  [provider-id native-id request]
-  (let [{:keys [body content-type headers request-context]} request
-        concept (api-core/body->concept!
-                 :subscription provider-id native-id body content-type headers)
-        _ (lt-validation/validate-launchpad-token request-context)
-        _ (api-core/verify-provider-exists request-context provider-id)
-        subscription-user (get (json/decode (:metadata concept)) "SubscriberId")
-        token-user (api-core/get-user-id request-context headers)
-        bypass-acls (subscription-user-matches-token? concept request-context headers)
-        _ (when-not bypass-acls
-                    (info "ACLs were checked because the token user"
-                          token-user
-                          "is not the same as the subscription user"
-                          subscription-user)
-                    (acl/verify-ingest-management-permission
-                      request-context :update :provider-object provider-id)
-                    (acl/verify-subscription-management-permission
-                      request-context :update :provider-object provider-id))
-        _ (common-enabled/validate-write-enabled request-context "ingest")
-        concept (validate-and-prepare-subscription-concept concept)
-        concept-with-user-id (api-core/set-user-id concept request-context headers)
-        ;; Log the ingest attempt
+  (let [validated-concept (validate-and-prepare-subscription-concept concept)
+        concept-with-user-id (api-core/set-user-id validated-concept
+                                                   request-context
+                                                   headers)
+        ;; Log the ingest attempt before the save
         _ (info (format "Ingesting subscription %s from client %s"
                         (api-core/concept->loggable-string concept-with-user-id)
                         (:client-id request-context)))
-        save-subscription-result (ingest/save-subscription request-context concept-with-user-id)]
+        save-subscription-result (ingest/save-subscription request-context
+                                                           concept-with-user-id)]
     ;; Log the successful ingest, with the metadata size in bytes.
-
-    ; there is a worry that subscriptions could be created neferiusly, log the
-    ; action. When no longer needed (log) also remove the variables from the let.
-    (when bypass-acls
-      (warn "ACLs were bypassed because the requesting account" token-user
-        "matched the subscription user" subscription-user ""))
-
     (api-core/log-concept-with-metadata-size concept-with-user-id request-context)
     (api-core/generate-ingest-response headers save-subscription-result)))
+
+(defn ingest-subscription
+  "Processes a request to create or update a subscription. Note, this will allow
+  unlimited subscriptions to be created by users for themselfs. Revisit if this
+  Becomes a problem"
+  [provider-id native-id request]
+  (let [{:keys [body content-type headers request-context]} request]
+    (common-enabled/validate-write-enabled request-context "ingest")
+    (lt-validation/validate-launchpad-token request-context)
+    (api-core/verify-provider-exists request-context provider-id)
+    (let [concept (api-core/body->concept! :subscription provider-id native-id
+                                           body
+                                           content-type
+                                           headers)
+          subscription-user (:SubscriberId (json/decode (:medata concept) true))
+          token-user (api-core/get-user-id request-context headers)]
+      (if (= token-user subscription-user)
+        (warn (format (str "ACLs were bypassed because the token account %s "
+                      "matched the subscription user %s in the metadata.")
+                      token-user
+                      subscription-user))
+        (do
+          (info (format (str "ACLs were checked because the token user %s is "
+                        "not the same as the subscription user %s in the "
+                        "metadata.")
+                        token-user
+                        subscription-user))
+          (acl/verify-ingest-management-permission request-context
+                                                   :update
+                                                   :provider-object
+                                                   provider-id)
+          (acl/verify-subscription-management-permission request-context
+                                                         :update
+                                                         :provider-object
+                                                         provider-id)))
+      (perform-subscription-ingest concept request-context headers))))
 
 (defn delete-subscription
   "Deletes the subscription with the given provider id and native id."
--- a/ingest-app/src/cmr/ingest/api/subscriptions.clj
+++ b/ingest-app/src/cmr/ingest/api/subscriptions.clj
@@ -1,6 +1,7 @@
 (ns cmr.ingest.api.subscriptions
   "Subscription ingest functions in support of the ingest API."
   (:require
+   [cheshire.core :as json]
    [cmr.acl.core :as acl]
    [cmr.common-app.api.enabled :as common-enabled]
    [cmr.common-app.api.launchpad-token-validation :as lt-validation]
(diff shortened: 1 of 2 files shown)


ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -39,6 +55,13 @@
                         (:client-id request-context)))
         save-subscription-result (ingest/save-subscription request-context concept-with-user-id)]
     ;; Log the successful ingest, with the metadata size in bytes.
+
+    ; there is a worry that subscriptions could be created neferiusly, log the
+    ; action. When no longer needed (log) also remove the variables from the let.
```
mschmele, line 60: This comment is worded strangely and is hard to make sense of

ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -26,10 +34,18 @@
...
         _ (lt-validation/validate-launchpad-token request-context)
         _ (api-core/verify-provider-exists request-context provider-id)
-        _ (acl/verify-ingest-management-permission
-            request-context :update :provider-object provider-id)
-        _ (acl/verify-subscription-management-permission
-            request-context :update :provider-object provider-id)
+        subscription-user (get (json/decode (:metadata concept)) "SubscriberId")
+        token-user (api-core/get-user-id request-context headers)
+        bypass-acls (subscription-user-matches-token? concept request-context headers)
+        _ (when-not bypass-acls
```
mschmele, line 40: I know you didn't start this convention, but I _really_ don't like this abuse of the `_` binding in this `let`. This feels very much like a code smell and I would love to see this cleaned up significantly

mschmele (changes requested):

ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -26,10 +34,18 @@
                  :subscription provider-id native-id body content-type headers)
         _ (lt-validation/validate-launchpad-token request-context)
         _ (api-core/verify-provider-exists request-context provider-id)
-        _ (acl/verify-ingest-management-permission
-            request-context :update :provider-object provider-id)
-        _ (acl/verify-subscription-management-permission
-            request-context :update :provider-object provider-id)
+        subscription-user (get (json/decode (:metadata concept)) "SubscriberId")
+        token-user (api-core/get-user-id request-context headers)
+        bypass-acls (subscription-user-matches-token? concept request-context headers)
```
sxu123, line 39: I'm not sure about the logic.  When you UPDATE a subscription, I can understand that when the token is the same as the subscriber-id, you don't have to check the permission. But when you INGEST a brand new subscription, you will still have to check permission right? regardless if the subscriber himself is submitting the request or the operator is submitting the request on his behalf.

ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -18,6 +19,13 @@
     (v/validate-concept-metadata concept)
     concept))
 
+(defn subscription-user-matches-token?
+  "Check to see if the user for the supplied token matches the user in the subscription request"
+  [concept request-context headers]
+  (let [token-user (api-core/get-user-id request-context headers)
+        subscriber (get (json/decode (:metadata concept)) "SubscriberId")]
```
ygliuvt, line 26: We normally use symbols to parse out JSON. change this to:
```suggestion
        subscriber (:SubscriberId (json/decode (:metadata concept) true))]
```

ingest-app/src/cmr/ingest/api/subscriptions.clj
jceaser: I'm not sure about how to handle that section either. My first inclination is to end the let, hold an if and inside that do a new let.

ingest-app/src/cmr/ingest/api/subscriptions.clj
jceaser: I didn't think the key was a symbol, I thought it was a string.

ingest-app/src/cmr/ingest/api/subscriptions.clj
jceaser: If I understand you correctly, no, the idea was to allow people without this ACL (from Earthdata Search) to create their own. So I'm not sure I see what you mean.

ingest-app/src/cmr/ingest/api/subscriptions.clj
jceaser: how about:

;; In development developers pointed out that subscriptions could be created
;; neferiusly. To monitor this the action is logged so that we can decided
;; if a different approach is needed.  When the log is no longer needed,
;; also remove the variables from the let.

ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -18,30 +19,60 @@
...
+                        subscription-user))
+          (acl/verify-ingest-management-permission request-context
+                                                   :update
+                                                   :provider-object
+                                                   provider-id)
+          (acl/verify-subscription-management-permission request-context
+                                                         :update
+                                                         :provider-object
+                                                         provider-id)))
+      (perform-subscription-ingest concept request-context headers))))
```
mschmele, line 74: Nested `let`s are a code smell and this could easily be broken off into another function. I'm really not okay with this abuse of the `let` block and am happy to pair with you if you'd like

mschmele (changes requested):

ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -18,30 +19,60 @@
...
+(defn ingest-subscription
+  "Processes a request to create or update a subscription. Note, this will allow
+  unlimited subscriptions to be created by users for themselfs. Revisit if this
+  Becomes a problem"
+  [provider-id native-id request]
+  (let [{:keys [body content-type headers request-context]} request]
+    (common-enabled/validate-write-enabled request-context "ingest")
+    (lt-validation/validate-launchpad-token request-context)
+    (api-core/verify-provider-exists request-context provider-id)
+    (let [concept (api-core/body->concept! :subscription provider-id native-id
```
ygliuvt, line 49: I don't mind the multiple levels of `let` blocks. we should however put the validation related code into a function called something like `check-subscription-ingest-permission`.

ingest-app/src/cmr/ingest/api/subscriptions.clj
sxu123: Even if Earthdata Search requires no ACL on subscription creation, what about the clients using CMR without going through Earthdata Search?  Do we want to allow anyone to create subscriptions without any ACLs?

ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -18,30 +19,61 @@
...
-        save-subscription-result (ingest/save-subscription request-context concept-with-user-id)]
+        save-subscription-result (ingest/save-subscription request-context
+                                                           concept-with-user-id)]
     ;; Log the successful ingest, with the metadata size in bytes.
     (api-core/log-concept-with-metadata-size concept-with-user-id request-context)
     (api-core/generate-ingest-response headers save-subscription-result)))
 
+(defn ingest-subscription
+  "Processes a request to create or update a subscription. Note, this will allow
+  unlimited subscriptions to be created by users for themselfs. Revisit if this
```
eereiter, line 42: themselves

ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -18,30 +19,61 @@
...
+    (common-enabled/validate-write-enabled request-context "ingest")
+    (lt-validation/validate-launchpad-token request-context)
+    (api-core/verify-provider-exists request-context provider-id)
+    (let [concept (api-core/body->concept! :subscription provider-id native-id
+                                           body
+                                           content-type
+                                           headers)
+          subscription-user (:SubscriberId (json/decode (:metadata concept) true))
+          token-user (api-core/get-user-id request-context headers)]
+      (if (and token-user ;; don't allow a bypass of ACLs just because of nil values
```
daniel-zamora, line 55: single semi-colon for same line comments

system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
```diff
@@ -262,3 +264,57 @@
...
+  (let [acls (ac-util/search-for-acls
+                (transmit-config/echo-system-token)
+                {:identity-type "provider"
+                 :provider "PROV1"
+                 :target "INGEST_MANAGEMENT_ACL"})
+        _ (echo-util/ungrant (system/context) (:concept_id (first (:items acls))))
+        supplied-concept-id "SUB1000-PROV1"
+        user1-token (echo-util/login (system/context) "user1")
+        concept (subscription-util/make-subscription-concept
+                        {:concept-id supplied-concept-id
```
daniel-zamora, line 282: fix indentation

system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
```diff
@@ -262,3 +264,57 @@
...
           (is (= 200 status))
           (is (= 3 revision-id)))))))
+
+(deftest roll-your-own-subscription-tests
+  """
+  Use cases coming from EarthData Search wanting to allow their users to create
+  subscriptions without the need to have any acls
+  """
+  (let [acls (ac-util/search-for-acls
+                (transmit-config/echo-system-token)
```
daniel-zamora, line 274: fix indentation

ingest-app/src/cmr/ingest/api/subscriptions.clj
jceaser: yes, anyone can create their own subscriptions. I tried to suggest a new interface just for earthdata search, or other limitations, they are not needed by the customer. So we are going to just log them for now. Several developers have suggested that we limit the number that a user can create to something like 100, that may happen in the future. You know what, I'll write a ticket to that effect and see what our ERB says.

jceaser: https://bugs.earthdata.nasa.gov/browse/CMR-6928 - I'm going to resolve this comment.

ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -18,30 +19,66 @@
...
+  [provider-id native-id request]
+  (let [{:keys [body content-type headers request-context]} request]
+    (check-subscription-ingest-permission request-context provider-id)
+    (let [concept (api-core/body->concept! :subscription provider-id native-id
+                                           body
+                                           content-type
+                                           headers)
+          subscription-user (:SubscriberId (json/decode (:metadata concept) true))
+          token-user (api-core/get-user-id request-context headers)]
+      (if (and token-user ; don't allow a bypass of ACLs just because of nil values
```
ygliuvt, line 60: My review comment was to move these user-id and ACL checks into a function called `check-subscription-ingest-permission`, not the three basic checks that you have moved. Those three basic checks are OK to be listed as is in this `ingest-subscription` function.

ygliuvt (approved):

PR #1108, opened 2020-11-13
CMR-6164

daniel-zamora:


daniel-zamora:
PR is a little pre mature, there is currently no tests for this job.

ingest-app/src/cmr/ingest/services/jobs.clj
```diff
@@ -263,37 +264,49 @@
...
+(defn- filter-gran-refs-by-subscriber-id
+  "Takes a list of granule references and a subscriber id and removes any granule that the user does
+   not have read access to."
+  [context gran-refs subscriber-id]
+  (let [concept-ids (map :concept-id gran-refs)
+        permissions (json/parse-string
+                     (access-control/get-permissions context {:user_id subscriber-id
+                                                              :concept_id (map :concept-id gran-refs)}))]
+    (filter (fn [granule-reference]
+              (some #{"read"} (get permissions (:concept-id granule-reference))))
```
ygliuvt, line 276: What about granules that are public?

ygliuvt (changes requested):
Need to consider public granules when filtering by subscriber id.

ygliuvt:
yes, we should add a test for this.

> PR is a little pre mature, there is currently no tests for this job.

jceaser (approved):
thank you for fixing my code alignment!

ingest-app/src/cmr/ingest/services/jobs.clj
```diff
@@ -263,42 +264,61 @@
...
```
ygliuvt, line 276: I want to confirm that this check should include any granules that are accessible by registered user. Are you sure this check will include those?

daniel-zamora: yes.

ygliuvt: cool. Make sure the test has a granule covers this case (i.e. accessible by registered user).

ygliuvt (approved):

PR #1110, opened 2020-11-16
CMR-6621 subscription processing based on last notification time

jaybarra:
Subscription processing now accounts for last notification time rather than fixed value.

### DIFF
--- a/metadata-db-app/src/cmr/metadata_db/data/oracle/search.clj
+++ b/metadata-db-app/src/cmr/metadata_db/data/oracle/search.clj
@@ -10,6 +10,7 @@
    [cmr.metadata-db.data.oracle.concept-tables :as tables]
    [cmr.metadata-db.data.oracle.concepts :as oc]
    [cmr.metadata-db.data.oracle.sql-helper :as sh]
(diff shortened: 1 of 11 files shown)


metadata-db-app/src/cmr/metadata_db/data/oracle/search.clj
```diff
@@ -142,25 +149,51 @@
...
+        stmt (gen-find-concepts-in-table-sql concept-type table fields sql-params)]
     (j/with-db-transaction
-     [conn db]
-     (doall
-      (mapv #(oc/db-result->concept-map concept-type conn (:provider_id %) %)
-            (su/query conn stmt))))))
+      [conn db]
+      (doall
+        (mapv
+          #(if (:include_last_notified_at params)
```
ygliuvt, line 191: I think we have talked about it and we want to query the database with one query and getting all fields back, not making multiple sql queries to get the last notified time. This code actually making two calls for each subscription concept to get the last notification time.

ygliuvt (changes requested):
Need to make changes to how we query subscriptions with last notification time.

PR #1112, opened 2020-11-17
CMR-6758: Update Elasticsearch version

mschmele:


daniel-zamora (approved):

ygliuvt:
We need to update cloud deployments code to update ES and Kibana version as well and it needs to be tested and deployed first before this can be formally merged.

jaybarra (approved):

jceaser (approved):

ygliuvt (changes requested):
There appears to be spatial related test failures in the build. We need to make sure all tests pass in the build with the ES upgrade. I am surprised 7.10.0 is available in `docker.elastic.co` because the listing is only up to `7.9.3`.

PR #1114, opened 2020-11-18
CMR-6621 subscription updates search back from their last notification date rather than fixed amount

jaybarra:


metadata-db-app/src/cmr/metadata_db/data/oracle/search.clj
```diff
@@ -82,6 +83,30 @@
     ;; add index hint to the generated sql statement
     (update-in stmt [0] #(string/replace % #"SELECT" (format "SELECT /*+ INDEX(%s) */" table)))))
 
+(defmethod gen-find-concepts-in-table-sql :subscription
```
mschmele, line 86: Might be worth a comment at the top of this function

metadata-db-app/src/cmr/metadata_db/data/oracle/search.clj
```diff
@@ -82,6 +83,30 @@
...
+  [concept-type table fields params]
+  (let [fields (cons :last_notified_at fields)
+        root-query (if (:include-all params)
```
mschmele, line 89: Nitpick: I usually try to break things up a bit when I have an `if` statement in a `let` block. If it makes sense to, maybe put this `if` into its own function

mschmele (approved):
The subscription job namespace is definitely looking cleaner - thank you for that 👏

daniel-zamora (approved):
Paired while reviewing, if comments during pairing are resolved we can merge this.

PR #1116, opened 2020-11-18
CMR-6894

ygliuvt:
Updated revision-date for deleted expired concepts.

### DIFF
--- a/indexer-app/src/cmr/indexer/data/concepts/attribute.clj
+++ b/indexer-app/src/cmr/indexer/data/concepts/attribute.clj
@@ -2,8 +2,8 @@
   "Contains functions for converting attributes into a elastic documents"
   (:require
     [camel-snake-kebab.core :as csk]
-    [clj-time.format :as f]
-    [clojure.string :as str]
+    [clojure.string :as string]
+    [cmr.common.date-time-parser :as p]
     [cmr.common.services.errors :as errors]
     [cmr.common.util :as util]
     [cmr.umm-spec.additional-attribute :as aa]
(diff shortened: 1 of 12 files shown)


jaybarra (approved):

indexer-app/src/cmr/indexer/data/concepts/attribute.clj
```diff
@@ -2,8 +2,8 @@
   "Contains functions for converting attributes into a elastic documents"
   (:require
     [camel-snake-kebab.core :as csk]
-    [clj-time.format :as f]
-    [clojure.string :as str]
+    [clojure.string :as string]
```
mschmele, line 5: 👏

system-int-test/test/cmr/system_int_test/search/collection_sorting_search_test.clj
```diff
@@ -2,10 +2,10 @@
   "Tests searching for collections using basic collection identifiers"
   (:require
    [clj-time.core :as t]
-   [clj-time.format :as f]
    [clojure.string :as str]
    [clojure.test :refer :all]
    [cmr.common-app.services.search.messages :as msg]
+   [cmr.common.date-time-parser :as p]
```
mschmele, line 8: Nitpick: maybe name this alias something like `date-time-parser`?

system-int-test/test/cmr/system_int_test/search/collection_temporal_search_test.clj
```diff
@@ -2,8 +2,8 @@
   "Integration test for CMR collection temporal search"
   (:require
     [clj-time.core :as t]
-    [clj-time.format :as f]
     [clojure.test :refer :all]
+    [cmr.common.date-time-parser :as p]
```
mschmele, line 6: Same as above nitpick

mschmele (approved):
Nitpick comment applies to all of those aliases. It's not a big deal, though

system-int-test/test/cmr/system_int_test/search/collection_sorting_search_test.clj
ygliuvt: We have been alias this namespace as `p` throughout CMR, I think I will keep it this way to be consistent. If we want to use a different alias, we should do it across CMR. It didn't bother to change the `p` alias because `date-time-parser` is too long and the function that we call is pretty verbose already.

PR #1117, opened 2020-11-19
CMR-6807 Adds subscriber-id lookup to acl bypass for updating subscriptions

daniel-zamora:
…ptions

### DIFF
--- a/ingest-app/src/cmr/ingest/api/subscriptions.clj
+++ b/ingest-app/src/cmr/ingest/api/subscriptions.clj
@@ -48,9 +48,21 @@
 (defn- check-subscription-ingest-permission
   "All the checks needed before starting to process an ingest of subscriptions"
   [concept request-context headers provider-id]
-  (let [subscription-user (:SubscriberId (json/decode (:metadata concept) true))
+  (let [old-concept-subscriber (-> (mdb/find-concepts request-context
+                                                      {:provider-id provider-id
+                                                       :native-id (:native-id concept)
+                                                       :exclude-metadata false
+                                                       :latest true}
+                                                      :subscription)
+                                   first
+                                   :extra-fields
+                                   :subscriber-id)
+        subscription-user (if old-concept-subscriber
+                            old-concept-subscriber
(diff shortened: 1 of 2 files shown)


ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -48,8 +48,20 @@
...
+                                                    :native-id (:native-id concept)
+                                                    :exclude-metadata false
+                                                    :latest true}
+                                                   :subscription)
+                                first
+                                :extra-fields
+                                :subscriber-id)
+                            (:SubscriberId (json/decode (:metadata concept) true)))
         token-user (api-core/get-user-id request-context headers)]
+    (proto-repl.saved-values/save 2)
```
jaybarra, line 64: Remove this, looks good otherwise.

jaybarra (approved):

ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -48,7 +48,18 @@
...
+                            (-> (mdb/find-concepts request-context
+                                                   {:provider-id provider-id
+                                                    :native-id (:native-id concept)
+                                                    :exclude-metadata false
+                                                    :latest true}
+                                                   :subscription)
+                                first
+                                :extra-fields
+                                :subscriber-id)
+                            (:SubscriberId (json/decode (:metadata concept) true)))
```
sxu123, line 62: Let me try to understand this: when concept-id exists, it's an update, otherwise it's a create.
When try to update, the token user needs to be the same as the subscriber-id of the existing subscription fetched, rather than the subscriber-id in the metadata because it could be different from the original subscriber, that looks good.  When ingesting a brand new subscription, any registered user can create a subscription as long as the subscriber-id is the same as the token-user. Correct?

sxu123 (approved):

ygliuvt:
Question: Do we want to update the API doc to mention the particularity of the ingest permissions required for subscription ingest? I almost want to hide it.

ygliuvt (approved):

PR #1118, opened 2020-11-19
CMR-6805: Fixed the last variable/association old endpoint related tests.

sxu123:
1. Fixed all the variable/association tests related to the old endpoints.
2. Modified the old variable ingest endpoint - 
     only allows update. Ingesting new variable  needs to use the new endpoint.
3. Removed the old variable association/dissociation endpoints.


jaybarra (approved):

daniel-zamora (approved):
Please fix all argument indentation to be 1 space instead of 2.  Otherwise approved.

ygliuvt (approved):

PR #1119, opened 2020-12-01
CMR-6657

daniel-zamora:
This PR prevents subscription ingest for a user who doesn't have the appropriate ACL to view the collection.  Primary use is through MMT, preventing admin users from accidentally subscribing users without collection access, to a collection.

### DIFF
--- a/ingest-app/src/cmr/ingest/api/subscriptions.clj
+++ b/ingest-app/src/cmr/ingest/api/subscriptions.clj
@@ -6,9 +6,11 @@
    [cmr.common-app.api.enabled :as common-enabled]
    [cmr.common-app.api.launchpad-token-validation :as lt-validation]
    [cmr.common.log :refer [debug info warn error]]
+   [cmr.common.services.errors :as errors]
    [cmr.ingest.api.core :as api-core]
    [cmr.ingest.services.ingest-service :as ingest]
    [cmr.ingest.validation.validation :as v]
+   [cmr.transmit.access-control :as access-control]
    [cmr.transmit.metadata-db :as mdb]))
 
 (defn- validate-and-prepare-subscription-concept
@@ -23,7 +25,7 @@
 (defn- perform-subscription-ingest
   "This function assumes all checks have already taken place and that a
   subscription is ready to be saved"
-  [concept request-context headers]
+  [request-context concept headers]
   (let [validated-concept (validate-and-prepare-subscription-concept concept)
         concept-with-user-id (api-core/set-user-id validated-concept
                                                    request-context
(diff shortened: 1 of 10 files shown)


mock-echo-app/src/cmr/mock_echo/client/echo_util.clj
```diff
@@ -239,6 +239,19 @@
...
+  ([context params]
+   (ungrant-by-search context params (config/echo-system-token)))
+  ([context params token]
+   (doseq [acl (:items
+                (ac/search-for-acls context
+                                    params
+                                    {:token token}))
+           :let [concept-id (:concept_id acl)]
+           :when concept-id]
+    (ungrant context concept-id))))
```
ygliuvt, line 253: indentation is off.

system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
```diff
@@ -472,3 +468,73 @@
 
           "Delete subscription as admin"
           ingest/delete-concept [concept2 {:token admin-user-token}] 200 nil)))))
+
+(deftest subscription-ingest-subscriber-collection-permission-check-test
```
ygliuvt, line 472: We are only testing user-type, not user's group id in this test. 
In general, we always test for both cases in terms of ACL permissions. I am debating if we have to do it for both cases every single test as eventually it goes down to the same function that check for user id based permissions.
I will leave it up to you if you want to add the ACL check based on user's group id.

ygliuvt (approved):

system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
daniel-zamora: done

jaybarra (approved):

ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -20,11 +22,30 @@
...
+                                                         :user_id subscriber-id})
+                        json/decode
+                        (get concept-id))]
+    (when-not (some #{"read"} permissions)
+      (errors/throw-service-error
+       :unauthorized
+       (format "Subscriber-id [%s] does not have permission to view collection [%s]."
+               subscriber-id
+               concept-id)))))
+
```
sxu123, line 42: I remember this was a pain point before that we can't just check the permission on the collection because the subscription is really on the granules.  and the permission on the granules are not guaranteed just because the permission on the collection is granted.

daniel-zamora: This isn't a problem.

PR #1120, opened 2020-12-02
CMR-6941 removing not needed intermittened failed test.

eereiter:
As per Leo: Our tests are taking too long to run. I don’t think we need to test everything just for the sake of testing. The unit test above it is good enough for the default case.  So it was decided to delete the test. I have successfully built the branch 3 times without any errors. Building another one while this pull request is open.

### DIFF
--- a/search-app/test/cmr/search/test/services/humanizers/humanizer_range_facet_service.clj
+++ b/search-app/test/cmr/search/test/services/humanizers/humanizer_range_facet_service.clj
@@ -49,24 +49,6 @@
      :reportable false,
      :order 0}))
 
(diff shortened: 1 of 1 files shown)


jaybarra (approved):

jaybarra:
> As per Leo: Our tests are taking too long to run. I don’t think we need to test everything just for the sake of testing.
I'm not opposed to reviewing test coverage to make sure that we are testing the functionality of the system thoroughly.

mschmele (approved):
What Jay said. Assuming we're actually testing our system, I'm okay with trimming some extras

PR #1132, opened 2020-12-30
CMR-6943: separate openndap and download links in atom response

mschmele:


jaybarra (changes requested):
Please add in a normal unit test to hit the `service-url?` and `service-urls` functions.

jaybarra (approved):

umm-spec-lib/src/cmr/umm_spec/related_url.clj
```diff
@@ -21,11 +21,18 @@
      (or (get sub-type-titles sub-type)
          (get sub-type-titles "default")))))
 
+(defn service-url?
+  "Returns true if the related-url is a service url such as OPeNDAP"
+  [related-url]
+  (and (not= "Text/csv" (:MimeType related-url))
+       (= "OPENDAP DATA" (:Subtype related-url))))
```
mitchstartzel, line 28: Unsure on this point - are OPeNDAP service URLs the only type of service URL we plan to support? Function might be better named as `opendap-url?` or something similar

mitchstartzel (approved):

umm-spec-lib/src/cmr/umm_spec/related_url.clj
mschmele: I'm leaving it open for other service types in the future such as harmony

mitchstartzel: Looking at your other changes, I agree it would be better practice to add a new condition to this function rather than create new functions for each service type, if we need it in the future

PR #1133, opened 2020-12-31
CMR-7015: Update jackson-databind in message-queue-lib/metadata-db-app

mitchstartzel:
* Updated jackson-databind in message-queue-lib/metadata-db-app
* Also updated jackson-core in these microservices to support new databind version where necessary


metadata-db-app/project.clj
```diff
@@ -12,7 +12,7 @@
   :dependencies [[cheshire "5.8.1"]
                  [clj-time "0.15.1"]
                  [com.fasterxml.jackson.core/jackson-core "2.9.8"]
```
jaybarra, line 14: We should try to keep versions consistent across modules. Let's update jackson-core to match the 2.10.5 version.

jaybarra (changes requested):

metadata-db-app/project.clj
mitchstartzel: Still trying to get a stable version/build, but this will be updated to be consistent with whatever jackson version is chosen.

ygliuvt (approved):

jaybarra (approved):

PR #1138, opened 2021-01-07
CMR 6849: Change tag association/dissociation partial failure status code to 400

yiduoke:


search-app/test/cmr/search/test/api/tags.clj
```diff
@@ -0,0 +1,38 @@
...
+    }
+])
+
+(deftest tag-api-response-status-code-test
+  (testing "Making sure `tag-api-response` returns status code 400 when there's errors"
+      (do
+            (is
+                  (= 400 (:status (tags/tag-api-response dictionary-with-error))))
+            (is
+                  (= 200 (:status (tags/tag-api-response dictionary-with-no-error)))))))
```
mschmele, line 37: You shouldn't need to have this `do` block. Each `is` should run independently. The spacing is also a little off
```suggestion
  (testing "Making sure `tag-api-response` returns status code 400 when there's errors"
      (is (= 400 (:status (tags/tag-api-response dictionary-with-error))))
      (is (= 200 (:status (tags/tag-api-response dictionary-with-no-error)))))))
```

search-app/test/cmr/search/test/api/tags.clj
```diff
@@ -0,0 +1,38 @@
+(ns cmr.search.test.api.tags
+  "Tests to verify functionality in cmr.search.api.tags namespace."
+  (:require
+   [clojure.test :refer :all]
+   [cmr.search.api.tags :as tags]))
+
+(def dictionary-with-error [
```
mschmele, line 7: Nitpick with both `def`s: the object should only be indented 2 spaces. Not a big deal, but it's our convention. Let me know if you can't find an example and I'd be happy to help

search-app/src/cmr/search/api/tags.clj
```diff
@@ -20,10 +20,15 @@
...
+(defn- has-error?
+  "Whether there's an error in [data]."
+	[data]
+  (some (fn [element] (contains? element "errors")) data))
+
+(defn tag-api-response
   "Creates a successful tag response with the given data response"
   ([data]
-   (tag-api-response 200 data))
+    (if (has-error? data) (tag-api-response 400 data) (tag-api-response 200 data)))
```
jaybarra, line 31: whitespace nitpick, put the `if` cases on new lines for readability

mschmele, line 31: Fix this formatting like so:
`if <condition>`
`statement to evaluate if true`
`statement to evaluate if false`
```suggestion
    (if (has-error? data) 
      (tag-api-response 400 data)
      (tag-api-response 200 data)))
```

search-app/test/cmr/search/test/api/tags.clj
```diff
@@ -0,0 +1,38 @@
...
+        }
+        "tagged_item" {
+            "concept_id" "C1200000005-PROV1"
+        }
+    }
+])
+
+(deftest tag-api-response-status-code-test
+  (testing "Making sure `tag-api-response` returns status code 400 when there's errors"
+      (do
```
jaybarra, line 33: Unnecessary `do`, the test runner will handle both `is` forms without the extra wrapper. And the `=` can be on the same line as the `is`

mschmele (changes requested):
Logic looks good - formatting just needs some fixes

jaybarra (changes requested):
We should also add a test case in the system-int-test suite.

system-int-test/src/cmr/system_int_test/utils/tag_util.clj
```diff
@@ -231,9 +231,13 @@
    (let [{:keys [status body errors]} response
          expected-tas (map #(coll-tag-association->expected-tag-association % error?)
                            coll-tag-associations)]
-     (is (= [200
-             (set (comparable-tag-associations expected-tas))]
-            [status (set (comparable-tag-associations body))])))))
+     (if (some (fn [element] (contains? element :errors)) (vals coll-tag-associations))
+       (is (= [400
```
jaybarra, line 235: Let's remove the `if` and duplicate this to make an `assert-tag-association-response-error?` function to handle the 400 case.

jaybarra (changes requested):
Tests look good but let's just update the error handler function.

jaybarra (approved):

mschmele (approved):

PR #1143, opened 2021-01-14
CMR-7042 Generate codecov reports

jaybarra:
Generate codecov.json reports to use with CI buidls instead of HTML reports during CI builds.

Also set reporter to be documentation for all builds for consistency.

### DIFF
--- a/spatial-lib/tests.edn
+++ b/spatial-lib/tests.edn
@@ -4,6 +4,7 @@
 
  :plugins [:kaocha.plugin.alpha/info
            ;; TODO: modify testing to let cloverage work
+           ;; the primitive-math library interferes with cloverage
            ;; :kaocha.plugin/cloverage
            :capture-output
            :junit-xml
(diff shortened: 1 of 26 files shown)


spatial-lib/tests.edn
```diff
@@ -15,11 +15,14 @@
  :color? #profile {:default true
                    :ci false}
 
- :reporter #profile {:default kaocha.report.progress/report
-                     :ci kaocha.report/documentation}
+ :reporter kaocha.report/documentation
 
  :kaocha.plugin.junit-xml/target-file "target/junit.xml"
 
+ ;; re-enable when cloverage is functional for this module
```
ygliuvt, line 22: Be more specific on why cloverage is not working for spatial-lib in the comment here.

ygliuvt (approved):
Minor comment on comment.

spatial-lib/tests.edn
jaybarra: There is already a separate ticket tracking this problem.

mschmele (approved):

PR #1144, opened 2021-01-14
CMR-7030: Fixing slow ingest of Service concept when there exist many collection associations

sxu123:
The purpose of updating all the service associations when the Service concept is updated is to trigger the associated collections to be reindexed since service info is indexed into the associated collections.  This is where most of the time are spent when updating a Service. Through testing, it seems not necessary to create new revisions of all the associations, we only need to publish the association update events for each association, which would successfully trigger the reindexing of all the associated collections - Existing tests in cmr.system-int-test.search.service.collection-service-search-test already cover the functionality we need to test, for instance, "update service affect collection search has-formats field", therefore no new tests are needed.

### DIFF
--- a/metadata-db-app/src/cmr/metadata_db/services/concept_service.clj
+++ b/metadata-db-app/src/cmr/metadata_db/services/concept_service.clj
@@ -873,9 +873,11 @@
                          msg/concept-does-not-exist
                          concept-id)))))
 
-(defn- update-service-associations
-  "Create a new revision for the non-tombstoned service associations that is related to the
-  given concept; Does noting if the given concept is not a service concept."
+(defn- publish-service-association-update-events
+  "Publish a concept-update-event for each non-tombstoned service associations that is related to the
+  given concept; This is to trigger the reindexing of the associated collections in elastic
+  search when service is updated because service info is indexed into the associated collections.
+  Does noting if the given concept is not a service concept."
   [context concept-type concept-id]
   (when (= :service concept-type)
     (let [search-params (cutil/remove-nil-keys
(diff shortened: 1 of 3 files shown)


metadata-db-app/src/cmr/metadata_db/services/concept_service.clj
```diff
@@ -906,12 +907,9 @@
           associations (filter #(= false (:deleted %))
                                (search/find-concepts context search-params))]
       (doseq [association associations]
-        (save-concept-revision
-         context
-         (-> association
-             (dissoc :revision-id :revision-date :transaction-id)
-             ;; set user-id to cmr to indicate the association is created by CMR
-             (assoc :user-id "cmr")))))))
+        (ingest-events/publish-event
```
ygliuvt, line 910: Good catch!

system-int-test/test/cmr/system_int_test/search/service/collection_service_search_test.clj
```diff
@@ -407,7 +407,69 @@
         [coll1 coll2] "service1" {:ignore-case true}
 
         "ignore-case false"
-        [] "service1" {:ignore-case false}))
+        [] "service1" {:ignore-case false})
+
+      ;; update service1's Name. Make sure the collection is reindexed
```
ygliuvt, line 412: This test change is unnecessary. The service update affect collection indexed fields has been tested in other test cases (e.g. `update service affect collection search has-formats field` etc.). If you really want to add a test where the service name is changed during service update. It should be added in a different test case. This test case is for testing different options when searching collections by service name. Adding all these tests for service update is too much noise such that it diminished the goal of this test -- to make sure service info is updated in collection index when service is updated. I am ok with no making any test changes.

system-int-test/test/cmr/system_int_test/search/tool/collection_tool_search_test.clj
```diff
@@ -90,7 +90,69 @@
         [coll1 coll2] "tool1" {:ignore-case true}
 
         "ignore-case false"
-        [] "tool1" {:ignore-case false}))
+        [] "tool1" {:ignore-case false})
+
+      ;; update tool1's Name. Make sure the collection is reindexed
+      ;; so that it can be searched using the new tool Name.
+      (tool-util/ingest-tool-with-attrs
```
ygliuvt, line 97: Similar comment as I made in `collection_service_search_test.clj`. I don't think this test is in the right place or necessary.

ygliuvt (changes requested):
Need to revise the test changes. 
We should deploy a build of this branch to UAT and verify that the ingest performance problem is resolved.

system-int-test/test/cmr/system_int_test/search/service/collection_service_search_test.clj
sxu123: Removed the test changes

system-int-test/test/cmr/system_int_test/search/tool/collection_tool_search_test.clj
sxu123: Removed the test changes.

ygliuvt (approved):

jaybarra (approved):

PR #1145, opened 2021-01-15
CMR-6834 - updating copyrights in license and contributing files

jceaser:
Mitch found one file I missed, using better grep chain and found 3 files in all.


jceaser:
grep -ir "copyright " * | grep -v 2021 | grep -v cmr-exchange | grep -v gml | grep -v xml | grep 20

PR #1154, opened 2021-01-26
CMR-6930 Create subscriptions via POST with optional native-id

jaybarra:
Created POST routes in ingest to create subscriptions. If a native-id is not supplied, it will be created. If the native id is specified, that value will be used.

Updated test groups to balance CI job build times.


jceaser (approved):

ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -107,19 +109,20 @@
...
 
 (defn ingest-subscription
   "Processes a request to create or update a subscription. Note, this will allow
   unlimited subscriptions to be created by users for themselves. Revisit if this
   Becomes a problem"
-  [provider-id native-id request]
-  (let [{:keys [body content-type headers request-context]} request]
+  [provider-id opt-native-id request]
+  (let [{:keys [body content-type headers request-context]} request
+        native-id (or opt-native-id (.toString (UUID/randomUUID)))]
```
daniel-zamora, line 122: Should we check for a collision here?

daniel-zamora (approved):

ygliuvt:
We should return the `native-id` in the response.

ingest-app/src/cmr/ingest/api/subscriptions.clj
ygliuvt: I think it should already be handled in metadata-db like all the other concepts.

ygliuvt (changes requested):
We should return the `native-id` in the response body.

ingest-app/src/cmr/ingest/api/subscriptions.clj
ygliuvt: But, now we can't just fail the ingest because of a collision. @daniel-zamora is correct. We should not only check for collisions, but also handle it properly. The error should not go back to the provider.

jaybarra: Not if it's a generated ID though. Generated IDs should have a retry

daniel-zamora (approved):

sxu123 (approved):

ygliuvt (commented):
I was late in reviewing this after the latest change and before this was merged. I think we need to make the following changes:

1) We should only allow ingest without native-id in the specific POST route, not the other routes where native-id is requested. While the API doc says "If `native-id` is not specified in the request it will be generated. This is only supported for POST requests.", the implementation does not support it.
2) Even though it is highly unlikely that we will have conflicts on the auto-generated native-id, we should do a pre-check to make sure that the auto-generated native-id is not already in use. This pre-check is not fool proof against race conditions and I have thought about making it fully fool proof by doing retries on conflict errors. But given the very remote possibility of that race condition to actually occur, I think we can go with just a pre-check.

We have to make the above changes. Currently, the existing subscription API is broken with this PR merged.

PR #1162, opened 2021-02-02
CMR-7089

ygliuvt:
Create ingest queue for granule bulk updates.

### DIFF
--- a/ingest-app/src/cmr/ingest/services/event_handler.clj
+++ b/ingest-app/src/cmr/ingest/services/event_handler.clj
@@ -32,6 +32,11 @@
 (defmethod handle-provider-event :default
   [_ _])
 
+(defn- handle-gran-bulk-update-event
+  [context msg]
+  ;; place holder function to be implemented later
+  )
+
 (defn subscribe-to-events
   "Subscribe to event messages on various queues"
   [context]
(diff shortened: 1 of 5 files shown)


mschmele (approved):
I'd like to merge this ASAP so I can build on top of it - I think my tickets would require me to add tests that would cover any of the code added here

ingest-app/src/cmr/ingest/services/event_handler.clj
```diff
@@ -32,11 +32,20 @@
 (defmethod handle-provider-event :default
   [_ _])
 
+(defn- handle-gran-bulk-update-event
```
jaybarra, line 35: Should move this code to the branch that does the implementation instead of introducing code that doesn't do anything

jaybarra (changes requested):

ingest-app/src/cmr/ingest/services/event_handler.clj
mschmele: I'm going to use this immediately so one way or another it's not going to be sitting

jaybarra: Ok,

jaybarra (approved):

ingest-app/src/cmr/ingest/services/event_handler.clj
ygliuvt: This function definition cannot be removed. It is used to set up event handler for the queue. See `subscribe-to-events` function.

PR #1168, opened 2021-02-06
Bug - CMR-7071 - URL encoded subscription queries fail to run

jceaser:
It looks like the ingest of subscription was not decoding queries, this was causing the processing code to try to run the raw subscription, which it also encoded before sending over the wire. My approach was to assume "garbage in" and clean on ingest to allow for all downstream users of the data to operate as they have.

Also in this change request are some formatting changes caused by Beautify (blame it not me) and some fixes to tests discovered in the now defunct CMR-6658, such as an unreliable use of forcing a lazy seq was converted to an advance-time in subscription_processing_test.clj, also added a test for the subscription->time-constraint function because its operation is very important to the success of subscriptions query time constraints and having the test makes me feel more confident that the code is working.

### DIFF
--- a/ingest-app/src/cmr/ingest/api/subscriptions.clj
+++ b/ingest-app/src/cmr/ingest/api/subscriptions.clj
@@ -41,20 +41,20 @@
                      (json/decode true))
         concept-id (:CollectionConceptId metadata)
         subscriber-id (:SubscriberId metadata)]
-      (try
-        (let [permissions (-> (access-control/get-permissions request-context
-                                                              {:concept_id concept-id
-                                                               :user_id subscriber-id})
-                              json/decode
-                              (get concept-id))]
-          (when-not (some #{"read"} permissions)
-            (subscriber-collection-permission-error
-             subscriber-id
-             concept-id)))
-        (catch Exception e
+    (try
+      (let [permissions (-> (access-control/get-permissions request-context
+                                                            {:concept_id concept-id
+                                                             :user_id subscriber-id})
+                            json/decode
+                            (get concept-id))]
+        (when-not (some #{"read"} permissions)
           (subscriber-collection-permission-error
            subscriber-id
-           concept-id)))))
+           concept-id)))
+      (catch Exception e
+        (subscriber-collection-permission-error
+         subscriber-id
+         concept-id)))))
 
 (defn- perform-subscription-ingest
   "This function assumes all checks have already taken place and that a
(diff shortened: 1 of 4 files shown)


ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -124,6 +125,18 @@
...
+(defn- decode-query-in-concept
+  "Query can be URL encoded which will cause requests for provider=PROV1&options[spatial][or]=true
+         to fail. Extracts the Query, decodes it and then puts it back in."
+  [concept]
+  (if-let [metadata (json/parse-string (:metadata concept))]
+    (let [decoded-query (codec/form-decode-str (get metadata "Query"))
+          metadata-fixed (assoc metadata "Query" decoded-query)
+          metadata-as-str (json/generate-string metadata-fixed)
+          concept-fixed (assoc concept :metadata metadata-as-str)]
+      concept-fixed)
```
mschmele, line 137: You can use `as->` here and remove this let: https://clojuredocs.org/clojure.core/as-%3E

```suggestion
    (as-> metadata metadata
          (codec/form-decode-str (get metadata "Query")
          (assoc metadata "Query" metadata)
          (json/generate-string metadata)
          (assoc concept :metadata metadata-as-str))
```

ingest-app/test/cmr/ingest/api/subscriptions_test.clj
```diff
@@ -1,10 +1,31 @@
...
    [clojure.test :refer :all]
    [cmr.ingest.api.core :as api-core]
    [cmr.ingest.api.subscriptions :as subscriptions]))
 
+(defn- concept->query
+  "extract just the Query out of a result from process-result->hash"
+  [result]
+  (let  [meta-raw (:metadata result)
+         meta-json (json/parse-string meta-raw)]
+    (get meta-json "Query")))
```
mschmele, line 14: Nitpick: use thread-last here instead

```suggestion
  (let [meta-json (->> result
                       :metadata
                       json/parse-string]
    (get meta-json "Query")))
```

system-int-test/test/cmr/system_int_test/ingest/subscription_processing_test.clj
```diff
@@ -24,7 +30,28 @@
...
-                                                      [:read :update])]))
+                                                      [:read :update])
+    (dev-system/freeze-resume-time-fixture)]))
+
+(defn- result->query
+  "extract just the Query out of a result from process-result->hash"
+  [result]
+  (let  [meta-raw (get-in result [:subscription :metadata])
+         meta-json (json/parse-string meta-raw)]
+    (get meta-json "Query")))
```
mschmele, line 41: Nitpick: you can thread-first here

```suggestion
  (-> result 
      (get-in [:subscription :metadata])
      json/parse-string]
      (get "Query")))
```

system-int-test/test/cmr/system_int_test/ingest/subscription_processing_test.clj
```diff
@@ -24,7 +30,28 @@
...
+
+(defn- process-result->hash
+  "Deconstruct a subscription process response from the function
+       trigger-process-subscriptions into a hash table so that tests can references
+       each value by name and not have to use (nth x #). The order is defined in
+       email_processing.clj"
+  [response]
+  (map (fn [x]
```
mschmele, line 49: Nitpick: can you pick a more descriptive name for this than `x`?

ygliuvt (changes requested):
Because url decode cannot reliably detect if a string has already been url encoded, we can't just decode any query provided as url encoded without updating our documentation. 

Since the request format is `application/json`, I don't see why we can't just specify that the `Query` field in subscription be url decoded original string form and let EDSC fix their requests. Would requesting the `Query` be url encoded fix the problem or create more problems for us? 

Regardless the solution, we need to update the ingest API doc to specify the `Query` field url encoding. Also we should always save the original request in database, not the modified form -- i.e. we cannot modify the query then save it to db.

jceaser:
It has been decided by group that this change is not to be added to CMR.

PR #1181, opened 2021-02-24
CMR-6931: Improved error handling and permissions checking

mitchstartzel:


mock-echo-app/src/cmr/mock_echo/api/urs.clj
```diff
@@ -46,6 +46,16 @@
...
 
+(defn get-user-info
+  "Returns mock URS info for a user"
+  [user-id]
+  (if-not (= "null" user-id)
+    {:status 200 :body {:first_name "mock"
+                        :last_name "name"
+                        :uid user-id
+                        :email_address (str user-id"@example.com")}}
+    {:status 500 :body "<html></Stub-HTML-Page></html>"}))
```
jceaser, line 57: what is that syntax, xhtml allows for <br /> is that something your after?

transmit-lib/src/cmr/transmit/urs.clj
```diff
@@ -69,12 +70,15 @@
...
                                                           :raw? true})]
-   (when-not (= 200 status)
-     (errors/internal-error!
-      (format "Cannot get info for username [%s] in URS. Failed with status code [%d].
-               EDL error message: [%s]"
-              user status (pr-str body))))
-   body))
+    (when-not (= 200 status)
+      (info (format "Cannot get info for username [%s] in URS. Failed with status code [%d].
+        EDL error message: [%s]" user status (pr-str body)))
```
jceaser, line 75: what's in (pr-str body), there is nothing like a token right?

jceaser (changes requested):

mock-echo-app/src/cmr/mock_echo/api/urs.clj
mitchstartzel: The reason I give a bogus html response like this is that, if you hit this URS endpoint with a non-existent username (`/api/users/foobar_user`) it throws a 500, and the body is just a huge chunk of HTML. So, this was just meant to mock the actual response a bit more accurately - if the status _is_ 500, we do no processing with the body regardless.

transmit-lib/src/cmr/transmit/urs.clj
mitchstartzel: Related to above answer - the response is a big HTML block, `There has been an error processing your request. ...` So no, no user or request info.

mock-echo-app/src/cmr/mock_echo/api/urs.clj
mitchstartzel: Your point stands, it's a bad HTML response. However, the `body` in this response could be removed entirely, there would be no change in functionality. Purely a formality.

jceaser: yea, but that syntax is not valid html or xhtml, xml, or anything, <Stub-HTML-PAGE /> at least looks like it is possible.

mitchstartzel: Fair point, updated

jceaser (approved):

ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -164,7 +164,11 @@
                      (api-core/get-user-id-from-token context)
                      subscriber)
         email (if-not email
```
ygliuvt, line 166: use `(if email email ....)` over `(if-not email ... email)`

mschmele (approved):

ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -164,7 +164,11 @@
...
-                (urs/get-user-email context subscriber)
+                (if-not subscriber
```
ygliuvt, line 167: similar to comment above, you `if` over `if-not`. Don't know why you add this check here. The permission check should have happened before this function is called.

mitchstartzel: This was implemented to handle the specific case where a user tries to ingest a subscription without including a token in the header. In this case, the `subscriber` defined directly before is `nil`, which means we try to get URS user info for user `null` in the urs request.

mitchstartzel: The permission check happens after we call this function, so we would error out by way of URS responding with a 500 before we get to the ingest permission check. Specifically wanted to avoid that case

ygliuvt: Didn't realize that you are responding to my comments as I was typing it. :)
We could make `get-user-email` more robust by adding handling of a nil subscriber. After some thought, I think throwing exception here is OK if you add a comment.

mitchstartzel: Comment added

ingest-app/src/cmr/ingest/api/subscriptions.clj
mitchstartzel: 👍

ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -160,12 +160,19 @@
...
-                email)]
+        subscriber (if subscriber
+                     subscriber
+                     (api-core/get-user-id-from-token context))
+        email (if email
+                email
+                (if subscriber
+                  (urs/get-user-email context subscriber)
+                  ;; An error is thrown here to handle the case in which a subscription with no
+                  ;; subscriber-id is supplied, AND there is no token in the headers. In that case,
```
ygliuvt, line 171: Change `In that case, subscriber would be nil, and we don't want to hit URS for info on username 'nil'` to something like: `We will never allow a guest user to CRUD a subscription, so return a 401 error here.`

system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
```diff
@@ -198,7 +206,7 @@
                     {:accept-format :json
                      :raw? true})
           {:keys [errors]} (ingest/parse-ingest-body :json response)]
-      (is (re-find #"required key \[Name\] not found" (first errors)))))
+      (is (re-find #"You do not have permission to perform that action." (first errors)))))
```
ygliuvt, line 209: This test needs to be updated. It is not testing what it is supposed to test.

system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
```diff
@@ -207,7 +215,7 @@
                     {:accept-format :xml
                      :raw? true})
           {:keys [errors]} (ingest/parse-ingest-body :xml response)]
-      (is (re-find #"required key \[Name\] not found" (first errors))))))
+      (is (re-find #"You do not have permission to perform that action." (first errors))))))
```
ygliuvt, line 218: This test needs to be updated. It is not testing what it is supposed to test.

ingest-app/src/cmr/ingest/api/subscriptions.clj
mitchstartzel: 👍

system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
mitchstartzel: I disagree here - the point of this test is to make sure the error response comes back in the format specified in the accept header. Due to new logic, we now error our earlier than before (with a new message), but it is still coming back in the desired format.

system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
ygliuvt: ignore my comment. I think the test is just testing any error in response. Even though the error message in this case is awful, I guess it is OK.

system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
ygliuvt: ignore my comment. I think the test is just testing any error in response. Even though the error message in this case is awful, I guess it is OK.

ygliuvt: We should probably return a more reasonable error message when the request body is empty though. Return a permission error is really confusing.

ygliuvt: We should either fix this test in this PR or file a ticket to get a more meaningful error message in this case.

mitchstartzel: If a token was supplied with the request, then a the other error message would be returned - the schema validation doesn't occur until after permission ingest checks, though.

ygliuvt: ok, let's add a test for empty body with token for completeness.

ygliuvt: We should just delete these two tests that you modified. They are unnecessary if just checking for any error message because we have plenty tests that tests on specific error messages.

ygliuvt: nm. These are for testing accept headers. keep them.

mitchstartzel: Agreed, two new assertions added.

ygliuvt: Man, it's hard to review tests in an excerpt. :)

ygliuvt (approved):

PR #1185, opened 2021-03-04
CMR-7071 - Lookup subscriber email on demand

jceaser:
Changes to ignore the email address in the subscription and instead look it up based on the user id, which in turn can come from either metadata or the token. The very first commit to this request also includes some cleanup from past tickets which were rejected, these include an improved test and breaking out the subscription job code from the jobs.clj.

### DIFF
--- a/ingest-app/src/cmr/ingest/api/subscriptions.clj
+++ b/ingest-app/src/cmr/ingest/api/subscriptions.clj
@@ -155,14 +155,14 @@
       native-id)))
 
 (defn- add-id-and-email-to-metadata-if-missing
-   "If SubscriberId is provided, use it. Else, get it from the token.
-   If subscriber EmailAddress is provided, use it. Else, use the value supplied by EDL for the user."
+   "If SubscriberId is provided, use it. Else, get it from the token."
   [context metadata]
   (let [{subscriber :SubscriberId} metadata
         subscriber (if-not subscriber
                      (api-core/get-user-id-from-token context)
                      subscriber)]
     (assoc metadata :SubscriberId subscriber)))
+
 
 (defn add-id-and-email-if-missing
   "Parses and generates the metadata, such that add-id-and-email-to-metadata-if-missing
@@ -185,9 +185,9 @@
                             content-type
                             headers)
           native-id (get-unique-native-id request-context tmp-subscription)
-          subscriber-id (get-subscriber-id tmp-subscription)
           new-subscription (assoc  tmp-subscription :native-id native-id)
-          newer-subscription (add-id-and-email-if-missing request-context new-subscription)]
+          newer-subscription (add-id-and-email-if-missing request-context new-subscription)
+          subscriber-id (get-subscriber-id newer-subscription)]
       (check-ingest-permission request-context provider-id subscriber-id)
       (perform-subscription-ingest request-context newer-subscription headers))))
 
@@ -209,8 +209,8 @@
                             body
                             content-type
                             headers)
-          subscriber-id (get-subscriber-id tmp-subscription)
-          new-subscription (add-id-and-email-if-missing request-context tmp-subscription)]
+          new-subscription (add-id-and-email-if-missing request-context tmp-subscription)
+          subscriber-id (get-subscriber-id new-subscription)]
       (check-ingest-permission request-context provider-id subscriber-id)
       (perform-subscription-ingest request-context new-subscription headers))))
 
@@ -236,8 +236,8 @@
                                            :latest true}
                                           :subscription))]
                          (get-in original-subscription [:extra-fields :subscriber-id]))
-        new-subscriber (get-subscriber-id tmp-subscription)
-        new-subscription (add-id-and-email-if-missing request-context tmp-subscription)]
+        new-subscription (add-id-and-email-if-missing request-context tmp-subscription)
+        new-subscriber (get-subscriber-id new-subscription)]
     (check-ingest-permission request-context provider-id new-subscriber old-subscriber)
     (perform-subscription-ingest request-context new-subscription headers)))
 
--- a/ingest-app/src/cmr/ingest/api/subscriptions.clj
+++ b/ingest-app/src/cmr/ingest/api/subscriptions.clj
@@ -158,15 +158,11 @@
    "If SubscriberId is provided, use it. Else, get it from the token.
    If subscriber EmailAddress is provided, use it. Else, use the value supplied by EDL for the user."
   [context metadata]
-  (let [{subscriber :SubscriberId
-         email :EmailAddress} metadata
+  (let [{subscriber :SubscriberId} metadata
         subscriber (if-not subscriber
                      (api-core/get-user-id-from-token context)
-                     subscriber)
-        email (if-not email
-                (urs/get-user-email context subscriber)
-                email)]
-    (assoc metadata :SubscriberId subscriber :EmailAddress email)))
+                     subscriber)]
+    (assoc metadata :SubscriberId subscriber)))
 
 (defn add-id-and-email-if-missing
   "Parses and generates the metadata, such that add-id-and-email-to-metadata-if-missing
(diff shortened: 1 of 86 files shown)


ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -158,15 +158,11 @@
    "If SubscriberId is provided, use it. Else, get it from the token.
    If subscriber EmailAddress is provided, use it. Else, use the value supplied by EDL for the user."
   [context metadata]
-  (let [{subscriber :SubscriberId
-         email :EmailAddress} metadata
+  (let [{subscriber :SubscriberId} metadata
         subscriber (if-not subscriber
                      (api-core/get-user-id-from-token context)
```
mitchstartzel, line 163: In accordance with Leo's comments on my last PR - no reason not to simplify this `if-not` to an `if`

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts/subscription.clj
```diff
@@ -28,15 +27,14 @@
...
                  collection-concept-id]} :extra-fields
          user-id :user-id
          provider-id :provider-id} concept
         [cols values] (concepts/concept->common-insert-args concept)]
     [(concat cols ["provider_id" "user_id" "subscription_name"
-                   "subscriber_id" "email_address" "collection_concept_id"])
+                   "subscriber_id" "collection_concept_id"])
      (concat values [provider-id user-id subscription-name
-                     subscriber-id email-address collection-concept-id])]))
+                     subscriber-id collection-concept-id])]))
```
mitchstartzel, line 37: I think ignoring the email at ingest time is probably the right call, if for nothing other than simplifying the logic. There is one potential problematic scenario I see though - what if an admin subscribes a user to a collection, but maybe misspells the user's name? We don't do any username validation; we wouldn't catch that the user has no email until processing time. And at that point, how would we notify the admin that the subscription is no good?

ingest-app/src/cmr/ingest/services/jobs_subscriptions.clj
```diff
@@ -0,0 +1,199 @@
...
+      [sub-id subscriber-filtered-gran-refs subscriber-id subscription])))
+
+(defn- send-subscription-emails
+  "Takes processed processed subscription tuples and sends out emails if applicable."
+  [context subscriber-filtered-gran-refs-list]
+  (doseq [subscriber-filtered-gran-refs-tuple subscriber-filtered-gran-refs-list
+          :let [[sub-id subscriber-filtered-gran-refs subscriber-id subscription] subscriber-filtered-gran-refs-tuple]]
+    (when (seq subscriber-filtered-gran-refs)
+      (let [gran-ref-locations (map :location subscriber-filtered-gran-refs)
+            email-address (urs/get-user-email context subscriber-id)
```
mitchstartzel, line 183: At a certain point, we just have to trust that URS is giving us good data back, but as I mention in my comment below - should there be some method by which we return to an admin if the subscription they set up fails? Could happen due to misspelled username, for instance. Maybe we should check that a user exists in EDL at ingest time?

ingest-app/src/cmr/ingest/api/subscriptions.clj
mitchstartzel: This is also where the merge conflict exists - just use your code, it supersedes mine

system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
```diff
@@ -89,7 +93,7 @@
             response (ingest/ingest-concept concept {:token "mock-echo-system-token"})
             ingested-concept (mdb/get-concept (:concept-id response))
             parsed-metadata (json/parse-string (:metadata ingested-concept) true)]
-       (is (= "mark@example.com" (:EmailAddress parsed-metadata)))))))
+       (is (= nil (:EmailAddress parsed-metadata)))))))
```
jaybarra, line 96: nitpick, can be simplified to `nil?`

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts/subscription.clj
jceaser: I suggest we add a new ticket to add validation of EDL account an ingest time, would that work?

ingest-app/src/cmr/ingest/services/jobs_subscriptions.clj
jceaser: well, email accounts could dead, URS accounts could be deactivated. I suggest we investigate account deactivations from URS as we know it will happen to everyone at some point.

system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
jceaser: your right, I can do that!

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts/subscription.clj
mitchstartzel: Sure, I fine with that.

ingest-app/src/cmr/ingest/services/jobs_subscriptions.clj
jceaser: https://bugs.earthdata.nasa.gov/browse/CMR-7196

system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
```diff
@@ -63,6 +63,10 @@
         (is (= 201 (:status response)))
         (is (= "user1" (:SubscriberId parsed-metadata)))))))
 
+;; note, this test should be removed after email is removed from the schema and
+;;database. Until that time, test that the subscription no lonoger holds an email
+;;since it was removed from use as of CMR-7110 but not from the schema, to allow
+;;clients to continue to send it
 (deftest subscription-ingest-no-email-test
```
jceaser, line 70: remove this now

mitchstartzel (approved):

ingest-app/src/cmr/ingest/services/jobs_subscriptions.clj
```diff
@@ -0,0 +1,199 @@
...
+                           "Since this query was last run at "
+                           sub-start-time
+                           ", the following granules have been added or updated:\n\n"
+                           (email-granule-url-list gran-ref-location)
+                           "\n\nTo unsubscribe from these notifications, or if you have any questions, please contact us at [cmr-support@earthdata.nasa.gov](mailto:cmr-support@earthdata.nasa.gov).\n"))}]}))
+
+(defn- add-updated-since
+  "Pull out the start and end times from a time-constraint value and associate them to a map"
+  [raw time-constraint]
+  (let [parts (clojure.string/split time-constraint, #",")
```
daniel-zamora, line 102: clojure.string is aliased as string, I think we should use it here. I know this is probably old but now would be a good time to fix it.

daniel-zamora (approved):

ingest-app/src/cmr/ingest/services/jobs_subscriptions.clj
```diff
@@ -0,0 +1,199 @@
...
+                               (email-subscription-processing-lookback))
+              subscription (add-updated-since raw-subscription time-constraint)
+              subscriber-id (get-in subscription [:extra-fields :subscriber-id])
+              sub-id (get subscription :concept-id)
+              coll-id (get-in subscription [:extra-fields :collection-concept-id])
+              query-string (-> (:metadata subscription)
+                               (json/decode true)
+                               :Query)
+              query-params (create-query-params query-string)
+              params1 (merge {:created-at time-constraint
```
ygliuvt, line 162: There is no need to search for `created-at` as `revision-date` alone should be able to cover the use case.

jceaser: wow, this made it through so many reviews.  :) Let me give it a try and see what it does to my tests.

PR #1190, opened 2021-03-09
CMR-7071 - fixing from a merge conflict, adding these tests back

jceaser:
Introduced a bug from resolving a merge conflict, the code is fine, but we neglected to check the tests. The test needed to change because the error message was removed from the code.


mschmele (approved):

ygliuvt (changes requested):
If we don't return "You do not have permission to perform that action." error message. There is no need to have these tests here. The fix I put in my PR is what we need.

jceaser:
this is not needed seeing the explanation for why the test is not needed

PR #1191, opened 2021-03-09
CMR-6931: Moved schema changes to v1.1 instead of altering 1.0 directly

mitchstartzel:
* In my 6931 PR, the UMM-Sub v1.0 PR was edited directly. This was a mistake. Instead, should have created a v1.1 - this ticket does just that.

### DIFF
--- a/ingest-app/src/cmr/ingest/config.clj
+++ b/ingest-app/src/cmr/ingest/config.clj
@@ -43,7 +43,7 @@
 (defconfig subscription-umm-version
   "Defines the latest subscription umm version accepted by ingest - it's the latest official version.
    This environment variable needs to be manually set when newer UMM version becomes official"
-  {:default "1.0"})
+  {:default "1.1"})
 
 (defn ingest-accept-umm-version
   "Returns the latest umm version accepted by ingest for the given concept-type."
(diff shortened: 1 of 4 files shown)


ygliuvt (changes requested):
The latest supported subscription version needs to be updated to `1.1` in `cmr.ingest.config`.

umm-spec-lib/resources/json-schemas/subscription/umm/v1.0/umm-sub-json-schema.json
```diff
@@ -26,18 +26,20 @@
       "description": "The collection concept id of the granules subscribed.",
       "type": "string",
       "minLength": 1,
-      "maxLength": 255
+      "maxLength": 255
```
jaybarra, line 29: Check your editors whitespace cleanup setting. Looks like it's adding spaces

jaybarra (approved):

umm-spec-lib/resources/json-schemas/subscription/umm/v1.0/umm-sub-json-schema.json
mitchstartzel: I tried to figure this out before - will look more into it.

sxu123:
Do we need migration between 1.0 and 1.1?

ygliuvt (approved):

ygliuvt:
How are you handling the existing subscription record? Are we delete them all?

mitchstartzel:
> How are you handling the existing subscription record? Are we delete them all?

Is the alternative to this, "build a proper migration?" If so, I'm thinking that is the preferred option - but we should discuss!

jceaser (changes requested):
no documentation needed to be changed?

mitchstartzel:
Going to rework this with a proper migration, and link it to the correct ticket number as well.

PR #1200, opened 2021-03-18
CMR-7216: New has_granules_or_opensearch collection search parameter

mitchstartzel:
* This new parameter very heavily mirrors implementation for `has_granules_or_cwic`, as this parameter is meant to replace it.

### DIFF
--- a/search-app/docs/api.md
+++ b/search-app/docs/api.md
@@ -82,7 +82,7 @@ Join the [CMR Client Developer Forum](https://wiki.earthdata.nasa.gov/display/CM
     * [Additional Attribute](#c-additional-attribute)
     * [Author](#c-author)
     * [With/without granules](#c-has-granules)
-    * [With/without granules Or Cwic](#c-has-granules-or-cwic)
+    * [With/without granules Or OpenSearch](#c-has-granules-or-opensearch)
     * [OPeNDAP service URL](#c-has-opendap-url)
   * [Sorting Collection Results](#sorting-collection-results)
   * [Retrieving all Revisions of a Collection](#retrieving-all-revisions-of-a-collection)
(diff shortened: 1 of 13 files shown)


search-app/docs/api.md
```diff
@@ -1859,11 +1859,11 @@ When `has_granules` is set to "true" or "false", results will be restricted to c
 
     curl "%CMR-ENDPOINT%/collections?has_granules=true"
 
-#### <a name="c-has-granules-or-cwic"></a> Find collections with or without granules, or the collection is tagged with the configured CWIC tag.
+#### <a name="c-has-granules-or-opensearch"></a> Find collections with or without granules, or the collection is tagged with the configured OpenSearch tag.
 
-The `has_granules_or_cwic` parameter can be set to "true" or "false". When true, the results will be restricted to collections with granules or with the configured CWIC tag.  When false, results will be restricted to collections without granules.
+The `has_granules_or_opensearch` parameter can be set to "true" or "false". When true, the results will be restricted to collections with granules or with the configured OpenSearch tag (opensearch.granules.osdd).  When false, results will be restricted to collections without granules.
```
ygliuvt, line 1864: Can you clarify `When false, results will be restricted to collections without granules.` in the doc? is that all collections without granules or all collections that are tagged with Opensearch tag and also are without granules?

ygliuvt: I think it is neither. It is collections not tagged and also without granules.

mitchstartzel: The functionality for `false` is collections which either:
- have no granules, OR 
- have the configured OpenSearch tag

I considered changing this, but I ultimately decided on keeping functionality this way to match `has_granules_or_cwic`. Although, I would not be opposed to changing the functionality above to an AND instead of an OR, if others agree that would be preferable. If not, I will just clarify the docs.

ygliuvt:
General comment: Who is using the `has_granules_or_cwic` search parameter right now? If this is just a case where user will switch to a different tag to tag the collections? Can we just make the configuration change to use the new tag and keep the same search parameter? The fact that you removed `has_granules_or_cwic` parameter completely from API doc is an indication that is the exact use case. If we really want the new `has_granules_or_opensearch` parameter, we can just make that an alias to `has_granules_or_cwic` and without changing almost all the code in this PR.

mitchstartzel:
Closed - we are going to go with a much simpler, aliasing-based solution.

PR #1201, opened 2021-03-18
CMR-6932 - JWT and Heritage token support

jceaser:
Improved the Launchpad test by adding checks for several of the different types of ECHO and EDL tokens.

Legacy Tokens: Old style ECHO tokens
Heritage Tokens: New tokens made to function like legacy tokens, but are identifiable.
JWT Tokens: Long, parsable tokens

### DIFF
--- a/common-app-lib/src/cmr/common_app/api/launchpad_token_validation.clj
+++ b/common-app-lib/src/cmr/common_app/api/launchpad_token_validation.clj
@@ -5,7 +5,6 @@
    [clojure.data.codec.base64 :as base64]
    [clojure.string :as string]
    [cmr.common-app.config :as config]
-   [cmr.common.log :refer [debug info warn error]]
    [cmr.common.services.errors :as errors]
    [cmr.transmit.config :as transmit-config]))
 
@@ -28,7 +27,6 @@
    and it came from EarthDataLogin (EDL).
    Note: Similar code exists at gov.nasa.echo.kernel.service.authentication."
   [raw-token]
-  (warn "JWT token test:" (subs raw-token 0 (/ (count raw-token) 2)))
   (let [token (if (string/starts-with? raw-token "Bearer ")
                 (subs raw-token 7)
                 raw-token)]
--- a/common-app-lib/src/cmr/common_app/api/launchpad_token_validation.clj
+++ b/common-app-lib/src/cmr/common_app/api/launchpad_token_validation.clj
@@ -1,17 +1,66 @@
 (ns cmr.common-app.api.launchpad-token-validation
   "Validate Launchpad token."
   (:require
+   [cheshire.core :as json]
+   [clojure.data.codec.base64 :as base64]
+   [clojure.string :as string]
    [cmr.common-app.config :as config]
    [cmr.common.services.errors :as errors]
    [cmr.transmit.config :as transmit-config]))
 
+;; Note: Similar code exists at gov.nasa.echo.kernel.service.authentication
 (def URS_TOKEN_MAX_LENGTH 100)
+
+(defn is-heritage-token?
+  "Heritage token, meaning not the original style legacy token, but a modern
+   token made to be identifiable but also comply with the legacy standard. These
+   tokens are prefixed with 'EDL+'.
+   Note: Similar code exists at gov.nasa.echo.kernel.service.authentication."
+  [token]
+  (string/starts-with? token "EDL+"))
+
+(defn is-legacy-token?
+  "A check of last resort, only used if all other checks have not passed.
+   Note: Similar code exists at gov.nasa.echo.kernel.service.authentication."
+  [token]
+  (and (not (is-heritage-token? token))
+       (<= (count token) URS_TOKEN_MAX_LENGTH)))
+
+(defn is-jwt-token?
+  "Check if a token matches the JWT pattern (Base64.Base64.Base64) and if it
+   does, try to look inside the header section and verify that the token is JWT
+   and it came from EarthDataLogin (EDL).
+   Note: Similar code exists at gov.nasa.echo.kernel.service.authentication."
+  [token]
+  (if (some? (re-find #"[A-Za-z0-9+/=_-]+\.[A-Za-z0-9+/=_-]+\.[A-Za-z0-9+/=_-]+" token))
+    (let [token-parts (string/split token #"\.")
+          token-header (first token-parts)
+          ;token-payload (second token-parts) ;;future work - look at experation date
+          ;token-signiture (last token-parts) ;;future work - verify signature
+          header-raw (String. (.decode (java.util.Base64/getDecoder) token-header))]
+      ;; don't parse the data unless it is really needed to prevent unnecessary
+      ;; processing. Check first to see if the data looks like JSON
+      (if (and (string/starts-with? header-raw "{")
+               (string/ends-with? header-raw "}"))
+        (try
+          (if-let [header-data (json/parse-string header-raw true)]
+            (and (= "JWT" (:typ header-data))
+                 (= "Earthdata Login" (:origin header-data)))
+            false)
+          (catch com.fasterxml.jackson.core.JsonParseException e false))
+        false))
+    false))
 
 (defn is-launchpad-token?
   "Returns true if the given token is a launchpad token.
-   Currently we only check the length of the token to decide."
+   If the token is not a Legacy (ECHO), Heritage (EDL+), or JWT (newest) token,
+   then it must be a Launchpad token.
+   Note: Similar code exists at gov.nasa.echo.kernel.service.authentication."
   [token]
-  (> (count token) URS_TOKEN_MAX_LENGTH))
+  ;; note: ordered from least expensive to most
+  (not (or (is-heritage-token? token)
+           (is-legacy-token? token)
+           (is-jwt-token? token))))
 
 (defn validate-launchpad-token
   "Validate the token in request context is a Launchpad Token if launchpad token enforcement
--- a/common-app-lib/src/cmr/common_app/api/launchpad_token_validation.clj
+++ b/common-app-lib/src/cmr/common_app/api/launchpad_token_validation.clj
@@ -5,19 +5,12 @@
    [clojure.data.codec.base64 :as base64]
    [clojure.string :as string]
    [cmr.common-app.config :as config]
+   [cmr.common.log :refer [debug info warn error]]
    [cmr.common.services.errors :as errors]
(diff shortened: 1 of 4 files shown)


common-app-lib/src/cmr/common_app/api/launchpad_token_validation.clj
```diff
@@ -1,17 +1,66 @@
...
+  (and (not (is-heritage-token? token))
+       (<= (count token) URS_TOKEN_MAX_LENGTH)))
+
+(defn is-jwt-token?
+  "Check if a token matches the JWT pattern (Base64.Base64.Base64) and if it
+   does, try to look inside the header section and verify that the token is JWT
+   and it came from EarthDataLogin (EDL).
+   Note: Similar code exists at gov.nasa.echo.kernel.service.authentication."
+  [token]
+  (if (some? (re-find #"[A-Za-z0-9+/=_-]+\.[A-Za-z0-9+/=_-]+\.[A-Za-z0-9+/=_-]+" token))
```
mitchstartzel, line 35: Just a suggestion here - since it's so cheap to determine if a token is `heritage`, you could include a check in the beginning here - this could save you an expensive regex call. But, I don't know fully how this code will be used - if you already check for heritage type before ever checking for JWT type, then this is unnecessary.

common-app-lib/src/cmr/common_app/api/launchpad_token_validation.clj
```diff
@@ -1,17 +1,66 @@
...
+(defn is-heritage-token?
+  "Heritage token, meaning not the original style legacy token, but a modern
+   token made to be identifiable but also comply with the legacy standard. These
+   tokens are prefixed with 'EDL+'.
+   Note: Similar code exists at gov.nasa.echo.kernel.service.authentication."
+  [token]
+  (string/starts-with? token "EDL+"))
+
+(defn is-legacy-token?
+  "A check of last resort, only used if all other checks have not passed.
```
mitchstartzel, line 23: "A check of last resort"

If this is the last resort, then why does `is-launchpad-token?` check for `is-legacy-token?` 

I assume launchpad tokens are too large for one to ever get accidentally flagged as `false` from this check, but this docstring does seem to contradict the functionality of `is-launchpad-token?`

common-app-lib/src/cmr/common_app/api/launchpad_token_validation.clj
jceaser: well, I don't know, I like that the function documents exactly what it is that makes a JWT token a JWT token, and you need to do the pattern check. I also want these functions to be reusable as standalone checks. Also I was asked to do it last because you may be able to get out of the code faster with the easy checks done first

common-app-lib/src/cmr/common_app/api/launchpad_token_validation.clj
jceaser: no, that is not descriptive enough, when I had the tests in a different order, that comment was true, but the code is changed.

jaybarra (approved):

common-app-lib/src/cmr/common_app/api/launchpad_token_validation.clj
```diff
@@ -1,17 +1,66 @@
...
+   [clojure.data.codec.base64 :as base64]
+   [clojure.string :as string]
    [cmr.common-app.config :as config]
    [cmr.common.services.errors :as errors]
    [cmr.transmit.config :as transmit-config]))
 
+;; Note: Similar code exists at gov.nasa.echo.kernel.service.authentication
 (def URS_TOKEN_MAX_LENGTH 100)
 
+(defn is-heritage-token?
```
ygliuvt, line 14: I am hoping we don't invent new generic names for EDL tokens. How about calling this something like `is-prefixed-edl-token?`

common-app-lib/src/cmr/common_app/api/launchpad_token_validation.clj
```diff
@@ -1,17 +1,66 @@
...
+  "Heritage token, meaning not the original style legacy token, but a modern
+   token made to be identifiable but also comply with the legacy standard. These
+   tokens are prefixed with 'EDL+'.
+   Note: Similar code exists at gov.nasa.echo.kernel.service.authentication."
+  [token]
+  (string/starts-with? token "EDL+"))
+
+(defn is-legacy-token?
```
ygliuvt, line 22: The whole purpose of this function is to check the length of the token against a certain token length to rule out ECHO tokens and old EDL tokens and help us identify a Launchpad token. I would update the comment to say just that. And we should remove the check of `(is-heritage-token? token)` from the function. I don't see value in that check.

common-app-lib/src/cmr/common_app/api/launchpad_token_validation.clj
```diff
@@ -1,17 +1,66 @@
...
+  "A check of last resort, only used if all other checks have not passed.
+   Note: Similar code exists at gov.nasa.echo.kernel.service.authentication."
+  [token]
+  (and (not (is-heritage-token? token))
+       (<= (count token) URS_TOKEN_MAX_LENGTH)))
+
+(defn is-jwt-token?
```
ygliuvt, line 29: use `defn-` for private functions. This also apply to other functions defined above.

ygliuvt (changes requested):
Need some minor refactoring.

common-app-lib/src/cmr/common_app/api/launchpad_token_validation.clj
jceaser: I was going to argue for keeping it, but I took another look at the code and will combine two of the functions.

common-app-lib/src/cmr/common_app/api/launchpad_token_validation.clj
jceaser: I combined the heritage and legacy use case, so tokens are not always prefixed. These are tokens that are not JWT or Launchpad

common-app-lib/src/cmr/common_app/api/launchpad_token_validation.clj
mitchstartzel: I'm ok to keep as-is. Keeping these as standalone checks definitely has value, and is probably preferable as long as the check flow is done correctly.

jceaser:
Don't forget to review https://git.earthdata.nasa.gov/projects/CMR/repos/cmr-soap-services/pull-requests/228/overview as well. This is the legacy services change to match this change

common-app-lib/src/cmr/common_app/api/launchpad_token_validation.clj
```diff
@@ -1,17 +1,67 @@
...
+     (subs token (count prefix))
+     token)))
+
+(defn- is-legacy-token?
+  "There are two uses cases captured by this test, the Legacy token, the
+   Heritage token (meaning a token made to behave like a legacy token but
+   prefixed with EDL-).
+   Note: Similar code exists at gov.nasa.echo.kernel.service.authentication."
+  [token]
+  (or (string/starts-with? (remove-bearer token) "EDL-")
```
ygliuvt, line 28: why do we remove `Bearer` before checking for `EDL-`? This does not match what is stated in the comment of this function.

common-app-lib/src/cmr/common_app/api/launchpad_token_validation.clj
```diff
@@ -1,17 +1,63 @@
...
 (def URS_TOKEN_MAX_LENGTH 100)
 
+(defn- is-legacy-token?
+  "There are two uses cases captured by this test, the Legacy token, the
+   Heritage token (meaning a token made to behave like a legacy token but
+   prefixed with EDL-).
+   Note: Similar code exists at gov.nasa.echo.kernel.service.authentication."
+  [token]
+  (or (string/starts-with? token "EDL-")
+      (string/starts-with? token "Bearer EDL-")
```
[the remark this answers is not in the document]
ygliuvt: I think the size check on the next line would catch the old bearer tokens. That's why the overall check works. It makes senses to just check against `Bearer `. Also I would move the size check to the first check.

common-app-lib/test/cmr/common_app/test/api/launchpad_token_validation_tests.clj
```diff
@@ -0,0 +1,50 @@
...
+
+(deftest token-tests
+  ;; Security Note, these are valid looking tokens but are not and never were
+  ;; active tokens. Can be generated with something like SecureRandom.hex 32
+  ;; under rails
+  (let [token-legacy "78221B66-EF2B-9283-19B5-295D446B065B"
+        token-oauth "EDL-O84a9c44c0ad2c628300477aa2568ba39e2fa47190f5dc16f95fb24086f7"
+        token-user "EDL-U84a9c44c0ad2c628300477aa2568ba39e2fa47190f5dc16f95fb24086f7"
+        token-client "EDL-C84a9c44c0ad2c628300477aa2568ba39e2fa47190f5dc16f95fb24086f7"
+        token-ropc "EDL-R84a9c44c0ad2c628300477aa2568ba39e2fa47190f5dc16f95fb24086f7"
```
ygliuvt, line 33: I don't see Bearer token in the test.

common-app-lib/test/cmr/common_app/test/api/launchpad_token_validation_tests.clj
```diff
@@ -0,0 +1,52 @@
...
+        token-bearer (str "Bearer " token-user)
```
ygliuvt, line 34: Is it possible for us to have a Bearer token not starts with `Bearer EDL-`? I think that is possible. If so, I would like to see it in the test.

jceaser: I can add in the case where client id is passed in, I think that is what you mean, the EDL-U test is already there

jceaser: added new tests and I formatted the tests so that it, IMHO, it is easy to read.

common-app-lib/src/cmr/common_app/api/launchpad_token_validation.clj
```diff
@@ -1,17 +1,61 @@
...
```
ygliuvt, line 21: Change this line to `(string/starts-with? token "Bearer ")`, and move the line 22 size check to the first check.

ygliuvt (approved):

mitchstartzel (approved):

PR #1209, opened 2021-03-26
CMR-4865 Adding UseConstraints with LicenseURL and LicenseText to ECHO 10 schema.

eereiter:


jaybarra (approved):

mschmele (approved):

umm-spec-lib/test/cmr/umm_spec/test/umm_to_xml_mappings/echo10.clj
```diff
@@ -107,3 +107,64 @@
...
+    (let [actual-data {:UseConstraints {:LicenseURL {:Linkage "https://someurl.com"
+                                                     :Protocol "https"
+                                                     :ApplicationProfile "profile"
+                                                     :Name "License URL"
+                                                     :Description "License URL Description"
+                                                     :Function "information"
+                                                     :MimeType "text/html"}}}
+          result (echo10/umm-c-to-echo10-xml actual-data)
+          echo-result (first (select result "/Collection/UseConstraints"))]
+      (is (and (= "https://someurl.com" (value-of echo-result "LicenseURL/URL"))
```
ygliuvt, line 130: nit: rather than doing all fields check in one `is`, do each field in its own `is`. This way you will get the exact field name when the test fails.

ygliuvt (approved):

umm-spec-lib/test/cmr/umm_spec/test/umm_to_xml_mappings/echo10.clj
eereiter: I have updated the test.

PR #1210, opened 2021-03-26
CMR-7192: Redefine subscription uniqueness to use query as main parameter

mitchstartzel:
* Subscriptions must now be unique for given values of (Query, SubscriberId, Collection)

### DIFF
--- a/common-app-lib/src/cmr/common_app/services/ingest/subscription_common.clj
+++ b/common-app-lib/src/cmr/common_app/services/ingest/subscription_common.clj
@@ -5,7 +5,8 @@
    [clojure.string :as string]))
 
 (defn normalize-parameters-v1
-  "Take a list of URL parameters, and sort them."
+  "Returns a normalized url parameter string by splitting the string of parameters on '&' and
+   sorting them alphabetically"
   [parameter-string]
   (-> (if (string/starts-with? parameter-string "?")
         (subs parameter-string 1)
--- a/common-app-lib/src/cmr/common_app/services/ingest/subscription_common.clj
+++ b/common-app-lib/src/cmr/common_app/services/ingest/subscription_common.clj
@@ -1,0 +1,15 @@
+(ns cmr.common-app.services.ingest.subscription-common
+  "This contains the code for the scheduled subscription code to be shared
+   between metadata-db and ingest."
+  (:require
+   [clojure.string :as string]))
+
+(defn normalize-parameters-v1
+  "Take a list of URL parameters, and sort them."
+  [parameter-string]
+  (-> (if (string/starts-with? parameter-string "?")
+        (subs parameter-string 1)
+        parameter-string)
+      (string/split #"&")
+      sort
+      (as-> $ (string/join "&" $))))
--- a/common-app-lib/src/cmr/common_app/services/ingest/subscription_common.clj
+++ b/common-app-lib/src/cmr/common_app/services/ingest/subscription_common.clj
@@ -4,13 +4,15 @@
   (:require
    [clojure.string :as string]))
 
-(defn normalize-parameters-v1
+(defn normalize-parameters
   "Returns a normalized url parameter string by splitting the string of parameters on '&' and
    sorting them alphabetically"
   [parameter-string]
-  (-> (if (string/starts-with? parameter-string "?")
-        (subs parameter-string 1)
-        parameter-string)
-      (string/split #"&")
-      sort
-      (as-> $ (string/join "&" $))))
+  (when parameter-string
+    (-> (if (string/starts-with? parameter-string "?")
+          (subs parameter-string 1)
+          parameter-string)
+        (string/split #"&")
+        sort
+        (as-> $ (string/join "&" $))
+        (string/trim))))
--- a/common-app-lib/test/cmr/common_app/test/services/ingest/subscription.clj
+++ b/common-app-lib/test/cmr/common_app/test/services/ingest/subscription.clj
@@ -3,7 +3,6 @@
    [clojure.test :refer :all]
    [cmr.common-app.services.ingest.subscription-common :as sub-common]))
 
-
 (deftest normalize-parameters-test
   "Query normalization, should be sorted parameters"
   (testing "With a leading question mark"
--- a/common-app-lib/test/cmr/common_app/test/services/ingest/subscription.clj
+++ b/common-app-lib/test/cmr/common_app/test/services/ingest/subscription.clj
@@ -1,0 +1,23 @@
+(ns cmr.common-app.test.services.ingest.subscription
+  (:require
+   [clojure.test :refer :all]
+   [cmr.common-app.services.ingest.subscription-common :as sub-common]))
+
+
+(deftest normalize-parameters-test
+  "Query normalization, should be sorted parameters"
+  (testing "With a leading question mark"
+   (let [query "provider=PROV1&instrument=1B&instrument=2B"
+         expected "instrument=1B&instrument=2B&provider=PROV1"
+         actual (sub-common/normalize-parameters query)]
+     (is (= expected actual))))
+  (testing "Without a leading question mark"
+   (let [query "provider=PROV1&instrument=1B&instrument=2B"
+         expected "instrument=1B&instrument=2B&provider=PROV1"
+         actual (sub-common/normalize-parameters query)]
+     (is (= expected actual))))
+  (testing "Empty string"
+   (let [query ""
+         expected ""
+         actual (sub-common/normalize-parameters query)]
+     (is (= expected actual)))))
--- a/ingest-app/src/cmr/ingest/api/subscriptions.clj
+++ b/ingest-app/src/cmr/ingest/api/subscriptions.clj
@@ -59,18 +59,6 @@
            subscriber-id
            concept-id)))))
 
-(defn- subscription-duplicate-error
-  "Will Throw an error explaining that subscriptions can not be duplicated"
-  [subscriber-id concept-id, query]
-  (errors/throw-service-error
-   :conflict
-   (format (str "The subscriber id [%s] has already subscribed to the "
-                "collection with concept id [%s] using the query [%s]. "
-                "Subscribers must use unique queries for each Collection.")
-           subscriber-id
-           concept-id
-           query)))
-
 (defn- check-duplicate-subscription
   "The query used by a subscriber for a collection should be unique to prevent
    redundent emails from being sent to them. This function will check that a
@@ -84,12 +72,13 @@
         collection-id (:CollectionConceptId metadata)
         subscriber-id (:SubscriberId metadata)
         ;; Find concepts with matching collection-concept-id, normalized-query, and subscriber-id
-        duplicate-queries (mdb/find-latest-concepts
+        duplicate-queries (mdb/find-concepts
                            request-context
                            {:collection-concept-id collection-id
                             :normalized-query normalized-query
                             :subscriber-id subscriber-id
-                            :exclude-metadata false}
+                            :exclude-metadata true
+                            :latest true}
                            :subscription)
         ;;we only want to look at non-deleted subscriptions
         active-duplicate-queries (remove :deleted duplicate-queries)]
@@ -97,7 +86,12 @@
      ;;We need to make sure it has the same native-id, or else reject the ingest
      (when (and (> (count active-duplicate-queries) 0)
                 (every? #(not= native-id (:native-id %)) active-duplicate-queries))
-       (subscription-duplicate-error subscriber-id collection-id normalized-query))))
+       (errors/throw-service-error
+        :conflict
+        (format (str "The subscriber-id [%s] has already subscribed to the "
+                     "collection with concept-id [%s] using the query [%s]. "
+                     "Subscribers must use unique queries for each Collection.")
+                subscriber-id collection-id  normalized-query)))))
 
 (defn- get-subscriber-id
   "Returns the subscriber id of the given subscription concept by parsing its metadata."
@@ -199,10 +193,10 @@
 (defn- add-id-to-metadata-if-missing
    "If SubscriberId is provided, use it. Else, get it from the token."
   [context metadata]
-  (let [{subscriber :SubscriberId query :Query} metadata
-        subscriber (if-not subscriber
+  (let [subscriber-id (:SubscriberId metadata)
+        subscriber (if-not subscriber-id
                      (api-core/get-user-id-from-token context)
-                     subscriber)]
+                     subscriber-id)]
     (assoc metadata :SubscriberId subscriber)))
 
 (defn- add-fields-if-missing
--- a/ingest-app/src/cmr/ingest/api/subscriptions.clj
+++ b/ingest-app/src/cmr/ingest/api/subscriptions.clj
@@ -7,6 +7,7 @@
    [cmr.acl.core :as acl]
    [cmr.common-app.api.enabled :as common-enabled]
    [cmr.common-app.api.launchpad-token-validation :as lt-validation]
+   [cmr.common-app.services.ingest.subscription-common :as sub-common]
    [cmr.common.log :refer [debug info warn error]]
    [cmr.common.services.errors :as errors]
    [cmr.ingest.api.core :as api-core]
@@ -205,7 +206,7 @@
   [context subscription]
   (let [metadata (json/parse-string (:metadata subscription) true)
         new-metadata (add-id-to-metadata-if-missing context metadata)
-        normalized (jobs/normalize-subscription-search metadata)]
+        normalized (sub-common/normalize-parameters (:Query metadata))]
     (assoc subscription
            :metadata (json/generate-string new-metadata)
            :normalized-query normalized)))
--- a/ingest-app/src/cmr/ingest/api/subscriptions.clj
+++ b/ingest-app/src/cmr/ingest/api/subscriptions.clj
@@ -200,7 +200,7 @@
                      subscriber)]
     (assoc metadata :SubscriberId subscriber)))
 
-(defn- add-id-if-missing
+(defn- add-fields-if-missing
   "Parses and generates the metadata, such that add-id-to-metadata-if-missing
   can focus on insertion logic. Also adds normalized-query to the concept."
   [context subscription]
--- a/ingest-app/src/cmr/ingest/api/subscriptions.clj
+++ b/ingest-app/src/cmr/ingest/api/subscriptions.clj
@@ -11,9 +11,11 @@
    [cmr.common.services.errors :as errors]
    [cmr.ingest.api.core :as api-core]
    [cmr.ingest.services.ingest-service :as ingest]
+   [cmr.ingest.services.subscriptions-helper :as jobs]
    [cmr.ingest.validation.validation :as v]
    [cmr.transmit.access-control :as access-control]
    [cmr.transmit.metadata-db :as mdb]
+   [cmr.transmit.metadata-db2 :as mdb2]
    [cmr.transmit.urs :as urs])
   (:import
    [java.util UUID]))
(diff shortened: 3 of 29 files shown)


ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -57,6 +59,40 @@
...
+   subscriber-id, normalized-query."
+  [request-context subscription]
+  (let [native-id (:native-id subscription)
+        provider-id (:provider-id subscription)
+        metadata (-> (:metadata subscription) (json/decode true))
+        normalized-query (:normalized-query subscription)
+        collection-id (:CollectionConceptId metadata)
+        subscriber-id (:SubscriberId metadata)]
+    ;; There should be no other subscriptions matching this query ; nil is good
+    (when-not (nil? (mdb/find-latest-concept
```
ygliuvt, line 87: change this to (when (mdb/find-latest-concept...). Also fix the indentations in the lines below.

ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -154,23 +190,27 @@
...
+(defn- add-id-to-metadata-if-missing
    "If SubscriberId is provided, use it. Else, get it from the token."
   [context metadata]
-  (let [{subscriber :SubscriberId} metadata
+  (let [{subscriber :SubscriberId query :Query} metadata
         subscriber (if-not subscriber
                      (api-core/get-user-id-from-token context)
-                     subscriber)]
+                     subscriber)
+        query ()
```
ygliuvt, line 200: Remove these lines. Don't know why they are here.

common-app-lib/src/cmr/common_app/services/ingest/subscription_common.clj
```diff
@@ -0,0 +1,15 @@
+(ns cmr.common-app.services.ingest.subscription-common
+  "This contains the code for the scheduled subscription code to be shared
+   between metadata-db and ingest."
+  (:require
+   [clojure.string :as string]))
+
+(defn normalize-parameters-v1
+  "Take a list of URL parameters, and sort them."
```
ygliuvt, line 8: Change the docstring to say something like "Returns the normalized parameter url string by ....".

ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -154,23 +190,27 @@
...
+        normalized ()]
     (assoc metadata :SubscriberId subscriber)))
 
-
-(defn add-id-and-email-if-missing
+(defn- add-id-if-missing
   "Parses and generates the metadata, such that add-id-and-email-to-metadata-if-missing
```
ygliuvt, line 205: update the docstring.

metadata-db-app/src/cmr/metadata_db/migrations/074_add_normalized_to_subscription_table.clj
```diff
@@ -0,0 +1,46 @@
...
+    []
+    (println "cmr.metadata-db.migrations.074-update-subscription-table up...")
+
+    ;; The Normalized_Query column holds a version of the Subscription Query which
+    ;; has been normalized to a standard followed by both EDSC and CMR. Both
+    ;; applications can assume the normalized query is the same despite changes
+    ;; in query order or other such nominal differences. Simple string compair
+    ;; can be used on the normalized query.
+    (helper/sql "alter table cmr_subscriptions add normalized_query VARCHAR2(4000)")
+    (populate-new-column))
```
ygliuvt, line 40: We should add db index with collection_concept_id, subscriber_id and normalized_query so that the find subscription query will not be a full table scan.

system-int-test/test/cmr/system_int_test/ingest/subscription_processing_test.clj
```diff
@@ -115,12 +120,19 @@
...
+       (create-subscription-and-index coll1, "test_sub_prov1" "user2" "provider=PROV1")
        (is (= 2
              (->> (trigger-process-subscriptions)
                   (map #(nth % 1))
                   flatten
                   (map :concept-id)
-                  count)))))))
+                  count))))
+
+     (testing "check that duplicate subscriptions fail"
```
ygliuvt, line 133: For completeness, add a test to show using the same query parameters to create a new subscription on a deleted subscription is OK.

ygliuvt (changes requested):
Need to add db index and test.

common-app-lib/src/cmr/common_app/services/ingest/subscription_common.clj
mitchstartzel: Updated to, "Returns a normalized url parameter string by splitting the string of parameters on '&' and sorting them alphabetically"

system-int-test/test/cmr/system_int_test/ingest/subscription_processing_test.clj
```diff
@@ -115,9 +108,9 @@
      (testing "Deleting the subscription purges the suscription notification entry"
        ;; Delete and reingest the subscription. If the sub-notification was purged, then it
        ;; should look back 24 hours, as if the subscription were new.
-       (ingest/delete-concept subscription-concept)
-       (subscription-util/ingest-subscription subscription-concept
-                                              {:token "mock-echo-system-token"})
+       (let [concept {:provider-id "PROV1" :concept-type :subscription :native-id "test_sub_prov1"}]
+         (ingest/delete-concept concept))
+       (subscription-util/create-subscription-and-index coll1, "test_sub_prov1" "user2" "provider=PROV1")
```
jaybarra, line 113: unnecesary comma

ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -57,6 +59,46 @@
            subscriber-id
            concept-id)))))
 
+(defn- subscription-duplicate-error
```
ygliuvt, line 62: This function contains one function call and it is called just once in the code base. We don't need this function. Just call `(errors/throw-service-error ...)` in `check-duplicate-subscription` is more readable.

transmit-lib/src/cmr/transmit/metadata_db.clj
```diff
@@ -162,17 +162,25 @@
 
 (defn find-latest-concept
   "Searches metadata db for the latest concept matching the given parameters. Do not throw serivce
-  excpetion, returns the status and error message in a map in case of error."
+  exception, returns the status and error message in a map in case of error."
+  ([context params concept-type]
+   (find-latest-concept context params concept-type first))
+  ;"Allow for custom filtration of the results, instead of the default (first)"
+  ([context params concept-type filter-fn]
```
jaybarra, line 169: What is the purpose of adding a filter to `find-latest-concept` rather than using `find-concepts`?

ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -154,23 +196,25 @@
         (get-unique-native-id context subscription))
       native-id)))
 
-(defn- add-id-and-email-to-metadata-if-missing
+(defn- add-id-to-metadata-if-missing
    "If SubscriberId is provided, use it. Else, get it from the token."
   [context metadata]
-  (let [{subscriber :SubscriberId} metadata
+  (let [{subscriber :SubscriberId query :Query} metadata
```
ygliuvt, line 202: no need to introduce `query` variable if it is not used.

transmit-lib/src/cmr/transmit/metadata_db.clj
```diff
@@ -162,17 +162,25 @@
...
+         status (int status)]
+     (case status
+       200 (filter-fn (map mdb2/finish-parse-concept (json/decode body true)))
+       404 (errors/throw-service-error :not-found body)
+       ;; default
+       (errors/internal-error!
+         (format "%s search failed. status: %s body: %s"
+                 (str/capitalize (name concept-type)) status body))))))
+
+(defn find-latest-concepts
```
jaybarra, line 180: Will this ever return more than one since it doesn't allow passing in a filter to override the default `first` filter?

ingest-app/test/cmr/ingest/test/services/subscriptions_helper.clj
```diff
@@ -94,3 +94,18 @@
             expected (str start "," now)
             actual (#'jobs/subscription->time-constraint data now -1234)]
         (is (= expected actual))))))
+
+(deftest normalize-subscription-search-test
+  "Query normalization, should be sorted parameters"
+  (testing
```
ygliuvt, line 100: nit: CMR coding standard is to put description of the test on the same line of `testing` i.e. `(testing "With a leading question mark"`

ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -57,6 +59,46 @@
...
+        normalized-query (:normalized-query subscription)
+        collection-id (:CollectionConceptId metadata)
+        subscriber-id (:SubscriberId metadata)
+        ;; Find concepts with matching collection-concept-id, normalized-query, and subscriber-id
+        duplicate-queries (mdb/find-latest-concepts
+                           request-context
+                           {:collection-concept-id collection-id
+                            :normalized-query normalized-query
+                            :subscriber-id subscriber-id
+                            :exclude-metadata false}
```
ygliuvt, line 92: shouldn't :exclude-metadata be `true`?

ingest-app/src/cmr/ingest/services/subscriptions_helper.clj
```diff
@@ -29,6 +31,14 @@
 ;; SET NEXT_FIRE_TIME =(((cast (SYS_EXTRACT_UTC(SYSTIMESTAMP) as DATE) - DATE'1970-01-01')*86400 + 1200) * 1000)
 ;; WHERE trigger_name='EmailSubscriptionProcessing.job.trigger';
 
+(defn normalize-subscription-search
```
ygliuvt, line 34: There is no need for this function. We should work the `string/trim` into the existing function `normalize-parameters-v1`, and just call that function when needed.

metadata-db-app/src/cmr/metadata_db/migrations/074_add_normalized_to_subscription_table.clj
```diff
@@ -0,0 +1,52 @@
...
+    ;; can be used on the normalized query.
+    (helper/sql "alter table cmr_subscriptions add normalized_query VARCHAR2(4000)")
+    (populate-new-column)
+    (create-subscription-index))
+
+(defn down
+  "Migrates the database down from version 74."
+  []
+  (println "cmr.metadata-db.migrations.074-update-subscription-table down.")
+  (helper/sql "alter table cmr_subscriptions drop column normalized_query"))
```
ygliuvt, line 52: Drop the index first.

metadata-db-app/README.md
```diff
@@ -404,6 +404,7 @@ Supported combinations of concept type and parameters:
   * tag associations with concept-id, native-id, tag-key, associated-concept-id, associated-revision-id
   * services with concept-id, provider-id, native-id
   * service associations with concept-id, native-id, service-concept-id, associated-concept-id, associated-revision-id
+  * subscriptions with collection-concept-id, provider-id, subscriber-id, normalized-query
```
ygliuvt, line 407: add `concept-id` and `native-id`.

metadata-db-app/src/cmr/metadata_db/migrations/074_add_normalized_to_subscription_table.clj
```diff
@@ -0,0 +1,52 @@
...
+  (doseq [result (helper/query "SELECT id, concept_id, metadata FROM cmr_subscriptions")]
+    (let [id (:id result)
+          query (result->query result)
+          sql-statment (format "UPDATE cmr_subscriptions SET normalized_query='%s' WHERE id=%s" query id)]
+      (helper/sql sql-statment))))
+
+(defn- create-subscription-index
+  []
+  (helper/sql "CREATE INDEX cmr_subs_ccinqsi ON CMR_SUBSCRIPTIONS
+              (collection_concept_id, normalized_query, subscriber_id)"))
```
ygliuvt, line 32: change the order to `(collection_concept_id, subscriber_id, normalized_query)`

system-int-test/test/cmr/system_int_test/bootstrap/bulk_index/subscriptions_test.clj
```diff
@@ -48,12 +48,16 @@
      (let [coll1 (d/ingest-umm-spec-collection "PROV1" (data-umm-c/collection {:EntryTitle "E1"
                                                                                :ShortName "S1"
                                                                                :Version "V1"}))
-           sub1 (subscription/ingest-subscription-with-attrs {:provider-id "PROV1" :CollectionConceptId (:concept-id coll1)}
+           sub1 (subscription/ingest-subscription-with-attrs {:provider-id "PROV1"
+                                                              :Query "limit=1"
```
ygliuvt, line 52: nit: I wish the `Query` could be more realistic, nobody would create a subscription with a query of `limit=1`, something like `short_name=x` would be better.

system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
```diff
@@ -758,6 +763,7 @@
                      {:SubscriberId "post-user"
                       :Name "a different subscription with native-id"
                       :native-id "another-native-id"
+                      :Query "polygon=1,2,3"
```
ygliuvt, line 766: nit: throughout the PR, `polygon=1,2,3` is not a valid polygon value. Change it to something valid like `platform=x`

transmit-lib/src/cmr/transmit/metadata_db.clj
```diff
@@ -162,17 +162,25 @@
 
 (defn find-latest-concept
   "Searches metadata db for the latest concept matching the given parameters. Do not throw serivce
-  excpetion, returns the status and error message in a map in case of error."
+  exception, returns the status and error message in a map in case of error."
```
ygliuvt, line 165: Looks like you just want to add a function called `find-latest-concepts`. It is a lot easier to do that with the existing `find-concepts` function (just addd `:latest true` to the params would do it) rather than dealing with this function.

ygliuvt (changes requested):

transmit-lib/src/cmr/transmit/metadata_db.clj
mitchstartzel: Yes - since we we pass in the function`(fn [item] item)` as our `filter-fn`, find-latest-concept will return the latest revision of every matching concept, instead of just whichever concept happens to be `(first ..)`

transmit-lib/src/cmr/transmit/metadata_db.clj
mitchstartzel: By default, this function only returns the `(first...)` concept it finds, but for this purpose, we needed it to return all concepts it finds. So, the custom filter provided by the new `find-latest-concepts` is really just "no filter," although the filter could be more specific if ever needed for another purpose.

transmit-lib/src/cmr/transmit/metadata_db.clj
mitchstartzel: `find-concepts` does exactly what I need it to, using it will be much cleaner 👍

metadata-db-app/src/cmr/metadata_db/migrations/074_add_normalized_to_subscription_table.clj
mitchstartzel: I have no problems with this change, but I'm very curious why you prefer this order.

ingest-app/src/cmr/ingest/services/subscriptions_helper.clj
mitchstartzel: The idea with this function is that it serves to be a constant call for the future, even if we want to change the normalization algorithm with a future `normalize-parameters-v2`. That way, we could just plug the new v2 into this function without having to change all the function calls in code/tests.

system-int-test/test/cmr/system_int_test/ingest/subscription_processing_test.clj
mitchstartzel: Removed here, and a few other places where `create-subscription-and-index` was called

transmit-lib/src/cmr/transmit/metadata_db.clj
mitchstartzel: Update on this comment and the one below: I have swapped to using `find-concepts`, so `find-latest-concept` will be reverted, and the new function will be removed.

mitchstartzel: I have switched over to find-concepts - this is a much simpler solution, no arguments there.

transmit-lib/src/cmr/transmit/metadata_db.clj
mitchstartzel: Outdated - I have swapped over to `find-concepts`, so this comment no longer applies.

system-int-test/test/cmr/system_int_test/bootstrap/bulk_index/subscriptions_test.clj
mitchstartzel: I have done a pass on all the tests to use realistic queries.

metadata-db-app/src/cmr/metadata_db/migrations/074_add_normalized_to_subscription_table.clj
ygliuvt: Good question. The order in the index matters in that if you use partial fields to do a search, the index can still be used if you use the first couple fields in a multi-fields index. It is more likely for a user to search for subscriptions based on collection_concept_id and subscriber_id than collection_concept_id and normalized_query. That is why I want you to make that change. It is not something that we would definitely benefit from, but I just can't unseen it once I noticed it. :)

ingest-app/src/cmr/ingest/services/subscriptions_helper.clj
ygliuvt: I don't follow your argument. :) If we change the normalization algorithm in the future, we would just change the function `normalize-parameters-v1`. The test should go with that function rather than this one. I still recommend delete this function and move the tests to `normalize-parameters-v1`. Since we can't have more than one algorithms to normalize parameter, change the name of the function to `normalize-parameters`.

metadata-db-app/src/cmr/metadata_db/migrations/074_add_normalized_to_subscription_table.clj
mitchstartzel: I see - I will definitely keep this in mind for the future.

system-int-test/src/cmr/system_int_test/utils/subscription_util.clj
```diff
@@ -182,3 +183,15 @@
   "Verifies the subscription references"
   [subscriptions response]
   (d/refs-match? subscriptions response))
+
+(defn create-subscription-and-index
+  [collection name user query]
+  (let [concept {:Name (or name "test_sub_prov1")
+                :SubscriberId (or user "user2")
```
ygliuvt, line 190: indentation is off.

system-int-test/test/cmr/system_int_test/ingest/subscription_processing_test.clj
```diff
@@ -41,14 +41,13 @@
 (defn- create-granule-and-index
   "A utility function to reduce common code in these tests, Create a test granule and then wait for it to be indexed."
   [provider collection granule-ur]
-  (let [granule (data-core/ingest provider
-                    (data-granule/granule-with-umm-spec-collection collection
-                                                                   (:concept-id collection)
-                                                                   {:granule-ur granule-ur
-                                                                    :access-value 1})
-                    {:token "mock-echo-system-token"})]
+  (let [concept-id (:concept-id collection)
```
ygliuvt, line 44: I am not sure this is more readable than what we have before.

mitchstartzel: I find it more readable, if only because `result` is returned instead of `granule`, which signifies that there is ingest response info along with the granule.

ingest-app/src/cmr/ingest/services/subscriptions_helper.clj
mitchstartzel: After discussion with Thomas and Leo, decided to collapse to just one function, `normalize parameters`.

ygliuvt (approved):

common-app-lib/test/cmr/common_app/test/services/ingest/subscription.clj
```diff
@@ -0,0 +1,23 @@
+(ns cmr.common-app.test.services.ingest.subscription
+  (:require
+   [clojure.test :refer :all]
+   [cmr.common-app.services.ingest.subscription-common :as sub-common]))
+
+
```
jaybarra, line 6: extra whitespace nitpick

jaybarra (approved):

PR #1216, opened 2021-04-02
CMR-7191 update bulk granule update schema to have stricter rules

jaybarra:
Would like some additional feedback about the format of the error messages returned by the JSON schema validation library. Specifically the "#/updates/2" referring to the instruction level schema. Should they be transformed into something more readable or fine as-is?

### DIFF
--- a/ingest-app/resources/granule_bulk_update_schema.json
+++ b/ingest-app/resources/granule_bulk_update_schema.json
@@ -6,18 +6,29 @@
         "UPDATE_FIELD"
       ]
     },
+    "UpdateFieldEnum": {
+      "type": "string",
+      "enum": [
+        "OPeNDAPLink",
+        "S3Link"
+      ]
+    },
+    "UpdateTupleType": {
+      "type": "array",
+      "minItems": 2,
+      "maxItems": 2,
+      "items": {
+        "type": "string"
+      },
+      "examples": [
+        ["SL:AB_5DSno.008:30500511", "s3://aws.example.com"]
+      ]
+    },
     "UpdateArgumentsType": {
-      "title": "Items",
       "type": "array",
       "minItems": 1,
       "items": {
-        "title": "Arguments",
-        "type": "string",
-        "default": "",
-        "examples": [
-          "SL:AB_5DSno.008:30500511"
-        ],
-        "pattern": "^.*$"
+        "$ref": "#/definitions/UpdateTupleType"
       }
     }
   },
@@ -33,11 +44,9 @@
   "properties": {
     "name": {
       "$id": "#root/name",
-      "title": "Name",
       "type": "string",
-      "default": "",
       "examples": [
-        "Add opendap links"
+        "Add OPeNDAP links"
       ],
       "pattern": "^.*$"
     },
@@ -45,23 +54,10 @@
       "$ref": "#/definitions/OperationEnum"
     },
     "update-field": {
-      "$id": "#root/update-field",
-      "title": "Update-field",
-      "type": "string",
-      "default": "",
-      "examples": [
-        "OPeNDAPLink"
-      ],
-      "pattern": "^.*$"
+      "$ref": "#/definitions/UpdateFieldEnum"
     },
     "updates": {
-      "$id": "#root/updates",
-      "title": "Updates",
-      "type": "array",
-      "minItems": 1,
-      "items": {
-        "$ref": "#/definitions/UpdateArgumentsType"
-      }
+      "$ref": "#/definitions/UpdateArgumentsType"
     }
   }
 }
--- a/ingest-app/resources/granule_bulk_update_schema.json
+++ b/ingest-app/resources/granule_bulk_update_schema.json
@@ -21,7 +21,7 @@
         "type": "string"
       },
       "examples": [
-        ["SL:AB_5DSno.008:30500511", "s3://aws.example.com"]
+        ["SL:AB_5DSno.008:30500511", "https://example.com/opendap/30500511"]
       ]
     },
     "UpdateArgumentsType": {
--- a/ingest-app/src/cmr/ingest/services/granule_bulk_update_service.clj
+++ b/ingest-app/src/cmr/ingest/services/granule_bulk_update_service.clj
@@ -28,9 +28,34 @@
 (def granule-bulk-update-schema
  (js/json-string->json-schema (slurp (io/resource "granule_bulk_update_schema.json"))))
 
-(defn- validate-granule-bulk-update
- [json]
- (js/validate-json! granule-bulk-update-schema json))
+(defn- validate-granule-bulk-update-json
+  "Validate the request against the schema and return the parsed request."
+  [json]
+  (js/validate-json! granule-bulk-update-schema json))
+
+(defn- validate-granule-bulk-update-job-name
+  "Validates that if a name is provided for a job, it is unique to the provider."
+  [context request provider-id]
+  (when-let [job-name (:name request)]
+    (let [tasks (seq (data-granule-bulk-update/get-granule-tasks-by-provider context provider-id))
+          task-names (->> tasks
+                          (map :name)
+                          (map #(string/split % #":"))
+                          first)]
+      (when (some? (some #{job-name} task-names))
+        (errors/throw-service-errors
+         :bad-request
+         [(format (str "Granule bulk update task name needs to be unique within the provider. "
+                       "A granule bulk update job with the name [%s] already exist for provider [%s].")
+                  job-name
+                  provider-id)])))))
+
+(defn- duplicates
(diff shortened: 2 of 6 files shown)


system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_granule_update_schema_test.clj
```diff
@@ -0,0 +1,74 @@
...
+         response (json/parse-string body true)]
+
+     (is (= 400 status))
+     (is (seq (filter #(= err-msg %) (:errors response)))
+         (format "Error message containing [%s] was not found in [%s]"
+                 err-msg
+                 (pr-str response))))
+
+   "insufficient bulk operation targets"
+   (fn [m] (assoc m :updates []))
```
mschmele, line 42: I'd like a comment block explaining the what/why with the `fn` being passed, or for the tests to be made simpler (even if it makes them verbose)

mschmele (commented):
The code all looks good to me. I'll approve as soon as the tests are either more thoroughly commented or simplified

ingest-app/src/cmr/ingest/services/granule_bulk_update_service.clj
```diff
@@ -66,11 +73,22 @@
...
                     (let [message (.getMessage e)
-                          user-facing-message (if (string/includes? message "GBUT_PN_I")
-                                                "Granule bulk update task name needs to be unique within the provider."
-                                                message)]
+                          user-facing-message
+                          (condp #(string/includes? %2 %1) message
+                            "GBUT_PN_I"
+                            "Granule bulk update task name needs to be unique within the provider."
+
+                            "ORA-00001: unique constraint (CMR_INGEST.BUGS_PK)"
```
ygliuvt, line 81: This check should happen before any of the database operation takes place, i.e. we don't want a granule bulk update task created if it is invalid.

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_granule_update_schema_test.clj
```diff
@@ -0,0 +1,76 @@
...
+   [cmr.system-int-test.utils.index-util :as index]
+   [cmr.system-int-test.utils.ingest-util :as ingest]))
+
+(use-fixtures :each (ingest/reset-fixture {"provguid1" "PROV1"}))
+
+(def base-request {:name "sample bulk update"
+                   :operation "UPDATE_FIELD"
+                   :update-field "OPeNDAPLink"
+                   :updates
+                   [["ur_1" "s3://aws.foo.com"]
```
ygliuvt, line 20: This `base-request` is a bad request itself. It provided a `s3` link to an OPeNDAPLink update. We should probably add a new validation to flag this error as well. We should use a base request that is valid and show that it passes the validation.

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_granule_update_schema_test.clj
```diff
@@ -0,0 +1,76 @@
...
+   "update entries: fewer than 2 (1)"
+   (fn [req] (update req :updates conj ["ur_3"]))
+   "#/updates/2: expected minimum item count: 2, found: 1"
+
+   "update entries: fewer than 2 (0)"
+   (fn [req] (update req :updates conj []))
+   "#/updates/2: expected minimum item count: 2, found: 0"
+
+   "duplicate granule_ur in request"
+   (fn [req] (update req
```
ygliuvt, line 60: I would rather see the duplicate urs be listed explicitly here in the test rather than doing a `conj` that made thing opaque.

ygliuvt (changes requested):
Need to move the duplicate granule urs validation to before the db insertion.

ingest-app/resources/granule_bulk_update_schema.json
```diff
@@ -6,18 +6,29 @@
...
+    },
+    "UpdateTupleType": {
+      "type": "array",
+      "minItems": 2,
+      "maxItems": 2,
+      "items": {
+        "type": "string"
+      },
+      "examples": [
+        ["SL:AB_5DSno.008:30500511", "s3://aws.example.com"]
```
ygliuvt, line 24: The examples should be consistent among the fields. We shouldn't use `OPeNDAPLink` in Operation, but an S3 link in the url value.

ingest-app/src/cmr/ingest/services/granule_bulk_update_service.clj
jaybarra: Should the check of the name of the job being unique per provider also be done prior to insertion?

ingest-app/src/cmr/ingest/services/granule_bulk_update_service.clj
```diff
@@ -64,14 +109,12 @@
...
-                                                "Granule bulk update task name needs to be unique within the provider."
-                                                message)]
+                    (error "An exception occurred saving a bulk granule update request:" e)
+                    (let [message (.getMessage e)]
                       (errors/throw-service-errors
-                       :invalid-data
-                       [(str "error creating granule bulk update task: " user-facing-message)]))))]
+                       :internal-error
+                       [(str "There was a problem saving a bulk granule update request."
+                             "Please try again, if the problem persists please contact cmr@nasa.gov.")]))))]
```
jaybarra, line 117: Since the exception message is likely from the database directly, we may not want to expose it to the user. If we get to this spot we will likely need to figure out what is causing this and have ops or dev support to diagnose and fix the problem. Error is hidden from the user but available to developers and ops in logging.

ingest-app/src/cmr/ingest/services/granule_bulk_update_service.clj
```diff
@@ -47,15 +72,35 @@
         event-type (string/lower-case (str operation ":" update-field))]
     (map (partial update->instruction event-type) updates)))
 
+(defn- validate-granule-bulk-update-no-duplicates
+  "Validate no duplicate URs exist within the list of instructions"
+  [request]
+  (when-let [duplicate-urs (->> request
+                                request->instructions
+                                (pmap :granule-ur)
```
ygliuvt, line 80: change `pmap` to `map`.

ingest-app/src/cmr/ingest/services/granule_bulk_update_service.clj
ygliuvt, line 117: I don't think `cmr@nasa.gov` is a valid email. I think it should be `cmr-support@earthdata.nasa.gov`. Please ask John to confirm.

ingest-app/src/cmr/ingest/services/granule_bulk_update_service.clj
```diff
@@ -28,9 +28,34 @@
...
 
-(defn- validate-granule-bulk-update
- [json]
- (js/validate-json! granule-bulk-update-schema json))
+(defn- validate-granule-bulk-update-json
+  "Validate the request against the schema and return the parsed request."
+  [json]
+  (js/validate-json! granule-bulk-update-schema json))
+
+(defn- validate-granule-bulk-update-job-name
```
ygliuvt, line 36: Do we have this requirement to enforce unique job name within a provider? From what @mschmele  coded, he is actually allowing this to happen by appending task id to the name if I remember correctly.

jaybarra: If that's the case I'll remove the check. The other message from oracle should have been removed before then.

ygliuvt: We should keep the message from Oracle. That is guarded against race condition which is different from checking the task name from user input.

system-int-test/test/cmr/system_int_test/ingest/granule_bulk_update/granule_bulk_update_schema_test.clj
```diff
@@ -0,0 +1,79 @@
...
+
+(def base-request {:operation "UPDATE_FIELD"
+                   :update-field "OPeNDAPLink"
+                   :updates
+                   [["ur_1" "https://aws.foo.com"]
+                    ["ur_2" "https://aws.bar.com"]]})
+
+(deftest bulk-granule-schema-validation-test
+  (are3
+   [mutate err-msg]
```
ygliuvt, line 24: nit: This mutate way of doing the test cases is too verbose and hard to follow. I think if we just list the updates as the argument, it would be easier to follow.

ygliuvt: the operations and update-field validations can be separated out.

ygliuvt (approved):

mschmele (approved):

PR #1218, opened 2021-04-05
CMR-7145: Added latest umm schema versions to the api doc.

sxu123:


littleidiot40 (approved):

mschmele (changes requested):
We have a lot of templating code that should handle this kind of logic. I don't like the idea of slurping up the `.md` files, doing a string replace, and then spitting it back out. Unless it's not feasible to do it any other way, I'd really rather use the patterns already in place

sxu123:
@mschmele, Please point out exactly what you think should be used.

mschmele:
@sxu123 use selmer. It is our templating library and I believe we have several examples of injecting values into static pages in the search-app

sxu123:
@mschmele it's nothing more than using "(static/generate"  api and that's exactly what i'm using.

ingest-app/docs/api.md
```diff
@@ -280,6 +284,11 @@ A native-id is an identifier, unique per provider, used to identify concepts wit
 
 ## <a name="metadata-ingest"></a> Metadata Ingest
 
+### <a name="latest-umm-versions"></a> Latest UMM Schema Versions
+
+<Placeholder for UMM Schema Versions>
+[//]: # (Please don't modify the Placeholder line above. It will be replaced by the latest acceptable UMM schema versions at api.html generation time.)
```
mschmele, line 290: I think this can be done via Selmer's `variables and tagging` here. Something like this. You'll need to create a collection of tuples or maps, and pass them through as part of the `data` object that gets passed to the render
```suggestion
<ul>
{% for format in data.metadata-formats %}
    <li>{{format.format-name}}: {{format.format-version-number}}</li>
{% endfor %}
</ul>
```

sxu123: Mark, I've tried. You can not reference the variables at the api.md level.  These version variables can only be referenced in ingest-docs-static.html. I could make it work that way but it's kind of a hybrid which I don't really like.  
1. I have to add only the UMM Schema Version section in api.md 
### Latest UMM Schema Versions

  * [Latest UMM schema versions](#latest-umm-versions)

2. Then I will add the real content in ingest-docs-static.html, which means it's gonna be at very end of the api.html:

{% block main-content %}
{{ page-content|safe }}
<h3><a name="latest-umm-versions"></a> Latest UMM Schema Versions</h3>
<p>The following are the latest acceptable UMM schema versions for metadata ingest:</p>
<p>UMM-C: {{ umm-c }}<br/>UMM-G: {{ umm-g }}<br/>UMM-S: {{ umm-s }}<br/>UMM-T: {{ umm-t }}<br/>UMM-SUB: {{ umm-sub }}<br/>UMM-VAR: {{ umm-var }}</p>
{% endblock %}

Honestly I think this is more confusing because when you look at the api.md, you don't see latest-umm-versions section because it's generated in the ingest-docs-static.html.

Here is what I modified the code to make the above approach work:
(defn generate-api-docs
  "Generate CMR Ingest API docs."
  []
  (let [collection-umm-version (common-config/collection-umm-version)
        granule-umm-version (ingest-config/granule-umm-version)
        variable-umm-version (ingest-config/variable-umm-version)
        service-umm-version (ingest-config/service-umm-version)
        tool-umm-version (ingest-config/tool-umm-version)
        subscription-umm-version (ingest-config/subscription-umm-version)]
  (static/generate
   "resources/public/site/docs/ingest/api.html"
   "templates/ingest-docs-static.html"
   (merge
    (data/base-static)
    {:site-title "CMR Ingest"
     :page-title "API Documentation"
     :page-content (static/md-file->html "docs/api.md")
     :umm-c collection-umm-version
     :umm-g granule-umm-version
     :umm-s service-umm-version
     :umm-t tool-umm-version
     :umm-sub subscription-umm-version
     :umm-var variable-umm-version}))))

ingest-app/src/cmr/ingest/site/static.clj
```diff
@@ -2,21 +2,38 @@
   "The functions of this namespace are specifically responsible for returning
   ready-to-serve pages."
   (:require
+   [clojure.string :as string]
```
mschmele, line 5: It looks like `string` isn't being used any more, so we can remove it

ingest-app/docs/api.md
```diff
@@ -280,6 +284,13 @@ A native-id is an identifier, unique per provider, used to identify concepts wit
 
 ## <a name="metadata-ingest"></a> Metadata Ingest
 
+### <a name="latest-umm-versions"></a> Latest UMM Schema Versions
+
+The following are the latest acceptable UMM schema versions for metadata ingest:
+
+<p>UMM-C: {{ umm-c }}<br/>UMM-G: {{ umm-g }}<br/>UMM-S: {{ umm-s }}<br/>UMM-T: {{ umm-t }}<br/>UMM-SUB: {{ umm-sub }}<br/>UMM-VAR: {{ umm-var }}</p>
```
mschmele, line 291: Nitpick: we could just do this an an unordered list:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ul
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/li
```suggestion
<ul class="umm-versions">
  <li>UMM-C: {{ umm-c }}</li>
  <li>UMM-G: {{ umm-g }}</li>
  <li>UMM-S: {{ umm-s }}</li>
  <li>UMM-T: {{ umm-t }}</li>
  <li>UMM-SUB: {{ umm-sub }}</li>
  <li>UMM-VAR: {{ umm-var }}</li>
</ul>
```

mschmele (approved):

PR #1222, opened 2021-04-13
CMR-7272: add CSV formatting support to collection search results

mschmele:


system-int-test/test/cmr/system_int_test/search/collection_search_format_test.clj
```diff
@@ -624,14 +624,20 @@
...
           (dk/assert-collection-kml-results-match [coll1] results))))
 
-    (testing "csv is not supported"
-      (is (= {:errors ["The mime type [text/csv] is not supported for collections."],
-              :status 400}
-             (search/find-concepts-csv :collection {})))
+    (testing "csv"
+      (let [response (search/find-concepts-csv :collection {})]
+        (is (= (:status response) 200))
+        (is (string/starts-with?
```
mschmele, line 630: I found that the search results here were changing between test runs, hence `starts-with`

search-app/src/cmr/search/results_handlers/csv_results_handler.clj
```diff
@@ -1,16 +1,58 @@
...
+  [context query elastic-result]
+  (let [{start-date :start-date
+         provider-id :provider-id
+         processing-level :processing-level-id
+         entry-title :entry-title
+         platforms :platform
+         version :version-id
+         short-name :collection-short-name} (:_source elastic-result)
+        start-date (when start-date (str/replace (str start-date) #"\+0000" "Z"))]
+    (merge {:row [provider-id short-name version entry-title processing-level (first platforms) start-date]}
```
mschmele, line 42: `platforms` are to my knowledge the only thing coming back from a CSV result set that are a sequence. Should I take the first item like I already am, or combine the values somehow? We're just expecting the short name

system-int-test/test/cmr/system_int_test/search/collection_search_format_test.clj
ygliuvt: We have to check on the granule csv values in the test. Try add sort parameter to make the granules come back in a predictable order, so that we can verify the values.

search-app/src/cmr/search/results_handlers/csv_results_handler.clj
ygliuvt: We should return the platforms as comma separated string when there are multiple. Refer to granule `Online Access URLs` and `Browse URLs` for examples.

search-app/src/cmr/search/results_handlers/csv_results_handler.clj
```diff
@@ -1,16 +1,58 @@
...
     [java.io StringWriter]))
 
-(def CSV_HEADER
+(def COLLECTION_CSV_HEADER
+  ["Data Provider"
+   "Short Name"
+   "Version"
+   "Entry Title"
+   "Processing Level"
+   "Platform"
```
ygliuvt, line 21: We should change this to `Platforms`.

ygliuvt (changes requested):
Need to change Platforms values in CSV response, update test and API doc.

search-app/docs/api.md
```diff
@@ -458,9 +458,15 @@ __Example__
...
 #### <a name="csv"></a> CSV
 
-The comma separated value (CSV) format is only supported for granules.
+The comma separated value (CSV) format is supported for collections and granules.
 
-__Example__
+__Examples__
+
+```csv
+Data Provider,Short Name,Version,Entry Title,Processing Level,Platforms,Start Time
```
ygliuvt, line 466: nit: make it clear that this is the collection csv response, and the other one is the granule csv response.

search-app/docs/api.md
```diff
@@ -458,9 +458,15 @@ __Example__
...
+PROV1,short,v2,dataset-id,2,platform-1,2015-06-05T20:10:39.166Z
+PROV2,very-short,v1,data-id,1,,2019-08-03T20:11:59.143Z
```
ygliuvt, line 468: Platform is a required field in UMM-C. We will never have a case when there is no platform for a collection. We should fix this example.

system-int-test/test/cmr/system_int_test/search/collection_search_format_test.clj
```diff
@@ -695,6 +687,41 @@
...
+                                    :beginning-date-time "2000-01-01T12:00:00Z"
+                                    :platforms [{:short-name "platform #21"
+                                                 :long-name "platform #21"
+                                                 :type "good platform"}
+                                                {:short-name "platform #32"
+                                                 :long-name "platform #32"
+                                                 :type "very good platform"}]}))
+  (index/wait-until-indexed)
+
+  (let [response (search/find-concepts-csv :collection {:sort-key "provider"})]
```
ygliuvt, line 714: nit: with the fields defined, you don't need to explicitly sort by provider now.

ygliuvt (approved):
Approved assuming some minor comments will be addressed.

jaybarra (approved):

system-int-test/test/cmr/system_int_test/search/collection_search_format_test.clj
```diff
@@ -695,6 +687,42 @@
...
+                                              :platforms [{:short-name "platform #21"
+                                                           :long-name "platform #21"
+                                                           :type "good platform"}
+                                                          {:short-name "platform #32"
+                                                           :long-name "platform #32"
+                                                           :type "very good platform"}]}))
+        g1 (d/ingest "PROV1" (dg/granule c1))])
+  (index/wait-until-indexed)
+
+  (let [response (search/find-concepts-csv :collection {:include-granule-counts "true"})]
```
ygliuvt, line 715: We have `granule-count` always as a field in CSV response. Its value should always be included in the result rather than via a different flag.

ygliuvt, line 715: remove `include-granule-counts` option.

system-int-test/test/cmr/system_int_test/search/collection_search_format_test.clj
```diff
@@ -695,6 +687,42 @@
...
+    (is (= 200 (:status response)))
+    (is (= (str "Data Provider,Short Name,Version,Entry Title,Processing Level,Platforms,Start Time,End Time\n"
+                "PROV1,,V1,I am #1,L1,platform #1,1970-01-01T12:00:00.000Z,\nPROV2,,V2,lorem ipsum,L2,\"platform #21,platform #32\",2000-01-01T12:00:00.000Z,\n")
```
ygliuvt, line 718: Looks like the `Short Name` is not exported correctly in the csv result.

ygliuvt (changes requested):
Need to fix the `Short Name` value in csv result.

system-int-test/test/cmr/system_int_test/search/collection_search_format_test.clj
mschmele: Fixed

search-app/src/cmr/search/results_handlers/csv_results_handler.clj
```diff
@@ -1,16 +1,83 @@
...
+                  (str version)
+                  entry-title
+                  (str processing-level)
+                  platform-short-names
+                  start-date
+                  end-date]
+            :id collection-id}
+           (acl-rhh/parse-elastic-item :collection elastic-result))))
+
+(defmethod gcrf/query-results->concept-ids :csv
```
ygliuvt, line 64: I don't think this function is used. We should remove it.

ygliuvt: hmm. On second thought, this function is needed, but the implementation should be raising an error saying `include_granule_counts` is not supported for csv format. We should have a test for `include_granule_counts` in csv format.

ygliuvt: The standard error message currently used by CMR for other formats is: 
e.g. KML:
`Collections search in kml format is not supported with include_granule_counts option`

ygliuvt (approved):

PR #1224, opened 2021-04-19
CMR-7203 partitioning granule update messages into smaller chunks to support very large requests

jaybarra:
This is a draft to review the basic strategy of handling very large bulk granule update requests
Things to consider for large size processing updates

* size of request that can be converted and held in memory in JSON->data
* size of message published to queue
* will anything age off the queue before it gets to processing in the case of very large queue sizes
* is the intermediary partitioning step necessary

### DIFF
--- a/ingest-app/src/cmr/ingest/services/granule_bulk_update_service.clj
+++ b/ingest-app/src/cmr/ingest/services/granule_bulk_update_service.clj
@@ -64,7 +64,7 @@
 (defn- validate-granule-bulk-update-no-duplicate-urs
   "Validate no duplicate URs exist within the list of instructions"
   [urs]
-  (when-let [duplicate-urs (duplicates urs)]
+  (when-let [duplicate-urs (seq (duplicates urs))]
     (errors/throw-service-errors
      :bad-request
      [(format (str "Duplicate granule URs are not allowed in bulk update requests. "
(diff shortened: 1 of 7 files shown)


ingest-app/src/cmr/ingest/services/granule_bulk_update_service.clj
```diff
@@ -97,10 +125,15 @@
...
+
+    ;; Queue the granules bulk update events
+    (publish-instructions-partitioned
      context
-     (ingest-events/granules-bulk-event provider-id task-id user-id instructions))
+     provider-id
+     user-id
+     task-id
+     instructions
+     (granule-bulk-update-chunk-size))
```
jaybarra, line 136: The previous limiter seemed to be the message size. Partition the message into smaller bites to process

ygliuvt: I don't think parsing out the original request and queue the granule events in chunks is necessary.

ygliuvt: I like the way you queue the granule bulk update events.

system-int-test/test/cmr/system_int_test/ingest/granule_bulk_update/granule_bulk_update_volume_test.clj
```diff
@@ -0,0 +1,72 @@
...
+                   :update-field "OPeNDAPLink"
+                   :updates []})
+
+(defn update-instruction
+  "Generate an update instruction."
+  [ur]
+  [ur (str "https://file.example.nasa.gov/" ur)])
+
+;; This test is potentially unstable on systems with smaller amounts of
+;; memory and can take multiple minutes to complete. Test with caution
```
ygliuvt, line 29: There is no need to test with `501` granules. Testing with `granule-bulk-update-chunk-size` of 2 and 3 granules is sufficient.

ygliuvt (changes requested):
I like the way you queue the granule bulk update events. The only risk is would this queuing of chunked requests take too long for the original request to return to client in a timely manner. This is something we will have to find out in WL testing.
Request change due to testing with large requests unnecessarily. We shouldn't put real large requests in our integration test. This is a source of long and unstable build. We should reserve WL for these kind of tests.

daniel-zamora (approved):

PR #1227, opened 2021-04-23
Cmr 7250 index wrong gcmd science keywords

sxu123:
Clarification:
1. This change applies to iso19115 only. isosmap shouldn't be impacted.
2. ScienceKeywords for iso19115 should be the descriptiveKeywords with "theme" keyword type (current behavior)
 AND with gmd:thesaurusName's CharacterString value including "GCMD".  (new feature)  This is because NOAA uses the same type of theme for their Science keywords, as well as their WMO keywords.
3. We should get the values from both gco:CharacterString(current behavior) and gmx:Anchor (new feature)

### DIFF
--- a/umm-spec-lib/src/cmr/umm_spec/xml_to_umm_mappings/iso_smap.clj
+++ b/umm-spec-lib/src/cmr/umm_spec/xml_to_umm_mappings/iso_smap.clj
@@ -119,6 +119,8 @@
   Category of each theme descriptive keyword to determine if it is a science keyword."
   [data-id-el sanitize?]
   (if-let [science-keywords (seq
+                              ;; kws/parse-science-keywords is shared by iso19115 and isosmap
+                              ;; "true" indicates it's isosmap case.
                               (->> (kws/parse-science-keywords data-id-el sanitize? true)
                                    (filter #(.contains kws/science-keyword-categories (:Category %)))))]
     science-keywords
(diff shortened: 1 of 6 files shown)


umm-spec-lib/src/cmr/umm_spec/xml_to_umm_mappings/iso_smap.clj
```diff
@@ -119,7 +119,7 @@
   Category of each theme descriptive keyword to determine if it is a science keyword."
   [data-id-el sanitize?]
   (if-let [science-keywords (seq
-                              (->> (kws/parse-science-keywords data-id-el sanitize?)
+                              (->> (kws/parse-science-keywords data-id-el sanitize? true)
```
eereiter, line 122: It would be nice here if you defined a variable like smap? as true in this function and sent that instead of true, or put in a comment describing what true is.  I had to go to the parse-science-keywords function to figure out what true meant.

eereiter (approved):
Looks great!  Please fix the one comment that I have left and then I approve.

ygliuvt (approved):

umm-spec-lib/src/cmr/umm_spec/xml_to_umm_mappings/iso_smap.clj
sxu123: Added comment

PR #1229, opened 2021-04-23
CMR-7214: Add support for additional keyword searching

mitchstartzel:
This solves an issue where keywords such as `(TMPA-RT)` were decomposed into `["(TMPA-RT)", "tmpa", "rt"]` which prevents keyword searching for `TMPA-RT` since elasticsearch doesn't return inexact matches. To solve this, we add all existing keywords like before, but also as the contents of parens or brackets as well, if applicable - for example, in the above case, the new set of keywords would be `["(TMPA-RT)", "tmpa-rt", "tmpa", "rt"]`.

### DIFF
--- a/indexer-app/src/cmr/indexer/data/concepts/keyword_util.clj
+++ b/indexer-app/src/cmr/indexer/data/concepts/keyword_util.clj
@@ -30,8 +30,8 @@
              (= \] last-char)))))
 
 (defn- prepare-keyword-field
-  [field-value]
   "Convert a string to lowercase then separate it into keywords"
+  [field-value]
   (when field-value
     (let [field-value (string/lower-case field-value)]
       (if (wrapped-keyword? field-value)
--- a/indexer-app/src/cmr/indexer/data/concepts/keyword_util.clj
+++ b/indexer-app/src/cmr/indexer/data/concepts/keyword_util.clj
@@ -20,8 +20,8 @@
   "Checks if the field-value about to be processed is a term surrounded by parens or brackets.
    If it is, we want to add the usual keywords, but also add the unwrapped field-value on its own."
   [field-value]
-  (let [first-char (first field-value)]
-       [last-char (last field-value)]
+  (let [first-char (first field-value)
+        last-char (last field-value)]
     (and (or (= \( first-char)
              (= \{ first-char)
              (= \[ first-char))
--- a/indexer-app/src/cmr/indexer/data/concepts/keyword_util.clj
+++ b/indexer-app/src/cmr/indexer/data/concepts/keyword_util.clj
@@ -16,12 +16,28 @@
   "Defines Regex to split strings with special characters into multiple words for keyword searches."
   #"[!@#$%^&()\-=_+{}\[\]|;'.,\\\"/:<>?`~* ]")
 
+(defn- wrapped-keyword?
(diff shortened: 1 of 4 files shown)


indexer-app/test/cmr/indexer/test/data/concepts/keyword_util.clj
```diff
@@ -453,18 +453,19 @@
...
+                "analysis and visualization data center contact datacenterurl description "
+                "documentation doe earth earth science services edg engineering eosdis eros esip "
+                "example general general documentation geographic geographic information systems "
+                "get get service health home home page imagery imaging information infrared "
+                "institute iris iris/passcal john lp lp daac lpdaac marsh microwave microwave "
+                "imagery of page passcal publicationurl radar radiometer related related-url "
+                "description. related-url-example-two.com related-url-example.com rt science "
+                "science cat 3 science contact science term 3 science topic 3 service services "
+                "spectral spectral/engineering spock suite surface surface winds systems team team "
+                "spock technical technical contact term test tmpa tmpa-rt topic two url usgs "
```
mitchstartzel, line 467: see the new term on this line - `tmpa-rt`

indexer-app/src/cmr/indexer/data/concepts/keyword_util.clj
```diff
@@ -16,12 +16,28 @@
   "Defines Regex to split strings with special characters into multiple words for keyword searches."
   #"[!@#$%^&()\-=_+{}\[\]|;'.,\\\"/:<>?`~* ]")
 
+(defn- wrapped-keyword?
+  "Checks if the field-value about to be processed is a term surrounded by parens or brackets.
+   If it is, we want to add the usual keywords, but also add the unwrapped field-value on its own."
+  [field-value]
+  (let [first-char (first field-value)]
```
ygliuvt, line 23: This is a syntax error.

indexer-app/src/cmr/indexer/data/concepts/keyword_util.clj
```diff
@@ -16,12 +16,28 @@
...
+    (and (or (= \( first-char)
+             (= \{ first-char)
+             (= \[ first-char))
+         (or (= \) last-char)
+             (= \} last-char)
+             (= \] last-char)))))
+
 (defn- prepare-keyword-field
   [field-value]
   "Convert a string to lowercase then separate it into keywords"
```
ygliuvt, line 34: Move the doc-string up.

ygliuvt (changes requested):
Need to fix the syntax error and I don't think the fix will handle the case "Mitch made a (merry-go-round)."

indexer-app/src/cmr/indexer/data/concepts/keyword_util.clj
mitchstartzel: That's embarassing. I'll blame it on it being a Friday afternoon

mitchstartzel:
> Need to fix the syntax error and I don't think the fix will handle the case "Mitch made a (merry-go-round)."

This case is handled, we add whatever is inside the parens or brackets exactly, without splitting it up. I have added such a term to the tests (`(TMPA-RT-MULTI-TERM)`) to verify this.

ygliuvt (approved):

jaybarra (approved):

PR #1238, opened 2021-04-29
CMR-7203 bulk granule update scalability updates

jaybarra:
Removed bottlenecks to processing by moving bulk granule task status updates to polling job instead of after ever granule update operation.
The new update job runs every 300 seconds by default and is configurable by environmental configuration. The job queries the ingest database for a list of tasks not marked as complete, and publishes a message for each to run an update on the job.

The logic for updating the status was relocated from the granule update code to a new function to be triggered independently of the granule updates. 

A new status update endpoint was created to allow for an unscheduled update to be executed. This endpoint checks for ingest-management permissions.

### DIFF
--- a/ingest-app/docs/api.md
+++ b/ingest-app/docs/api.md
@@ -46,6 +46,10 @@ Join the [CMR Client Developer Forum](https://wiki.earthdata.nasa.gov/display/CM
     * [POST - Collection bulk update](#collection-bulk-update)
   * /providers/\<provider-id\>/bulk-update/granules
     * [POST - Granule bulk update](#granule-bulk-update)
+  * /granule-bulk-update/status
+    * [POST - Granule bulk update](#granule-bulk-update)
+  * /granule-bulk-update/status/\<task-id\>
+    * [GET - Granule bulk update status](#granule-bulk-update)
 
 ***
 
@@ -1209,3 +1213,11 @@ curl -i -H "Echo-Token: XXXX" -H "Cmr-Pretty:true" %CMR-ENDPOINT%/granule-bulk-u
 }
 ```
 Granule bulk update tasks and statuses are available for 90 days.
+
+### Refresh Granule Bulk Update Status
+Granule bulk update task statuses can be refreshed manually, provided the user has the appropriate permissions, with the following command.
+
+```
+curl -XPOST -i -H "Echo-Token: XXXX" %CMR-ENDPOINT%/granule-bulk-update/status
+```
+
--- a/ingest-app/docs/api.md
+++ b/ingest-app/docs/api.md
@@ -1215,7 +1215,8 @@ curl -i -H "Echo-Token: XXXX" -H "Cmr-Pretty:true" %CMR-ENDPOINT%/granule-bulk-u
 Granule bulk update tasks and statuses are available for 90 days.
 
 ### Refresh Granule Bulk Update Status
-Granule bulk update task statuses can be refreshed manually, provided the user has the appropriate permissions, with the following command.
+
+By default the bulk granule update jobs are checked for completion every 5 minutes. However granule bulk update task statuses can be refreshed manually, provided the user has the ingest-management permission, with the following command.
 
 ```
 curl -XPOST -i -H "Echo-Token: XXXX" %CMR-ENDPOINT%/granule-bulk-update/status
--- a/ingest-app/docs/api.md
+++ b/ingest-app/docs/api.md
@@ -42,9 +42,9 @@ Join the [CMR Client Developer Forum](https://wiki.earthdata.nasa.gov/display/CM
     * [POST - Translate collection metadata.](#translate-collection)
   * /translate/granule
     * [POST - Translate granule metadata.](#translate-granule)
-  * /providers/<provider-id>/bulk-update/collections
+  * /providers/\<provider-id\>/bulk-update/collections
     * [POST - Collection bulk update](#collection-bulk-update)
-  * /providers/<provider-id>/bulk-update/granules
+  * /providers/\<provider-id\>/bulk-update/granules
     * [POST - Granule bulk update](#granule-bulk-update)
 
 ***
--- a/ingest-app/src/cmr/ingest/data/granule_bulk_update.clj
+++ b/ingest-app/src/cmr/ingest/data/granule_bulk_update.clj
@@ -84,145 +84,130 @@
   cmr.oracle.connection.OracleStore
 
   (get-granule-tasks-by-provider-id
-   [db provider-id]
-   (jdbc/with-db-transaction
-    [conn db]
-    ;; Returns a list of bulk update tasks for the provider
-    (let [stmt (sql-utils/build
-                (sql-utils/select
-                 [:created-at :name :task-id :status :status-message :request-json-body]
-                 (sql-utils/from "granule_bulk_update_tasks")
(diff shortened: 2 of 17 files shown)


ingest-app/docs/api.md
```diff
@@ -42,9 +42,9 @@ Join the [CMR Client Developer Forum](https://wiki.earthdata.nasa.gov/display/CM
     * [POST - Translate collection metadata.](#translate-collection)
   * /translate/granule
     * [POST - Translate granule metadata.](#translate-granule)
-  * /providers/<provider-id>/bulk-update/collections
```
jaybarra, line 45: Fixes rendering issue on ingest api docs page

mitchstartzel (approved):
Only pressing issue I see is that we still aren't checking that the message we're putting on the queue might be too large (please correct me if we are), but that might make more sense to put in new ticket.

mschmele (approved):

ingest-app/src/cmr/ingest/data/granule_bulk_update.clj
```diff
@@ -254,6 +239,77 @@
...
+            [:granule-ur :status]
+            (sql-utils/from "bulk_update_gran_status")
+            (sql-utils/where `(and (= :status "PENDING")
+                                   (= :task-id ~task-id))))))))
+
+(defn-timed mark-task-complete
+  "Marks a granule bulk task as COMPLETE and sets the status message.
+  It will throw an exception if there still granules marked as PENDING."
+  [context task-id]
+  (validate-task-exists context task-id)
```
ygliuvt, line 289: No need to validate if task exists again here.

ingest-app/src/cmr/ingest/services/jobs.clj
```diff
@@ -216,6 +221,14 @@
        context
        (ingest-events/provider-autocomplete-suggestion-reindexing-event provider)))))
 
+(defn trigger-bulk-granule-task-status-updates
+  [context]
+  (when-let [incomplete-tasks (seq (granule-bulk-update/get-incomplete-granule-task-ids context))]
+    (doseq [task-id incomplete-tasks]
+      (ingest-events/publish-gran-bulk-update-event
```
ygliuvt, line 228: There is no need to update task status via a message queue event. We should just call the function that does the work here.

ygliuvt (changes requested):
Need to change the update task status job to just call the process function directly rather than going through message queue events.

ingest-app/src/cmr/ingest/services/granule_bulk_update_service.clj
```diff
@@ -26,15 +27,20 @@
    [cmr.umm-spec.versioning :as versioning]))
 
 (def granule-bulk-update-schema
- (js/json-string->json-schema (slurp (io/resource "granule_bulk_update_schema.json"))))
+  (js/json-string->json-schema (slurp (io/resource "granule_bulk_update_schema.json"))))
+
+(defconfig granule-bulk-update-chunk-size
+  "Default size to partition granule-bulk-update instructions into."
+  {:default 500
```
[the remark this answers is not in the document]
jaybarra: Yes, in the volume test, slightly a misnomer now, the value is dropped to `2`

ingest-app/docs/api.md
```diff
@@ -1209,3 +1213,11 @@ curl -i -H "Echo-Token: XXXX" -H "Cmr-Pretty:true" %CMR-ENDPOINT%/granule-bulk-u
 }
 ```
 Granule bulk update tasks and statuses are available for 90 days.
+
+### Refresh Granule Bulk Update Status
+Granule bulk update task statuses can be refreshed manually, provided the user has the appropriate permissions, with the following command.
```
ygliuvt, line 1218: Add more description on exactly what this endpoint does and specify the exact permission needed to perform this task.

ingest-app/src/cmr/ingest/services/event_handler.clj
```diff
@@ -1,9 +1,10 @@
 (ns cmr.ingest.services.event-handler
- (:require
-  [cmr.ingest.config :as config]
-  [cmr.ingest.services.bulk-update-service :as bulk-update]
-  [cmr.ingest.services.granule-bulk-update-service :as granule-bulk-update-service]
-  [cmr.message-queue.queue.queue-protocol :as queue-protocol]))
+  (:require
+   [cmr.common.log :refer (debug info warn error)]
```
ygliuvt, line 3: Remove this line.

ingest-app/src/cmr/ingest/services/granule_bulk_update_service.clj
```diff
@@ -259,8 +307,21 @@
...
-     (errors/throw-service-error
-      :error
-      [(format "Exception caught while attempting to cleanup granule bulk update tasks: %s" e)]))))
+    (data-granule-bulk-update/cleanup-bulk-granule-tasks context)
+    (catch Exception e
+      (errors/throw-service-error
+       :error
+       [(format "Exception caught while attempting to cleanup granule bulk update tasks: %s" e)]))))
+
+(defn update-bulk-granule-task-status
```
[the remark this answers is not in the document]
ygliuvt: Also the function name should be changed to something more reasonable if it is only used to update task status to complete.

jaybarra:
> Only pressing issue I see is that we still aren't checking that the message we're putting on the queue might be too large (please correct me if we are), but that might make more sense to put in new ticket.

I reduced the default chunk size to 100 and ran that in the test environment

ingest-app/src/cmr/ingest/services/jobs.clj
```diff
@@ -216,6 +221,14 @@
        context
        (ingest-events/provider-autocomplete-suggestion-reindexing-event provider)))))
 
+(defn update-completed-bulk-granule-task-statuses
```
ygliuvt, line 224: This function is exactly the same as what is in `update-completed-granule-task-statuses` function in the api bulk.clj file. Move this function to `granule-bulk-update-service`, and call it from this namespace and the api bulk namespace.

system-int-test/test/cmr/system_int_test/ingest/granule_bulk_update/granule_bulk_update_volume_test.clj
```diff
@@ -0,0 +1,68 @@
...
+
+       (are3 [urs]
+             (let [request (->> urs
+                                (map update-instruction)
+                                (assoc base-request :updates))
+                   {:keys [body status]} (ingest/bulk-update-granules
+                                          "PROV1"
+                                          request
+                                          bulk-update-options)
+                   response (json/parse-string body true)]
```
ygliuvt, line 59: `response` is not used in the test, we can delete this line.
Also this test is not really complete, it just verified the bulk ingest request is submitted successfully. There is no verification that the granules are actually updated correctly.

ygliuvt: In other words, this test is not testing the queuing of chunks of granules actually works.

jaybarra: That verification does happen in other bulk tests, but I will expand these as well

ygliuvt (approved):

PR #1241, opened 2021-05-10
CMR-7386

ygliuvt:
As a user I want the collection JSON endpoint to return platforms

### DIFF
--- a/system-int-test/src/cmr/system_int_test/data2/atom.clj
+++ b/system-int-test/src/cmr/system_int_test/data2/atom.clj
@@ -280,7 +280,8 @@
   [collection]
   (let [{{:keys [short-name version-id processing-level-id collection-data-type]} :product
          :keys [concept-id format-key has-variables has-formats has-transforms
-                has-spatial-subsetting has-temporal-subsetting services variables tools]} collection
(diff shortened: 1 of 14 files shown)


system-int-test/src/cmr/system_int_test/data2/atom.clj
```diff
@@ -280,7 +280,8 @@
   [collection]
   (let [{{:keys [short-name version-id processing-level-id collection-data-type]} :product
          :keys [concept-id format-key has-variables has-formats has-transforms
-                has-spatial-subsetting has-temporal-subsetting services variables tools]} collection
+                has-spatial-subsetting has-temporal-subsetting
+                services variables tools platforms]} collection
```
jaybarra, line 284: This `platforms` is not referenced

jaybarra (changes requested):

mschmele (approved):

system-int-test/src/cmr/system_int_test/data2/atom.clj
ygliuvt: removed.

jaybarra (approved):

PR #1243, opened 2021-05-11
CMR-7356 + CMR-7355 + CMR-7251

mitchstartzel:
CMR-7356: Reject subscription ingests with invalid query parameters
CMR-7355: Subscription documentation revisions/cleanup
CMR-7251: Improved subscription processing job logging

### DIFF
--- a/common-app-lib/src/cmr/common_app/services/ingest/subscription_common.clj
+++ b/common-app-lib/src/cmr/common_app/services/ingest/subscription_common.clj
@@ -18,3 +18,13 @@
         (as-> $ (string/join "&" $))
         string/trim
         digest/md5)))
+
+(defn create-query-params
+  "Create query parameters using the query string like
+  \"polygon=1,2,3&concept-id=G1-PROV1\""
+  [query-string]
+  (let [query-string-list (string/split query-string #"&")
+        query-map-list (map #(let [a (string/split % #"=")]
+                               {(first a) (second a)})
+                             query-string-list)]
+     (apply merge query-map-list)))
--- a/common-app-lib/src/cmr/common_app/services/ingest/subscription_common.clj
+++ b/common-app-lib/src/cmr/common_app/services/ingest/subscription_common.clj
@@ -18,14 +18,3 @@
         (as-> $ (string/join "&" $))
         string/trim
         digest/md5)))
-
-(defn create-query-params
-  "Create query parameters using the query string like
-  \"polygon=1,2,3&concept-id=G1-PROV1\""
-  [query-string]
-  (when query-string
-    (let [query-string-list (string/split query-string #"&")
-          query-map-list (map #(let [a (string/split % #"=")]
-                                 {(first a) (second a)})
-                               query-string-list)]
-       (apply merge query-map-list))))
--- a/common-app-lib/src/cmr/common_app/services/ingest/subscription_common.clj
+++ b/common-app-lib/src/cmr/common_app/services/ingest/subscription_common.clj
@@ -23,8 +23,9 @@
   "Create query parameters using the query string like
   \"polygon=1,2,3&concept-id=G1-PROV1\""
   [query-string]
-  (let [query-string-list (string/split query-string #"&")
-        query-map-list (map #(let [a (string/split % #"=")]
-                               {(first a) (second a)})
-                             query-string-list)]
-     (apply merge query-map-list)))
+  (when query-string
+    (let [query-string-list (string/split query-string #"&")
+          query-map-list (map #(let [a (string/split % #"=")]
+                                 {(first a) (second a)})
+                               query-string-list)]
+       (apply merge query-map-list))))
--- a/common-app-lib/test/cmr/common_app/test/services/ingest/subscription.clj
+++ b/common-app-lib/test/cmr/common_app/test/services/ingest/subscription.clj
@@ -24,4 +24,4 @@
 (deftest create-query-params
   (is (= {"polygon" "-78,-18,-77,-22,-73,-16,-74,-13,-78,-18"
           "concept-id" "G123-PROV1"}
-         (#'sub-common/create-query-params "polygon=-78,-18,-77,-22,-73,-16,-74,-13,-78,-18&concept-id=G123-PROV1"))))
+         (sub-common/create-query-params "polygon=-78,-18,-77,-22,-73,-16,-74,-13,-78,-18&concept-id=G123-PROV1"))))
--- a/common-app-lib/test/cmr/common_app/test/services/ingest/subscription.clj
+++ b/common-app-lib/test/cmr/common_app/test/services/ingest/subscription.clj
@@ -20,8 +20,3 @@
          expected "d41d8cd98f00b204e9800998ecf8427e"
          actual (sub-common/normalize-parameters query)]
      (is (= expected actual)))))
-
-(deftest create-query-params
-  (is (= {"polygon" "-78,-18,-77,-22,-73,-16,-74,-13,-78,-18"
-          "concept-id" "G123-PROV1"}
-         (sub-common/create-query-params "polygon=-78,-18,-77,-22,-73,-16,-74,-13,-78,-18&concept-id=G123-PROV1"))))
--- a/common-app-lib/test/cmr/common_app/test/services/ingest/subscription.clj
+++ b/common-app-lib/test/cmr/common_app/test/services/ingest/subscription.clj
@@ -20,3 +20,8 @@
          expected "d41d8cd98f00b204e9800998ecf8427e"
          actual (sub-common/normalize-parameters query)]
      (is (= expected actual)))))
+
+(deftest create-query-params
+  (is (= {"polygon" "-78,-18,-77,-22,-73,-16,-74,-13,-78,-18"
+          "concept-id" "G123-PROV1"}
+         (#'sub-common/create-query-params "polygon=-78,-18,-77,-22,-73,-16,-74,-13,-78,-18&concept-id=G123-PROV1"))))
--- a/ingest-app/src/cmr/ingest/api/subscriptions.clj
+++ b/ingest-app/src/cmr/ingest/api/subscriptions.clj
@@ -66,12 +66,10 @@
   "Performs a granule search using the provided query parameters. If the query is no good,
    we throw a service error."
   [context params]
-  (try
-    (search/find-granule-references context params)
-    (catch Exception e
-      (errors/throw-service-error :bad-request
-       (str "Error ingesting subscription: invalid query parameters [" (dissoc params :token) "]\n\n"
-              (.getMessage e))))))
+  (if-let [errors (search/find-granule-references-return-errors context params)]
+    (errors/throw-service-error :bad-request
+      (str "Error ingesting subscription with query parameters [" (dissoc params :token) "]: " errors))))
+
 
 (defn- validate-query
   "Performs a granule search using subscription query parameters for purposes of validation"
@@ -79,9 +77,8 @@
   (let [metadata (-> (:metadata subscription) (json/decode true))
         collection-id (:CollectionConceptId metadata)
         query-string (:Query metadata)
-        query-params (sub-common/create-query-params query-string)
+        query-params (jobs/create-query-params query-string)
         search-params (merge {:collection-concept-id collection-id
-                              :page-size 0
                               :token (config/echo-system-token)}
                              query-params)
         gran-refs (search-gran-refs-with-sub-params request-context search-params)]))
--- a/ingest-app/src/cmr/ingest/api/subscriptions.clj
+++ b/ingest-app/src/cmr/ingest/api/subscriptions.clj
@@ -68,7 +68,7 @@
   [context params]
   (when-let [errors (search/find-granule-references-return-errors context params)]
     (errors/throw-service-error :bad-request
-      (str "Subscription query validation failed with the following error(s): " (:body errors)))))
+      (str "Subscription query validation failed with the following error(s): " errors))))
 
 (defn- validate-query
   "Performs a granule search using subscription query parameters for purposes of validation"
--- a/ingest-app/src/cmr/ingest/api/subscriptions.clj
+++ b/ingest-app/src/cmr/ingest/api/subscriptions.clj
@@ -68,8 +68,7 @@
     (search/find-granule-references context params)
     (catch Exception e
       (errors/throw-service-error :bad-request
-       (str "Error ingesting subscription: invalid query parameters ["
-            (dissoc params :token) "]\n\n" (.getMessage e))))))
+       (str "Error ingesting subscription: invalid query parameters [" (dissoc params :token) "]")))))
 
 (defn- validate-query
   "Checks that the subscription query returns a valid response"
--- a/ingest-app/src/cmr/ingest/api/subscriptions.clj
+++ b/ingest-app/src/cmr/ingest/api/subscriptions.clj
@@ -70,7 +70,8 @@
     (search/find-granule-references context params)
     (catch Exception e
       (errors/throw-service-error :bad-request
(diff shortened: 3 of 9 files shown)


jaybarra (changes requested):

common-app-lib/test/cmr/common_app/test/services/ingest/subscription.clj
```diff
@@ -20,3 +20,8 @@
          expected "d41d8cd98f00b204e9800998ecf8427e"
          actual (sub-common/normalize-parameters query)]
      (is (= expected actual)))))
+
+(deftest create-query-params
+  (is (= {"polygon" "-78,-18,-77,-22,-73,-16,-74,-13,-78,-18"
+          "concept-id" "G123-PROV1"}
+         (#'sub-common/create-query-params "polygon=-78,-18,-77,-22,-73,-16,-74,-13,-78,-18&concept-id=G123-PROV1"))))
```
ygliuvt, line 27: `create-query-params` is public now, you can remove the `#'`

ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -224,12 +248,13 @@
                             content-type
                             headers)
           native-id (get-unique-native-id request-context tmp-subscription)
-          new-subscription (assoc  tmp-subscription :native-id native-id)
-          newer-subscription (add-fields-if-missing request-context new-subscription)
-          subscriber-id (get-subscriber-id newer-subscription)]
+          sub-with-native-id (assoc  tmp-subscription :native-id native-id)
+          final-sub (add-fields-if-missing request-context sub-with-native-id)
+          subscriber-id (get-subscriber-id final-sub)]
+      (validate-query request-context final-sub)
```
ygliuvt, line 254: Move the `validate-query` call below `check-ingest-permission` call.

ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -224,12 +248,13 @@
                             content-type
                             headers)
           native-id (get-unique-native-id request-context tmp-subscription)
-          new-subscription (assoc  tmp-subscription :native-id native-id)
-          newer-subscription (add-fields-if-missing request-context new-subscription)
-          subscriber-id (get-subscriber-id newer-subscription)]
+          sub-with-native-id (assoc  tmp-subscription :native-id native-id)
```
ygliuvt, line 251: remove the extra space before `tmp-subscription`.

ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -249,11 +274,12 @@
                             body
                             content-type
                             headers)
-          new-subscription (add-fields-if-missing request-context tmp-subscription)
-          subscriber-id (get-subscriber-id new-subscription)]
+          final-sub (add-fields-if-missing request-context tmp-subscription)
+          subscriber-id (get-subscriber-id final-sub)]
+      (validate-query request-context final-sub)
```
ygliuvt, line 279: same as above. move this call down.

ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -60,6 +62,28 @@
...
            concept-id)))))
 
+(defn- search-gran-refs-with-sub-params
+  "Performs a granule search using the provided query parameters. If the query is no good,
+   we throw a service error."
+  [context params]
+  (try
+    (search/find-granule-references context params)
+    (catch Exception e
+      (errors/throw-service-error :bad-request
```
ygliuvt, line 72: The error should contain the specific error message returned from the search-app.

ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -60,6 +62,28 @@
...
+       (str "Error ingesting subscription: invalid query parameters [" (dissoc params :token) "]")))))
+
+(defn- validate-query
+  "Performs a granule search using subscription query parameters for purposes of validation"
+  [request-context subscription]
+  (let [metadata (-> (:metadata subscription) (json/decode true))
+        collection-id (:CollectionConceptId metadata)
+        query-string (:Query metadata)
+        query-params (sub-common/create-query-params query-string)
+        search-params (merge {:collection-concept-id collection-id
```
ygliuvt, line 82: add a `:page-size 0` to limit the request size.

ygliuvt (changes requested):

ingest-app/src/cmr/ingest/api/subscriptions.clj
mitchstartzel: Now this is smart.

ingest-app/src/cmr/ingest/api/subscriptions.clj
mitchstartzel: I like this change - my original thought process was that if the query is bad, it's not worth wasting time checking anything else, so we look for that first. But the same is true if the user has no permissions, which is cheaper to check, so I agree it should go first.

common-app-lib/src/cmr/common_app/services/ingest/subscription_common.clj
```diff
@@ -18,3 +18,14 @@
         (as-> $ (string/join "&" $))
         string/trim
         digest/md5)))
+
+(defn create-query-params
```
eereiter, line 22: why was this moved into common?  Is another micro service other than ingest going to use it?  right now I only see ingest using it.   If so then great, if not it probably should remain in ingest.

ingest-app/src/cmr/ingest/api/subscriptions.clj
eereiter: Oh!! I deleted the wrong comment, and I don't know how to get it back.  Sorry Mitch - he stated: "now this is smart." and I agree with him.

common-app-lib/src/cmr/common_app/services/ingest/subscription_common.clj
mitchstartzel: I'm open to discussion on this - my reasoning was that while it _is_ only used in Ingest, it was previously only used in ingest/services while I now needed it in ingest/api. Moving it to common_app/service/ingest seemed like the smart maneuver to cut down on unnecessary imports, since there's no microservice-level common files, but I concede I am not sure if this violates precedent. If it does, I have no issues reverting it.

ygliuvt: Agree with Erich. Functions used only in ingest should stay in ingest code base. `ingest/api` is supposed to call functions within `ingest/services`, so it is logical to put this function in `ingest/services`.

ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -60,6 +62,30 @@
            subscriber-id
            concept-id)))))
 
+(defn- search-gran-refs-with-sub-params
+  "Performs a granule search using the provided query parameters. If the query is no good,
+   we throw a service error."
+  [context params]
+  (try
+    (search/find-granule-references context params)
```
ygliuvt, line 70: Since this function is called on every subscription ingest, we should not use the try catch to handle a normal operation scenario. Let's create a new function (maybe called `find-granule-references-return-errors`) to return the error from search regularly rather than throwing exception, and call this new function here.

system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
```diff
@@ -167,6 +167,25 @@
...
+                  {:token "mock-echo-system-token"})
+          metadata {:CollectionConceptId (:concept-id coll1)
+                    :Query "options%5Bspatial%5D%5Bor%5D=true"}
+          concept (subscription-util/make-subscription-concept metadata)
+          response (ingest/ingest-concept
+                    concept
+                    {:accept-format :json
+                     :raw? true})
+          {:keys [errors]} (ingest/parse-ingest-body :json response)]
+      (is (re-find #"Error ingesting subscription: invalid query parameters" (first errors))))))
```
ygliuvt, line 186: We should check the error message has the specific error message search-app returned about the invalid query parameter.

system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
```diff
@@ -167,6 +167,26 @@
...
+          concept (subscription-util/make-subscription-concept metadata)
+          response (ingest/ingest-concept
+                    concept
+                    {:accept-format :json
+                     :raw? true})
+          {:keys [errors]} (ingest/parse-ingest-body :json response)]
+      (is (re-find #"Error ingesting subscription with query parameters" (first errors)))
+      (is (re-find #"Granule search failed. status: 400 body:" (first errors))))))
+
+
```
jaybarra, line 189: extra whitespace nitpick

ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -60,6 +62,27 @@
...
 
+(defn- search-gran-refs-with-sub-params
+  "Performs a granule search using the provided query parameters. If the query is no good,
+   we throw a service error."
+  [context params]
+  (if-let [errors (search/find-granule-references-return-errors context params)]
+    (errors/throw-service-error :bad-request
+      (str "Error ingesting subscription with query parameters [" (dissoc params :token) "]: " errors))))
+
+
```
jaybarra, line 73: extra whitespace nitpick

ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -60,6 +62,27 @@
            subscriber-id
            concept-id)))))
 
+(defn- search-gran-refs-with-sub-params
+  "Performs a granule search using the provided query parameters. If the query is no good,
+   we throw a service error."
+  [context params]
+  (if-let [errors (search/find-granule-references-return-errors context params)]
```
jaybarra, line 69: There's no 'else' case, so this can be a 'when-let'

jaybarra (changes requested):

transmit-lib/src/cmr/transmit/search.clj
```diff
@@ -89,6 +89,29 @@
...
+        header (ch/context->http-headers context)
+        response (client/post request-url
+                              (merge
+                                (config/conn-params conn)
+                                {:body (codec/form-encode request-body)
+                                 :content-type mt/form-url-encoded
+                                 :throw-exceptions false
+                                 :headers (if token (assoc header config/token-header token) header)}))
+        {:keys [status body]} response]
+    (if (= status 400)
```
ygliuvt, line 111: this line should be changed to `(when (not= 200 status)`

ygliuvt: then you don't need line 113.

system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
```diff
@@ -167,6 +167,26 @@
...
+          metadata {:CollectionConceptId (:concept-id coll1)
+                    :Query "options%5Bspatial%5D%5Bor%5D=true"}
+          concept (subscription-util/make-subscription-concept metadata)
+          response (ingest/ingest-concept
+                    concept
+                    {:accept-format :json
+                     :raw? true})
+          {:keys [errors]} (ingest/parse-ingest-body :json response)]
+      (is (re-find #"Error ingesting subscription with query parameters" (first errors)))
+      (is (re-find #"Granule search failed. status: 400 body:" (first errors))))))
```
ygliuvt, line 187: I meant check the exact error message that is returned from search on the validation error. As you can tell already that the whole error message is not very user friendly. The `Granule search failed. status: 400 body:" part is confusing to the end user. That is something only cmr dev cares. To the end user, the message should say something like "Subscription query failed validation with error ....".

transmit-lib/src/cmr/transmit/search.clj
```diff
@@ -89,6 +89,28 @@
...
+        response (client/post request-url
+                              (merge
+                                (config/conn-params conn)
+                                {:body (codec/form-encode request-body)
+                                 :content-type mt/form-url-encoded
+                                 :throw-exceptions false
+                                 :headers (if token (assoc header config/token-header token) header)}))
+        {:keys [status body]} response]
+    (when-not (= status 200)
+      response)))
```
ygliuvt, line 112: change `response` to `body`.

ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -60,6 +62,26 @@
            subscriber-id
            concept-id)))))
 
+(defn- search-gran-refs-with-sub-params
+  "Performs a granule search using the provided query parameters. If the query is no good,
+   we throw a service error."
+  [context params]
+  (when-let [errors (search/find-granule-references-return-errors context params)]
+    (errors/throw-service-error :bad-request
+      (str "Subscription query validation failed with the following error(s): " (:body errors)))))
```
ygliuvt, line 71: change `(:body errors)` to `errors`.

transmit-lib/src/cmr/transmit/search.clj
```diff
@@ -89,6 +89,28 @@
       (errors/internal-error!
         (format "Granule search failed. status: %s body: %s" status body)))))
 
+(defn-timed find-granule-references-return-errors
+  "Find granules by parameters in a post request. If the search fails, returns the failure
```
ygliuvt, line 93: change to `......, returns the response body.`

transmit-lib/src/cmr/transmit/search.clj
mitchstartzel: I was thinking that since we don't know what exactly the status code will be (only that it isn't 200) it might be useful to have the whole response, but since I'm not looking for it at all in the validation code, I can't argue with this change.

transmit-lib/src/cmr/transmit/search.clj
```diff
@@ -89,6 +89,27 @@
       (errors/internal-error!
         (format "Granule search failed. status: %s body: %s" status body)))))
 
+(defn-timed find-granule-references-return-errors
```
ygliuvt, line 92: When I recommended this current name, I wasn't expecting you to change this function in a way that it does not function with valid searches. Since you have modified the function in a way that it does not return any result when the search is valid, we should rename this function to something like `validate-granule-search-params`.

transmit-lib/src/cmr/transmit/search.clj
```diff
@@ -89,6 +89,27 @@
...
+  "Find granules by parameters in a post request. If the search fails, returns the response body."
+  [context params]
+  (let [conn (config/context->app-connection context :search)
+        request-url (str (conn/root-url conn) "/granules.json")
```
ygliuvt, line 96: change `/granules.json` to `/granules`.

transmit-lib/src/cmr/transmit/search.clj
```diff
@@ -89,6 +89,27 @@
...
+        request-body (-> params
+                         (dissoc :token)
+                         (assoc :page_size 0))
```
ygliuvt, line 99: Don't expect you put this here, but if we change the function name, we are ok.

transmit-lib/src/cmr/transmit/search.clj
```diff
@@ -89,6 +89,27 @@
       (errors/internal-error!
         (format "Granule search failed. status: %s body: %s" status body)))))
 
+(defn-timed find-granule-references-return-errors
+  "Find granules by parameters in a post request. If the search fails, returns the response body."
```
ygliuvt, line 93: update the docstring to reflect what this function is doing now.

ygliuvt (approved):

jaybarra (approved):

PR #1249, opened 2021-05-25
CMR-7380: Added support for collection keyword quoted phrase search

sxu123:
1. Added a new index keyword2 of text-field-keyword-mapping for collections. It's indexing the same fields as the keyword, except that it is using "keyword" analyzer instead of "whitespace" analyzer and it's indexing individual keyword field as is, rather than indexing individual words from all the keyword fields. It also index all the fields the existing index keyword indexed. Retire the current keyword index because once the mapping is changed, you can't reuse the existing one.
2. Keep the existing keyword search api. When doing the keyword search for collections, All the search is routed to search on keyword2 index. The difference on quoted and unquoted keyword search is the quoted keyword search uses span query and the unquoted one uses query-string, and no longer uses whitespace analyzer, it uses keyword analyzer.
3. Currently we only support single keyword phrase search.  Mix of keyword and keyword phrases search is not supported, multiple keyword phrase search is not supported. It's documented in the search api doc.


common-app-lib/src/cmr/common_app/services/search/query_to_elastic.clj
```diff
@@ -201,10 +201,13 @@
...
-   (let [field (query-field->elastic-field field concept-type)]
+   (let [field (query-field->elastic-field field concept-type)
+         analyzer (if (= :keyword-phrase field)
+                    :keyword
+                    :whitespace)]
      {:query_string {:query (escape-query-string query-str)
-                     :analyzer :whitespace
-                     :default_field field
+                     :analyzer analyzer 
+                     :default_field field
```
jceaser, line 210: nitpick: is space after words looked down on by the other CMR developers?

indexer-app/src/cmr/indexer/data/concepts/collection/keyword.clj
```diff
@@ -1,12 +1,13 @@
...
+   [clojure.string :as str]
+   [cmr.common.concepts :as concepts]
+   [cmr.common.util :as util]
+   [cmr.indexer.data.concepts.keyword-util :as keyword-util]))
 
 (defn create-keywords-field
-  [concept-id collection other-fields]
+  [concept-id collection other-fields to-keyword-text?]
   "Create a keyword field for keyword searches by concatenating several other fields
   into a single string"
```
jceaser, line 12: The comment should be updated to indicate why one would send true or false for the 3rd parameter.

search-app/src/cmr/search/data/query_to_elastic.clj
```diff
@@ -302,11 +302,29 @@
...
+(defn- remove-quotes-from-keyword-query-string
+  "When keyword query string is double quoted string in a condition, we want remove the quotes and change
+  the field from :keyword to :keyword-phrase so that the search can be against :keyword-phrase index."
+  [condition]
+  (let [query-str (:query-str condition)
+        trimmed-query-str (when query-str
+                            (str/trim query-str))]
+    (if (and query-str
+             (= :keyword (:field condition))
+             (str/starts-with? trimmed-query-str "\"")
```
jceaser, line 314: why do we not care about single quotes?

jceaser (changes requested):

system-int-test/test/cmr/system_int_test/search/collection_keyword_search_test.clj
```diff
@@ -183,8 +185,10 @@
           (and parameter-matches? json-matches?))
 
         ;; search by contact persons
-        "Bob Hope" [coll26]
-        "bob.hope@nasa.gov" [coll26]
+        "Bob Hope" [coll26 coll26-1]
+        "\"Bob Hope\"" []
+        "bob.hope@nasa.gov" [coll26 coll26-1]
```
jceaser, line 190: While not a deciding factor for me on approving this change, I dislike the use of actual email address with real enough names that could be possible. I looked it up, there is a Richard (by middle name) Hope that works for nasa, but lucky for us he uses his first name so this email is not active. bob.home@example.gov would be better as example.* is the official fake domain name. Same goes for Mr Freeze below. We do have people of the Fries family in NASA, NSIDC could as well even though a google search for this exact email returns no hits, it is possible. I would recommend victor.fries@example.org. Just a suggestion, and maybe it should be in the back of our mind when we create new examples.

system-int-test/test/cmr/system_int_test/search/collection_keyword_search_test.clj
```diff
@@ -205,12 +209,21 @@
 
         ;; entry title
         "coll1" [coll1]
+        "\"coll1\"" [coll1]
+        "coll26" [coll26 coll26-1]
+         "\"coll26\"" [coll26]
         "Mitch made a (merry-go-round)" [coll2]
+        "\"Mitch made a (merry-go-round)\"" [coll2]
         "(merry-go-round)" [coll2]
+        "\"(merry-go-round)\"" []
```
ygliuvt, line 218: I don't think this is the desired behavior when people searching with quotes.

ygliuvt (changes requested):
This change does not offer the desired search behavior when searching with quotes. If a user searches with `"Sea Ice"`, he/she is expecting to see a collection with keyword `Huge Sea Ice` returned. The current implementation does not return the collection with `Huge Sea Ice`, only collections that has the exact keyword `Sea Ice`. This is not what Chris Lynnes wants when filing this ticket.

indexer-app/src/cmr/indexer/data/index_set.clj
```diff
@@ -468,7 +468,7 @@
           :personnel m/string-field-mapping
 
           ;; analyzed field for keyword searches
-          :keyword m/text-field-mapping
+          :keyword2 m/text-field-keyword-mapping
```
eereiter, line 471: Why did you have to create a keyword2? Why can't we just use keyword? We just need to re-index the collections, which takes 10 minutes and we do that everyday anyway and the old data won't exist anymore.

eereiter (changes requested):

indexer-app/src/cmr/indexer/data/index_set.clj
eereiter: I got your feedback on slack that you can't change the type of data that exists in the keyword.

eereiter (approved):

indexer-app/src/cmr/indexer/data/concepts/collection.clj
```diff
@@ -403,11 +403,11 @@
             :created-at created-at
             :coordinate-system coordinate-system
 
-            ;; fields added to support keyword searches
-            :keyword (k/create-keywords-field concept-id collection
-                                              {:platform-long-names platform-long-names
-                                               :instrument-long-names instrument-long-names
-                                               :entry-id entry-id})
+            ;; fields added to support keyword searches including the quoted string case.
+            :keyword2 (k/create-keywords-field concept-id collection
```
jaybarra, line 407: Would recommend renaming as `:keyword-extra-chars` or something slightly more verbose to disambiguate `:keyword` and `:keyword2`

jaybarra (changes requested):
Would agree with @ygliuvt . Possibly create a new analyzer with a [character group tokenizer](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-chargroup-tokenizer.html)

I have created analyzers before, so might be able to help.

common-app-lib/src/cmr/common_app/services/search/query_to_elastic.clj
```diff
@@ -202,10 +202,27 @@
...
+       ;; because we can not modify the existing index mapping.
+       (and (= :keyword-phrase field) (= :collection concept-type))
+         {:span_near
+           {:clauses [{:span_multi
+                        {:match
+                          {:wildcard
+                            {:keyword2 (escape-query-string query-str)}}}}]
+            :slop 0
+            :in_order true}}
+       (and (= :keyword field) (= :collection concept-type))
```
ygliuvt, line 216: No need to have this clause. We can use the `concept-type->field-mappings` function or something similar to map the keyword search field to the correct elasticsearch field and use the `else` clause handle anything that is not a keyword phrase.

indexer-app/src/cmr/indexer/data/index_set.clj
ygliuvt: correct, you can't update the mapping of an existing ES field.

system-int-test/test/cmr/system_int_test/search/collection_keyword_search_test.clj
```diff
@@ -205,18 +209,32 @@
 
         ;; entry title
         "coll1" [coll1]
+        "\"coll1\"" [coll1]
+        "coll26" [coll26 coll26-1]
+         "\"coll26\"" [coll26 coll26-1]
         "Mitch made a (merry-go-round)" [coll2]
```
ygliuvt, line 215: Add a test case to search with combination of regular word and quotes, e.g. `Mitch "a (merry-go-round)"`

ygliuvt: I think this is a test case that we need to support.

common-app-lib/src/cmr/common_app/services/search/query_to_elastic.clj
```diff
@@ -201,11 +201,22 @@
   cmr.common_app.services.search.query_model.TextCondition
   (condition->elastic
    [{:keys [field query-str]} concept-type]
-   (let [field (query-field->elastic-field field concept-type)]
-     {:query_string {:query (escape-query-string query-str)
-                     :analyzer :whitespace
-                     :default_field field
-                     :default_operator :and}}))
+   (let [elastic-field (query-field->elastic-field field concept-type)]
+     (cond
```
ygliuvt, line 205: nit: use `if` rather than `cond`.

search-app/docs/api.md
```diff
@@ -1578,13 +1578,23 @@ Supports ignore_case and the following aliases for "NEAR\_REAL\_TIME": "near\_re
...
 #### <a name="c-keyword"></a> Find collections by keyword (free text) search
 
-Keyword searches are case insensitive and support wild cards ? and *.
-There is a limit of 30 wild cards allowed in keyword searches. Within 30 wild cards, there's also limit on the max keyword
+Keyword searches are case insensitive and support wild cards ? and *, in which '\*' matches zero or more characters and '?' matches any single character. There is a limit of 30 wild cards allowed in keyword searches. Within 30 wild cards, there's also limit on the max keyword
 string length. The longer the max keyword string length, the less number of keywords with wild cards allowed.
 
+The following searches on "alpha", "beta" and "g?mma" individually and returns the collections that contain all these individual words
+in the keyword fields that are indexed. Note: these words don't have to exist in the same keyword field, but they have to exsit as a
+space(or one of special character delimiter CMR uses) delimited word.
```
ygliuvt, line 1586: add a space before `(`.

search-app/docs/api.md
```diff
@@ -1578,13 +1578,23 @@ Supports ignore_case and the following aliases for "NEAR\_REAL\_TIME": "near\_re
...
+
     curl "%CMR-ENDPOINT%/collections?keyword=alpha%20beta%20g?mma"
 
-The following fields are indexed for keyword search:
+We also support keyword phrase search. The following searches on "*alpha beta g?mma*" as a phrase and returns the collections with
```
ygliuvt, line 1590: It should be `"alpha beta g?mma"`

system-int-test/test/cmr/system_int_test/search/collection_keyword_search_test.clj
```diff
@@ -205,18 +209,37 @@
...
+        "\"merry-go-round\"" [coll2]
         "merry" [coll2]
+        "\"merry\"" [coll2]
         "merry go round" [coll2]
+        "\"merry go round\"" []
         "merry-go" []
+        "\"merry-go\"" []
+
+        ;; mix of keyword and keyword phrase search: not supported yet.
+        "Mitch \"a (merry-go-round)\"" []
```
ygliuvt, line 229: I think we should return a validation error on the un-supported parameter combinations rather than silently return nothing. Let's add a validation rule for keyword search parameters and handle these two cases.

ygliuvt (approved):

PR #1250, opened 2021-05-25
CMR-7311

ygliuvt:
As a CMR developer, I would like to set up a lambda function to consume the CMR collection indexing events to populate graph db.


graph-db/indexer/handler.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+module.exports.index = async event => {
```
ygliuvt, line 3: This function will be completely replaced with the actual indexing code that is coming when we work on CMR-7307. So ignore what's happening inside here.

graph-db/indexer/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "dependencies": {
+    "serverless": "^2.43.0"
+  },
+  "name": "graph-lambda",
```
mschmele, line 5: I'd name this something more descriptive like `cmr-graphdb-indexer`

mschmele (commented):
We should either keep jest/eslint configs at the top-level directory, or copy them to each project. It may be worth it to do that in the next ticket, though. Your code additions look good to me, but I'm not sure how we actually want to organize this project

graph-db/indexer/handler.js
```diff
@@ -0,0 +1,16 @@
...
+  console.log(`Ryan, Sarah and Mark will see ingest event: [${JSON.stringify(event, null, 2)}].`)
```
jaybarra, line 4: Should remove personal identifiers

graph-db/indexer/handler.js
```diff
@@ -0,0 +1,16 @@
...
+  return {
+    statusCode: 200,
+    body: JSON.stringify(
+      {
+        message: 'Processed ingest event successfully!',
+        input: event,
```
jaybarra, line 10: What is the purpose of returning the event? If you are returning more than a 200 should that jut be the request body?

jaybarra (changes requested):

graph-db/indexer/handler.js
mschmele: I'm going to be modifying this, so it shouldn't actually matter what is here

graph-db/indexer/handler.js
ygliuvt: See my comment: `This function will be completely replaced with the actual indexing code that is coming when we work on CMR-7307. So ignore what's happening inside here.`

graph-db/indexer/handler.js
ygliuvt: See my comment: This function will be completely replaced with the actual indexing code that is coming when we work on CMR-7307. So ignore what's happening inside here.

mschmele (approved):
There will be some more moving around that's needed in the next PR for common util functions, but this PR satisfies the ticket - looks good to me

jaybarra (approved):

PR #1252, opened 2021-05-27
CMR-7248 updated instructions for bootstrapping db

jaybarra:
Single top-level lein command for setting up the database

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,6 +284,11 @@ lein modules itest --skip-meta :in-memory-db
 If you want to run tests against Oracle, bring up the Oracle VM and execute
 the following to create the users and run the migrations:
 
+```sh
+lein migrate
+```
+
+Deprecated commmand
 ``` sh
 cmr setup db
 ```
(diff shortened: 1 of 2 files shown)


daniel-zamora (approved):

README.md
```diff
@@ -284,6 +284,11 @@ lein modules itest --skip-meta :in-memory-db
 If you want to run tests against Oracle, bring up the Oracle VM and execute
 the following to create the users and run the migrations:
 
+```sh
+lein migrate
+```
+
+Deprecated commmand
```
eereiter, line 291: https://wiki.earthdata.nasa.gov/display/CMR/Oracle+DB+setup+for+local+CMR+development is the wiki page that includes instructions on how to setup the oracle database for development.  This wiki page states that we should use cmr setup db.  Why is it being deprecated?  - if the way to setup the database is changing, please update the documentation so that developers down the road can do it easily.

eereiter (changes requested):
Please update the oracle setup documentation.  It states that we should use cmr setup db, but in your comments you are deprecating it.

ygliuvt (changes requested):
I don't think this change is addressing the issue documented in CMR-7248. The ticket itself is somewhat not clearly worded on the solution. We cannot run `lein migrate` on a deployed environment currently. I think the fix should be to fix CMR so that a new service can come up from a fresh database.

jaybarra:
OBE

PR #1254, opened 2021-05-28
CMR-7307: Reorganize project and allow indexing of single CMR collection from SQS event

mschmele:


ygliuvt (changes requested):
Need to fix the following issues:
1) Currently, the Documentation vertices are not bootstrapped because the add Documentation Vertex call is constructed, but not executed in the code.
2) The RelatedUrl description is used as `title` in the Documentation vertex which cannot be NULL, but there is no checking on description to not be NULL. We need to decide if we want to keep the `title` property on the Documentation vertex. If we do, we need to fill in a default value when there is no description on the RelatedUrl.
3) The bootstrap process should be idempotent. We should not recreate the edges between Documentation vertex and Dataset vertex if they already exist. Currently, we are recreating them every time bootstrap is run.

graph-db/commonUtils/cmr/__mocks__/getSecureParam.js
```diff
@@ -0,0 +1,2 @@
+// eslint-disable-next-line no-unused-vars
+exports.getSecureParam = async (_param) => '1234-very-good-token'
```
mschmele, line 2: Just remove params for all mocked functions

graph-db/commonUtils/indexing/indexCmrCollection.js
```diff
@@ -25,29 +29,38 @@ exports.indexCmrCollection = async (collection, gremlin) => {
     datasetName = doiUrl
   }
 
+  console.log(`RelatedUrls for concept [${conceptId}]: ${relatedUrls}`)
```
mschmele, line 32: remove this

graph-db/commonUtils/indexing/indexRelatedUrl.js
```diff
@@ -0,0 +1,62 @@
+const gremlin = require('gremlin')
+
+// eslint-disable-next-line no-underscore-dangle
+const __ = gremlin.process.statics
```
mschmele, line 4: Rename this something useful

graph-db/package.json
```diff
@@ -8,7 +8,9 @@
     "test": "jest --config jest.config.js",
     "test-coverage": "jest --coverage",
     "ci-test": "jest --coverage --config jest.config.junit_output.js",
-    "bootstrap": "serverless invoke local -f bootstrapGremlinServer -e CMR_ROOT=\"http://cmr.uat.earthdata.nasa.gov\" -e PAGE_SIZE=2000 -e GREMLIN_URL=\"ws://localhost:8182/gremlin\""
+    "bootstrap": "serverless invoke local -f bootstrapGremlinServer -e CMR_ROOT=\"http://cmr.uat.earthdata.nasa.gov\" -e PAGE_SIZE=2000 -e GREMLIN_URL=\"ws://localhost:8182/gremlin\" --stage=sit",
+    "index-collection": "serverless invoke local -f indexCmrCollection -e CMR_ROOT=\"http://cmr.uat.earthdata.nasa.gov\" -e GREMLIN_URL=https://cmr.sit.earthdata.nasa.gov/graphdb --stage=sit --path testData.json",
```
mschmele, line 12: gremlin url here should be localhost

graph-db/testData.json
```diff
@@ -0,0 +1 @@
+{ "Records": [{ "body": "{\"concept-id\": \"C1237293909-SCIOPSTEST\" }"}]}
```
mschmele, line 1: We don't need this

graph-db/commonUtils/indexing/indexCmrCollection.js
```diff
@@ -1,12 +1,15 @@
...
 
 /**
  * Given a collection from the CMR, index it into Gremlin
  * @param {JSON} collection collection object from `items` array in cmr response
- * @param {Gremlin Traversal Object} gremlin connection to gremlin server
+ * @param {Gremlin Traversal Object} g connection to gremlin server
  * @returns
  */
-exports.indexCmrCollection = async (collection, gremlin) => {
+exports.indexCmrCollection = async (collection, g) => {
```
ygliuvt, line 12: The collection update event looks something like:
```
{
  "action": "concept-update",
  "concept-id": "C1200371048-DEMO_PROV",
  "revision-id": 10
}
```
We should properly handle at least the action type in this PR. We can write a separate ticket to handle the revision-id.

graph-db/package.json
```diff
@@ -8,7 +8,8 @@
     "test": "jest --config jest.config.js",
     "test-coverage": "jest --coverage",
     "ci-test": "jest --coverage --config jest.config.junit_output.js",
-    "bootstrap": "serverless invoke local -f bootstrapGremlinServer -e CMR_ROOT=\"http://cmr.uat.earthdata.nasa.gov\" -e PAGE_SIZE=2000 -e GREMLIN_URL=\"ws://localhost:8182/gremlin\""
+    "bootstrap": "serverless invoke local -f bootstrapGremlinServer -e CMR_ROOT=\"http://cmr.uat.earthdata.nasa.gov\" -e PAGE_SIZE=2000 -e GREMLIN_URL=\"ws://localhost:8182/gremlin\" --stage=sit",
```
ygliuvt, line 11: The GREMLIN_URL setting does not match the `stage` sit setting. I think the stage should be local dev here.

graph-db/serverless/src/indexCmrCollection/__test__/handler.test.js
```diff
@@ -0,0 +1,22 @@
...
+})
+
+describe('indexCmrCollection handler', () => {
+  test('test index of single collection', async () => {
+    const event = { Records: [{ body: '{"concept-id": "C1237293909-SCIOPS" }' }] }
+
+    const indexed = await indexCmrCollection(event)
+
+    const { body, statusCode } = indexed
+    expect(body).toBe('Collection [C1237293909-SCIOPS] indexed sucessfully: true')
```
ygliuvt, line 19: It would be nice to verify that the collection vertex actually exists in graph db, rather than the response says thing are indexed. The code could not work at all and returns a response that says things are indexed successfully.

ygliuvt: similar comments on related url vertex and the `documents` edge. We should verify they exists in the graph db.

abbottry (changes requested):
Your method of mocking works, but wont as soon as you attempt to run more than one test in a single file. It's not necessarily what that method was created for, its more for mocking libraries.

graph-db/package.json
mschmele: Not in this case. This is for bootstrapping a local gremlin server, so I renamed the command to `bootstrap-local` to make that more clear

graph-db/commonUtils/indexing/indexRelatedUrl.js
```diff
@@ -0,0 +1,61 @@
...
+    || urlContentType !== 'PublicationURL') {
+    // We only care about documentation at the moment.
+    // Checking the URLContentType is the most efficient way to find its type.
+    // Return early if it isn't some kind of documentation.
+    return
+  }
+
+  try {
+    // Use `fold` and `coalesce` to check existance of vertex, and create one if none exists.
+    const documentationVertex = await g
```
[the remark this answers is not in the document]
mschmele: Yeah, I like that better. The convention is to use `g` because that's what the gremlin console uses, but I hate that convention, at least in code.

graph-db/package.json
mschmele: So that the serverless config doesn't throw errors when it's trying to fill parameter values

graph-db/README.md
```diff
@@ -49,13 +49,54 @@ Clone the graphexp repository at https://github.com/bricaud/graphexp
...
-* Indexer
+### Build
+```
+npm install
+```
+
+### Run
+To invoke the bootstrap function and load data into your Gremlin server from the CMR run the following command:
+```
+npm run bootstrap
```
ygliuvt, line 64: `bootstrap` no longer exists. You changed it to `bootstrap-local`. We need to update the description above to clarify what is being run.

graph-db/serverless/src/utils/getConceptType.js
```diff
@@ -0,0 +1,17 @@
+const conceptTypes = {
+  C: 'collection',
+  G: 'granule',
+  S: 'service',
+  V: 'variable'
```
mitchstartzel, line 5: Are there actually plans to index things like services and variables in the far future? I know we've seen things like ingest of new service revisions take painfully long when the service is associated with many collections - I wonder if graphdb tech can be applied there.

graph-db/commonUtils/indexing/indexRelatedUrl.js
```diff
@@ -0,0 +1,67 @@
+const gremlin = require('gremlin')
+
+const gremlinStatistics = gremlin.process.statics
```
mitchstartzel, line 3: gremlin.process._statics_? Looks like this should be gremlin.process.statistics

mitchstartzel: This also appears in `indexCmrCollection.js`

graph-db/commonUtils/indexing/indexCmrCollection.js
```diff
@@ -0,0 +1,64 @@
...
+ * @param {Gremlin Traversal Object} gremlinConnection connection to gremlin server
+ * @returns
+ */
+exports.indexCmrCollection = async (collection, gremlinConnection) => {
+  const {
+    meta: { 'concept-id': conceptId },
+    umm: {
+      EntryTitle: entryTitle,
+      DOI: { DOI: doiDescription },
+      RelatedUrls: relatedUrls
```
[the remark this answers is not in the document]
mitchstartzel: Ah, I see below you are checking for collections with documentation in common - very clever.

mitchstartzel (approved):
A few comments and questions. I want to keep up to speed on graphdb, seems like interesting tech.

graph-db/package.json
```diff
@@ -8,7 +8,8 @@
     "test": "jest --config jest.config.js",
     "test-coverage": "jest --coverage",
     "ci-test": "jest --coverage --config jest.config.junit_output.js",
-    "bootstrap": "serverless invoke local -f bootstrapGremlinServer -e CMR_ROOT=\"http://cmr.uat.earthdata.nasa.gov\" -e PAGE_SIZE=2000 -e GREMLIN_URL=\"ws://localhost:8182/gremlin\""
+    "bootstrap-local": "serverless invoke local -f bootstrapGremlinServer -e CMR_ROOT=\"http://cmr.uat.earthdata.nasa.gov\" -e PAGE_SIZE=2000 -e GREMLIN_URL=\"ws://localhost:8182/gremlin\" --stage=sit",
```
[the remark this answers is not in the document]
mschmele: This is a testing command, and UAT tends to have more reliable data

graph-db/serverless/src/utils/getConceptType.js
mschmele: In the near-term no, but it's certainly not out of the question for the future

PR #1263, opened 2021-06-17
CMR-7443

ygliuvt:
This PR is for work done mostly by @abbottry and @macrouch to add integration test support and tidy up the jtest tests and codebase.


graph-db/serverless.yml
```diff
@@ -82,3 +76,10 @@ resources:
         PolicyName: cmrParamStorePolicy-${opt:stage}
         Roles:
           - Ref: IamRoleGraphLambdaExecution
+custom:
+  variables:
+    # Default values for environment variables
+    CMR_ROOT: ${env:CMR_ROOT, 'https://cmr.sit.earthdata.nasa.gov'}
```
jaybarra, line 82: I would suggest this is just blank if no default is set to make sure its being done explicitly as part of a deployment

graph-db/serverless/src/utils/cmr/__test__/fetchPageFromCmr.test.js
```diff
@@ -1,44 +1,36 @@
...
-})
-
-afterEach(() => {
-  // Restore any ENV variables overwritten in tests
-  process.env = OLD_ENV
-})
+import * as indexPageOfCmrResults from '../indexPageOfCmrResults'
 
 describe('fetchPageFromCMR', () => {
+  // beforeAll(() => {
```
jaybarra, line 8: Can the comments be removed?

jaybarra (approved):

graph-db/serverless.yml
ygliuvt: I am going to change it to be deduced from `stage`.

macrouch (approved):

graph-db/serverless.yml
ygliuvt: Actually, we do want the ability to set a `CMR_ROOT` that is different from what `stage` maps to. And I don't want to have to put in `CMR_ROOT` every time for the default deployment `sit`. So I am going to keep it the way it was.

PR #1264, opened 2021-06-18
CMR-7309

ygliuvt:
As a CMR client user, I would like to delete collection in graph db when it is deleted from CMR

### DIFF
--- a/graph-db/serverless/src/indexCmrCollection/__test__/handler.test.js
+++ b/graph-db/serverless/src/indexCmrCollection/__test__/handler.test.js
@@ -7,7 +7,7 @@ import indexCmrCollection from '../handler'
 
 import { updateCollection, deleteCollection } from '../../testUtil/indexCollection'
 
-import { verifyExistInGraphDb, verifyNotExistInGraphDb, verifyDatasetNotExistInGraphDb } from '../../testUtil/verifyGraphDb'
+import { verifyExistInGraphDb, verifyNotExistInGraphDb } from '../../testUtil/verifyGraphDb'
 
 beforeEach(() => {
   jest.clearAllMocks()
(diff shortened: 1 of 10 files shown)


jaybarra (approved):

graph-db/serverless/src/testUtil/verifyGraphDb.js
```diff
@@ -24,3 +24,17 @@ export const verifyGraphDb = async (datasetTitle, docName) => {
   const { value: { id: edgeId } } = record
   expect(edgeId).not.toBe(null)
 }
+
+export const verifyNotExistInGraphDb = async (datasetTitle, docName) => {
+  // verify the dataset vertex with the given title does not exist
+  const dataset = await global.testGremlinConnection
+    .V().has('dataset', 'title', datasetTitle).next()
```
macrouch, line 31: in `deleteCmrCollection.js` you have chained calls like this on new lines for every call. like this: 
```
.V()
.has(...)
.next()
```

That is easier to read than having them all on the same line. I'd switch all of the gremlin calls in this file to match the example above with each chained method call being on a new line.

graph-db/serverless/src/indexCmrCollection/__test__/handler.test.js
```diff
@@ -2,64 +2,61 @@ import nock from 'nock'
...
 
     const indexed = await indexCmrCollection(event)
 
+    expect(consoleMock).toBeCalledTimes(1)
+    expect(consoleMock).toBeCalledWith('Skip indexing of collection [C123755555-TESTPROV] as it is not found in CMR')
+
     const { body, statusCode } = indexed
 
-    expect(body).toBe('Successfully indexed 1 collection(s)')
+    expect(body).toBe('Successfully indexed 0 collection(s)')
```
macrouch, line 47: A better message for this example might be `Successfully indexed 0 collection(s). Skipped 1 collection(s)`. 

In `indexCmrCollection/handler.js` you could have a separate counter `skippedCount` that you only increment when `items.length === 0`, next to that console.log on line 48

macrouch (approved):

PR #1265, opened 2021-06-18
CMR-6936: Fixed the 500 error for temporal facet search when there are no granules

sxu123:


search-app/src/cmr/search/services/query_execution/facets/hierarchical_links_helper.clj
```diff
@@ -48,7 +48,7 @@
   "Returns a sequence of keys that have multiple values where at least one of values matches the
   passed in value. Looks for matches case insensitively."
   [query-params value]
-  (let [value (str/lower-case value)]
+  (let [value (when value (str/lower-case value))]
```
jaybarra, line 51: I'm not opposed to adding a nil check here to prevent a null pointer exception, but can we filter what goes into this beforehand to stop nulls getting here in the first place?

jaybarra (changes requested):
Can you add a test case for this?

mitchstartzel (approved):
Approved assuming a test (which should have caught this!) is added.

jceaser (approved):

jaybarra (approved):

PR #1267, opened 2021-06-18
CMR-7308

ygliuvt:
As a CMR client user, I would like to update collection in graph db.

### DIFF
--- a/graph-db/serverless/src/utils/cmr/indexCmrCollection.js
+++ b/graph-db/serverless/src/utils/cmr/indexCmrCollection.js
@@ -2,6 +2,7 @@ import gremlin from 'gremlin'
 import 'array-foreach-async'
 
 import indexRelatedUrl from './indexRelatedUrl'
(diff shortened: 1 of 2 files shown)


macrouch (approved):

jaybarra (approved):

graph-db/serverless/src/utils/cmr/indexCmrCollection.js
```diff
@@ -25,6 +26,9 @@ export const indexCmrCollection = async (collection, gremlinConnection) => {
     }
   } = collection
 
+  // delete the collection first so that we can clean up its related documentation vertices
+  await deleteCmrCollection(conceptId, gremlinConnection)
```
mschmele, line 30: This may make bootstrapping quite a bit more expensive, which could result in that lambda timing out. It would only be a one-time problem, though

mschmele (approved):

PR #1280, opened 2021-07-14
CMR-7528

ygliuvt:
Create Campaign vertices and relate them to Collections.


graph-db/serverless/src/utils/cmr/indexCampaign.js
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Given a project object, Gremlin connection, and associated collection, index campaign and build relationships for any GENERAL campaign that exists in common with other nodes in the graph database
+ * @param {JSON} project UMM project object
+ * @param {Connection} gremlinConnection a connection to the gremlin server
+ * @param {Graph Node} dataset the parent collection vertex in the gremlin server
+ * @returns null
+ */
+const indexCampaign = async (project, gremlinConnection, dataset, conceptId) => {
+  // CMR UMM-C actually put the campaign info in the Projects ShortName field rather than the Campaigns field for historical reasons
```
ygliuvt, line 13: I have confirmed this with Erich. This is something that the UMM powerhouse team will eventually fix and we will update it here when that happens.

graph-db/README.md
```diff
@@ -125,19 +125,9 @@ To bootstrap all collections in a specific provider (e.g LPDAAC_TS2), send the f
...
-* concept-id - Collection concept id
-* name - Url to the collection landing page
-* title - Entry title of the collection
-* doi - DOI link of the collection
-
-The documentation vertex has the following properties:
-
-* name - documentation url
-* title - description of the documentation url
+![CMR Collection GraphDB Diagram](images/cmr_collection_graphdb_diagram.png)
```
mitchstartzel, line 130: 💯

graph-db/serverless/src/utils/cmr/deleteCmrCollection.js
```diff
@@ -8,22 +8,45 @@ const { P: { lte } } = gremlin.process
...
+  // drop all the campaign vertices that are connected to and only connected to the dataset vertex
+  let success
+  success = await deleteLinkedVertices(conceptId, gremlinConnection, 'campaign', 'includedIn')
+  if (success === false) {
+    return false
+  }
 
+  // drop all the documentation vertices that are connected to and only connected to the dataset vertex
+  success = await deleteLinkedVertices(conceptId, gremlinConnection, 'documentation', 'documentedBy')
+  if (success === false) {
```
jaybarra, line 49: Can the early outs be simplified to `if (!success)`

graph-db/serverless/src/utils/cmr/indexCmrCollection.js
```diff
@@ -65,6 +67,12 @@ export const indexCmrCollection = async (collection, gremlinConnection) => {
   const { value = {} } = dataset
   const { id: datasetId } = value
 
+  if (projects && projects.length > 0) {
```
jaybarra, line 70: Is the wrapper and check for elements necessary or can forEachAsync handle an empty array without it?

jaybarra (approved):

graph-db/serverless/src/utils/cmr/deleteCmrCollection.js
ygliuvt: `!success` will be `true` if `success` is undefined or `false`. I may not matter in this particular case, but I think it is better to check for `false` specifically as that is what is returned from the function.

graph-db/serverless/src/utils/cmr/indexCmrCollection.js
ygliuvt: it will not work if projects is null without the check.

PR #1281, opened 2021-07-14
CMR-7556: Bumped rhino version

mitchstartzel:
* Bumped rhino, nested dependency of `[metosin/ring-swagger "0.22.14"]`


eereiter (approved):

jaybarra:
Do you need to add an `excludes` to swagger to make sure the older version is not included?

jaybarra (approved):

mitchstartzel:
> Do you need to add an `excludes` to the swagger dependency to make sure the older version is not included?

Lein is smart enough here - the old rhino version takes its place in the `.pom` file

PR #1286, opened 2021-07-23
CMR-7481, CMR-7529

ygliuvt:
Create platformInstrument vertices to relate platforms and instruments to collections.


graph-db/serverless/src/utils/cmr/indexPlatform.js
```diff
@@ -0,0 +1,99 @@
...
+const gremlinStatistics = gremlin.process.statics
+
+/**
+ * create the acquiredBy edge between platform/instrument and collection
+ * @param {JSON} piVertex the platformInstrument vertex
+ * @param {Connection} gremlinConnection a connection to the gremlin server
+ * @param {Graph Node} dataset the parent collection vertex in the gremlin server
+ * @returns null
+ */
+const createAcquiredByEdge = async (piVertex, gremlinConnection, dataset) => {
```
mschmele, line 12: Normally each of these functions would be their own file to make testing easier. I think it's worth sticking to that pattern

ygliuvt: I think it is more readable to keep this private function in this file.

mschmele: I disagree because it's an inconsistency in how we organize our code

macrouch: I agree with @mschmele

ygliuvt: Reluctantly refactored code. :)

mschmele (approved):
Thanks for making those changes

macrouch (approved):

PR #1287, opened 2021-07-26
CMR-7506: Bulk Granule Update for Checksum in ECHO10

mitchstartzel:
* Targets XML element at `Granule/DataGranule/Checksum/<Value/Algorithm>
* If one value is supplied, only value is updated. If two values are supplied, they are assigned to value and algorithm respectively
* If the granule has no `<DataGranule>` element, then the update fails.

### DIFF
--- a/ingest-app/src/cmr/ingest/services/granule_bulk_update_service.clj
+++ b/ingest-app/src/cmr/ingest/services/granule_bulk_update_service.clj
@@ -254,21 +254,21 @@
    :invalid-data [(format "Appending s3 urls is not supported for format [%s]" (:format concept))]))
 
 (defmulti update-checksum
-  "Add s3 url to the given granule concept."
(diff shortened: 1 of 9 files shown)


ingest-app/src/cmr/ingest/services/granule_bulk_update_service.clj
```diff
@@ -318,6 +338,31 @@
   (modify-s3-link*
    context concept bulk-update-params user-id append-s3-url))
 
+(defn- modify-checksum*
+  "Modify the S3Link data for the given concept with the provided URLs
```
ygliuvt, line 342: Change S3 link references to checksum.

ygliuvt (approved):

jaybarra (approved):

yiduoke (approved):

PR #1288, opened 2021-07-27
CMR-7592

ygliuvt:
Rename dataset vertex name property to landing-page.


mschmele (commented):
My only question is why start to throw errors after catching them?

ygliuvt:
> My only question is why start to throw errors after catching them?

This is to make sure we terminate the indexing in case of an error.

mschmele (approved):
Discussed whether or not we terminate or continue indexing on errors in slack. Looks good to me

macrouch (approved):

PR #1301, opened 2021-08-06
Cmr 7489 progressive collection update

sxu123:
Ingest validations include 3 parts: 1. Schema validation, 2. umm-spec collection validation 3. Business rule validation.  Out of these 3 validations, we now allow the users to update a collection with #2 and #3 validation errors, if these errors exist in the previous revision collection. If new validation errors are introduced, the update will fail. Updates with schema validation errors, existing or new, will always fail.

### DIFF
--- a/ingest-app/docs/api.md
+++ b/ingest-app/docs/api.md
@@ -329,7 +329,7 @@ curl -i -XPOST -H "Content-type: application/echo10+xml" \
 
 Collection metadata can be created or updated by sending an HTTP PUT with the metadata to the URL `%CMR-ENDPOINT%/providers/<provider-id>/collections/<native-id>`. The response will include the [concept id](#concept-id) and the [revision id](#revision-id). The metadata that is uploaded is validated for XML well-formedness, XML schema validation, and against UMM validation rules. Keyword validation can be enabled with the [keyword validation header](#validate-keywords-header). If there is a need to retrieve the native-id of an already-ingested collection for updating, requesting the collection via the search API in UMM-JSON format will provide the native-id.
 
-Note: we now provide progressive collection update feature, which allows a collection to be updated with non-schema related validation errors that are existing validation errors for the previous collection revision. Only newly introduced validation errors will fail the update. Schema validation errors always fail the update.
+Note: we now provide progressive collection update feature through a new configuration parameter CMR_PROGRESSIVE_UPDATE_ENABLED, which is turned on by default. It allows a collection to be updated with non-schema related validation errors that are existing validation errors for the previous collection revision. Only newly introduced validation errors will fail the update. Schema validation errors always fail the update.
 
 ```
 curl -i -XPUT \
--- a/ingest-app/docs/api.md
+++ b/ingest-app/docs/api.md
@@ -329,6 +329,8 @@ curl -i -XPOST -H "Content-type: application/echo10+xml" \
 
 Collection metadata can be created or updated by sending an HTTP PUT with the metadata to the URL `%CMR-ENDPOINT%/providers/<provider-id>/collections/<native-id>`. The response will include the [concept id](#concept-id) and the [revision id](#revision-id). The metadata that is uploaded is validated for XML well-formedness, XML schema validation, and against UMM validation rules. Keyword validation can be enabled with the [keyword validation header](#validate-keywords-header). If there is a need to retrieve the native-id of an already-ingested collection for updating, requesting the collection via the search API in UMM-JSON format will provide the native-id.
 
+Note: we now provide progressive collection update feature, which allows a collection to be updated with non-schema related validation errors that are existing validation errors for the previous collection revision. Only newly introduced validation errors will fail the update. Schema validation errors always fail the update.
+
 ```
 curl -i -XPUT \
   -H "Content-type: application/echo10+xml" \
--- a/ingest-app/src/cmr/ingest/config.clj
+++ b/ingest-app/src/cmr/ingest/config.clj
@@ -7,6 +7,10 @@
    [cmr.oracle.config :as oracle-config]
    [cmr.oracle.connection :as conn]))
 
+(defconfig progressive-update-enabled
+  "Flag for whether or not collection progressive update is enabled."
+  {:default true :type Boolean})
+
 (defconfig bulk-update-cleanup-minimum-age
   "The minimum age(in days) of the rows in bulk-update-task-status table that can be cleaned up"
   {:default 90
--- a/ingest-app/src/cmr/ingest/validation/validation.clj
+++ b/ingest-app/src/cmr/ingest/validation/validation.clj
@@ -2,6 +2,7 @@
   "Provides functions to validate concept"
   (:require
     [cheshire.core :as json]
+    [clojure.data :as data]
     [clojure.string :as str]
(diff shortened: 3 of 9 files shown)


system-int-test/test/cmr/system_int_test/ingest/collection_progressive_update_test.clj
```diff
@@ -0,0 +1,89 @@
...
+              {:keys [errors warnings concept-id revision-id]} response]
+          (is (= concept-id (:concept-id coll)))
+          (is (= revision-id (+ 1 (:revision-id coll))))
+          (is (= nil errors))
+          (is (= expected-warnings warnings))))
+
+        (testing "Progressive update collection successful case, real existing errors"
+        (let [additional-attributes [(data-umm-cmn/additional-attribute {:Name "int" :DataType "STRING" :ParameterRangeBegin 1 :ParameterRangeEnd 10})]
```
jaybarra, line 52: Whitespace nitpick, this should be indented

system-int-test/test/cmr/system_int_test/ingest/collection_progressive_update_test.clj
```diff
@@ -0,0 +1,89 @@
...
+          (is (= nil errors))
+          (is (= expected-warnings warnings))))
+        
+        (testing "Progressive update collection failed case, introducing new data-dates errors"
+        (let [additional-attributes [(data-umm-cmn/additional-attribute {:Name "int" :DataType "STRING" :ParameterRangeBegin 1 :ParameterRangeEnd 10})]
+              temporal-extents [(data-umm-cmn/temporal-extent {:beginning-date-time "2001-01-01T12:00:00Z"
+                                        :ending-date-time "2000-05-11T12:00:00Z"})]
+              data-dates [(umm-cmn/map->DateType {:Date (time-core/date-time 2000) :Type "DELETE"})]
+              ;; now that the previous concept containing validation errors, we don't need to pass options to mimic existing errors.
+              response (try (data-core/ingest-umm-spec-collection "PROV1" (data-umm-c/collection
```
jaybarra, line 78: The try/catch can be replaced with a 
```
(is (thrown-with-msg? 
       java.lang.Exception
       #"Ingest failed" 
       (data-core/ingest-umm-spec-collection ...
```

ingest-app/src/cmr/ingest/config.clj
```diff
@@ -7,6 +7,10 @@
    [cmr.oracle.config :as oracle-config]
    [cmr.oracle.connection :as conn]))
 
+(defconfig progressive-update-enabled
+  "Flag for whether or not collection progressive update is enabled."
+  {:default true :type Boolean})
```
jaybarra, line 12: Can you make a note in the deployment repo instructions that this flag exists now?

jaybarra (approved):

ingest-app/docs/api.md
```diff
@@ -329,6 +329,8 @@ curl -i -XPOST -H "Content-type: application/echo10+xml" \
 
 Collection metadata can be created or updated by sending an HTTP PUT with the metadata to the URL `%CMR-ENDPOINT%/providers/<provider-id>/collections/<native-id>`. The response will include the [concept id](#concept-id) and the [revision id](#revision-id). The metadata that is uploaded is validated for XML well-formedness, XML schema validation, and against UMM validation rules. Keyword validation can be enabled with the [keyword validation header](#validate-keywords-header). If there is a need to retrieve the native-id of an already-ingested collection for updating, requesting the collection via the search API in UMM-JSON format will provide the native-id.
 
+Note: we now provide progressive collection update feature, which allows a collection to be updated with non-schema related validation errors that are existing validation errors for the previous collection revision. Only newly introduced validation errors will fail the update. Schema validation errors always fail the update.
```
ygliuvt, line 332: Need to specify the new configuration parameter and state that it is turn on by default.

ygliuvt (changes requested):
This comment is on the design approach of this ticket. This implementation affects the whole ingest workflow by always retrieving the previous collection revision and doing the validation on both the previous and the current revisions at the same time on every single collection ingest. This is not ideal/correct. 

Almost all the collection ingests in the real world are valid collections, we should not add the penalty of having to retrieve the previous revision and see if the previous version is valid to every collection ingest. We should only do that additional check when and only when the current revision is invalid on the granule related extra checks.

ingest-app/src/cmr/ingest/config.clj
sxu123: Sure.

ingest-app/src/cmr/ingest/validation/validation.clj
```diff
@@ -181,18 +182,44 @@
...
+                                       (when (:validate-keywords? validation-options)
+                                         [(keyword-validations context)]))))
+               ;; get the newly introduced validation errors
+               new-err-messages (seq (first (data/diff (set err-messages) (set prev-err-messages))))]
+            (if new-err-messages
+              (errors/throw-service-errors :invalid-data new-err-messages)
+              ;; when there is no newly introduced errors, err-messages contains only existing errors.
+              [{:existing-errors (map #(str (:path %) " " (:errors %)) err-messages)}]))
+         (errors/throw-service-errors :invalid-data err-messages))
+       (do
```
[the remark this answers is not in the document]
sxu123: This was there before and I think it's required because it's multiple statements.

ingest-app/docs/api.md
sxu123: I'm not sure that we should do that.  This configuration parameter was suggested just in case it's not working well or if there is a bug so that we can turn it off while fixing things, eventually we want to get rid of it. so this is more for the operator to take a note.

system-int-test/test/cmr/system_int_test/ingest/collection_progressive_update_test.clj
sxu123: Sorry I can't put the right message in to make this work.

system-int-test/test/cmr/system_int_test/ingest/collection_progressive_update_test.clj
sxu123: done

ygliuvt (approved):

ingest-app/src/cmr/ingest/validation/validation.clj
```diff
@@ -181,18 +182,45 @@
...
+             (not warn?))
+       ;; whenever it's time to throw errors, we want to check if it's an collection update and
+       ;; it's not bulk-update and progressive-update-enabled is true. If so, we want to throw
+       ;; errors only when new errors are introduced, otherwise return all the existing errors as
+       ;; error-warnings.
+       (if (and (config/progressive-update-enabled)
+                (not (:bulk-update? validation-options)) 
+                prev-collection)
+         (let [prev-err-messages (if (and (:test-existing-errors? validation-options)
+                                          (= "mock-echo-system-token" (:token context)))
```
ygliuvt, line 203: This should be the configurable system token check, `cmr.transmit.config/echo-system-token?`.

ygliuvt (changes requested):
Need to change token check.

ygliuvt (approved):

PR #1304, opened 2021-08-10
CMR-7524: Bulk Granule Update for Files/FilePackages in UMM-G

mitchstartzel:
* Addition of AdditionalFile update type for UMM-G, allowing update of Size, Format, and Checksum in File/FilePackages:
	- Size and SizeUnit, SizeInBytes
	- Format, FormatType, and MimeType
	- Checksum (Value and Algorithm)
   Rules can be found in the updated api doc.

### DIFF
--- a/ingest-app/docs/api.md
+++ b/ingest-app/docs/api.md
@@ -1293,11 +1293,11 @@ curl -i -XPOST \
   ]
 }'
 ```
-In the above request, `Example_Granule_UR_1` recieves updates to two elements: The FilePackage `ZippedFilePackage` has SizeInBytes, Size/SizeUnit, and Format updated, while File `GranuleFileName1` has MimeType and Checksum updated. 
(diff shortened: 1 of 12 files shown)


ingest-app/docs/api.md
```diff
@@ -1236,6 +1236,74 @@ Example granule bulk update response:
...
+      ]
+    }
+  ]
+}'
+```
+In the above request, `Example_Granule_UR_1` recieves updates to two elements: The FilePackage `ZippedFilePackage` has SizeInBytes, Size/SizeUnit, and Format updated, while File `GranuleFileName1` has MimeType and Checksum updated. 
+
+Note that specifying whether an element is a File or FilePackage is unecessary. Providing the `name` for an element is sufficient to locate and update it.
+
+`Example_Granule_UR_2` also receives updates on the contained `GranuleZipFile`, on its Size/SizeUnit and SizeInBytes fields. This also displays a special use special case for Size-related updates: When a file update is requested with Size `0`, then the Size and SizeUnit fields will be removed from the resulting file. The same applies for SizeInBytes, which will be removed on its own if a value of `0` is supplied.
```
ygliuvt, line 1300: change `special use special case` to `special use case`.

ygliuvt (approved):

jaybarra (approved):

PR #1309, opened 2021-08-12
CMR-7605 Fixing a null pointer exception when translating UMM-G granules to ECHO 10

eereiter:


umm-lib/src/cmr/umm/echo10/granule.clj
```diff
@@ -57,13 +57,12 @@
...
                   checksum]} data-granule
-          day-night (if day-night day-night "UNSPECIFIED")
-          size-mb (get-size-in-mb size size-unit)]
+          day-night (if day-night day-night "UNSPECIFIED")]
       (x/element :DataGranule {}
                  (when size-in-bytes
                    (x/element :DataGranuleSizeInBytes {} size-in-bytes))
                  (when size
-                   (x/element :SizeMBDataGranule {} size-mb))
+                   (x/element :SizeMBDataGranule {} (get-size-in-mb size size-unit)))
```
ygliuvt, line 65: I don't see any real code change in this file. Is this ticket OBE?

umm-lib/test/cmr/umm/test/echo10/granule.clj
```diff
@@ -445,4 +446,18 @@
...
+(deftest generate-data-granule-test
+  (testing "Testing the generate-data-granule method when the size doesn't exist - making sure that
+            a null pointer exception isn't thrown. CMR-7605"
+    (let [granule-data (g/generate-data-granule
+                         (umm-g/map->DataGranule
+                           {:size-in-bytes 71938553
+                            :size-unit "KB"
+                            :producer-gran-id "0000000.0000001.hdf"
+                            :day-night "NIGHT"}))]
+      (is (and (= (cx/string-at-path granule-data [:DataGranuleSizeInBytes])
```
ygliuvt, line 460: Change this line to two separate `is` verification rather than an `and`.

umm-lib/test/cmr/umm/test/echo10/granule.clj
```diff
@@ -8,14 +8,15 @@
...
-            [cmr.umm.test.generators.granule :as gran-gen]
+            [clojure.string :as string]
             [cmr.common.date-time-parser :as p]
+            [cmr.common.xml :as cx]
             [cmr.umm.echo10.granule :as g]
             [cmr.umm.echo10.echo10-core :as echo10]
             [cmr.umm.umm-collection :as umm-c]
             [cmr.umm.umm-granule :as umm-g]
-            [cmr.umm.test.echo10.echo10-collection-tests :as tc]))
+            [cmr.umm.test.echo10.echo10-collection-tests :as tc]
```
ygliuvt, line 18: nit: the imports are not sorted correctly and indentation is wrong.

ygliuvt (changes requested):
There is no real code change in this PR. How does it fix the bug?

umm-lib/src/cmr/umm/echo10/granule.clj
eereiter: No, the code change is moving the (get-size-in-mb size size-unit) function from the let statement to be wrapped in a when clause after line 64.  This prevents the function getting called if the size is nil, whereas before it was being called and could cause a null pointer exception.

umm-lib/test/cmr/umm/test/echo10/granule.clj
eereiter: I wanted them to be anded together to signify that both have to be true, but making it two separate is accomplishes the same thing.

eereiter: fixed.

umm-lib/test/cmr/umm/test/echo10/granule.clj
eereiter: fixed.

umm-lib/src/cmr/umm/echo10/granule.clj
ygliuvt: Got it. I missed the `when` clause.

ygliuvt (approved):

umm-lib/src/cmr/umm/echo10/granule.clj
```diff
@@ -57,13 +57,12 @@
                   size-unit
                   size-in-bytes
                   checksum]} data-granule
-          day-night (if day-night day-night "UNSPECIFIED")
-          size-mb (get-size-in-mb size size-unit)]
+          day-night (if day-night day-night "UNSPECIFIED")]
```
jaybarra, line 60: This can be simplified to `(or day-night "UNSPECIFIED")`

jaybarra (approved):

umm-lib/src/cmr/umm/echo10/granule.clj
eereiter: sure - not my code, but I'll make the improvement.

PR #1311, opened 2021-08-13
CMR-7474: replacing Echo-Token header usage with Authorization

yiduoke:


mschmele (approved):
Code looks good. In the future, a more descriptive PR title would be helpful

yiduoke:
> Code looks good. In the future, a more descriptive PR title would be helpful

ok! edited the title

transmit-lib/src/cmr/transmit/echo/tokens.clj
```diff
@@ -48,7 +48,7 @@
 
     (let [[status parsed body] (r/rest-post context "/tokens/get_token_info"
                                             {:headers {"Accept" mt/json
-                                                       "Echo-Token" (transmit-config/echo-system-token)}
+                                                       "Authorization" (transmit-config/echo-system-token)}
                                              :form-params {:id token}})]
```
mitchstartzel, line 52: I think this change specifically is going to break CMR when deployed - this is the main call that passes tokens to Legacy-Services, and Legacy-Services I predict will not be expecting an `Authorization` header. I think you should dig around Legacy-Services to find where it is accepting token headers to make sure this will be accepted (or for now, just pass in both headers)

mitchstartzel (approved):
Looks good for merge, assuming my comment is resolved (or ruled unnecessary)

PR #1314, opened 2021-08-17
CMR-7492: Added logging when collection update succeeds with existing…

sxu123:
… errors.

### DIFF
--- a/ingest-app/src/cmr/ingest/services/ingest_service/collection.clj
+++ b/ingest-app/src/cmr/ingest/services/ingest_service/collection.clj
@@ -1,6 +1,8 @@
 (ns cmr.ingest.services.ingest-service.collection
   (:require
    [clojure.string :as string]
+   [cmr.common.api.context :as common-context]
+   [cmr.common.log :as log :refer (warn)]
    [cmr.common.util :refer [defn-timed]]
    [cmr.ingest.config :as config]
    [cmr.ingest.services.helper :as ingest-helper]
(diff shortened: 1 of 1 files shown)


eereiter (approved):

ingest-app/src/cmr/ingest/services/ingest_service/collection.clj
```diff
@@ -59,10 +61,13 @@
...
                                   collection-warnings)
-         warnings (concat err-warnings warnings collection-warnings)]
+         non-err-warnings (concat warnings collection-warnings)
+         warnings (concat err-warnings non-err-warnings)]
      ;; The sanitized UMM Spec collection is returned so that ingest does not fail
      {:collection sanitized-collection
-      :warnings warnings})))
+      :warnings warnings
+      :err-warnings err-warnings
+      :non-err-warnings non-err-warnings})))
```
mschmele, line 70: These names make this code very confusing. Would something like `all-warnings`, `user-visible-warnings` and `admin-only-warnings` make more sense? How are `warnings`, `err-warnings` and `non-err-warnings` different? At the very least, I'd like to see a comment here

mschmele (changes requested):

ingest-app/src/cmr/ingest/services/ingest_service/collection.clj
sxu123: We have non-err-warnings(The pure warnings we have now),  then we have err-warnings, introduced in progressive update feature(They are actually validation errors, but since they exist in previous revision of the collection, we now allow the ingest to pass, but we still need to return to the user these pre-existing errors, so they are returned as part of warnings.).  Now warnings = non-err-warnings + err-warnings.   I have to separate them up because the requirement is that I log the existing errors and the pure warnings separately.

mschmele (approved):

PR #1319, opened 2021-08-19
ENV specific KMS URLs

jceaser:
Adding code to change the "version" parameter used when sending requests to KMS in SIT and UAT and preserving the old way for production.

### DIFF
--- a/transmit-lib/src/cmr/transmit/kms.clj
+++ b/transmit-lib/src/cmr/transmit/kms.clj
@@ -39,38 +39,20 @@
    :related-urls :uuid
    :granule-data-format :uuid})
 
-(defn- if-not-production
-  "Return true if CMR is not currently in production. This function is passed to
-   env-varient. If new rules are needed, then write a similar file. The function
-   requires that a function return true if the KMS URL is to be appended to"
-  []
-  (not= "prod" (System/getenv "ENVIRONMENT")))
-
-(defn- env-varient
-  "Return the KMS URL fragment that is specific to the envirment CMR is currently
-   running in. Use this function to update the URL parameter list that will be
-   sent to KMS.
-   Take three parameters:
-   * resource-name : URL fragment found in keyword-scheme->gcmd-resource-name
-   * env-fn : function that returns true if URL is to be appended to
-   * varient : URL parameters to append if env-fn is true"
-  [resource-name env-fn varient]
-  (let [url resource-name]
-    (if (env-fn)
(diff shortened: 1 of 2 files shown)


jceaser:
This change is needed to allow CMR to use the correct data in SIT and UAT while allowing PROD to remain on it's data set.

NOTE: this is an issue because all three environments use the same KMS server due to the way AWS and EBNET networks are configured.

eereiter (approved):

transmit-lib/src/cmr/transmit/kms.clj
```diff
@@ -51,7 +84,7 @@
    :measurement-name "measurementname?format=csv"
    :concepts "idnnode?format=csv"
    :iso-topic-categories "isotopiccategory?format=csv"
-   :related-urls "rucontenttype?format=csv"
+   :related-urls (env-varient "rucontenttype?format=csv" if-not-production "&version=DRAFT")
```
sxu123, line 87: Seems that the code doesn't need to be this fancy.  the logic is so simple
(if production?
  url
  (str url varient)) 
We don't need to pass if-not-production as a function, and have an additional env-variant function.

sxu123 (approved):

transmit-lib/src/cmr/transmit/kms.clj
```diff
@@ -39,6 +39,39 @@
    :related-urls :uuid
    :granule-data-format :uuid})
 
+(defn- if-not-production
```
mitchstartzel, line 42: nitpick: this fn would sound more clojurey if the fn name was `not-production?`

transmit-lib/src/cmr/transmit/kms.clj
```diff
@@ -39,6 +39,39 @@
...
+      url)))
+
+(comment
+ "The following line of code is used for trasitioning CMR from SIT->UAT->PROD.
+  These three envirments all talk to production KMS due to AWS->EBNET network
+  limitations. Because we need to be able to publish data for all three envirments
+  with one host, it was decided to use the different version capabilities in KMS
+  to isolate production from the other two envirments. Specificly: rucontenttype
+  is a three level tree in SIT, but not PROD. This process is only needed durring
+  the transition and should be deleted once in production."
```
mitchstartzel, line 70: What exactly is being deleted once in production here?

mitchstartzel (approved):

PR #1321, opened 2021-08-24
CMR-7525: BGU for OPeNDAP type in UMM-G

mitchstartzel:


ingest-app/src/cmr/ingest/services/granule_bulk_update/opendap/opendap_util.clj
```diff
@@ -31,27 +31,28 @@
   https://opendap.*.earthdata.nasa.gov/* and no more than one that does not match the pattern.
   Returns the parsed urls in a map under keys :cloud and :on-prem with the corresponding urls."
   [url]
```
mitchstartzel, line 33: The difference checker went wild here - all that changed is that the function body is now wrapped in a `(when url ...)`

ingest-app/src/cmr/ingest/services/granule_bulk_update/opendap/umm_g.clj
```diff
@@ -36,6 +37,14 @@
        first
        (update-opendap-url* (first (opendap-type url-map)))))
 
+(defn- updated-type-related-url
+  "Updates types"
+  [related-url]
+  (if (re-find #"opendap" (string/lower-case (:URL related-url)))
```
mitchstartzel, line 43: Curious what @eereiter thinks here - I opted to identify any link containing the string `opendap` as an opendap link. Meaning, any link that matches will have its Type and Subtype updated.

Are there any alternatives we should consider? Could this be overzealous in identifying opendap links?

PR #1332, opened 2021-09-02
CMR-7040: As a client, I would like to search CMR with search_after parameter for deep paging

ygliuvt:
See the API doc in this PR for description of the new search-after feature.

### DIFF
--- a/common-app-lib/src/cmr/common_app/services/search/params.clj
+++ b/common-app-lib/src/cmr/common_app/services/search/params.clj
@@ -274,12 +274,14 @@
   [context concept-type params]
   (let [[params query-attribs] (parse-query-level-params concept-type params)
         options (u/map-keys->kebab-case (get params :options {}))
-        scroll-id (:scroll-id context)
+        {:keys [scroll-id search-after]} context
         params (dissoc params :options)
         query (if (empty? params)
                 ;; matches everything
                 (qm/query query-attribs)
                 ;; Convert params into conditions
                 (params->query-conditions context concept-type options params query-attribs))]
-    ;; add the scroll-id if present
-    (merge query (when scroll-id {:scroll-id scroll-id}))))
+    ;; add the scroll-id and search-after if present
+    (merge query
+           (when scroll-id {:scroll-id scroll-id})
+           (when search-after {:search-after search-after}))))
--- a/search-app/docs/api.md
+++ b/search-app/docs/api.md
@@ -14,6 +14,7 @@ CMR Legacy Services' ECHO tokens will be deprecated soon. Please use EDL tokens 
     * [CORS Header support](#cors-header-support)
     * [Query Parameters](#query-parameters)
     * [Paging Details](#paging-details)
+    * [Search After](#search-after)
     * [Scrolling Details](#scrolling-details)
     * [Parameter Options](#parameter-options)
     * [Collection Result Feature Parameters](#collection-result-features)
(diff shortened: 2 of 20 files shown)


search-app/docs/api.md
```diff
@@ -244,8 +245,47 @@ The CMR contains many more results than can be returned in a single response so
...
+
+When all the results have been returned, the subsequent search will return an empty result set and no `CMR-Search-After` header in the response.
+
+Different from scrolling requests, each search-after request needs to supply all the search parameters, and the `CMR-Search-After` header needs to be updated with the new value returned from the previous search to page through the whole result set. Although user can change the search parameters and still get results back as long as the sort order of the search is unchanged, it breaks the rationale of paging and offers no real use case. Thus user should always supply the same search parameters while using Search After requests to page through a large result set.
```
chris-durbin, line 258: One of the features we have been taking advantage of is not having to know the query that passed to CMR to initiate a scroll session. We have a different application initiating the scroll session than the one that consumes the results. Having to pass the same query in can be difficult (especially with a large POST body or with a shapefile).

Can you think of a way of supporting Search-After without having to pass in the query again?

ygliuvt: This would be completely against the stateless nature of ES search-after feature. I would like to keep the CMR search-after feature as is to utilize the native ES search-after feature. What you are asking for can be done, but I would like to write a separate ticket to implement (maybe) a new endpoint for the session based search.

search-app/src/cmr/search/services/query_service.clj
```diff
@@ -200,38 +200,52 @@
...
+                            (:hits results) (name concept-type) total-took (:client-id context)
+                            (rfh/printable-result-format (:result-format query)) (pr-str params))]
+    (info (if scroll-id
+            (format "%s, scroll-id: %s." log-message (str (hash scroll-id)))
+            (if search-after
+              (format "%s, search-after: %s, new search-after: %s."
+                      log-message
+                      (json/encode search-after)
+                      (:search-after results))
+              (format "%s." log-message))))
```
mschmele, line 221: Not sure how you feel about `cond`, but something like this reads more clearly to me than a nested `if`

```suggestion
    (info (cond
            scroll-id (format "%s, scroll-id: %s." log-message (str (hash scroll-id)))
            search-after (format "%s, search-after: %s, new search-after: %s."
                                 log-message
                                 (json/encode search-after)
                                 (:search-after results))
            :else (format "%s." log-message))))
```

mschmele (approved):
For some reason I expected this to be much more complicated than it is - that's a nice surprise! If we don't already have a ticket written to deprecate scrolling, we should probably do that, unless we want to leave that as an option

common-app-lib/src/cmr/common_app/services/search/params.clj
```diff
@@ -274,12 +274,14 @@
...
+        {:keys [scroll-id search-after]} context
         params (dissoc params :options)
         query (if (empty? params)
                 ;; matches everything
                 (qm/query query-attribs)
                 ;; Convert params into conditions
                 (params->query-conditions context concept-type options params query-attribs))]
-    ;; add the scroll-id if present
-    (merge query (when scroll-id {:scroll-id scroll-id}))))
+    ;; add the scroll-id and search-after if present
```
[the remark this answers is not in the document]
ygliuvt: Only one will exist once the validation is complete. At here, we just add the field if it exists.

jaybarra (approved):

PR #1337, opened 2021-09-10
CMR-6745

ygliuvt:
Fixed the spatial OR search. 
Cleaned up spatial OR and granule count result feature code.


mschmele (approved):
I missed this being broken, but the fix is cleaner than my implementation so I like it 🙂. Having access to that `options` object seems to help simplify a lot of things

daniel-zamora (approved):

PR #1344, opened 2021-09-16
CMR-7713 updating versions

jaybarra:


eereiter (approved):

eereiter (approved):
is the package-lock.json file generated:  

19,190 additions, 2,049 deletions not shown because the diff is too large. Please use a local Git client to view these changes.  seems really big.

daniel-zamora (approved):

mitchstartzel (approved):
I'm pretty sure if you bump serverless further (current is 2.59?) you don't need the adm-zip resolution. But, if that causes any hassle, just ship it.

graph-db/package.json
```diff
@@ -24,7 +24,7 @@
     "gremlin": "^3.4.10",
     "lodash": "^4.17.21",
     "node-fetch": "^2.6.1",
-    "serverless": "^2.38.0",
+    "serverless": "^2.46.0",
```
ygliuvt, line 27: Why don't we update to the latest serverless version at `2.59.0`?

ygliuvt: That will have fixes for all the vulnerabilities.

jaybarra: 2.59.0 introduces a different issue

graph-db/package.json
```diff
@@ -38,5 +38,10 @@
     "nock": "^13.1.1",
     "serverless-offline": "^6.9.0",
     "serverless-pseudo-parameters": "^2.5.0"
+  },
+  "resolutions": {
+    "adm-zip": "^0.5.2",
```
ygliuvt, line 43: With the serverless package updated to the latest version, we don't need this anymore.

graph-db/package.json
```diff
@@ -24,7 +24,7 @@
     "gremlin": "^3.4.10",
     "lodash": "^4.17.21",
     "node-fetch": "^2.6.1",
-    "serverless": "^2.38.0",
+    "serverless": "^2.59.0",
     "serverless-webpack": "^5.5.1",
```
ygliuvt, line 28: Let's update `serverless-webpack` to `^5.5.4` as well.

graph-db/package.json
ygliuvt: by this, I mean the whole resolutions block.

PR #1348, opened 2021-09-23
CMR-7416 fixes single string multi bucket lists, adds tests

daniel-zamora:
This PR attempts address issue where bucket lists in metadata are ingested as a single json string array.  We want the buckets/prefixes to be returned in a consistent manner even when this is the case.  It might be better to add validation to prevent these type of values in the metadata, or ask data providers to refrain from ingesting values in this way.  This howerver fixes how the data, that currently exists, is returned.

### DIFF
--- a/access-control-app/src/cmr/access_control/services/acl_service.clj
+++ b/access-control-app/src/cmr/access_control/services/acl_service.clj
@@ -2,7 +2,7 @@
   (:require
    [clojure.edn :as edn]
    [clojure.set :as set]
-   [clojure.string :as str]
+   [clojure.string :as string]
    [cmr.access-control.data.access-control-index :as index]
    [cmr.access-control.data.acl-json-results-handler :as result-handler]
    [cmr.access-control.data.acl-schema :as schema]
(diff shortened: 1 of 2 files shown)


access-control-app/src/cmr/access_control/services/acl_service.clj
```diff
@@ -390,6 +408,7 @@
          :items
          (map :s3-bucket-and-object-prefix-names)
          flatten
+         parse-single-string-multi-valued-bucket-lists
          distinct
          (remove nil?))))
```
mitchstartzel, line 413: The `(remove nil?)` gets applied after `parse-single-string-multi-valued-bucket-lists` - is it possible a nil value can be passed to the latter, causing it to choke on the `str/trim` on line 371?

access-control-app/src/cmr/access_control/services/acl_service.clj
```diff
@@ -364,6 +364,24 @@
...
       (auth-util/get-sids context (tokens/get-user-id context user-token)))))
 
+(defn- fix-single-string-with-multiple-values
+  "Check if string has embedded array, if not then return result conj'd with results,
+   if so then convert it into an array of strings and concat that with results."
+  [results result]
+  (let [result (str/trim result)]
+    (if (str/starts-with? result "[")
+      (as-> (str/replace result #"\[|\]|\\|\"" "") result
+            (str/split result #",")
```
mitchstartzel, line 374: nit - I thought the CMR standard was to import `clojure.string :as string`

mitchstartzel (approved):
Looks good, minor comments

ygliuvt:
We should file a ticket to validate the S3 prefixes and fail the ingest for multiple S3 prefix as an array in string. To fix that ticket is more important than this one. To me, this ticket does not need to be fixed in code, we should just fix the bad data in CMR.

ygliuvt (approved):
Approved only because you have made the code change. We want to add the validation to not allow the single string multi bucket lists being ingested and clean up the bad data in CMR. After that, we can revert this change.

PR #1351, opened 2021-09-28
CMR-7591: Updated the static KMS sample files and added the full files.

sxu123:


ygliuvt:
`kms_examples/README` file needs update.

ygliuvt:
The ticket mentioned: `Also, many of the KMS concepts are now delivered through a paging mechanism which is currently not supported by CMR. CMR works in legacy mode in SIT/UAT/PROD but the testing environment does not match these environments. `. I don't see any code changes addressing paging or legacy mode. Are these not in the scope of this PR?

I don't see the value in dumping all the full GCMD keywords in CMR and not using them in our tests. We are not a repository of GCMD keywords, our test data should be small as long as it covers all test cases. It doesn't need to be real or full.

ygliuvt (changes requested):

sxu123:
Ticket is updated with clarifications.  This ticket only addresses updating the static KMS files and fix tests.  A new ticket CMR-7755 is filed to address the issue that CMR code is not supporting new KMS paging.

ygliuvt (approved):

PR #1354, opened 2021-10-05
CMR-7734: Update ACL ingest validation on group id to use new EDL groups

mitchstartzel:
Allows CRUD on ACLs with EDL Group names has group_id. According to URS docs, EDL Group names "must be comprised of letters and numbers with no spaces."

SID lookup based on EDL groups will be implemented in a future ticket. This ticket is merely a validation change. It is already possible to create/update an ACL with a bogus group_id (as long as it looks like the a group concept id) so allowing a new flavor of group_id (which will always be bogus, for now) follows the precedent.

### DIFF
--- a/access-control-app/int-test/cmr/access_control/int_test/acl_crud_test.clj
+++ b/access-control-app/int-test/cmr/access_control/int_test/acl_crud_test.clj
@@ -1567,3 +1567,18 @@
               {:user_id "user1"
                :system_object "DASHBOARD_ARC_CURATOR"}
               {:token token}))))))
+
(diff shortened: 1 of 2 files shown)


jaybarra (approved):

daniel-zamora (changes requested):
The current validation doesn't actually do much if its or'ing an AG concept id validation with any string value.  We should only do this validation if a feature toggle indicates we should, and the validation should remain the same.

daniel-zamora (approved):

ygliuvt (approved):

PR #1359, opened 2021-10-13
Feature: CMR-7172 Adds a "consortium" column to the provider table.

jceaser:
Consortium is a "project", group, or effort for which a provider belongs to, like CWIC. Storing these lists of groups in the provider table allows future code to auto apply indexes and remove the need for tags.

This change adds provider to the database. In SQL the value can be null if never provided, no default is needed.


jceaser:
Waiting for bamboo to turn green

PR #1375, opened 2021-11-02
CMR-7720: As a CMR user, I would like the STAC format to be supported for granules

ygliuvt:


search-app/src/cmr/search/results_handlers/stac_results_handler.clj
```diff
@@ -218,3 +223,63 @@
 (defmethod qs/single-result->response [:granule :stac]
   [context query results]
   (single-result->response context query results))
+
+(defn- get-fc-links
+  "Returns the links for feature collection"
+  [context query hits]
+  (let [{:keys [page-size offset]} query
+        page-num (if offset
+                   (inc (/ offset page-size))
```
jaybarra, line 232: Would `(inc (quot offset page-size))` be safer?

jaybarra (approved):

search-app/src/cmr/search/results_handlers/stac_results_handler.clj
ygliuvt: the `offset` here is the calculated result of page-size * page-num, so there is no need to worry about what if it does not fully divid.

search-app/src/cmr/search/results_handlers/stac_results_handler.clj
```diff
@@ -218,3 +223,63 @@
...
+  (let [{:keys [page-size offset]} query
+        page-num (if offset
+                   (inc (quot offset page-size))
+                   1)
+        item-count (if offset
+                     (+ offset page-size)
+                     page-size)
+        allowed-hits (min hits MAX_RESULT_WINDOW)
+        match-1 (str "page_num=" page-num)
+        match-2 (str "page-num=" page-num)
```
eereiter, line 239: Why have match-2?  it is the same as match-1 and I don't see it changing.  for page-num-pattern it looks the same if you used match-1 match-1 match-1 match-1.

ygliuvt: It is matching `page_num` and `page-num` both and in beginning of the query string or end of the query string.

search-app/src/cmr/search/results_handlers/stac_results_handler.clj
```diff
@@ -218,3 +223,63 @@
...
+
+(defn- get-fc-links
+  "Returns the links for feature collection"
+  [context query hits]
+  (let [{:keys [page-size offset]} query
+        page-num (if offset
+                   (inc (quot offset page-size))
+                   1)
+        item-count (if offset
+                     (+ offset page-size)
```
eereiter, line 235: I don't understand this - is seems wrong - is the offset the number of items per page?  If so, shouldn't this be multiplication and not addition?  Or is offset the total number of items seen so far?  Then adding page seems wrong.  If offset is 100 and page-size is 15 then the item-count is 115.  first page - item-count grows by 1, second page, it grows by 2, and so on.

search-app/test/cmr/search/test/results_handlers/stac_results_handler.clj
```diff
@@ -138,3 +138,117 @@
...
+(defn- self-link
+  "Returns the self link with the given page-num"
+  [base-string page-num]
+  {:rel "self"
+   :href (href base-string page-num)})
+
+(def ^:private root-link
+  "Defines the root link"
+  {:rel "root"
+   :href "http://localhost:3003/"})
```
eereiter, line 156: instead of hardcoding these can we use the values that are set by environment?  OK, well never mind. I see that a lot of our tests have this data hardcoded.

search-app/test/cmr/search/test/results_handlers/stac_results_handler.clj
```diff
@@ -138,3 +138,117 @@
...
+   :method "GET"
+   :href (href base-string page-num)}))
+
+(deftest get-fc-links
+  (testing "Get feature colection links"
+    (are3 [query-string page-size page-num base-string expected-nav]
+      (let [context (merge {:system {:public-conf {:protocol "http"
+                                                   :host "localhost"
+                                                   :port "3003"}}}
+                                   {:query-string query-string})
```
eereiter, line 180: Spacing seems off here.

search-app/src/cmr/search/results_handlers/stac_results_handler.clj
ygliuvt: We are using the current page-num used by STAC request to figure out if there is a previous page and next page. So if the total item returned at this page is greater than the total match or the maximum items allowed, there won't be a next page.

search-app/test/cmr/search/test/results_handlers/stac_results_handler.clj
ygliuvt: this is only for test, there won't be anything other than localhost.

search-app/test/cmr/search/test/results_handlers/stac_results_handler.clj
ygliuvt: fixed indentation.

search-app/src/cmr/search/results_handlers/stac_results_handler.clj
eereiter: Thanks.  I got it now.

eereiter (approved):

PR #1378, opened 2021-11-04
CMR-7753: Updated UMM-C schema to 1.16.6.

sxu123:
Part1: Update UMM-C schema to version 1.16.6, which adds FreeAndOpenData to UseConstraints.
Part2: Make use of the UseConstraints fields, including the new FreeAndOpenData field to alter the consortiums values before indexing.


eereiter (approved):

indexer-app/src/cmr/indexer/data/concepts/collection.clj
```diff
@@ -180,6 +180,39 @@
...
+  [use-constraints]
+  (let [description (:Description use-constraints)
+        linkage (get-in use-constraints [:LicenseURL :Linkage])
+        license-text (:LicenseText use-constraints)]
+    (or (some #(str/includes? description %) geoss-url-list)
+        (some #(str/includes? linkage %) geoss-url-list)
+        (some #(str/includes? license-text %) geoss-url-list))))
+
+(defn- alter-consortiums
+  "Alter the consortiums list based on use-constraints."
```
ygliuvt, line 202: Add comment on why we add `GEOSS` to consortiums when there is a consortium of `EOSDIS` and no `use-constraints`.

ygliuvt: why don't we do the same when there is `use-constraints`, but no value for `FreeAndOpenData`?

system-int-test/test/cmr/system_int_test/search/collection_consortium_geoss_search_test.clj
```diff
@@ -0,0 +1,78 @@
...
+        coll6 (d/ingest-umm-spec-collection
+               "PROV6"
+               (umm-c/collection 1
+                                 {:UseConstraints {:Description "PRO6" :LicenseURL {:Linkage "http://not-geoss-url"}}})
+               {:format :umm-json})]
+
+    (index/wait-until-indexed)
+
+    (testing "consortium parameter geoss search"
+      (are3 [items consortium options]
```
ygliuvt, line 72: there is only one test case, no need to use `are3` here.

indexer-app/src/cmr/indexer/data/concepts/collection.clj
sxu123: When there if consortium of EOSDIS, the provider is considered EOSDIS provider and they want to add GEOSS to the list.  I don't know the reason behind I just code according to the requirement Erich put in the ticket.

system-int-test/test/cmr/system_int_test/search/collection_consortium_geoss_search_test.clj
sxu123: If you look at the description Erich put in there. When there is use-constraints, EOSDIS provider or not doesn't play a role in it.

indexer-app/src/cmr/indexer/data/concepts/collection.clj
sxu123: Comment is not gonna help more than the code. EOSDIS provider requires GEOSS consortium when there is no constraint. That's exactly what the code says. I don't have any more info about it.

ygliuvt (approved):

PR #1389, opened 2021-11-18
CMR-7893: Fixed a bug.

sxu123:


search-app/src/cmr/search/data/complex_to_simple_converters/has_granules_or_cwic.clj
```diff
@@ -18,9 +18,15 @@
     (let [has-granules-or-cwic-map (has-granules-or-cwic-base/get-has-granules-or-cwic-map context)
           has-granules-map (has-granules-base/get-has-granules-map context)
           concept-ids (map key (filter val has-granules-map))
-          condition (cqm/string-conditions :concept-id concept-ids true)
+          condition (if (seq concept-ids)
+                      (cqm/string-conditions :concept-id concept-ids true)
+                      ;; when concept-ids are empty, string-conditions throw internal errors.
+                      ;; replace it with a non-existing concept-id to make sure nothing is returned using the condition..
+                      (cqm/string-conditions :concept-id ["non-existing"] true))
```
eereiter, line 25: This seems to be a hack. Isn't this check in place so that users don't do an all granule search?  What do you get back when (cqm/string-conditions :concept-id ["non-existing"] true)) is called?

sxu123: You will get back nothing.

sxu123: This is a collection search, not granules search.

eereiter: When I run that call, this is what I get back: 
{:case-sensitive? true,
 :pattern? false,
 :value "non-existing",
 :field :concept-id}

Looking further down the code, the only place condition is used is if "this" is not has-granules-or-cwic.  Can you just catch the exception and return something that the caller can use?

sxu123: This function constructs a query for ES.  If we pass null for the value, and the catch the exception, we still have to return a query that will guarantee NOT to return anything, which is just like the one you printed out above.

sxu123: If you want to catch exception in the code that calls reduce-query-condition, one level up is reduce-query in cmr.common-app.services.search.complex-to-simple, you still have to produce a query that returns nothing(same problem)  Another level up, (defmethod execute-query :elasticsearch in cmr.common-app.services.search.query-execution calls reduce-query, but this function is used by many type of query conditions. I can't just catch an exception and return query result with 0 hit for all cases(I don't know what other issues there might be).  In short, it has to be handled locally in cmr.search.data.complex-to-simple-converters.has-granules-or-cwic, and since we know when concept-ids are empty, there is exception thrown, there is no point causing that exception.

eereiter: Thanks, I guess I am fine with it.

eereiter (approved):

search-app/src/cmr/search/data/complex_to_simple_converters/has_granules_or_cwic.clj
```diff
@@ -18,9 +18,15 @@
...
           or-cwic-concept-ids (map key (filter val has-granules-or-cwic-map))
-          or-cwic-condition (cqm/string-conditions :concept-id or-cwic-concept-ids true)]
+          or-cwic-condition (if (seq or-cwic-concept-ids)
+                              (cqm/string-conditions :concept-id or-cwic-concept-ids true)
+                              (cqm/string-conditions :concept-id ["non-existing"] true))]
```
jaybarra, line 29: Why not use `when` if the second condition will always return true? If there are no values, we shouldn't add the condition to query just to force something

jaybarra (changes requested):

search-app/src/cmr/search/data/complex_to_simple_converters/has_granules_or_cwic.clj
sxu123: when won't work, but match-none will.  Thanks.

jaybarra (approved):

PR #1390, opened 2021-11-19
CMR-7876 Fixing bug with FreeAndOpenData with a false value.

eereiter:


umm-spec-lib/test/cmr/umm_spec/test/umm_to_xml_mappings/echo10.clj
```diff
@@ -118,12 +118,18 @@
       (is (= "Description" (value-of echo-result "Description")))
       (is (= nil (value-of echo-result "FreeAndOpenData")))))
 
-  (testing "echo10 use constraints FreeAndOpenData test"
+  (testing "echo10 use constraints FreeAndOpenData test true"
```
jaybarra, line 121: Could combine with an `are3`

jaybarra (approved):

mitchstartzel (approved):

umm-spec-lib/test/cmr/umm_spec/test/umm_to_xml_mappings/echo10.clj
eereiter: I could, there are several old tests that I could have changed, but I chose not to in this case, because of time.

PR #1395, opened 2021-11-24
CMR-6848: Read CWIC from consortium for has_granules_or_cwic.

sxu123:


eereiter (approved):

search-app/docs/api.md
```diff
@@ -1578,7 +1578,7 @@ Find collections that match all of the 'project' param values
 This supports `pattern`, `ignore_case` and option `and`.
 
 Find collections matching 'consortium' param value
-CWIC, FEDEO, CEOS, EOSDIS
```
jceaser, line 1581: why remove example values, it's going to be hard enough to explain what a consortium is, having these names would help.

indexer-app/src/cmr/indexer/data/concepts/collection.clj
```diff
@@ -211,11 +211,11 @@
     (let [free-and-open (:FreeAndOpenData use-constraints)]
       (case free-and-open
         true (distinct (conj consortiums "GEOSS"))
-        false (remove #(= "GEOSS" (str/upper-case %)) consortiums)
+        false (remove #(= "GEOSS" %) consortiums)
```
jceaser, line 214: why do we want to make any code case sensitive? I see the change on line 232, but do we ever expect to have a 'geoss' that is different from 'GEOSS'?

sxu123: It's not case sensitive. consortiums is already upper-cased in the index.

search-app/docs/api.md
sxu123: This is just to fix the bug Matt mentioned in the documentation. Somehow the example in the line below it didn't show the escaped bracket, even though in the code the escape was there. so I just cleaned up the line and made it exactly the same as project example above.  I didn't really want to put these values there I forgot why I did. it's not complete and not directly describing the example below.

jceaser: I see there are some URLs with example values, that should work for my needs

jceaser (approved):

PR #1397, opened 2021-12-01
Cmr 6847 read open search from consortiums

sxu123:


search-app/src/cmr/search/services/query_execution/has_granules_or_cwic_results_feature.clj
```diff
@@ -38,9 +46,32 @@
...
             [coll-id 1]))))
 
+(defn get-opensearch-collections
+  "Returns the collection granule count by searching elasticsearch by aggregation"
+  [context provider-ids]
+  (let [condition (gc/or-conds [(qm/string-conditions :consortiums ["CWIC"])
+                                (qm/string-conditions :consortiums ["FEDEO"])
+                                (qm/string-conditions :consortiums ["GEOSS"])
+                                (qm/string-conditions :consortiums ["CEOS"])
+                                (qm/string-conditions :consortiums ["EOSDIS"])])
```
jaybarra, line 56: Possibly add a `defconfig` someplace with a default list of consortiums instead of having to list them out individually all the time

jaybarra (approved):

search-app/docs/api.md
```diff
@@ -2119,7 +2119,7 @@ The `has_granules_or_cwic` parameter can be set to "true" or "false". When true,
 
 #### <a name="c-has-granules-or-opensearch"></a> Find collections with or without granules, or the collection is tagged with the configured OpenSearch tag.
 
-The `has_granules_or_opensearch` parameter can be set to "true" or "false". When true, the results will be restricted to collections with granules or with the configured OpenSearch tag.  When false, will return any collections without granules.
+The `has_granules_or_opensearch` parameter can be set to "true" or "false". When true, the results will be restricted to collections with granules or with any of the configured OpenSearch consortiums, which are CWIC,FEDEO,GEOSS,CEOS and EOSDIS.  When false, will return any collections without granules.
```
ygliuvt, line 2122: I don't think when `has_granules_or_opensearch` is false, we will return any collections that are without granules. It should have returned any collections that are without granules and are also not opensearch. What is the logic behind your decision here?

ygliuvt (changes requested):
Need clarification on what should be returned when `has_granules_or_opensearch` is false and make the code change accordingly.

search-app/docs/api.md
sxu123: I'm sorry Leo, but this has been the behavior for has_granules_or_cwic, i.e. when false, it returns all the collections without granules, regardless if they have cwic or not. and has_granules_or_opensearch has been the alias for has_granules_or_cwic. I have questioned car-dev for this because this is so against common logic.  But Daniel, who was the initial developer for this has pointed me to the documentation that specifically mentioned this and the system integration testing confirmed it.  He remember he was very annoyed by this requirement.   So, I think if you want this changed, you should file a new ticket after finding out why this was done this way before.

ygliuvt: Can you file a ticket to find out what the reason is? We should at least document the reason in the API doc to explain why it is done in this unreasonable way.

ygliuvt: I am ok with approving this PR if you put the ticket number here.

sxu123: CMR-7919

ygliuvt (approved):
Approved as we filed CMR-7919 to address the illogical behavior of has_granules_or_opensearch=false scenario.

PR #1412, opened 2022-01-06
Cmr 7840 allow non space delimiters

sxu123:


jaybarra (approved):

ygliuvt (changes requested):
I don't see the point of supporting all non-word characters as delimiter for consortiums, and the rule is not documented anywhere is unacceptable. I think we should clearly document the format of consortium string in the README where we document the provider creation/update API. 

My recommendation is a comma separated string of consortiums. CMR can trim out the leading/trailing whitespace in each consortium in the input and index them.

sxu123:
@ygliuvt, Currently we use space as a delimiter, you can use as many spaces as you want. As to where to document the delimiter info, we are not sure.  Erich thinks some people might want to use comma, others might want to use semicolon; etc....  According to @eereiter, all the consortium values will only consist of alphabet, numbers and underscores, which implies anything else could be used as delimiters, doesn't mean the operators will use them, but if they do, they can all be parsed correctly. I will leave this to @eereiter to decide.  I'm all for this approach.  Using comma as a delimiter is no difference of using space.

PR #1415, opened 2022-01-14
CMR-8009 - renaming configuration variable name and adding a log

jceaser:
making the KMS schema override setting a standard defconf instead of trying to use a direct ENV variable. Adding some logs to show what CMR is doing in SIT.

### DIFF
--- a/transmit-lib/src/cmr/transmit/config.clj
+++ b/transmit-lib/src/cmr/transmit/config.clj
@@ -96,14 +96,14 @@
 (defconfig kms-scheme-override-json
   "Allow for KMS schema urls to be overriden by setting this value to a JSON
    strings such as:
-   {\"platforms\":\"file://frozen/platforms.csv\","
-    "\"mime-type\":\"mimetype?format=csv&version=special\"}
+   {\"platforms\":\"file://frozen/platforms.csv\",
+    \"mime-type\":\"mimetype?format=csv&version=special\"}
     The reason this is needed is because CMR can only talk to the production KMS
     server. SIT, UAT, and Production may all need to be changed to KMS schema
     versions needed for these specific envirnments to overcome this limitation
     and to present testing data as required."
-  :default ""
-  :parser #(json/parse-string % true))
+  {:default ""
+  :parser #(json/parse-string % true)})
 
 (defconfig urs-username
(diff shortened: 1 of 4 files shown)


transmit-lib/src/cmr/transmit/kms.clj
```diff
@@ -239,10 +239,14 @@
   ;;     "{\"platforms\":\"file://frozen/platforms.csv\"}"
   (let [gcmd-resource-name (keyword-scheme->kms-resource keyword-scheme)]
     (if (str/starts-with? gcmd-resource-name "file://")
-      (->> (str/replace gcmd-resource-name #"^file://" "")
+      (do
+        (warn (format "Loading static KMS resource from %s for %s."
```
mitchstartzel, line 243: I think this `warn` as well as the one on line 258 would make more sense as (`info`...)

transmit-lib/src/cmr/transmit/config.clj
```diff
@@ -92,6 +92,17 @@
...
+(defconfig kms-scheme-override-json
+  "Allow for KMS schema urls to be overriden by setting this value to a JSON
+   strings such as:
+   {\"platforms\":\"file://frozen/platforms.csv\","
+    "\"mime-type\":\"mimetype?format=csv&version=special\"}
+    The reason this is needed is because CMR can only talk to the production KMS
+    server. SIT, UAT, and Production may all need to be changed to KMS schema
+    versions needed for these specific envirnments to overcome this limitation
+    and to present testing data as required."
+  "")
```
mitchstartzel, line 104: The definition for `defmacro` states that the `:default` option is required. Additionally, maybe you can make use of a `:parser` here?

```
(defmacro defconfig
...
Options:

  * :default - Sets the default value if none is provided. Required.
  * :type - Sets the type of value the configuration parameter will have. The type will be used
  to determine how to parse the environment variable value.
  * :parser - A custom parser function to use to parse the environment variable value.
 ```

mitchstartzel (changes requested):

transmit-lib/src/cmr/transmit/config.clj
jceaser: o I missed the default part, and a I'll try the :parser setting.

eereiter (approved):

PR #1422, opened 2022-01-21
cmr-8009 - redesign/simplification - pushing more behavior to configuration setting

jceaser:
After a review with another CMR member it was decided to simplify this code by removing the second override case and pushing those settings to configuration. The "frozen" use case has been renamed to "static" and no longer uses the string replace function which could cause unexpected behaviors. Instead internal paths are built based on know values. CMR now has only the production/default paths for KMS and allows external overrides as needed for other environments.

Static file also moved from transmit app to indexer app where they can be found by the indexer cache refresh process.


eereiter (approved):

transmit-lib/src/cmr/transmit/config.clj
```diff
@@ -96,7 +96,7 @@
 (defconfig kms-scheme-override-json
   "Allow for KMS schema urls to be overriden by setting this value to a JSON
    strings such as:
-   {\"platforms\":\"file://frozen/platforms.csv\",
+   {\"platforms\":\"static\",
```
ygliuvt, line 99: Add comment on what `static` means, i.e. this indicates the schema is loaded from CMR source code statically and also where the static files are located.

transmit-lib/src/cmr/transmit/kms.clj
```diff
@@ -50,27 +50,18 @@
...
-
-  ;; parse and return a map
-  ([json-string] (json/parse-string json-string true)))
-
-(comment
- ;; testing with an env is not convenient, use this to diretly test what the
- ;; function would do in the real world with a valid value
- (scheme-overrides (str "{\"platforms\":\"file://frozen/platforms.csv\","
-                        "\"mime-type\":\"mimetype?format=csv&version=special\"}")))
+  "CMR will Allow for any KMS resource URL to be overriden by a config
```
ygliuvt, line 61: Update the comment to say like "Returns the KMS schema override map as configured in CMR_KMS_SCHEMA_OVERRIDE_JSON."

ygliuvt (approved):
You need to resolve the merge conflict before you can merge though.

PR #1431, opened 2022-01-26
CMR-7714: Documented the missing error format.

sxu123:


ingest-app/docs/api.md
```diff
@@ -226,11 +226,14 @@ Requests could fail for several reasons when communicating with the CMR as descr
 
 Ingest validation errors can take one of two shapes. General error messages will be returned as a list of error messages like the following:
```
ygliuvt, line 227: I think `Ingest validation errors can take one of two forms.` should be above the General Error section.

sxu123: done.

ygliuvt (approved):

PR #1432, opened 2022-01-26
Cmr 6537 4

isja17:
Updated json to return with underscores instead of hyphens, updated json schema, updated api.


search-app/test/cmr/search/test/results_handlers/subscriptions_json_results_handler.clj
```diff
@@ -0,0 +1,26 @@
+(ns cmr.search.test.results-handlers.subscriptions-json-results-handler
```
ygliuvt, line 1: Not see much value in this test. Delete it.

system-int-test/test/cmr/system_int_test/search/subscription/subscription_search_test.clj
```diff
@@ -527,3 +527,29 @@
       "Sort by name then provider id descending order"
       ["-name" "-provider"]
       [subscription2 subscription1 subscription4 subscription3])))
+
+(deftest json-result-format-test
```
ygliuvt, line 531: We could delete this test and fold the key checking into the `cmr.system-int-test.utils.subscription-util/search-json` function by modifying `search/process-response`.

search-app/resources/schema/subscription_resource_schema.json
```diff
@@ -0,0 +1,56 @@
+{
```
ygliuvt, line 1: The subscription search result JSON schema should be defined in: [EOSDIS Metadata Format Depot](https://git.earthdata.nasa.gov/projects/EMFD/repos/unified-metadata-model/browse/subscription), e.g. v1.1 is [here](https://git.earthdata.nasa.gov/projects/EMFD/repos/unified-metadata-model/browse/subscription/v1.1/umm-sub-search-results-json-schema.json). Remove this file.

ygliuvt (changes requested):
Need to push the result JSON schema to EOSDIS Metadata Format Depot and address the review comments.

search-app/test/cmr/search/test/results_handlers/subscriptions_json_results_handler.clj
jaybarra: The unit level tests are good for development and without needing the entire system running.

ygliuvt: This test basically tests that I used these keys to construct a map and verify that these keys exist in the map that is returned. It is too trivial. And these keys in the map is in certain format does not guarantee that the final JSON response key in that format. That is why I recommended doing the check in the system-int-test rather than in the unit test. If there is test in system-int-test, this unit test will become even more unnecessary. Only when we don't add the verification of key format in the system-int-test, does this test have some value.

jaybarra: System integration tests do not count toward code coverage. Unit tests do, this assists with maintaining required code coverage levels.

system-int-test/test/cmr/system_int_test/search/subscription/subscription_search_test.clj
jaybarra: process-response specifically modifies the response to use kebab-cases for all map keys. Having that be aware of what type of query makes less sense than comparing an unprocessed response available.

search-app/test/cmr/search/test/results_handlers/subscriptions_json_results_handler.clj
ygliuvt: If our goal of higher code coverage is to have a higher code coverage number, I have nothing. :)

search-app/resources/schema/subscription_resource_schema.json
isja17: done, those changes have been merged there.

system-int-test/src/cmr/system_int_test/utils/subscription_util.clj
```diff
@@ -110,18 +111,34 @@
...
+  (is (= #{:collection_concept_id
+           :concept_id
+           :name
+           :native_id
+           :provider_id
+           :revision_id
+           :subscriber_id}
+         (set (keys subscription))))
+  ;; verifying no dash in keys
+  (is (not-any? #(string/includes? % "-") (map name (keys subscription)))))
```
ygliuvt, line 126: If we have the stricter check on the line above, we don't really need this line here. My original suggestion is to only have this check, not he check above. That way, we don't need to make code change when say in the future, we add `additional_attribute` in the subscription response.

ygliuvt (approved):

jaybarra (approved):

system-int-test/src/cmr/system_int_test/utils/subscription_util.clj
ygliuvt: I want you to delete line 117.

PR #1437, opened 2022-01-31
CMR-7798 updated ingest documentation to say UMM JSON version is required.

isja17:
…ired.

### DIFF
--- a/ingest-app/docs/api.md
+++ b/ingest-app/docs/api.md
@@ -87,7 +87,7 @@ Content-Type is a standard HTTP header that specifies the content type of the bo
 | application/iso:smap+xml          | ISO 19115 SMAP    | collection, granule |
 | application/vnd.nasa.cmr.umm+json | UMM JSON          | collection, granule, variable, service, subscription, tool |
 
-Note: UMM JSON accepts an additional version parameter for both `Content-Type` and `Accept` headers. Like charset, it is appended with a semicolon (;). If no version is appended, the latest version is assumed.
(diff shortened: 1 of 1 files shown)


jaybarra (approved):

ingest-app/docs/api.md
```diff
@@ -87,7 +87,7 @@ Content-Type is a standard HTTP header that specifies the content type of the bo
 | application/iso:smap+xml          | ISO 19115 SMAP    | collection, granule |
 | application/vnd.nasa.cmr.umm+json | UMM JSON          | collection, granule, variable, service, subscription, tool |
 
-Note: UMM JSON accepts an additional version parameter for both `Content-Type` and `Accept` headers. Like charset, it is appended with a semicolon (;). If no version is appended, the latest version is assumed.
+Note: UMM JSON accepts an additional version parameter for both `Content-Type` and `Accept` headers. Like charset, it is appended with a semicolon (;). UMM JSON version is required.
```
ygliuvt, line 90: I know you didn't do this, but I don't think we should mention `Accept` header in the `Content-Type` header section at all. This note is only about the `Content-Type` header.

isja17: I can remove this. In the accept header section, do you want me to say version is required?

ygliuvt (approved):

ingest-app/docs/api.md
ygliuvt: We need to put a similar note on the Accept header section. The only difference is UMM JSON version is not required in Accept header. If no version is appended, the latest version is assumed.

PR #1440, opened 2022-02-03
CMR-7854: Do not update subsription last-notified-at when manually calling subscription job

brianalexander:
Previous PR: #1426 
The last pull request had the side effect of updating last-notified-at when run.  This could move the last-notified-at to a time that is before its current value.  Moving last-notified-at backwards will result in duplicate subscription emails being sent on the next run of the subscription job.  To avoid this, last-notified-at is no longer updated when called manually.

additional changes:

No longer mocking `send-subscription-emails` as I want to test that functionality.  I abstracted `postal-core/send-message` into the local function `send-email` so that it can be mocked.  This simplifies and shrinks the mock space to its smallest possible amount.

`send-subscription-emails` now also returns the subscriber granule reference list `subscriber-filtered-gran-refs-list ` which is necessary to keep up with previous behavior in other tests.


brianalexander:
#<1426>

system-int-test/test/cmr/system_int_test/ingest/subscription_processing_test.clj
```diff
@@ -40,19 +48,12 @@
     (index/wait-until-indexed)
     result))
 
-(defn- mock-send-subscription-emails
+(defn- mock-send-email
   "This function is used along with with-redefs to avoid sending emails in
-   integration tests. If send-subscription-emails is called in tests without being mocked,
+   integration tests. If send-subscription-emails is called in tests without send-email being mocked,
    errors while be returned when attempting to connect to the mail server in
```
ygliuvt, line 54:
```suggestion
   errors will be returned when attempting to connect to the mail server in 
```

ygliuvt (approved):

jaybarra (approved):

PR #1446, opened 2022-02-04
CMR-8022

ygliuvt:
Add support for authenticated requests to the Preview Page.

### DIFF
--- a/search-app/docs/api.md
+++ b/search-app/docs/api.md
@@ -388,10 +388,15 @@ out will include the `CMR-Time-Out` header set to `true`.
 
 #### <a name="html"></a> HTML
 
-The HTML response format is supported for collections. It allows a single collection record to be viewed in a web browser. HTML is only supported for retrieving a single collection at a time with a URL of the format:
+The HTML response format is intended to be used only in a web browser to view a single collection record. The collection html representation can be viewed by putting the following in the web browser address bar.
 
 ```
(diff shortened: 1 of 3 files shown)


search-app/docs/api.md
```diff
@@ -388,11 +388,16 @@ out will include the `CMR-Time-Out` header set to `true`.
...
-The HTML response format is supported for collections. It allows a single collection record to be viewed in a web browser. HTML is only supported for retrieving a single collection at a time with a URL of the format:
+The HTML response format is intended to be used only in a web browser to view a single collection record. The collection html representation can be viewed by putting the following in the web browser address bar.
 
 ```
 %CMR-ENDPOINT%/concepts/<concept-id>
 ```
+For private collection, an EDL bearer token can be used to grant permission. e.g.
+
+```
+%CMR-ENDPOINT%/concepts/<concept-id>?token=EDL-xxxxxx
```
jaybarra, line 399: Do we need to add filtering to catch this use case to obfuscate tokens in request logs?

jaybarra (approved):

search-app/docs/api.md
ygliuvt: We don't do this for any CMR use cases where token is provided in params. Maybe we should add that filtering.

eereiter (approved):

PR #1449, opened 2022-02-04
CMR-8046 - a bug fix to isolate parse errors which can happen if CSV …

jceaser:
Background: If KMS changes it's structure and CMR team does not get notified, CMR will fail to parse the new CSV file. When this happens, CMR use to throw an exception from inside a for loop where CMR integrates over schema names (platforms, instruments, science keywords) to download CSV files to process. When this happens nothing catches the exception and the for loop stops processing schemes meaning all schemes with valid data which happens latter in the loop do not get processed.

Fix: Don't throw exception. Instead, return true/false in a tester function and handle that state latter on. I also added some tests since we need to increase code coverage anyways.

### DIFF
--- a/transmit-lib/src/cmr/transmit/kms.clj
+++ b/transmit-lib/src/cmr/transmit/kms.clj
@@ -63,7 +63,12 @@
    Return value is the KMS schema override map as configured in AWS as
    CMR_KMS_SCHEMA_OVERRIDE_JSON."
   []
-  (json/parse-string (config/kms-scheme-override-json) true))
+  (try
+    (json/parse-string (config/kms-scheme-override-json) true)
+    (catch com.fasterxml.jackson.core.JsonParseException e
+      (error "Failed to parse Scheme Override JSON while loading KMS resource [%s]: %s",
+             (config/kms-scheme-override-json)
+             (.getMessage e)))))
 
 ;; These are the default locations in KMS for all supported schemes
 (def keyword-scheme->gcmd-resource-name
@@ -166,7 +171,7 @@
   "Number of lines which contain header information in csv files (not the actual keyword values)."
   2)
 
-(defn- validate-subfield-names
+(defn- subfield-names-valid?
   "Validates that the provided subfield names match the expected subfield names for the given
   keyword scheme. Returns true if they match, false otherwise."
   [keyword-scheme subfield-names]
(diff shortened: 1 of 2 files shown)


transmit-lib/src/cmr/transmit/kms.clj
```diff
@@ -168,42 +168,45 @@
 
 (defn- validate-subfield-names
```
ygliuvt, line 169: change the function name to something like `subfield-names-valid?`

ygliuvt (approved):

eereiter (approved):

transmit-lib/test/cmr/transmit/test/kms.clj
```diff
@@ -77,13 +77,57 @@
           actual (#'cmr.transmit.kms/parse-entries-from-csv :mime-type sample-csv-mimetype)]
       (is (= expected actual))))
 
-  (testing "Invalid subfield names in the CSV throws an exception"
-    (is (thrown-with-msg?
-          Exception
-          #"Expected subfield names for instruments to be"
-          (#'cmr.transmit.kms/parse-entries-from-csv :instruments sample-csv)))))
+  (testing "Invalid subfield names in the CSV returns nothing"
+    (is nil? (#'cmr.transmit.kms/parse-entries-from-csv :platforms sample-csv))))
```
jaybarra, line 81:
```suggestion
  (testing "Invalid subfield names in the CSV returns nothing"
    (is (nil? (#'cmr.transmit.kms/parse-entries-from-csv :platforms sample-csv)))))
```
```

**Review by jaybarra — CHANGES_REQUESTED** [review]

File: transmit-lib/test/cmr/transmit/test/kms.clj [hunk 3]
```diff
@@ -77,13 +77,57 @@
...
+    (cmr.transmit.config/set-kms-scheme-override-json! "{}")
+    (let [result (#'cmr.transmit.kms/scheme-overrides)]
+      (is (= {} result))))
+
+  (testing "No configuration value should return nil"
+    ;; doing this test last so that the setting goes back to the default for
+    ;; tests that follow this call.
+    (cmr.transmit.config/set-kms-scheme-override-json! "")
+    (let [result (#'cmr.transmit.kms/scheme-overrides)]
+      (is (= nil result)))))
```
jaybarra, line 119:
```suggestion
      (is (nil? result)))))
```

transmit-lib/test/cmr/transmit/test/kms.clj
```diff
@@ -77,13 +77,57 @@
...
+
```
jaybarra, line 120: Whitespace, extra newline

jaybarra, line 120:
```suggestion
```

transmit-lib/test/cmr/transmit/test/kms.clj
jceaser: I rewrote this section to use are3

transmit-lib/test/cmr/transmit/test/kms.clj
jceaser: thanks

jaybarra (approved):

PR #1452, opened 2022-02-08
CMR-6928 Subscription limit validation and tests

dpesall:
This change limits the number subscriptions an individual user can create to 100 active subscriptions. If the user is already at the subscription limit, they will not be able to successfully create a new subscription. This will help prevent undue load or abuse. The tests validate a user can create subscriptions if they are below the limit and that they will be rejected if they are already at the limit.

### DIFF
--- a/ingest-app/src/cmr/ingest/api/subscriptions.clj
+++ b/ingest-app/src/cmr/ingest/api/subscriptions.clj
@@ -94,17 +94,16 @@
                        {:subscriber-id (:SubscriberId metadata)
                         :latest true}
                        :subscription)
-        active-subscriptions (remove :deleted subscriptions)]
-     (def subscription-revision false)
-     (doseq [element active-subscriptions]
-      (if (and (= (:native-id element) (:native-id subscription))
-               (= (:provider-id element) (:provider-id subscription)))
-          (def subscription-revision true)))
-     (when (and (>= (count active-subscriptions) (jobs/subscriptions-limit)) (not subscription-revision))
-       (errors/throw-service-error
-        :conflict
-        (format "The subscriber-id [%s] has already reached the subscription limit."
-                subscriber-id)))))
+        active-subscriptions (remove :deleted subscriptions)
+        at-limit (>= (count active-subscriptions) (jobs/subscriptions-limit))]
+    (when at-limit
+      (let [pred #(and (= (:native-id %) (:native-id subscription)) (= (:provider-id %) (:provider-id subscription)))
+            is-not-revision (empty? (filter pred active-subscriptions))]
+        (when is-not-revision
+          (errors/throw-service-error
+           :conflict
+           (format "The subscriber-id [%s] has already reached the subscription limit."
+                   subscriber-id)))))))
 
 (defn- check-duplicate-subscription
   "The query used by a subscriber for a collection should be unique to prevent
--- a/ingest-app/src/cmr/ingest/api/subscriptions.clj
+++ b/ingest-app/src/cmr/ingest/api/subscriptions.clj
@@ -82,6 +82,24 @@
                              query-params)
         gran-refs (search-gran-refs-with-sub-params request-context search-params)]))
 
+(defn- check-subscription-limit
+  "Given the  configuration for subscription limit, this valdiates that the
+  user has less than the limit before we allow more subscriptions to be
+  ingested by that user."
+  [request-context subscription]
+  (let [metadata (-> (:metadata subscription) (json/decode true))
+        subscriber-id (:SubscriberId metadata)
+        subscriptions (mdb/find-concepts
+                       request-context
+                       {:subscriber-id (:SubscriberId metadata)}
+                       :subscription)
+        active-subscriptions (remove :deleted subscriptions)]
+     (when (>= (count active-subscriptions) (jobs/subscriptions-limit))
+       (errors/throw-service-error
+        :conflict
+        (format "The subscriber-id [%s] has already reached the subscription limit."
+                subscriber-id)))))
+
 (defn- check-duplicate-subscription
   "The query used by a subscriber for a collection should be unique to prevent
    redundent emails from being sent to them. This function will check that a
@@ -111,9 +129,9 @@
                 (every? #(not= native-id (:native-id %)) active-duplicate-queries))
        (errors/throw-service-error
         :conflict
-        (format (str "The subscriber-id [%s] has already subscribed to the "
-                     "collection with concept-id [%s] using the query [%s]. "
-                     "Subscribers must use unique queries for each Collection.")
+        (format "The subscriber-id [%s] has already subscribed to the "
+                "collection with concept-id [%s] using the query [%s]. "
+                "Subscribers must use unique queries for each Collection."
                 subscriber-id collection-id  normalized-query)))))
 
 (defn- get-subscriber-id
@@ -252,6 +270,7 @@
       (check-ingest-permission request-context provider-id subscriber-id)
       (validate-query request-context final-sub)
       (check-duplicate-subscription request-context final-sub)
+      (check-subscription-limit request-context final-sub)
       (perform-subscription-ingest request-context final-sub headers))))
 
 (defn create-subscription-with-native-id
@@ -277,6 +296,7 @@
       (check-ingest-permission request-context provider-id subscriber-id)
       (validate-query request-context final-sub)
       (check-duplicate-subscription request-context final-sub)
+      (check-subscription-limit request-context final-sub)
       (perform-subscription-ingest request-context final-sub headers))))
 
 (defn create-or-update-subscription-with-native-id
@@ -306,6 +326,7 @@
     (check-ingest-permission request-context provider-id new-subscriber old-subscriber)
     (validate-query request-context final-sub)
     (check-duplicate-subscription request-context final-sub)
+    (check-subscription-limit request-context final-sub)
     (perform-subscription-ingest request-context final-sub headers)))
 
 (defn delete-subscription
--- a/ingest-app/src/cmr/ingest/api/subscriptions.clj
+++ b/ingest-app/src/cmr/ingest/api/subscriptions.clj
@@ -82,6 +82,14 @@
                              query-params)
         gran-refs (search-gran-refs-with-sub-params request-context search-params)]))
 
+(defn is-subscription-revision?
+  "True if the subscription is a revision, otherwise false."
+  [subscription active-subscriptions]
+  (let [pred #(and
+                (= (:native-id %) (:native-id subscription))
+                (= (:provider-id %) (:provider-id subscription)))]
+    (boolean (seq (filter pred active-subscriptions)))))
+
 (defn- check-subscription-limit
   "Given the  configuration for subscription limit, this valdiates that the
   user has less than the limit before we allow more subscriptions to be
@@ -95,15 +103,13 @@
                         :latest true}
                        :subscription)
         active-subscriptions (remove :deleted subscriptions)
-        at-limit (>= (count active-subscriptions) (jobs/subscriptions-limit))]
-    (when at-limit
-      (let [pred #(and (= (:native-id %) (:native-id subscription)) (= (:provider-id %) (:provider-id subscription)))
-            is-not-revision (empty? (filter pred active-subscriptions))]
-        (when is-not-revision
-          (errors/throw-service-error
-           :conflict
-           (format "The subscriber-id [%s] has already reached the subscription limit."
-                   subscriber-id)))))))
+        at-limit (>= (count active-subscriptions) (jobs/subscriptions-limit))
+        is-not-revision #(not (is-subscription-revision? subscription active-subscriptions))]
+    (when (and at-limit (is-not-revision))
+      (errors/throw-service-error
+       :conflict
+       (format "The subscriber-id [%s] has already reached the subscription limit."
+               subscriber-id)))))
 
 (defn- check-duplicate-subscription
   "The query used by a subscriber for a collection should be unique to prevent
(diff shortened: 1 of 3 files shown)


ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -111,9 +129,9 @@
                 (every? #(not= native-id (:native-id %)) active-duplicate-queries))
        (errors/throw-service-error
         :conflict
-        (format (str "The subscriber-id [%s] has already subscribed to the "
-                     "collection with concept-id [%s] using the query [%s]. "
-                     "Subscribers must use unique queries for each Collection.")
+        (format "The subscriber-id [%s] has already subscribed to the "
+                "collection with concept-id [%s] using the query [%s]. "
+                "Subscribers must use unique queries for each Collection."
                 subscriber-id collection-id  normalized-query)))))
```
jaybarra, line 135: format needs to take a single string as the first argument with the replacements passed in.

jaybarra (changes requested):

ingest-app/src/cmr/ingest/api/subscriptions.clj
dpesall: Thank Jay, should be formatted properly now.

jaybarra (approved):

system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
```diff
@@ -43,6 +44,58 @@
...
+               (data-umm-c/collection
+                {:ShortName "coll-exceed-limit-2"
+                 :EntryTitle "entry-title-exceed-limit-2"})
+               {:token "mock-echo-system-token"})
+        coll3 (data-core/ingest-umm-spec-collection
+               "PROV1"
+               (data-umm-c/collection
+                {:ShortName "coll-exceed-limit-3"
+                 :EntryTitle "entry-title-exceed-limit-3"})
+               {:token "mock-echo-system-token"})]
```
brianalexander, line 66: I'm pretty sure it's possible to have multiple subscriptions on the same collection (as long as they have different queries).  I think it could be cleaner to drop these 2 collections and then just specify queries when calling `make-subscription-concept` (if you search in the same file for ":Query" you can see an example).  It's not actually important what the query is for the tests as long as they are valid queries.

system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
```diff
@@ -43,6 +44,58 @@
...
+     (testing "ingest on PROV1 with subscription request for subscriber-id below subscription limit"
+       (let [concept (subscription-util/make-subscription-concept {:provider-id "PROV1"
+                                                                   :Name "Sub1"
+                                                                   :CollectionConceptId (:concept-id coll1)
+                                                                   :SubscriberId "user1"
+                                                                   :EmailAddress "foo@example.com"})
```
brianalexander, line 72: We can drop `:provider-id`, `:EmailAddress`, and `:Name` from each call to `make-subscription-concept` as they aren't necessary/relevant for the test.

dpesall: I dropped most of the fields, but I found when I dropped :Name as well it would appear no longer create new subscriptions but rather modify previous subscriptions. Without :Name the tests to create a subscription returned code 200 OK, whereas they properly returned 201 Created with :Name specified.

system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
dpesall: Thanks Brian, I removed coll2 and coll3 and just added all the test subscriptions to coll1 with their own queries.

system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
```diff
@@ -43,6 +44,58 @@
...
+             user1-token (echo-util/login (system/context) "user1")
```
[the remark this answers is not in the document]
dpesall: Refactored and looks better!

ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -82,6 +82,24 @@
...
+  "Given the  configuration for subscription limit, this valdiates that the
+  user has less than the limit before we allow more subscriptions to be
+  ingested by that user."
+  [request-context subscription]
+  (let [metadata (-> (:metadata subscription) (json/decode true))
+        subscriber-id (:SubscriberId metadata)
+        subscriptions (mdb/find-concepts
+                       request-context
+                       {:subscriber-id (:SubscriberId metadata)}
+                       :subscription)
```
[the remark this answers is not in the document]
dpesall: Created test `subscription-count-limit-ignores-old-subscriptions-test` and you were spot on with thinking revised subscriptions would count against their subscription count. Updated my query in `subscriptions.clj` to only specify `latest true`.

`subscription-count-limit-ignores-old-subscriptions-test` validates that revised queries don't count against the subscription limit. I initially had this inside the other test `subscription-count-exceeds-limit-test` but I felt the clarity of the test was lost a bit so I split them up, but you can let me know if you think they should be in one test.

system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
```diff
@@ -43,6 +44,74 @@
...
+        (is (= 200 (:status response)))
+        (is (mdb/concept-exists-in-mdb? (:concept-id response) (:revision-id response)))))
+     (testing "ingest on PROV1 with subscription request for subscriber-id below subscription limit"
+       (let [concept (subscription-util/make-subscription-concept {:Query "platform=NOAA-7"
+                                                                   :Name "sub2"
+                                                                   :CollectionConceptId (:concept-id coll1)
+                                                                   :SubscriberId "user1"})
+             response (ingest/ingest-concept concept {:token user1-token})]
+        (is (= 201 (:status response)))
+        (is (mdb/concept-exists-in-mdb? (:concept-id response) (:revision-id response)))))))
```
brianalexander, line 79: I think we only need 1 `testing` in here.  Something like.. `(testing "revisions are not included when calculating subscription count" ...)`

You could also move it `subscription-count-exceeds-limit-test` after the first test.  Create a revision of "sub1" and since you've already set the subscription limit to 1, it should succeed.   You can verify both revisions exist using `(is (mdb/concept-exists-in-mdb? (:concept-id response) 1))` and `(is (mdb/concept-exists-in-mdb? (:concept-id response) 2))` (I think revision-id starts from 1).

system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
```diff
@@ -43,6 +44,74 @@
...
+(deftest subscription-count-exceeds-limit-test
+  (side/eval-form `(jobsub/set-subscriptions-limit! 1))
+  (let [coll1 (data-core/ingest-umm-spec-collection
+               "PROV1"
+               (data-umm-c/collection
+                {:ShortName "coll-exceed-limit-1"
+                 :EntryTitle "entry-title-exceed-limit-1"})
+               {:token "mock-echo-system-token"})
+        user1-token (echo-util/login (system/context) "user1")]
+     (testing "ingest on PROV1 with subscription request for subscriber-id below subscription limit" ;zzz
```
brianalexander, line 90: remove `;zzz`

how about renaming to "succeeds when ingesting subscription for user below their subscription limit"

system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
```diff
@@ -43,6 +44,74 @@
...
+       (let [concept (subscription-util/make-subscription-concept {:Query "platform=NOAA-7"
+                                                                   :Name "sub1"
+                                                                   :CollectionConceptId (:concept-id coll1)
+                                                                   :SubscriberId "user1"})
+             response (ingest/ingest-concept concept {:token user1-token})]
+         (is (= 201 (:status response)))
+         (is (mdb/concept-exists-in-mdb? (:concept-id response) (:revision-id response)))))
+     (testing "ingest on PROV1 with subscription request for subscriber-id already at subscription limit"
```
brianalexander, line 98: how about renaming to "fails with correct response when ingesting subscription for user beyond their subscription limit"

system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
```diff
@@ -43,6 +44,74 @@
...
+       (let [concept (subscription-util/make-subscription-concept {:Query "platform=NOAA-9"
+                                                                   :Name "sub2"
+                                                                   :CollectionConceptId (:concept-id coll1)
+                                                                   :SubscriberId "user1"})
+             response (ingest/ingest-concept concept {:token user1-token})]
+        (is (= 409 (:status response)))
+        (is (= "The subscriber-id [user1] has already reached the subscription limit." (first (:errors response))))))
+    (side/eval-form `(jobsub/set-subscriptions-limit! 100))
+    (testing "ingest on PROV1 with subscription request for subscriber-id below updated subscription limit"
```
brianalexander, line 107: I'm not sure this test provides value.  

When the limit is 1, you proved... 

1.  it succeeds at ingesting 1 subscription
2. it succeeds when you have 1 subscription with multiple revisions
3. it fails when you try to create more subscriptions than the limit.

Could probably get rid of it.

brianalexander: You could add a test to prove that after hitting the limit and then deleting a subscription, that a user can then create a new subscription

ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -82,6 +82,25 @@
                              query-params)
         gran-refs (search-gran-refs-with-sub-params request-context search-params)]))
 
+(defn- check-subscription-limit
+  "Given the  configuration for subscription limit, this valdiates that the
+  user has less than the limit before we allow more subscriptions to be
+  ingested by that user."
+  [request-context subscription]
+  (let [metadata (-> (:metadata subscription) (json/decode true))
```
jceaser, line 90: is -> needed in this case?

dpesall: It's possible I did something else wrong in removing it, but it seemed to break when I tried without the thread.

system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
dpesall: I think we intended this last test to validate that updating `subscriptions-limit` to 100 was actually changing the limit and allowing the user to create more subscriptions. If you think it should still be removed I can get rid of it though.

system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
dpesall: Definitely sounds much better that way, reworded.

system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
dpesall: Removed, and reworded.

system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
dpesall: Combined the tests in to the single `subscription-count-exceeds-limit-test`. I went with your suggestion and tested a revision of "sub1" after hitting the limit of 1 after the first ingest.

system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
```diff
@@ -43,6 +44,48 @@
...
+(deftest subscription-count-exceeds-limit-test
+  (side/eval-form `(jobsub/set-subscriptions-limit! 1))
+  (let [coll1 (data-core/ingest-umm-spec-collection
+               "PROV1"
+               (data-umm-c/collection
+                {:ShortName "coll-exceed-limit-1"
+                 :EntryTitle "entry-title-exceed-limit-1"})
+               {:token "mock-echo-system-token"})
+        user1-token (echo-util/login (system/context) "user1")]
+     (testing "ingest on PROV1 succeeds when ingesting subscription for user below their subscription limit"
```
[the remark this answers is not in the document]
dpesall: Removed on all my tests

system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
```diff
@@ -43,6 +44,48 @@
...
+         (is (= 200 (:status response)))
+         (is (mdb/concept-exists-in-mdb? (:concept-id response) (:revision-id response)))))
+     (testing "ingest on PROV1 fails with correct response when ingesting subscription for user beyond their subscription limit"
+       (let [concept (subscription-util/make-subscription-concept {:Query "platform=NOAA-9"
+                                                                   :Name "sub2"
+                                                                   :CollectionConceptId (:concept-id coll1)
+                                                                   :SubscriberId "user1"})
+             response (ingest/ingest-concept concept {:token user1-token})]
+        (is (= 409 (:status response)))
+        (is (= "The subscriber-id [user1] has already reached the subscription limit." (first (:errors response))))))
```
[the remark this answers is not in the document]
dpesall: Added a test that deletes a subscription when at the subscription limit and then creates a new subscription successfully to validate that deleted subscriptions do not count against the subscription limit.

brianalexander (approved):

PR #1454, opened 2022-02-10
CMR-8025: Added latency index, search and v2 facet

sxu123:


indexer-app/src/cmr/indexer/data/concepts/collection.clj
```diff
@@ -397,6 +402,10 @@
             :access-value access-value
             :processing-level-id processing-level-id
             :processing-level-id-lowercase (util/safe-lowercase processing-level-id)
+            :latency latency
+            :latency-lowercase (util/safe-lowercase latency)
+            :latency-priority {:value latency
+                               :priority 0}
```
ygliuvt, line 408: Why do we have a `priority` field if we don't actually use it?

search-app/docs/api.md
```diff
@@ -1589,6 +1590,18 @@ Find collections that match all of the 'consortium' param values
 
      curl "%CMR-ENDPOINT%/collections?consortium\[\]=CWIC&consortium\[\]=FEDEO&consortium\[\]=CEOS&options\[consortium\]\[and\]=true"
 
+#### <a name="c-latency"></a> Find collections by latency
```
ygliuvt, line 1593: There is no way anyone using this API will know what the 1 of 3 `latency` values to use to search for collections. There should be introductions on the search parameter. 
Also I think we should just use the term `collection_data_type` in Collection metadata as the parameter name and the values that are defined in the schema for the search. `latency` is something we can use in facet result to help user understand it, but not in searching.

ygliuvt (changes requested):
Need to address review comments.

search-app/docs/api.md
sxu123: I think we need to add a new one. Otherwise the facet will show not only "1 to 3 hours" but also "NEAR_REAL_TIME".   We only want to show "1 to 3 hours."

system-int-test/test/cmr/system_int_test/search/collection_latency_search_test.clj
```diff
@@ -0,0 +1,66 @@
+(ns cmr.system-int-test.search.collection-latency-search-test
+  "Integration tests for collection consortium search"
```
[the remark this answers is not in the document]
sxu123: Done.

search-app/docs/api.md
ygliuvt: The latency search parameter does not make sense to the clients. We can keep it internal for v2 facets and not show it in API doc similar to what we did with another v2 facets field `horizontal-data-resolution-range`. We can remove this section as `collection_data_type` parameter is already documented in API doc.

indexer-app/src/cmr/indexer/data/concepts/collection.clj
sxu123: Most of our facets are using the priority one. I tried different kinds and couldn't make it work.

search-app/docs/api.md
sxu123: Sure. That makes sense.  Removed.

sxu123: We've decided to keep the new latency index.

ygliuvt (approved):

ygliuvt:
Please resolve merge conflict before merging.

indexer-app/src/cmr/indexer/data/concepts/collection.clj
```diff
@@ -248,6 +248,11 @@
         spatial-keywords (lk/location-keywords->spatial-keywords-for-indexing
                           (:LocationKeywords collection))
         access-value (get-in collection [:AccessConstraints :Value])
+        latency (cond
+                  (= "NEAR_REAL_TIME" collection-data-type) "1 to 3 hours"
+                  (= "LOW_LATENCY" collection-data-type) "3 to 24 hours"
+                  (= "EXPEDITED" collection-data-type) "1 to 4 days"
+                  :else nil)
```
jaybarra, line 255: Style guide recommends case over cond when possible
```suggestion
        latency (case collection-data-type
                  "NEAR_REAL_TIME" "1 to 3 hours"
                  "LOW_LATENCY" "3 to 24 hours"
                  "EXPEDITED" "1 to 4 days"
                  nil)
```

jaybarra (approved):

PR #1456, opened 2022-02-10
Cmr 7979 cache echo token in browse scalar 

isja17:
instead of retrieving the echo token from parameter store every time, we now cache it.


brianalexander:
Here are my thoughts:

1. Why use Redis?  
We don't need to cache the token in Redis.  Doing the cacheing in memory will be more performant, save resources, and be more secure.
2. What happens if the token changes?  
Right now, the token is being cached in Redis and is only updated when the token expires (approx 24 hours).  What happens if the token changes? Is it possible for them to be out of sync?

A good flow would be:
1. Get the token when the application starts
2. If a request fails because the token is bad (Authorization 401 error?), request a new token, update the in-memory token, re-run the failed request.

brianalexander (commented):
Here are my thoughts:

1. Why use Redis?  
We don't need to cache the token in Redis.  Doing the cacheing in memory will be more performant, save resources, and be more secure.
2. What happens if the token changes?  
Right now, the token is being cached in Redis and is only updated when the token expires (approx 24 hours).  What happens if the token changes? Is it possible for them to be out of sync?

A good flow would be:
1. Get the token when the application starts
2. If a request fails because the token is bad (Authorization 401 error?), request a new token, update the in-memory token, re-run the failed request.

brianalexander (approved):

jaybarra (approved):

browse-scaler/src/cmr.js
```diff
@@ -1,27 +1,35 @@
 const fetch = require('node-fetch');
 const { getSecureParam } = require('./util');
-
 const config = require ('./config');
+const { setValue, getValue } = require ('./in-memory-cache')
+var authorizationToken;
```
ygliuvt, line 5: remove this line.

browse-scaler/src/in-memory-cache.js
```diff
@@ -0,0 +1,15 @@
...
+    function getInstance() {
+        return _data;
+    }
+
+    return {
+      getInstance: getInstance
+    };
+  }());
+
+exports.setValue = (key, item) => { InMemoryCache.getInstance() [key] = item;}
```
ygliuvt, line 13: change `item` to `value`.

browse-scaler/src/in-memory-cache.js
```diff
@@ -0,0 +1,15 @@
...
+
+exports.getValue = (id) => { return InMemoryCache.getInstance() [id]}
```
ygliuvt, line 15: change `id` to `key`.

ygliuvt (approved):
approved with minor change comments.

PR #1469, opened 2022-02-21
CMR-7658: Replaced the current message delimiter with a less commonly used one.

sxu123:


ingest-app/docs/api.md
```diff
@@ -220,10 +220,20 @@ Note: X-Request-Id response header always contains the same value as the CMR-Req
...
+     "revision-id" : 1,
+     "warnings" : null,
+     "existing-errors" : null}
 
-    {"concept-id":"C12345-PROV","revision-id":1,"warnings":"object has missing required properties ([\"ProcessingLevel\"])"}
+When there are existing errors allowed for progressive update, the response looks like the following:
 
+    {"concept-id" : "C1200000005-PROV1",
+     "revision-id" : 2,
+     "warnings" : [ "After translating item to UMM-C the metadata had the following issue(s): [:TilingIdentificationSystems] Tiling Identification Systems must be unique. This contains duplicates named [MODIS Tile EASE, MISR].;; [:AdditionalAttributes 2] Value [6] is not a valid value for type [DATETIME].;; [:AdditionalAttributes 3] Value [GHRSST Level 2P Global Subskin Sea Surface Temperature from the Advanced Microwave Scanning Radiometer 2 on the GCOM-W satellite] is not a valid value for type [DATETIME]." ],
```
jaybarra, line 232: Warning and existing-errors are already in an array, why not use that to our advantage and keep them as items in an array?

jaybarra (commented):
Warning and existing-errors are already in an array, why not use that to our advantage and keep them as items in an array? We are returning this as JSON so wouldn't it be more convenient to keep this as an array instead of a concatenated string?

sxu123:
@jaybarra We don't always return them in JSON. It could be in xml also.  I don't want to dramatically change the way we return things which might break the way clients handle it.

mitchstartzel (approved):
The real culprit seems to be the `.` after each error message, but this is a step in the right direction.

jaybarra (approved):
I believe a better long-term solution is to use the array, or return XML as a list of elements and can look at introducing that in a future ticket.

PR #1472, opened 2022-02-22
CMR-8004: Changed granuledataformat to dataformat in KMS urls.

sxu123:


mitchstartzel (commented):
Will this change require a simultaneous KMS change? In other words, if this were to go to PROD today, would the KMS refresh job break?

If the answer is yes, we should have the deployment dates for both KMS and this CMR change planned out before this PR gets merged.

ygliuvt:
> Will this change require a simultaneous KMS change? In other words, if this were to go to PROD today, would the KMS refresh job break?

I assume GCMD would keep both `granuledataformat` and `dataformat` endpoints available in the transition period and only retire the `granuledataformat` after CMR is switched to `dataformat`. If this is not GCMD's plan, it should be.

ygliuvt (approved):

sxu123:
KMS-387 is the ticket for GCMD to make the change.  Looks like they already retire granuledataformat in SIT. it's not in PROD yet, hopefully it's gonna be there before this one is.

eereiter:
That is a mistake we were told that the old name was going to be aliased.  Scott Ritz is going to talk to MDT to make sure granuledataformat and dataformat will both be supported.

mitchstartzel:
> KMS-387 is the ticket for GCMD to make the change. Looks like they already retire granuledataformat in SIT. it's not in PROD yet, hopefully it's gonna be there before this one is.

Good to see the KMS work has already been done. If this ticket gets merged this sprint, it will go to UAT March 2nd and PROD March 16th. Unless we have confirmation from KMS that their deployments will go out ahead of these dates, I strongly advise against merging this PR. We have been bitten more than once recently by merging CMR-sided changes too soon.

eereiter:
KMS can now handle both, the old name and the new in SIT.  We can merge this PR.  Their deployment schedule is on the same Wednesday as the CMR's. We just need to coordinate that the KMS deployment happens on the same day, but before the CMRs. Otherwise we will have to wait 2 more weeks.

eereiter (approved):

PR #1478, opened 2022-03-02
Cmr 8047 - Allow the CMR to handle the new KMS location (spatial) keyword structure while still not using it until the CMR is ready.

eereiter:


jceaser (approved):

ygliuvt:
Is it too much to ask KMS to create an endpoint that supports versioning? That way, CMR doesn't need to implement code like this.

eereiter:
EDSC has been asking the CMR the same question for a long time of why hasn't the CMR provided versioning for its API.  I think because it is a lot more work. KMS's code base and the amount of data they keep would increase greatly.  We could also change the code to read the CSV files and only take what the CMR uses instead of validating if the file structure doesn't match exactly. That is a bit more work, but I think it would remove the issue of changing the structure. Unless the names change for what the CMR is using.

ygliuvt:
> EDSC has been asking the CMR the same question for a long time of why hasn't the CMR provided versioning for its API.
We do offer UMM versioning which is parallel to KMS's versioning that I am asking for. I am not aware of the use case where EDSC wants versioning of CMR API and we have always maintained backwards compatibility of the APIs.

ygliuvt (approved):
I would have expected the implementation to be supporting both the new and old spatial keywords hierarchies and switch between them via a CMR config parameter. Given this code is temporary anyways, I am OK with this change.

PR #1486, opened 2022-03-07
Cmr 7786 get umm t related url and url from kms

sxu123:


ingest-app/src/cmr/ingest/services/ingest_service/tool.clj
```diff
@@ -1,20 +1,53 @@
...
+            kms-index
+            :related-urls
+            msg/url-content-type-type-subtype-not-matching-kms-keywords)}]))
+
+(defn- validate-all-fields
+  "Check all fields that need to be validated. Currently this is the Related URL Content Type, Type, and
+  Subtype fields. Throws an error if errors are found"
+  [context tool]
+  (let [errors (seq (umm-spec-validation/validate-tool tool (match-kms-content-type-type-and-subtype context)))]
+    (if-errors-throw :bad-request errors)))
```
daniel-zamora, line 43: weird, github is seeing if-errors-throw as a macro.

system-int-test/test/cmr/system_int_test/ingest/tool_ingest_test.clj
```diff
@@ -217,3 +218,50 @@
                              :native-id "tool_v_1.0"
                              :format "application/vnd.nasa.cmr.umm+json; version=1.0"})]
       (is (= 201 status)))))
+
+(deftest related-url-content-type-type-and-subtype-check
+  (testing
+    "Check valid and invalid related urls and url's url content type, type, and subtype.
```
daniel-zamora, line 224: Can you turn this into a comment and make the "testing" form more concise.

ingest-app/src/cmr/ingest/services/messages.clj
```diff
@@ -90,6 +90,14 @@
           (:Type related-url)
           (:Subtype related-url)))
 
+(defn url-content-type-type-subtype-not-matching-kms-keywords
+  "Error msg when URL Content Type, Type, and Subtype do not match a valid set in KMS."
+  [url]
+  (format "URL Content Type, Type, and Subtype [%s>%s>%s] are not a valid set together."
+          (:URLContentType url)
+          (:Type url)
+          (:Subtype url)))
```
daniel-zamora, line 99: If any of these are nil this will substitute "null" into the string.  I guess that is ok as long as its intended.

daniel-zamora (approved):

ingest-app/src/cmr/ingest/services/messages.clj
eereiter: I would put an if here to see if Subtype is nil and depending on the outcome do the format statement accordingly
```
(if subtype
  (format "URL Content Type, Type, and Subtype [%s>%s>%s] are not a valid set together."
               (:URLContentType url)
               (:Type url)
              (:Subtype url))
  (format "URL Content Type, Type, and Subtype [%s>%s] are not a valid set together."
               (:URLContentType url)
               (:Type url)))
```

system-int-test/test/cmr/system_int_test/ingest/tool_ingest_test.clj
sxu123: Sure. Thanks.

sxu123: Sure. Thanks.

ingest-app/src/cmr/ingest/services/messages.clj
sxu123: Thanks for your feedbacks.  But looking at the kms rucontenttype.csv file, you can see all the URLContentType, Type and Subtypes are there, when Subtypes are nil, it's just an empty string "".  So, we're comparing them all. I think it's clearer to show which three values are not valid, rather than just two sometimes. It would be confusing to me.

sxu123: I can change it to (format "URL Content Type, Type, and Subtype [%s>%s>%s] are not a valid set together."
           (:URLContentType url)
           (:Type url)
           (get url :Subtype ""))

eereiter: Sounds good.

eereiter (approved):

ingest-app/src/cmr/ingest/services/messages.clj
sxu123: Actually, RelatedURL's message is the same way. I think I'd better keep it consistent with RelatedURL's message since they compare the same thing. NULL is probably more clear than just an empty nothing there.

PR #1487, opened 2022-03-08
CMR-8021 Create an ingest test to be part of the system-validation-tests

dpesall:
This change creates new tests in system-validation-tests to test the ingest of a collection, searching for the collection, and then deleting the collection to help verify data flow in CMR so new changes do not break CMR.

### DIFF
--- a/system-validation-test/features/ingest_basic.feature
+++ b/system-validation-test/features/ingest_basic.feature
@@ -43,3 +43,11 @@ Feature: Basic Ingest API calls
     And I add extension "providers/PROV1/collections/TestColl1"
     When I submit a "DELETE" request
     Then the response status code is 200
+    And I wait "2.75" seconds for deletion to complete
+    Given I am searching for "collections"
+    And I clear the extension
+    And I want "json"
+    And I add a query param "short_name" of "TestColl1"
+    When I submit a "GET" request
+    Then the response status code is 200
+    And the response body is empty
--- a/system-validation-test/features/ingest_basic.feature
+++ b/system-validation-test/features/ingest_basic.feature
@@ -3,36 +3,43 @@ Feature: Basic Ingest API calls
 
   Background:
     Given I use the authorization token from environment variable "CMR_TOKEN"
+    And the provider "PROV1" exists
+    And set body to the following XML "<Collection><ShortName>TestColl1</ShortName><VersionId>Version01</VersionId><InsertTime>1999-12-31T19:00:00-05:00</InsertTime><LastUpdate>1999-12-31T19:00:00-05:00</LastUpdate><LongName>CMR8021COLLECTION</LongName><DataSetId>LarcDatasetId</DataSetId><Description>A minimal valid collection</Description><Orderable>true</Orderable><Visible>true</Visible></Collection>"
 
   @ingest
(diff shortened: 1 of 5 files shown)


system-validation-test/features/ingest_basic.feature
```diff
@@ -0,0 +1,45 @@
...
+    And I add extension "providers/PROV1/collections/TestColl1"
+    And I add header "Content-type=application/echo10+xml"
+    When I submit a "PUT" request
+    And I wait "1.75" seconds for ingest to complete
+    Given I am deleting on "ingest"
+    And I clear the extension
+    And I clear headers
+    And I add extension "providers/PROV1/collections/TestColl1"
+    When I submit a "DELETE" request
+    Then the response status code is 200
```
daniel-zamora, line 45: Is it possible to verify that the concept is deleted by attempting to retrieve it?

system-validation-test/features/ingest_basic.feature
```diff
@@ -0,0 +1,45 @@
+Feature: Basic Ingest API calls
+  Ingest collections for various formats
+
+  Background:
+    Given I use the authorization token from environment variable "CMR_TOKEN"
+    And the provider "PROV1" exists
```
daniel-zamora, line 6: So is this limited to only local tets?

system-validation-test/features/ingest_basic.feature
dpesall: Added a GET request after the deletion to ensure that nothing is found when searching for the deleted collection.

system-validation-test/features/ingest_basic.feature
dpesall: It can be run in different environments, but the requirements for running are:
- CMR_TOKEN and CMR_ROOT environment variables are configured for the proper environment (localhost, SIT, etc)
- A provider with name of PROV1 exists

If these are not both satisfied, instead of failing the test will rather be shown as pending and skipped.

daniel-zamora: Right, but does PROV1 exist?  And do we have an OK from ops to use that provider name if it doesn't?

dpesall: That's a good question. Who would I reach out to regarding that?

system-validation-test/features/step_definitions/rest_steps.rb
```diff
@@ -41,17 +41,51 @@ def update_query(query, key, value)
...
+      response = HTTParty.delete(resource_uri, options)
     else
       raise "#{method} is not supported yet"
     end
 
     response
   end
+
+  def check_provider(provider_name)
+    url = 'http://localhost:3002/providers'
```
daniel-zamora, line 59: This will only check locally.

dpesall: Fixed to now use get_url() to get the correct endpoint for the current environment.

system-validation-test/features/ingest_basic.feature
daniel-zamora: It might be worth making that configurable.

brianalexander (approved):

PR #1492, opened 2022-03-14
CMR-8021-2 Enhancement to 8021: Create an ingest test to be part of the system-validation-tests

dpesall:
This enhancement makes changes from CMR-8021 which fix the response requests to function properly both in SIT and local environments.

Parent:
[CMR-8021](https://github.com/nasa/Common-Metadata-Repository/pull/1487): [created new validation tests in system-validation-tests to perform an ingest operation to verify data-flow in CMR so new changes do not break CMR.](https://github.com/nasa/Common-Metadata-Repository/pull/1487)

### DIFF
--- a/system-validation-test/features/ingest_basic.feature
+++ b/system-validation-test/features/ingest_basic.feature
@@ -5,12 +5,12 @@ Feature: Basic Ingest API calls
     Given I use the authorization token from environment variable "CMR_TOKEN"
     And the provider from environment variable "CMR_TEST_PROV" exists
 
-    And set body to the following XML "<Collection><ShortName>TestCollection8021</ShortName><VersionId>Version01</VersionId><InsertTime>1999-12-31T19:00:00-05:00</InsertTime><LastUpdate>1999-12-31T19:00:00-05:00</LastUpdate><LongName>CMR8021COLLECTION</LongName><DataSetId>LarcDatasetId01</DataSetId><Description>A minimal valid collection</Description><Orderable>true</Orderable><Visible>true</Visible></Collection>"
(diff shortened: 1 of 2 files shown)


eereiter (approved):

system-validation-test/features/step_definitions/rest_steps.rb
```diff
@@ -119,6 +119,10 @@ def get_url(concept_type, cmr_root)
   @body = xml
 end
 
+Given(/^I clear the body$/) do
+  @body = ''
```
jaybarra, line 123: Would prefer if this was set to `nil` instead of an empty string since that could still be considered a payload

system-validation-test/features/ingest_basic.feature
```diff
@@ -4,50 +4,58 @@ Feature: Basic Ingest API calls
...
+    And I add extension "providers/DEMO_PROV/collections/TestCollection8021"
     And I add header "Content-type=application/echo10+xml"
     When I submit a "PUT" request
     Then the response status code is in "200,201"
 
   @ingest
   Scenario: Searching for ingested Collection
     Given I am ingesting a "Collection"
-    And I add extension "providers/PROV1/collections/TestColl1"
+    And I add extension "providers/DEMO_PROV/collections/TestCollection8021"
```
jaybarra, line 21: Could we make this a bit more precise wording wise? Like a catcher for `And I use path ...` since extension makes me thing of something like `.json` or `.xml`

system-validation-test/features/ingest_basic.feature
```diff
@@ -4,50 +4,58 @@ Feature: Basic Ingest API calls
...
     And I add header "Content-type=application/echo10+xml"
     When I submit a "PUT" request
-    And I wait "1.75" seconds for ingest to complete
+    And I wait "2.25" seconds for ingest to complete
```
jaybarra, line 24: 👍

jaybarra (changes requested):

system-validation-test/features/step_definitions/rest_steps.rb
dpesall: Fixed

system-validation-test/features/ingest_basic.feature
dpesall: That's a good point -- I made a new handler specifying adding/setting a url path for better clarity and updated the resource_url constructor to concat the extension and url path in the proper order.

jaybarra (approved):

PR #1500, opened 2022-03-22
MMT-2813 - Changing TEA config gen. from API Gateway to Application Load Balancer

jceaser:
When trying to publish to sit, it was found that NGAP could not configure a path to point to an API Gatway application. serverless.yml was changed to instead publish to the CMR application load balancer. The path parameter code had to change as no way exists to setup a lambda path parameter in the load balancer.

To support sandbox which does not have a load balancer, a new configuration was created which should look like the old solution.

### DIFF
--- a/tea-config-generator/README.md
+++ b/tea-config-generator/README.md
@@ -43,11 +43,39 @@ use as a build and deployment environment when running under a CI/CD system.
 
 ## Deploying to AWS
 
+Several methods are provided to publish this application to specific CMR AWS
+environments. These processes can be reused for other similar environments based
+on need.
+
+### Deployment
+
 The [serverless.yml](serverless.yml) file will create an IAM role for the lambda
-functions. It will use the following Role name:
+functions and install them under an Application Load Balancer. The ARN is to be
+provided. The IAM role will use the following Role name:
 `tea-config-generator-role-${self:custom.env}-${self:provider.region}` and will
 use `arn:aws:iam::${aws:accountId}:policy/NGAPShRoleBoundary` for a Permissions
-Boundary.
+Boundary. This setup is suitable for situations where a public URL is applied at
+the Loadbalancer level. 
+
+* **NOTE**: This is the file used for deployment to SIT, UAT, and OPS
+
+
+### Sandbox
+The [serverless-sandbox.yml](serverless-sandbox.yml) file will also create an IAM
+role and be bound to the same Permissions Boundary, however it will use an API 
+Gateway to publish the lambda functions. The URL for the API Gateway can be used
+for internal testing or for use within an AWS account. 
+
+* **NOTE**: This is the file used for testing on AWS but not SIT, UAT, or OPS.
+
+
+### Local Use
+The [serverless-local.yml](serverless-local.yml) file is ment for use locally by
+developers of the application. It will locally publish an API Gateway and allow
+access to the lambda functions through that. It also includes an S3 bucket for
+displaying the HTML API documentation. 
+
+* **NOTE**: for localhost use only.
 
 ## Testing
 Read the details in the ./run.sh script. There are functions which have all the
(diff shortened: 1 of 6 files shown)


tea-config-generator/serverless-sandbox.yml
```diff
@@ -0,0 +1,68 @@
+service: tea-config-generator
```
ygliuvt, line 1: I don't think we want a completely different deployment for sandbox. What would be the point? I think we can assume that CMR is deployed in Sandbox before TEA is deployed and document it properly.

jceaser: I can get rid of it, I just found it helpful for testing some things before messing with the SIT AWS configuration.

tea-config-generator/serverless.yml
```diff
@@ -24,16 +24,29 @@ functions:
   capabilities:
     handler: capabilities.capabilities
     events:
-      - http: GET ${self:custom.endPoint}/
-      - http: GET ${self:custom.endPoint}/capabilities
+      - alb:
+          listenerArn: ${self:custom.loadBalancer}
```
ygliuvt, line 28: This listernerArn can be defined as reference to CF outputs, i.e. `${cf:${opt:stage}.servicesLbListenerArn}` where stage is passed in as input to serverless deploy.

tea-config-generator/serverless.yml
```diff
@@ -61,6 +74,7 @@ plugins:
   - serverless-offline
   - serverless-s3-local
 custom:
+  loadBalancer: ${env:AWS_TEA_CONFIG_LOADBALANCER, 'arn:aws:elasticloadbalancing:us-east-1:832706493240:listener/app/cmr-services-sit/9bae549c0224bffd/e2610cc0267bd95e'}
```
ygliuvt, line 77: Remove this config.

tea-config-generator/serverless.yml
```diff
@@ -61,6 +74,7 @@ plugins:
...
   env: ${env:AWS_TEA_CONFIG_ENV, 'sit'}
```
ygliuvt, line 78: This should be pass in via the `--stage` serverless option.

tea-config-generator/serverless-sandbox.yml
ygliuvt: I see the value of being able to test TEA configuration outside of CMR in Sandbox though. So I am OK with keeping it, but please add some documentation in the README.

tea-config-generator/serverless.yml
jceaser: is this present by default? I don't need to do anything to support it?

tea-config-generator/serverless.yml
jceaser: I removed the value

tea-config-generator/serverless.yml
jceaser: fixed it.

charleshuang80 (approved):

tea-config-generator/README.md
```diff
@@ -43,11 +43,39 @@ use as a build and deployment environment when running under a CI/CD system.
...
+The [serverless-sandbox.yml](serverless-sandbox.yml) file will also create an IAM
+role and be bound to the same Permissions Boundary, however it will use an API 
+Gateway to publish the lambda functions. The URL for the API Gateway can be used
+for internal testing or for use within an AWS account. 
+
+* **NOTE**: This is the file used for testing on AWS but not SIT, UAT, or OPS.
+
+
+### Local Use
+The [serverless-local.yml](serverless-local.yml) file is ment for use locally by
```
zimzoom, line 73: typo. "ment" -> "meant"

zimzoom (approved):

tea-config-generator/README.md
jceaser: that was a test, you pass/

mitchstartzel (approved):

cgokey (approved):

PR #1512, opened 2022-03-28
MMT-2844 - changes needed to publish in both SIT and UAT

jceaser:


cgokey (approved):

tea-config-generator/serverless.yml
```diff
@@ -37,11 +36,11 @@ functions:
...
-    timeout: 1
+    timeout: 5
     events:
       - alb:
           listenerArn: ${cf:${self:provider.stage}.servicesLbListenerArn}
-          priority: 51
+          priority: 998
           conditions:
             path: ${self:custom.endPoint}/status
     vpc: ~
```
ygliuvt, line 46: What does `~` mean here? We should be able to delete this line.

jceaser: it means use the default, I was using it to document that I mean to not apply the setting at this level because it is not needed. If this is not valuable, then I can remove it. I would be required if I moved the vpc setting to the provider level.

ygliuvt (approved):

tea-config-generator/serverless.yml
ygliuvt: I am not sure what the standard is. I will leave it up to you to keep it or delete it.

PR #1517, opened 2022-03-31
CMR-8119 Adds POST/parameter search for stac format

daniel-zamora:
This PR allows for search-app concept search endpoint returning a stac format to use POST requests with parameters that are equivalent to GET requests with the same parameters.  This does not include JSON queries.

### DIFF
--- a/search-app/src/cmr/search/api/concepts_search.clj
+++ b/search-app/src/cmr/search/api/concepts_search.clj
@@ -216,7 +216,7 @@
         scroll-id (:scroll-id scroll-id-and-search-params)
         cached-search-params (:search-params scroll-id-and-search-params)
         search-after (get headers (string/lower-case common-routes/SEARCH_AFTER_HEADER))
-        ctx (assoc ctx :query-string body :scroll-id scroll-id)
+        ctx (assoc ctx :query-string body :scroll-id scroll-id :query-params params)
         params (core-api/process-params concept-type params path-w-extension headers mt/xml)
         result-format (:result-format params)
         _ (block-excessive-queries ctx concept-type result-format params)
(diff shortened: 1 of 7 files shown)


search-app/src/cmr/search/results_handlers/stac_results_handler.clj
```diff
@@ -313,24 +339,24 @@
...
                               :query-string
+                              (codec/form-decode)
+                              util/map-keys->snake_case
+                              (codec/form-encode)
                               (string/replace page-num-pattern ""))
         prev-num (when (> page-num 1)
                    (dec page-num))
         next-num (when (> allowed-hits item-count)
                    (inc page-num))]
+    (proto-repl.saved-values/save 6)
```
jaybarra, line 350:
```suggestion
```

jaybarra (approved):
Just need to remove the proto-repl reference

eereiter (approved):

search-app/src/cmr/search/api/concepts_search.clj
```diff
@@ -372,11 +368,11 @@
     (OPTIONS "/" req (common-routes/options-response))
     (GET "/"
       {params :params headers :headers ctx :request-context query-string :query-string}
-      (find-concepts ctx path-w-extension params headers query-string))
+      (find-concepts (merge ctx {:method "GET"}) path-w-extension params headers query-string))
```
ygliuvt, line 371: CMR standard is to use keyword `:get` rather than string `"GET"`

system-int-test/src/cmr/system_int_test/data2/stac.clj
```diff
@@ -6,26 +6,44 @@
...
+                (when prev-num
+                  {:rel "prev"
+                   :method "GET"
+                   :href (href query-string prev-num)})
+                (when prev-num
+                  {:rel "prev"
+                   :method "POST"
+                   :body body
+                   :merge true
+                   :href (href)}))
```
ygliuvt, line 35: Where is the `page_num` in POST? I don't think this works.

ygliuvt: Also what if there is combination of query string in the POT body and parameters?

ygliuvt (changes requested):
The `page_num` is not handled correctly in POST links. There is also a question of combination of query string and query params.

ygliuvt:
We should update the API doc for STAC POST.

ygliuvt (approved):

PR #1520, opened 2022-04-04
CMR-8093 - Update a KMS log

jceaser:
updating log so that ops can observe and confirm the override setting without needed to ask for help

### DIFF
--- a/transmit-lib/src/cmr/transmit/kms.clj
+++ b/transmit-lib/src/cmr/transmit/kms.clj
@@ -64,12 +64,14 @@
    Return value is the KMS schema override map as configured in AWS as
    CMR_KMS_SCHEMA_OVERRIDE_JSON."
   []
-  (try
-    (json/parse-string (config/kms-scheme-override-json) true)
-    (catch com.fasterxml.jackson.core.JsonParseException e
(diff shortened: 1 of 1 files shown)


jaybarra (approved):

transmit-lib/src/cmr/transmit/kms.clj
```diff
@@ -64,12 +64,14 @@
...
    CMR_KMS_SCHEMA_OVERRIDE_JSON."
   []
-  (try
-    (json/parse-string (config/kms-scheme-override-json) true)
-    (catch com.fasterxml.jackson.core.JsonParseException e
-      (error "Failed to parse Scheme Override JSON while loading KMS resource [%s]: %s",
-             (config/kms-scheme-override-json)
-             (.getMessage e)))))
+  (let [overrides (config/kms-scheme-override-json)]
+    (debug (format "Requested KMS schema overrides: %s." overrides))
```
jaybarra, line 68: May just want to make it `info` instead of debug

jaybarra (approved):

transmit-lib/src/cmr/transmit/kms.clj
jceaser: I was trying to be conservative, but I like the suggestion!

PR #1522, opened 2022-04-05
CMR-7765 add Format and MimeType to umm t umm s

sxu123:


system-int-test/test/cmr/system_int_test/ingest/service_ingest_test.clj
```diff
@@ -257,3 +257,77 @@
...
+      (is (= 201 status1))
+      (is (= 0 (count errors1)))
+      (is (= 400 status2))
+      (is (= 1 (count errors2)))
+      (is (= [{:path ["RelatedUrLs" 0 "Format"], :errors ["Format [invalid] was not a valid keyword."]}] errors2))
+      (is (= 400 status3))
+      (is (= 1 (count errors3)))
+      (is (= [{:path ["RelatedUrLs" 0 "MimeType"], :errors ["MimeType [invalid] was not a valid keyword."]}] errors3))
+      (is (= 400 status4))
+      (is (= 2 (count errors4)))
```
jaybarra, line 332: Could these be grouped into `testing` blocks or comments added to show what is being tested more clearly?
And is `RelatedUrLs` a typo? should the `r` be upper-case as well?

sxu123: I could add some comments.  RelatedUrLs is NOT a typo. All the existing tests have it spelled like that and it's what the validation error contains.  Not sure where it got that spelling from.

ygliuvt (approved):

ingest-app/src/cmr/ingest/services/messages.clj
```diff
@@ -67,7 +67,7 @@
   "Error message when relatedurl-mimetype is
   not in the KMS."
   [relatedurl-mime-type]
-  (format "Mime Type [%s] was not a valid keyword."
+  (format "MimeType [%s] was not a valid keyword."
```
jceaser, line 70: https://en.wikipedia.org/wiki/Media_type says it is formally called 'MIME type' as MIME is an acronym. Confirmed with https://en.wikipedia.org/wiki/MIME . I would normally cite something better then wikipedia, but error messages are suppose to be human readable, so we should match what humans expect.

system-int-test/test/cmr/system_int_test/ingest/variable_ingest_test.clj
```diff
@@ -590,7 +590,7 @@
         {:MimeType "application/x-hdf42"}
         [422]
         [{:path ["RelatedUrLs" 0 "MimeType"]
-          :errors ["Mime Type [application/x-hdf42] was not a valid keyword."]}])
+          :errors ["MimeType [application/x-hdf42] was not a valid keyword."]}])
```
jceaser, line 593: same comment as above

jceaser (approved):
approved, but consider my suggestion for the error message.

ingest-app/src/cmr/ingest/services/messages.clj
sxu123: I disagree.  We are trying to show the user the parameter in the RelatedURL that's wrong. So the name better matches the parameter name.  Like when there is something wrong with RelatedURLs, we don't separate them as Related URL...

PR #1526, opened 2022-04-07
CMR-7663 adjusts max points to below elasticseach clause limit.

daniel-zamora:
In testing we found that elasticsearch has a clause limit that won't allow our line parameter queries to go above a certain amount.  Adjusted the max point error threshold down to avoid 500s.


system-int-test/src/cmr/system_int_test/utils/search_util.clj
```diff
@@ -854,3 +856,37 @@
...
+       vals
+       (remove #(> (count %) 1))
+       (map reverse)))
+
+(defn- depointify
+  [point]
+  [(.lon point) (.lat point)])
+
+(defn make-excessive-points-without-dups
+  "Returns a list of n-ish unique points."
```
jaybarra, line 872: Would there be a way to guarantee N points?

daniel-zamora: I could make it continue adding points until the count was exactly n, instead of just make n points and removing the duplicates, but I didn't think the test needed that level of accuracty.

jaybarra (approved):

ygliuvt (approved):

PR #1530, opened 2022-04-11
CMR-8138: Refactor metrics-service for improved performance

brianalexander:
The major slow-down in the metrics-service is the elasticsearch queries made to verify that product/short-names actually exist.  As many items appear multiple times (differentiated by versions), I cached the results of the short-name queries to reduce the total number of calls.

Most other changes are refactoring for readability:
- Broke up large functions into smaller functions
- Removed references to the version (not required)
- Made sure that comprehensive is `false` by default and that it is working as anticipated. (According to the documentation, comprehensive is `false` by default, but it was being passed as `nil` which wasn't triggering the conditional correctly.)
- Made sure that the `current-metrics` where being used correctly.  Previous, the `short-name` was being looked up directly in the response for `current-metrics`, but the data structure was a list of maps and so there were never any hits.  `current-metrics` is now a set of `short-names` instead. (Will cause major speed-up).


search-app/src/cmr/search/services/community_usage_metrics/metrics_service.clj
```diff
@@ -130,29 +139,24 @@
...
 (defn get-community-usage-metrics
-  "Retrieves the set of community usage metrics from metadata-db."
+  "Retrieves the current community usage metrics from metadata-db."
   [context]
   (:community-usage-metrics (json/decode (:metadata (humanizer-service/fetch-humanizer-concept context)) true)))
 
+(defn- community-usage-metrics-list->set
+  "Convert list of {:short-name :access-count} maps into a set of short-names."
+  [metrics-list]
+  (into #{} (for [metric-map metrics-list]
```
jaybarra, line 149: This could also be just done with just `set`

jaybarra (approved):

search-app/src/cmr/search/services/community_usage_metrics/metrics_service.clj
```diff
@@ -167,14 +171,31 @@
    params
    [(partial cpv/validate-boolean-param :comprehensive)]))
 
+(defn- aggregate-usage-metrics
+  "Combine access-counts for entries with the same short-name."
+  [metrics]
+  (let [java-map (new java.util.HashMap)]
```
ygliuvt, line 177: I think this could be done a lot cleaner with `reduce`

ygliuvt (approved):

search-app/src/cmr/search/services/community_usage_metrics/metrics_service.clj
brianalexander: I reverted the function back to its previous version with minor modifications.  Can't tell you how it works, but it's definitely clojurish and provides the right outputs!

search-app/src/cmr/search/services/community_usage_metrics/metrics_service.clj
```diff
@@ -130,28 +140,33 @@
...
 (defn get-community-usage-metrics
-  "Retrieves the set of community usage metrics from metadata-db."
+  "Retrieves the current community usage metrics from metadata-db."
   [context]
   (:community-usage-metrics (json/decode (:metadata (humanizer-service/fetch-humanizer-concept context)) true)))
 
+(defn- community-usage-metrics-list->set
+  "Convert list of {:short-name :access-count} maps into a set of short-names."
+  [metrics-list]
+  (set (for [metric-map metrics-list]
```
ygliuvt, line 150: change this to `(set (map :short-name metrics-list))`.

ygliuvt, line 150: change this to `(set (map :short-name metrics-list))`.

search-app/src/cmr/search/services/community_usage_metrics/metrics_service.clj
```diff
@@ -130,28 +140,33 @@
...
          csv-lines)))
 
 (defn- aggregate-usage-metrics
-  "Combine access counts for entries with the same short-name and version number."
+  "Combine access counts for entries with the same short-name."
   [metrics]
-  ;; name-version-groups is map of [short-name, version] [entries that match short-name/version]
-  (let [name-version-groups (group-by (juxt :short-name :version) metrics)] ; Group by short-name and version
+  ;; name-groups is map of short-names [entries that match short-name]
+  (let [name-groups (group-by (juxt :short-name) metrics)] ; Group by short-name and version
```
ygliuvt, line 165: Are we supposed to drop the `version`? If so, we don't need the `juxt` call, just `(group-by :short-name metrics)` will do.

search-app/src/cmr/search/services/community_usage_metrics/metrics_service.clj
ygliuvt: We are not supposed to check in any code that we can't read. :) The above code is more efficient than `group-by`, but who cares.

PR #1535, opened 2022-04-15
CMR-8138-1: Update references to Version column to ProductVersion

brianalexander:
1. Current EMS usage data is reported with columns `Provider,Product,ProductVersion,Hosts` contrary to how the codebase and tests were using/anticipating `Provider,Product,Version,Hosts`.  This was changed sometime between 5 and 2 years ago.  The side-effect of this column renaming is that count-aggregations were done by short-name only.  Tests did not pick up this change because their dummy data still worked and there were no runtime errors because the endpoint was built in such a way as to handle a missing version column.
- All dummy data updated to reflect new column headers
- Reverted some changes in CMR-8138 that had removed references to the Version.
- Added an additional header validation to ensure that if the Version column changes in the future we'll be aware.
2. Version length is limited to 20 characters.  I'm not sure why that is, but it's part of the defined schema.  In the current community-usage-metrics, there are 6 products that have a version longer than 20 characters (all items share the same version.
```
PO DAAC,ASCATA-L2-25KM,Operational/Near-Real-Time,283
PO DAAC,ASCATA-L2-COASTAL,Operational/Near-Real-Time,201
PO DAAC,ASCATB-L2-25KM,Operational/Near-Real-Time,6023
PO DAAC,ASCATB-L2-COASTAL,Operational/Near-Real-Time,3980
PO DAAC,ASCATC-L2-25KM,Operational/Near-Real-Time,5943
PO DAAC,ASCATC-L2-COASTAL,Operational/Near-Real-Time,3796
```
- I made the decision that if an item has a Version longer than 20 characters, I would drop the version for that product.  This causes products that have versions longer than 20 characters to be summed over the entire product instead of by version.  The rankings will still work without the version.

4. Small bug fix from code change in CMR-8138.  If there were no humanizers stored in metadata-db, creating the first community-usage-metrics would fail because it tries to get the usage-metrics first then creates the new ones.
- Added `get-community-usage-metrics-or-empty-list` in `metrics_service.clj` - If no humanizer exists, an empty list is returned

### DIFF
--- a/search-app/src/cmr/search/services/community_usage_metrics/metrics_service.clj
+++ b/search-app/src/cmr/search/services/community_usage_metrics/metrics_service.clj
@@ -85,7 +85,7 @@
       (do
         (warn (format (str "While constructing community metrics humanizer, "
                            "could not find corresponding collection when searching for the term %s. "
-                           "Csv line entry: %s")
+                           "CSV line entry: %s")
                       product
                       csv-line))
         product))))
@@ -102,21 +102,20 @@
           (errors/throw-service-error :invalid-data
                                       (format (str "Error parsing 'Hosts' CSV Data. "
                                                    "Hosts must be an integer. "
-                                                   "Csv line entry: %s")
+                                                   "CSV line entry: %s")
                                               csv-line))))
       (errors/throw-service-error :invalid-data
                                   (format (str "Error parsing 'Hosts' CSV Data. "
                                                "Hosts may not be empty. "
-                                               "Csv line entry: %s")
+                                               "CSV line entry: %s")
                                           csv-line)))))
 
 (defn- get-version
   "Version must be 20 characters or less."
   [csv-line version-col]
   (let [reported-version (read-csv-column csv-line version-col)]
-    (if (<= (count reported-version) 20)
-      reported-version
-      nil)))
+    (when (<= (count reported-version) 20)
+      reported-version)))
 
 (defn- csv-entry->community-usage-metric
   "Convert a line in the csv file to a community usage metric. Only storing short-name (product)
@@ -160,7 +159,9 @@
   [context]
   (try
     (get-community-usage-metrics context)
-    (catch Exception e [])))
+    (catch Exception e 
+      (warn e)
+      [])))
 
 (defn- community-usage-metrics-list->set
   "Convert list of {:short-name :access-count} maps into a set of short-names."
--- a/search-app/src/cmr/search/services/community_usage_metrics/metrics_service.clj
+++ b/search-app/src/cmr/search/services/community_usage_metrics/metrics_service.clj
@@ -25,7 +25,7 @@
   [csv-header]
   (let [csv-header (mapv str/trim csv-header)]
     {:product-col (.indexOf csv-header "Product")
-     :version-col (.indexOf csv-header "Version")
+     :version-col (.indexOf csv-header "ProductVersion")
      :hosts-col (.indexOf csv-header "Hosts")}))
 
 (defn- read-csv-column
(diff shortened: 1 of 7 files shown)


search-app/src/cmr/search/services/community_usage_metrics/metrics_service.clj
```diff
@@ -102,20 +102,29 @@
           (errors/throw-service-error :invalid-data
                                       (format (str "Error parsing 'Hosts' CSV Data. "
                                                    "Hosts must be an integer. "
-                                                   "Csv line entry: [%s]")
+                                                   "Csv line entry: %s")
```
jaybarra, line 105: Can we also fix Csv -> CSV?

search-app/src/cmr/search/services/community_usage_metrics/metrics_service.clj
```diff
@@ -102,20 +102,29 @@
...
+                                               "Csv line entry: %s")
                                           csv-line)))))
 
+(defn- get-version
+  "Version must be 20 characters or less."
+  [csv-line version-col]
+  (let [reported-version (read-csv-column csv-line version-col)]
+    (if (<= (count reported-version) 20)
+      reported-version
+      nil)))
```
jaybarra, line 119:
```suggestion
    (when (<= (count reported-version) 20)
      reported-version)))
```

search-app/src/cmr/search/services/community_usage_metrics/metrics_service.clj
```diff
@@ -135,15 +144,24 @@
...
+  "Retrieves the current community usage metrics from metadata-db.
+   Returns an empty set if unable to retrieve metrics."
   [context]
   (:community-usage-metrics (json/decode (:metadata (humanizer-service/fetch-humanizer-concept context)) true)))
 
+(defn- get-community-usage-metrics-or-empty-list
+  [context]
+  (try
+    (get-community-usage-metrics context)
+    (catch Exception e [])))
```
jaybarra, line 163: If this is a json-decode exception we should still try to print what what wrong

system-int-test/test/cmr/system_int_test/search/humanizer/community_usage_metrics_test.clj
```diff
@@ -221,103 +221,102 @@
...
-               (humanizer-util/update-community-usage-metrics
-                admin-update-token
-                (format "Product,Version,Hosts\nAST_09XT,%s,4" long-value))))))
+    ;; (testing "Maximum version length validations"
+    ;;   (let [long-value (apply str (repeat 21 "x"))]
+    ;;     (is (= {:status 400
+    ;;             :errors ["#/0/version: expected maxLength: 20, actual: 21"]}
+    ;;            (humanizer-util/update-community-usage-metrics
+    ;;             admin-update-token
+    ;;             (format "Product,Version,Hosts\nAST_09XT,%s,4" long-value))))))
```
jaybarra, line 263: If this test is OBE it should be removed, if it's still valid we should leave it in and enabled

jaybarra (changes requested):
Does this PR have to coincide with an update to another application or be made live in the other environments at a specific time?

system-int-test/test/cmr/system_int_test/search/humanizer/community_usage_metrics_test.clj
brianalexander: Do you think the solution I chose is reasonable? I wasn't sure if that was the best way to go about it. I'm not really sure why there is this 20 character limit at all.  I could also remove the limit from the schema, but not sure if that would affect anything (possibly Earthdata Search can't support versions longer than 20 characters?)

I made the decision that if an item has a Version longer than 20 characters, I would drop the version for that product. This causes products that have versions longer than 20 characters to be summed over the entire product instead of by version. The rankings will still work without the version.

ygliuvt (approved):

jaybarra (approved):

PR #1539, opened 2022-04-21
CMR-7215: Bump UMM-Sub version from v1.0 to v1.1

brianalexander:


isja17 (approved):

ygliuvt (approved):
It's interesting that no test is broken with version at `1.0` for ingest. Something wasn't right.

PR #1543, opened 2022-04-25
Cmr 8197 fix iso19115 translation

sxu123:
CMR-8197 has been merged. Need to fix a bug in the translation from UMM-C to iso19115.

### DIFF
--- a/umm-spec-lib/test/cmr/umm_spec/test/generate_and_parse.clj
+++ b/umm-spec-lib/test/cmr/umm_spec/test/generate_and_parse.clj
@@ -12,6 +12,7 @@
    [cmr.umm-spec.iso-keywords :as kws]
    [cmr.umm-spec.iso19115-2-util :as iu]
    [cmr.umm-spec.json-schema :as js]
+   [cmr.umm-spec.migration.version.collection :as version-collection]
    [cmr.umm-spec.models.umm-collection-models :as umm-c]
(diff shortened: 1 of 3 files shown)


umm-spec-lib/test/cmr/umm_spec/test/generate_and_parse.clj
```diff
@@ -157,13 +161,19 @@
              (format "Parsing example file %s and converting to %s and then parsing again did not result in expected umm."
                      example-file target-format)))))))
 
-(deftest roundtrip-example-collection-record
-  (doseq [metadata-format tested-collection-formats]
+(deftest roundtrip-example-collection-record-non-iso19115
```
eereiter, line 164: Why were the ISO tests separated out - the code is the same?

sxu123: It's separated out because I need to test the case when UMM-C contains Footprints and doesn't contain SwathWidth. And that only works for iso19115. For echo10 and dif10, you have to manually remove the Footprints, convert the largest footprint to SwathWidth, it just looks messy.

sxu123: Modified the test.

eereiter: But I don't see what has changed other than it was separated out.

eereiter (approved):

PR #1547, opened 2022-04-27
CMR-8140: Indexed and validated StandardProduct.

sxu123:
There are two parts to this PR:
1. Indexing part: Done without testing which will be done in the search ticket CMR-8129.
2. Validation part: Done with testing.

### DIFF
--- a/indexer-app/src/cmr/indexer/data/concepts/collection.clj
+++ b/indexer-app/src/cmr/indexer/data/concepts/collection.clj
@@ -180,6 +180,13 @@
   (or (not (empty? (:DirectDistributionInformation collection)))
       (tag/has-cloud-s3-tag? tags)))
 
+(defn- standard-product?
+  "Test if the collection meets the criteria for being standard-product"
+  [collection tags]
+  (or (= true (:StandardProduct collection))
+      (and (not (:StandardProduct collection))
+           (tag/has-standard-product-tag? tags))))
+
 (def geoss-url-list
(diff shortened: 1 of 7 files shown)


indexer-app/src/cmr/indexer/data/concepts/collection.clj
```diff
@@ -180,6 +180,13 @@
   (or (not (empty? (:DirectDistributionInformation collection)))
       (tag/has-cloud-s3-tag? tags)))
 
+(defn- standard-product?
+  "Test if the collection meets the criteria for being standard-product"
+  [collection tags]
+  (or (= true (:StandardProduct collection))
+      (and (not (:StandardProduct collection))
+           (tag/has-standard-product-tag? tags))))
```
eereiter, line 188: You don't need (and (not (:StandardProduct collection)).
Just
(or (= true (:StandardProduct collection))
      (tag/has-standard-product-tag? tags)))

sxu123: I did it because you mentioned that when StandardProduct is false, we index it as false.

jceaser (approved):

eereiter (changes requested):

indexer-app/src/cmr/indexer/data/concepts/collection.clj
sxu123: Fixed!

eereiter (approved):

PR #1553, opened 2022-05-03
CMR-8129: Added StandardProduct search parameter.

sxu123:
This PR included the fix for CMR-8132 because the ingest in this test needs to index the collection in echo10 format. If StandardProduct is not translated, it can't be indexed.


jaybarra (approved):

eereiter (approved):

sxu123:
> Does this need a new field in elasticsearch?

Yes/ We have a new field

PR #1554, opened 2022-05-04
CMR-8170 Create tests to verify user can delete collection subscriptions with valid permissions

dpesall:
This change creates new tests to verify that a user with valid permissions can perform a delete on a collection subscription.

### DIFF
--- a/system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
+++ b/system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
@@ -546,31 +546,25 @@
               guest-token (echo-util/login-guest (system/context))
               response (ingest/delete-concept concept
                                               {:token guest-token})]
-          (is (= ["You do not have permission to perform that action."] (:errors response)))))))
-  (testing "delete a collection subscription"
-    (let [coll1 (data-core/ingest-umm-spec-collection
-                 "PROV1"
-                 (data-umm-c/collection
-                  {:ShortName "coll1"
-                   :EntryTitle "entry-title1"})
-                 {:token "mock-echo-system-token"})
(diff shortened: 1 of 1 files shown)


daniel-zamora (approved):

system-int-test/test/cmr/system_int_test/ingest/subscription_ingest_test.clj
```diff
@@ -528,6 +528,50 @@
...
+(deftest delete-collection-subscription-ingest-test
+  (mock-urs/create-users (system/context) [{:username "someSubId" :password "Password"}])
+  (testing "delete a collection subscription without valid permissions"
+    (let [coll1 (data-core/ingest-umm-spec-collection
+                  "PROV1"
+                   (data-umm-c/collection
+                    {:ShortName "coll1"
+                     :EntryTitle "entry-title1"})
+                   {:token "mock-echo-system-token"})]
+      (mock-urs/create-users (system/context) [{:username "someSubId" :password "Password"}])
```
TylerHeald1, line 540: You have this line twice, is one redundant?

dpesall: Yes it was partially redundant since the 2nd one was an identical declaration. I now removed the 2nd one and shifted the 2nd test to be nested inside the first let.

TylerHeald1: I should have clarified, I was referring to the mock-urs/create-users call. Its been a while since I've used that command, but it seems like you'd only need the one at the start of the deftest

dpesall: You're right, that 2nd use of it was completely redundant in that case!

PR #1559, opened 2022-05-12
CMR-6175: As a CMR Ingest Client, I Want a More Meaningful Error Status in the Event of Multiple, Overlapping Ingests.

ygliuvt:


jaybarra (approved):

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
```diff
@@ -421,14 +421,8 @@
...
-            error-code (cond
-                         (re-find #"unique constraint.*_CID_REV" error-message)
+            error-code (if (re-find #"unique constraint.* violated" error-message)
                          :revision-id-conflict
-
-                         (re-find #"unique constraint.*_CON_REV" error-message)
-                         :revision-id-conflict
-
-                         :else
                          :unknown-error)]
```
eereiter, line 426: How does this actual fix the problem?  I just looks like you made the same code better.

eereiter (approved):

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
ygliuvt: For granule ingest conflict, the constraint name in the error message is `METADATA_DB.ASF_GRANULES_CRI`, the new search pattern will catch it. The old pattern is too restricted for no reason.

PR #1560, opened 2022-05-12
Cmr 7856

TylerHeald1:


jaybarra:
Is this necessary? The default numeric type for clojure is `java.lang.Long` and not Integer. Long max value should be sufficient.

system-int-test/src/cmr/system_int_test/utils/search_util.clj
```diff
@@ -371,7 +371,7 @@
        (dj/parse-echo-json-result body)
        (if (= status 200)
          {:status status
-          :hits (Integer/parseInt (get headers "CMR-Hits"))
+          :hits (get headers "CMR-Hits")
```
jaybarra, line 374: 👍

transmit-lib/src/cmr/transmit/search.clj
```diff
@@ -51,7 +51,7 @@
         {:keys [headers body]} response
         status (int (:status response))]
     (case status
-      200 (Integer/parseInt (get headers "CMR-Hits"))
+      200 (get headers "CMR-Hits")
```
jaybarra, line 54: 👍

system-int-test/test/cmr/system_int_test/search/acls/collection_test.clj
```diff
@@ -360,7 +360,7 @@
                                     :collection {:token guest-token
                                                  :page-size 100
                                                  :concept-id (:concept-id coll5)})]
-        (is (= hits (count (:entries results))))
+        (is (= (bigint hits) (count (:entries results))))
```
jaybarra, line 363: The bigint casts may not be necessary since Clojure will default to Long. Would like to see a test that forces exceeds the Integer/MAX_VALUE

jaybarra (changes requested):

ygliuvt (changes requested):
I want to request change on this PR for a different reason (i.e. not adding test beyond Integer/MAX_VALUE). I think `bigint` is an overkill. We should just use `long` for the hits and I want the hits to be parsed out in the transmit library as before rather than put it on the caller of the utility function.

system-int-test/test/cmr/system_int_test/search/granule_counts_search_test.clj
```diff
@@ -752,3 +753,38 @@
 
               "granule count in json format"
               :atom (search/find-concepts-json :collection {:include-granule-counts false})))))
+
+
+(deftest over-max-int-granule-hits-test
```
ygliuvt, line 758: Remove or comment out this test. We may want to manually run a test to verify that we can go over the `Integer/MAX_VALUE` limit, but we certainly don't want this test in our integration test suite.

ygliuvt, line 758: Remove or comment out this test. We may want to manually run a test to verify that we can go over the `Integer/MAX_VALUE` limit, but we certainly don't want this test in our integration test suite.

system-int-test/test/cmr/system_int_test/search/acls/collection_test.clj
TylerHeald1: Going to opt for no automatic test in the test suite for this due to concerns over time and resource constraints

jaybarra (approved):

spatial-lib/src/cmr/spatial/point.clj
```diff
@@ -168,7 +169,14 @@
            :lat-rad lat-rad
            :options options
            :geodetic-equality geodetic-equality
-           not-found)))
+           not-found))
+  
+  java.lang.Iterable
```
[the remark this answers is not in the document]
TylerHeald1: When I was trying to make a test for the change, there was an error that kept being thrown about iterator() being an abstract method in the Point deftype. When I looked into it, clojure.lang.IPersistentMap inherits the method iterator() from java.lang.Iterable but that hadn't been defined in the deftype so I had to add that in. I don't know where it might be needed now, but I figured I'd leave it in despite no longer trying to make a test case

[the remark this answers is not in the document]
TylerHeald1: removed

spatial-lib/src/cmr/spatial/point.clj
```diff
@@ -138,6 +138,7 @@
   (equiv [this o]
          (and (isa? (class o) Point)
               (.equals this o)))
+  ;; The following functions need to be here because this type implements IPersistentMap, which inherits the functions.
```
ygliuvt, line 141: Remove this comment.

ygliuvt: I think the existing comment is saying we don't know why the error only shows up in CI, not in local testing.

ygliuvt (approved):

PR #1561, opened 2022-05-17
CMR-8236 Separated the two errors into their own functions and hence …

eereiter:
…the errors are now also separated with error messages an end user will understand. Also removed dead code and include statements.


ingest-app/src/cmr/ingest/validation/business_rule_validation.clj
```diff
@@ -89,11 +82,47 @@
...
+                              (mdb/get-providers context))
+        consortiums (when consortiums-str
+                      (remove empty? (string/split (string/upper-case consortiums-str) #" ")))]
+    (when (and (= true standard-product)
+               (not (some #(= "EOSDIS" %) consortiums)))
+      [(format (str "Standard product validation failed: "
+                    "Standard Product designation is only allowed for NASA data products. This collection is "
+                    "being ingested using a non-NASA provider which means the record is not a NASA record. "
+                    "Please remove the StandardProduct element from the record. "
+                    "The consortium designations include the following [%s] and EOSDIS is not one of them.")
```
jaybarra, line 102: Would suggest wording slightly differently if this is something the user can address by including the EOSDIS consortium. Such as "Standard Product designation is only allowed for NASA data products. To be designated as a NASA record and Standard Product this collection must include EOSDIS in its current list of consortiums: %s"

If the user can't fix this because its done at the provider level we can probably cut some of the message.

jaybarra (approved):

ingest-app/src/cmr/ingest/validation/business_rule_validation.clj
eereiter: The curator generally does not know about consortiums.  Also consortiums are part of the provider and not included in the metadata. Adding/defining consortiums is something that a CMR operator must do. I was just trying to add that into the message so that a CMR operator would know what to do.

eereiter: Ok, I got rid of the last sentence.

PR #1566, opened 2022-05-19
Cmr 8131 standard product translate dif10

sxu123:


umm-spec-lib/src/cmr/umm_spec/umm_to_xml_mappings/iso19115_2.clj
```diff
@@ -278,7 +278,7 @@
                   [:gmd:codeSpace [:gco:CharacterString "gov.nasa.esdis.umm.collectiondatatype"]]
                   [:gmd:description [:gco:CharacterString "Collection Data Type"]]]])
             (let [standard-product (:StandardProduct c)]
-              (when (not (nil? standard-product))
+              (when-not (nil? standard-product)
```
ygliuvt, line 281: nit: change this line to `(when (some? standard-product)`

ygliuvt (approved):

umm-spec-lib/resources/example-data/dif10/sample_collection.xml
```diff
@@ -139,5 +139,19 @@
...
-  </Extended_Metadata>
+    </Extended_Metadata>
+    <Extended_Metadata>
+    <Metadata>
+      <Group>gov.nasa.gsfc.gcmd.standardproduct</Group>
+      <Name>StandardProduct</Name>
+      <Value>true</Value>
+    </Metadata>
+    </Extended_Metadata>
+    <Extended_Metadata>
```
jceaser, line 150: why are these in here twice? Also, I think you can have multiple metadata tags in an extended_metadata tag.

umm-spec-lib/src/cmr/umm_spec/xml_to_umm_mappings/dif10.clj
```diff
@@ -317,6 +317,10 @@
    :TilingIdentificationSystems (spatial/parse-tiling doc)
    :ProcessingLevel {:Id (su/with-default (value-of doc "/DIF/Product_Level_Id") sanitize?)}
    :AdditionalAttributes (aa/xml-elem->AdditionalAttributes doc sanitize?)
+   :StandardProduct (last (for [metadata (select doc "/DIF/Extended_Metadata/Metadata")
```
jceaser, line 320: so if you put in more then one, your going to ignore all but the last? Also will this still work if there is one ExtendedMetadata with several Metadata tags instead of multiple ExtendedMetadata tags?

umm-spec-lib/resources/example-data/dif10/sample_collection.xml
jceaser: You have two in here to prove that it pulls the last I see, I retract my statement but ask that you confirm the code works with single extended_metatada tags containing multiple metadata tags.

jceaser (changes requested):
I'm afraid I need to request that a use case be created for metadata that contains one extended_metadata tag that contains multiple metadata tags instead of the existing test which has multiple extended_metadata tags containing a single metadata tag.

umm-spec-lib/src/cmr/umm_spec/xml_to_umm_mappings/dif10.clj
sxu123: Yes, it works.

umm-spec-lib/resources/example-data/dif10/sample_collection.xml
sxu123: Added unit tests.

jaybarra (approved):

jceaser (approved):

PR #1568, opened 2022-05-23
CMR-8163 handling multiple tokens

jaybarra:
CMR side of addressing case of multiple tokens passed in a request

### DIFF
--- a/acl-lib/src/cmr/acl/core.clj
+++ b/acl-lib/src/cmr/acl/core.clj
@@ -28,11 +28,18 @@
 (defn get-token
   "Returns the token the user passed in the headers or parameters"
   [params headers]
-  (let [non-empty-string #(when-not (str/blank? %) %)]
-    (or (non-empty-string (get headers tc/token-header))
-        (non-empty-string (:token params))
-        (when (allow-echo-token) 
-          (non-empty-string (get headers tc/echo-token-header))))))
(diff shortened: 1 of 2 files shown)


isja17 (approved):

acl-lib/src/cmr/acl/core.clj
```diff
@@ -28,11 +28,18 @@
 (defn get-token
   "Returns the token the user passed in the headers or parameters"
   [params headers]
-  (let [non-empty-string #(when-not (str/blank? %) %)]
-    (or (non-empty-string (get headers tc/token-header))
-        (non-empty-string (:token params))
-        (when (allow-echo-token) 
-          (non-empty-string (get headers tc/echo-token-header))))))
+  (let [non-empty-string #(when-not (str/blank? %) %)
```
ygliuvt, line 31: nit: Make `non-empty-string` a function rather than define it in the `let` block.

ygliuvt:
Is this the only code change that is needed to fix CMR-8163?

jaybarra:
> Is this the only code change that is needed to fix CMR-8163?

@ygliuvt No, there are additional changes in legacy-services. This prevents a corner-case where the token could be passed in twice.

ygliuvt (approved):

PR #1588, opened 2022-06-14
CMR-8076 Updated API docs for granules with multiple spatial features.

dpesall:
This change updates the spatial query documentation to specify how spatial queries for granules which have multiple spatial features (i.e. a Polygon and a Bounding Box) are handled by spatial filters.


jaybarra (approved):

ygliuvt (changes requested):
I think CMR `AND` different spatial conditions rather than `OR` for granule spatial search which is the same as collection spatial search. This documented the contrary.

ygliuvt (approved):

PR #1603, opened 2022-07-13
CMR-8379 - renaming migration file and restoring native id

jceaser:
Adding the native-id back into the table.

If you have used the previous version of this migration, then you will need to run (down) function.


sxu123:
I couldn't request changes for some reason.  You need to change 78 to 80.  It's not shown but if I look at the file, I see it.

sxu123 (changes requested):
You need to change 78 to 80.  It's not shown but if I look at the file, I see it.

sxu123 (approved):

zimzoom (approved):
Tested this migration, worked for me. Since I already had a previous version, had to manually call (down) then (up) from within the 080 migration namespace

PR #1605, opened 2022-07-14
CMR-8493 updated gem versions

jaybarra:
updating gems for system validation tests


jceaser (approved):
I like that we are removing the versions, maybe it will be less work in the future

dpesall (approved):

PR #1608, opened 2022-07-15
CMR-7849: Update CMR to check length of strings for text fields it attempts to ingest

dpesall:
Elasticsearch has built in limits for types it supports. CMR does not check length of strings for any text field it attempts to ingest. It is possible to attempt to save documents to elasticsearch with fields containing values longer than the supported maximum size `text` fields can support.


ygliuvt:
I am not sure this PR addresses the ticket. I am also a little puzzled by the ticket itself. I don't see anything wrong to just let Elasticsearch reject records with exceedingly long string values. 

We have various schemas defining the max length of fields in the metadata. If we want to make sure no string fields exceed the max length of a keyword type in ES, a good place to do that is in the metadata schemas. I believe 32776 is long enough for almost all the fields in metadata. In the rare cases where a certain field could have string value longer than 32776, we could use `text` type in ES, like we did with `keyword`. 

For fields like provider-id, short-name, entry-title, etc. we already have validations that they won't be longer than a certain size. e.g The max length of provider-id is only 10 characters. We don't need to have code in indexer to truncate them down. 

Again, I don't see anything wrong to just let Elasticsearch reject records with exceedingly long string values. Because:
1) We very rarely see this occur.
2) We should improve our metadata schemas to put in max length constraint.
3) It would be a better idea to let providers know by not indexing the record so that they cam make their data better rather than silently truncate it down without telling anyone.

dpesall:
> 

You make a lot of good points and I wish I would have asked Jay more details about what exactly he wanted from the ticket I suppose before he left for vacation since the acceptance criteria wasn't extremely specific. I also found it to be highly unlikely to have any text field with the exception of `:summary` (which was part of `keyword2`) ever be remotely close to the text limit but I still figured I would go ahead and limit the fields that could potentially have a long value set to it.

And I have a question regarding the current validations limiting fields like `short-name` -- the current validation only throws a warning (e.g. "ShortName: expected maxLength: 85, actual: 7346" or "#/AncillaryKeywords/0: expected maxLength: 255, actual: 12675"), but these currently do not prevent an ingest and still return a 200 status code. When these fields have character lengths beyond the elastic search limit, they will still return a 200 but will not be visible when searching for collections. I also agree with the fact that letting providers know about the limit rather than silently truncating is a better solution, but after trying a couple solutions that I felt were worse than what I currently have I decided to go with this approach but I'm open to different options.

PR #1613, opened 2022-07-25
CMR-8105: Changed reindexing of concepts after a date time to go through SQS.

ygliuvt:


jaybarra (approved):

bootstrap-app/src/cmr/bootstrap/data/bulk_index.clj
```diff
@@ -271,6 +271,29 @@
     (info "Deleted " total " concepts")
     total))
 
+(defn index-provider-data-later-than-date-time
+  "Index all concept revisions created later than or equal to the given date-time for a given provider."
+  [system provider-id date-time]
+  (info (format "Indexing concepts with revision-date later than [%s] for provider [%s] started."
+                date-time
+                provider-id))
+  (if (= "CMR" provider-id)
```
eereiter, line 280: We are hard coding CMR - are there other places where we use this string?  Should it be in a defconfig or something similar?

eereiter (approved):

bootstrap-app/src/cmr/bootstrap/data/bulk_index.clj
ygliuvt: This is the name of the provider-id that will be used to index system concepts. This will be documented in the README when it will be exposed to clients when we allow clients to use it to index system concepts in a different ticket CMR-8107. I will make it a `def` when working on that ticket.

PR #1615, opened 2022-07-26
CMR-7855 Subscription email contents are now logged along with sub-id…

eriknelson11:
… and email-address

### DIFF
--- a/ingest-app/src/cmr/ingest/services/subscriptions_helper.clj
+++ b/ingest-app/src/cmr/ingest/services/subscriptions_helper.clj
@@ -237,7 +237,8 @@
          (try
            (send-email email-settings email-content)
            (info (str "Successfully processed subscription [" sub-id "].
-                      Sent subscription email to [" email-address "]."))
+                      Sent subscription email to [" email-address "].
+                      \nSubscription email contents: [" email-content "]."))
            (when update-notification-time?
              (send-update-subscription-notification-time! context sub-id))
            (catch Exception e


eereiter (approved):

ingest-app/src/cmr/ingest/services/subscriptions_helper.clj
```diff
@@ -237,7 +237,8 @@
          (try
            (send-email email-settings email-content)
            (info (str "Successfully processed subscription [" sub-id "].
-                      Sent subscription email to [" email-address "]."))
+                      Sent subscription email to [" email-address "].
```
jceaser, line 240: The only thing else I can think of is to add `update-notification-time?` and/or start and stop times from the subscription.

eriknelson11: Okay, after running subscription_processing_test.clj, the logged message was not in the output so I wasn't able to see what the exact logged message would be from the tests. The ticket mentions they only care about the last week or so, I think `update-notification-time?` is a good idea.

daniel-zamora:
```
(user/set-logging-level! :info)
```
That should do it.  Default is :error.

eriknelson11: > The only thing else I can think of is to add `update-notification-time?` and/or start and stop times from the subscription. What do you think? I'm going with the philosophy that you never have the data you need in logs.
After looking further `update-notification-time?` is a boolean I believe and `email-content` is the returned object from the function `create-email-content` which contains the start and stop times. So I think just adding `email-content' to the log will provide the needed information.

jceaser (approved):
my questions were considered, approving the code.

tejusp (approved):

PR #1621, opened 2022-07-29
CMR-8376: Add feature toggle from new generic document pipeline

zimzoom:
These changes include just the feature toggle from the new document pipeline prototype, which is set by an environment variable that also serves as the list of approved document types. Also in these changes is a log info line that is activated when the environment variable is set. (If not set, it should default to an empty string and thus not trigger the log.)

### DIFF
--- a/common-lib/src/cmr/common/config.clj
+++ b/common-lib/src/cmr/common/config.clj
@@ -3,6 +3,7 @@
   through an environment variable. Configuration items should be added using the defconfig macro."
   (:require
    [camel-snake-kebab.core :as csk]
+   [cheshire.core :as json]
    [clojure.edn :as edn]
    [clojure.set :as set]
    [clojure.string :as str]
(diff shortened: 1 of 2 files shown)


common-lib/src/cmr/common/config.clj
```diff
@@ -234,3 +245,6 @@
                (pr-str unknown-vars))
          true)
        false))))
+
+(when (seq (approved-pipeline-documents))
+  (info "Prototype feature toggle detected"))
```
jaybarra, line 250: Should try to end with a newline. See if your editor has a config to add them automatically. Could we also name the prototype?

jaybarra (approved):

common-lib/src/cmr/common/config.clj
jceaser: actually this feature will stay forever, it should say something like "Generic documents support: %s" approved-pipeline-documents

zimzoom: thanks yall, i added the newline manually for now, for the editor config situation for VS Code has become complicated.

eereiter (approved):

PR #1622, opened 2022-07-29
CMR-8521 Fixing the generic concepts concept id prefix, so that each generic concept can have their own.

eereiter:
…generic concept can have their own.

### DIFF
--- a/transmit-lib/src/cmr/transmit/generic_documents.clj
+++ b/transmit-lib/src/cmr/transmit/generic_documents.clj
@@ -22,13 +22,14 @@
   ([conn provider-id]
    {:pre [conn provider-id]}
    (println "in concept ingest url")
-   (format "%s/generics/%s" (conn/root-url conn) (codec/url-encode provider-id)))
+   (format "%s/generics/%s/%s" (conn/root-url conn) (codec/url-encode provider-id) "nativeId"))
   ([conn provider-id concept-id]
    {:pre [conn concept-id]}
-   (format "%s/generics/%s/%s"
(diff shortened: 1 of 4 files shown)


transmit-lib/src/cmr/transmit/generic_documents.clj
```diff
@@ -22,13 +22,14 @@
   ([conn provider-id]
    {:pre [conn provider-id]}
    (println "in concept ingest url")
-   (format "%s/generics/%s" (conn/root-url conn) (codec/url-encode provider-id)))
+   (format "%s/generics/%s/%s" (conn/root-url conn) (codec/url-encode provider-id) "nativeId"))
```
ygliuvt, line 25: I don't understand why we are passing in a hardcoded `nativeId` in the routes here and below.

eereiter: Thats because we are in the process of putting native id back in.  currently in the feature/generic-document-pipeline native id is in metadata_db but not in ingest or the indexer app.  So I put it in there so that I could ingest 1 document so that I can continue on with my work.  I can remove it.  Thomas is working on putting in the native id back in the rest of the code, so by the end of this sprint this code will be removed.

zimzoom (approved):

transmit-lib/src/cmr/transmit/generic_documents.clj
ygliuvt: ok. It would be nice to have a comment with TODO.

ygliuvt (approved):

PR #1631, opened 2022-08-04
CMR-8244 Spatial search when entry has multiple spatial representations yields overly conservative results

dpesall:
The purpose of this ticket is to address the issue with a subset of granules that have multiple spatial types which includes a Bounding Rectangle that encompasses the entire swath of the granule's polygons -- this often results in global Bounding Rectangles (see granules in [this collection](https://search.earthdata.nasa.gov/search/granules?p=C2075141605-POCLOUD&pg[0][v]=f&pg[0][gsk]=-start_date&q=ascat%20POCLOUD&tl=1652128563.203!3!!&lat=-0.0703125)) This cause spatial searches to return granules when there is no data, because current functionality will return a granule if the spatial search matches _any_ of the geometries of the granule.

The two new functionalities from this ticket allow users to apply `every` and `ignore-br` to spatial polygon searches. The `every` flag will require the spatial polygon to match every single geometry within the granule. This can create issues for granules such as the one linked earlier that has a global Bounding Rectangle and then two Polygons, so it would only return if it matched both polygons which is usually not the preferred behavior. The `ignore-br` will ignore any bounding boxes in the granule and only check for intersections with other granule geometries such as polygons which addresses the use case of the ticket.


es-spatial-plugin/src/cmr/elasticsearch/plugins/spatial/script/core.clj
```diff
@@ -74,11 +117,16 @@
   (-> this .data :search-lookup (.setDocument doc-id)))
 
 (defn- -init [^Object intersects-fn ^Map params ^SearchLookup lookup ^DocReader doc-reader]
+  ;(.info (LogManager/getLogger) (str "====PARAMS" params))
```
TylerHeald1, line 120: Remove all the debug code

TylerHeald1 (changes requested):

es-spatial-plugin/src/cmr/elasticsearch/plugins/spatial/script/core.clj
dpesall: Fixed

system-int-test/test/cmr/system_int_test/search/granule_search_test.clj
```diff
@@ -553,3 +554,94 @@
...
+                                                     {:provider-id "PROV1"
+                                                      :concept-type :granule
+                                                      :concept-id "G23-PROV1"
+                                                      :native-id "G23-granule"
+                                                      :format-key :echo10})
+        intersects-none [29.8125 19.0954 36.5625 12.90915 53.15625 23.0321 40.5 35.6858 29.8125 19.0954]
+        intersects-bbox [-110.8125 -7.61796 -104.34375 -19.42808 -86.90625 -21.11524 -84.09375 -5.9308 -110.8125 -7.61796]
+        intersects-polygon [99 39.30683 67.5 33.12058 76.5 11.74989 104.625 21.31046 99 39.30683]
+        intersects-both [-150.75 -36.61535 162.5625 -29.30433 176.0625 -84.9806 -145.6875 -89.47969 -150.75 -36.61535]
+        __ (index/wait-until-indexed)]
```
daniel-zamora, line 575: a single _ works fine here

dpesall: Fixed

ygliuvt (changes requested):
Need to update API doc explaining what the new spatial parameters are and how to use them.

es-spatial-plugin/src/cmr/elasticsearch/plugins/spatial/script/core.clj
```diff
@@ -28,23 +30,64 @@
...
+
 (defn- get-from-fields
   [^LeafStoredFieldsLookup lookup key]
   (when (and lookup key (.containsKey lookup key))
     (when-let [^FieldLookup field-lookup (.get lookup key)]
       (seq (.getValues field-lookup)))))
 
+(defn any-true?
+  "Returns true if predicate f returns a truthy value against any of the items.
+  This is very similar to some but it's faster through it's use of reduce."
```
ygliuvt, line 53: `some` returns the first logical true value without going through the rest of the list. I don't see how this is better.

es-spatial-plugin/src/cmr/elasticsearch/plugins/spatial/script/core.clj
```diff
@@ -28,23 +30,64 @@
...
+  "Returns true if predicate f returns a truthy value against all of the items."
+  [f items]
+  (every? true?
+    (map f items)))
+
+(defn remove-br
+  "Removes bounding rectangles from granule spatial data."
+  [shapes]
+  (let [shapes-no-br (into [] (remove nil? (for [shape shapes]
+                                             (if (not (= 5 (count shape)))
```
ygliuvt, line 72: Add comment on why `(= 5 (count shape))` means it is a bounding rectangle. Also there is no need to have the `for` loop.

es-spatial-plugin/src/cmr/elasticsearch/plugins/spatial/script/core.clj
dpesall: This is a function pulled from `common-lib/util.clj`. I initially created a function from that file and added `every-true` to it as well but it was not detecting my changes to `util.clj` when I did a full rebuild so I added them to this file instead. I'm not sure how it is better either, I thought functionally it identical and much less readable than using `some`. I've replaced it with `some` now.

es-spatial-plugin/src/cmr/elasticsearch/plugins/spatial/script/core.clj
dpesall: Changed logic per your suggestion and added clarifying comment for the use of `(= 5 (count ...))`

dpesall: Changed to use `(instance)` instead of `(count)`.

dpesall:
> Need to update API doc explaining what the new spatial parameters are and how to use them.

API docs updated to reflect usage of new flags

search-app/src/cmr/search/data/complex_to_simple_converters/spatial.clj
```diff
@@ -156,6 +157,13 @@
                 (crossing-ranges->condition crossings)])))
           lat-ranges-crossings))))
 
+(defn- spatial-operator
+  "Returns the operator used to determine if spatial queries must match all polygons."
+  [params]
+  (if (= (type params) clojure.lang.PersistentArrayMap)
```
daniel-zamora, line 163: is there a specific reason `(map?)` isn't used?

dpesall: No good reason. Pushed a change to simplify it to use `(map?)`.

es-spatial-plugin/src/cmr/elasticsearch/plugins/spatial/script/core.clj
```diff
@@ -28,23 +30,57 @@
...
   [^LeafStoredFieldsLookup lookup key]
   (when (and lookup key (.containsKey lookup key))
     (when-let [^FieldLookup field-lookup (.get lookup key)]
       (seq (.getValues field-lookup)))))
 
+(defn any-true?
+  "Returns true if predicate f returns a truthy value against any of the items.
+  This is very similar to some but it's faster through it's use of reduce."
+  [f items]
+  (some true?
```
ygliuvt, line 55: This should be `(some f items)`.

es-spatial-plugin/src/cmr/elasticsearch/plugins/spatial/script/core.clj
```diff
@@ -28,23 +30,57 @@
...
+    (map f items)))
+
+(defn every-true?
+  "Returns true if predicate f returns a truthy value against all of the items."
+  [f items]
+  (every? true?
```
ygliuvt, line 61: This can be simplified as `(every? f itmes)`.

es-spatial-plugin/src/cmr/elasticsearch/plugins/spatial/script/core.clj
```diff
@@ -28,23 +30,57 @@
...
+    (map f items)))
+
+(defn remove-br
+  "Removes bounding rectangles from granule spatial data."
+  [shapes]
+  (let [shapes-no-br (into [] (remove #((instance? cmr.spatial.mbr.Mbr (first shapes)) shapes)))]
```
ygliuvt, line 67: I don't think this line is correct. Shouldn't this be `(remove #(instance? cmr.spatial.mbr.Mbr (first %))`?

es-spatial-plugin/src/cmr/elasticsearch/plugins/spatial/factory/lfactory.clj
```diff
@@ -2,7 +2,8 @@
   (:import
    (cmr.elasticsearch.plugins SpatialScript)
    (java.util Map)
-   (org.elasticsearch.script DocReader) 
+   (org.apache.logging.log4j Logger LogManager)
```
ygliuvt, line 5: This is not used, can be removed.

es-spatial-plugin/src/cmr/elasticsearch/plugins/spatial/script/core.clj
```diff
@@ -4,7 +4,9 @@
    [cmr.spatial.serialize :as srl])
   (:import
    (java.util Map)
-   (org.elasticsearch.script DocReader) 
+   (org.elasticsearch.script DocReader)
+   (org.apache.logging.log4j Logger LogManager)
```
ygliuvt, line 8: Not used. Remove.

es-spatial-plugin/src/cmr/elasticsearch/plugins/spatial/script/core.clj
dpesall: Fixed.

es-spatial-plugin/src/cmr/elasticsearch/plugins/spatial/script/core.clj
dpesall: Fixed.

es-spatial-plugin/src/cmr/elasticsearch/plugins/spatial/script/core.clj
dpesall: Fixed.

es-spatial-plugin/src/cmr/elasticsearch/plugins/spatial/factory/lfactory.clj
ygliuvt: If this is the case, I would like to see a commented out log line to show how logging is supposed to be done in the plugin.

es-spatial-plugin/src/cmr/elasticsearch/plugins/spatial/script/core.clj
```diff
@@ -28,23 +30,55 @@
...
   [^LeafStoredFieldsLookup lookup key]
   (when (and lookup key (.containsKey lookup key))
     (when-let [^FieldLookup field-lookup (.get lookup key)]
       (seq (.getValues field-lookup)))))
 
+(defn any-true?
+  "Returns true if predicate f returns a truthy value against any of the items.
+  This is very similar to some but it's faster through it's use of reduce."
+  [f items]
+  (some f items))
```
ygliuvt, line 55: This function is so trivial. It is better to just use the clojure function `some` in place where it is called.

es-spatial-plugin/src/cmr/elasticsearch/plugins/spatial/script/core.clj
```diff
@@ -28,23 +30,55 @@
...
+
+(defn every-true?
```
ygliuvt, line 57: This function is so trivial. It is better to just use the clojure function `every?` in place where it is called. Remove this function.

es-spatial-plugin/src/cmr/elasticsearch/plugins/spatial/factory/lfactory.clj
dpesall: Added

es-spatial-plugin/src/cmr/elasticsearch/plugins/spatial/script/core.clj
dpesall: Simplified

es-spatial-plugin/src/cmr/elasticsearch/plugins/spatial/script/core.clj
dpesall: Simplified

search-app/src/cmr/search/data/complex_to_simple_converters/spatial.clj
```diff
@@ -198,9 +206,12 @@
                          (orbital-condition context shape))
           mbr-cond (br->cond "mbr" (srl/shape->mbr shape))
           lr-cond (br->cond "lr" (srl/shape->lr shape))
-          spatial-script (shape->script-cond shape)
+          operator (spatial-operator (:polygon (:query-params context)))
```
ygliuvt, line 209: So this `ignore-br` option only works with search by `polygon`? What about searching by `bounding_box`, `line`, etc? I don't think this is enough.

ygliuvt: This search option should be in the form of options that apply to all spatial search parameters, not on `polygon`.

ygliuvt: I think the code change in spatial plugin supports all spatial types, it is just the way this option is defined in search API that needs to be changed.

search-app/docs/api.md
```diff
@@ -2447,7 +2447,23 @@ For granule additional attributes search, the default is searching for the attri
 
 #### <a name="g-spatial"></a> Find granules by Spatial
 The parameters used for searching granules by spatial are the same as the spatial parameters used in collections searches. (See under "Find collections by Spatial" for more details.)
+
 Note: When querying a granule which has multiple types of spatial features in the granule metadata (i.e. a Polygon and a Bounding Box), the granule will be returned if the spatial query matches at least one of the spatial types on the given granule (i.e. matches the granule's Polygon OR Bounding Box).
+This may not always be the desired behavior if a granule has a bounding rectangle which is less precise than the polygons. For this, you may use the 'ignore_br' or 'every' keywords when doing spatial searches. These flags can be applied to any type of spatial search (polygon, bounding box, circle, line, point).
+
+The 'ignore_br' flag will ignore the bounding rectangles of the granule and look to match at least one other geometry (i.e. one or more of the polygons).
+
+    curl "%CMR-ENDPOINT%/granules?collection_concept_id=%CMR-EXAMPLE-COLLECTION-ID%&polygon[ignore_br]=10,10,30,10,30,20,10,20,10,10"
```
ygliuvt, line 2456: what if there are multiple spatial areas in the search parameter? I think a better way to pass in the spatial search `ignore_br` is via the options, i.e. `options[spatial][ignore_br]=true`.

dpesall: If there are multiple spatial areas in the search parameter, the flag can currently be applied to each spatial area:

curl "http://localhost:3003/collections?polygon[ignore_br]=10,10,30,10,30,20,10,20,10,10&polygon[every]=11,11,31,11,31,21,11,21,11,11"

Passing it via options would mean all spatial areas would be restricted to the same behavior. I could still make the change if you prefer, but it would require _significant_ code change at this point to move it to options.

ygliuvt: I know it will require significant code change to switch to `options` parameter. It is unfortunate that we didn't settle on the best API before starting implementation. I still think the `options` parameter makes more sense than individual spatial parameters. I just don't see the use case of the example search that you give above. I don't even see the use case of the `every` option. I don't want to force you to rewrite everything. I would say bring this up with the team and get other people's opinions.

dpesall: Made the suggested change to API interaction to be of the format `options[spatial][ignore_br]=true`

es-spatial-plugin/src/cmr/elasticsearch/plugins/spatial/script/core.clj
```diff
@@ -28,23 +30,46 @@
...
 (defn doc-intersects?
   "Returns true if the doc contains a ring that intersects the ring passed in."
-  [^LeafStoredFieldsLookup lookup intersects-fn]
+  [^LeafStoredFieldsLookup lookup params intersects-fn]
   ;; Must explicitly return true or false or elastic search will complain
   (if-let [ords-info (get-from-fields lookup "ords-info")]
     (let [ords (get-from-fields lookup "ords")
-          shapes (srl/ords-info->shapes ords-info ords)]
+          shapes (srl/ords-info->shapes ords-info ords)
+          shapes-no-br (remove-br shapes)
```
ygliuvt, line 64: Change this line to `shapes-no-br (remove #(instance? cmr.spatial.mbr.Mbr %) shapes)`, and remove the function `remove-br`.

search-app/src/cmr/search/data/complex_to_simple_converters/spatial.clj
```diff
@@ -189,6 +190,18 @@
 
              (keys crossings-map))))))
 
+(defn- extract-operator
```
ygliuvt, line 193: I think it would be good to return validation error if the `operator` in `options[spatial][operator]` is not `exclude_br` or `every`.

es-spatial-plugin/src/cmr/elasticsearch/plugins/spatial/script/core.clj
dpesall: Fixed

search-app/src/cmr/search/data/complex_to_simple_converters/spatial.clj
dpesall: Currently if you try an invalid operator in options such as `options[spatial][abc]` then the API will return 
`{
"errors": [
"Option [asd] is not supported for param [spatial]"
]
}`

It will return this error for any spatial parameter not defined in `(def spatial-options #{:ignore-br :every})` in file `search.services.parameters.parameter-validation.clj`

Is this sufficient or should I implement an additional validation?

ygliuvt: This is enough. Maybe I missed it, but I don't see a validation test.

dpesall: Added validation test case for invalid spatial options.

ygliuvt (approved):

PR #1633, opened 2022-08-04
(feature branch only) CMR-8522 - adding native-id support in ingest

jceaser:


metadata-db-app/src/cmr/metadata_db/services/generic_documents.clj
```diff
@@ -27,20 +22,21 @@
...
    it into the expected response by pulling out generated fields like concept-id
    and putting them in the :Info field, then putting the metadata inside the
    :Metadata field without any of these extra fields."
   [raw-doc]
-  {:Info (assoc-these-from info-fields raw-doc)
+  {:Info (select-keys raw-doc info-fields)
    :Metadata (reduce #(dissoc %1 %2) raw-doc fields-to-drop)})
 
+;; drop this?
 (defn- create-error-map
```
jceaser, line 40: yes, lets get rid of this

transmit-lib/src/cmr/transmit/generic_documents.clj
```diff
@@ -23,12 +23,12 @@
    {:pre [conn provider-id]}
    (println "in concept ingest url")
    (format "%s/generics/%s" (conn/root-url conn) (codec/url-encode provider-id)))
```
jceaser, line 25: drop single url format as it is no longer needed

eereiter (approved):

metadata-db-app/src/cmr/metadata_db/services/generic_documents.clj
```diff
@@ -27,27 +22,21 @@
...
 
 (defn- raw-generic->response
   "Take the raw database object returned from either SQL or In-Memory and shape
    it into the expected response by pulling out generated fields like concept-id
    and putting them in the :Info field, then putting the metadata inside the
    :Metadata field without any of these extra fields."
   [raw-doc]
-  {:Info (assoc-these-from info-fields raw-doc)
+  {:Info (select-keys raw-doc info-fields)
    :Metadata (reduce #(dissoc %1 %2) raw-doc fields-to-drop)})
```
sxu123, line 37: Why are you not getting metadata directly like (select-keys raw-doc [:metadata]), instead of removing all the rest of the unwanted fields?

transmit-lib/src/cmr/transmit/generic_documents.clj
```diff
@@ -18,38 +18,27 @@
...
 ;; does an apply with an array of paramaters to resolve them. The calling
 ;; function must pass in the URL parameters as an array. Create can use the
 ;; single param function.
 (http-helper/defcreator2 create-generic :metadata-db concept-ingest-url {:use-system-token? true :raw? true})
-(http-helper/defgetter read-generic :metadata-db concept-ingest-url2 {:use-system-token? true :raw? true})
-(http-helper/defupdater update-generic :metadata-db concept-ingest-url2 {:use-system-token? true})
-(http-helper/defdestroyer delete-generic :metadata-db concept-ingest-url2 {:use-system-token? true})
+(http-helper/defgetter read-generic :metadata-db concept-ingest-url {:use-system-token? true :raw? true})
+(http-helper/defupdater update-generic :metadata-db concept-ingest-url {:use-system-token? true})
+(http-helper/defdestroyer delete-generic :metadata-db concept-ingest-url {:use-system-token? true})
```
sxu123, line 44: Good to see you removed url2.  As much as I like this clean code. It's hard to read. and the description doesn't really help much.  Could you give an example of what function it creates and how it's being called?

metadata-db-app/src/cmr/metadata_db/services/generic_documents.clj
```diff
@@ -27,27 +22,21 @@
...
+(def fields-to-drop (vec (concat info-fields [:id :deleted :transaction-id])))
 
 (defn- raw-generic->response
   "Take the raw database object returned from either SQL or In-Memory and shape
    it into the expected response by pulling out generated fields like concept-id
    and putting them in the :Info field, then putting the metadata inside the
    :Metadata field without any of these extra fields."
   [raw-doc]
-  {:Info (assoc-these-from info-fields raw-doc)
+  {:Info (select-keys raw-doc info-fields)
```
zimzoom, line 36: I agree with your choice to use select-keys here for readability, but just want to make sure -- your previous function included cutil/remove-nil-keys, do you not need that anymore? I don't think select-keys covers that but I may be wrong.

indexer-app/src/cmr/indexer/data/index_set_elasticsearch.clj
```diff
@@ -34,7 +34,7 @@
         ;; will know which one caused the issue by seing which index was the last
         ;; to start
         (when (= "dev" (cmr.common-app.config/release-version))
-          (println "Here in create-index function creating" index-name)
+          (println "This instance of CMR will publish to Elastic Index:" index-name)
```
zimzoom, line 37: Just a thought, do we want to make this an (info log?

zimzoom (approved):

transmit-lib/src/cmr/transmit/generic_documents.clj
jceaser: - concept-ingest-url is responsible for returning a URL for a service and it will take an array of parameters for that URL
- http-helper/def* macros all create a function which makes an HTTP call to an internal CMR service. There are some assumptions like getter will use GET and destroyer will use DELETE. The built function will pass in common headers a service like metadata-db would expect. Many of the functions above this block of code might be able to be converted over to using these macros. I found the macro being used in only a handful of places and decided that since I was creating new interfaces, I would try to use just these macros. An example usage is in `cmr.ingest.api.generic-documents/read-generic-document`.

indexer-app/src/cmr/indexer/data/index_set_elasticsearch.clj
jceaser: I'm going to forget, I'll do this now. Erich and I had the exact same conversation.

metadata-db-app/src/cmr/metadata_db/services/generic_documents.clj
jceaser: because it is not inside a metadata block, it was all mixed in with all the other fields, I don't know which ones are metadata at this point, but I do know which ones are not. Then, as you can see, I put them all under a map so that nothing outside of metadata-db will know what was done.

jceaser: I would have to go back and clean up the metadata code, but we mixed everything together to simplify our usage of the in memory database.

PR #1636, opened 2022-08-08
(Feature Branch) Cmr 8389 Adding in a more consistent API where we get rid of the term generics and put provider in query param.

eereiter:


ingest-app/src/cmr/ingest/api/routes.clj
```diff
@@ -250,13 +251,33 @@
...
   (context "/generics/provider/:provider-id/:native-id" [provider-id native-id]
     (POST "/" request (gen-doc/create-generic-document request))
     (GET "/" request (gen-doc/read-generic-document request))
     (PUT "/" request (gen-doc/update-generic-document request))
     (DELETE "/" request (gen-doc/delete-generic-document request))))
 
+(def generate-generic-concept-types-reg-ex
+  "This function creates a regular expresion for all of the generic concepts.  This is used to create the api endpoints.
+   An example string that is return looks like: \"dataqualitysummary|orderoption|serviceoption\" "
+  (let [x (-> (str (concepts/get-generic-concept-types-array))
```
jaybarra, line 265: Can we change the name from `x` to `rx`?

ingest-app/src/cmr/ingest/api/routes.clj
```diff
@@ -250,13 +251,33 @@
...
+              (clojure.string/replace #":|\]|\[" "")
+              (clojure.string/replace #" " "|"))]
+    x))
+
+(def generic-document-routes-2
```
jaybarra, line 270: This should be named `generic-document-aux` or something more descriptive than `2`?

jaybarra (changes requested):

ingest-app/src/cmr/ingest/api/routes.clj
eereiter: yes, I have made the change.

ingest-app/src/cmr/ingest/api/routes.clj
eereiter: Yes, I have renamed it generic-document-routes-auxiliary.  
FYI - There is another route function generic-document-routes, that sets the ingest URL to 
ingest/generics/provider/:provider-id/:native-id 

In this function I created routes such as 
ingest/dataqualitysummary/:native-id and provider is now passed in as a parameter.  provider being pass in as a parameter was requested by Ryan.

jaybarra (approved):

ingest-app/src/cmr/ingest/api/routes.clj
```diff
@@ -268,7 +289,9 @@
       ;; Add routes to create, update, delete, & validate concepts
       ingest-routes
 
+      ;; add routes to create, update, read, and delete generic concepts
       generic-document-routes
+      generic-document-routes-auxiliary
```
jceaser, line 294: so your suggesting we support both for a time?

ingest-app/src/cmr/ingest/api/generic_documents.clj
```diff
@@ -142,8 +142,35 @@
...
     ;; passed in to be resolved by a function.
-    (tgen/update-generic request-context [provider-id native-id] (json/generate-string concept))
-    {:status 204}))
+    (tgen/update-generic request-context [provider-id native-id] (json/generate-string concept))))
 
 (defn delete-generic-document [request]
   (println "stub function: delete " request))
+
+;; TODO: Generic work: This could be a candidate for a configuration file.
+(defn validate-and-get-required-parameters
```
jceaser, line 151: this only ever returns one parameter, the provider object (which I think is suppose to include the other provider fields but which I never needed in any code so I faked it by creating a map with only provider-id). I think you should return the provider object in a map under the key :provider (or something like that name) so that you can add new ones. Also, native-id is required right?

ingest-app/src/cmr/ingest/api/routes.clj
eereiter: Just until we and Ryan agree on what the URL will be.

ingest-app/src/cmr/ingest/api/generic_documents.clj
eereiter: Right now all this function does is grab a query parameter named :provider and checks to see if the provider exists.
/search/orderoption?provider=PROV1.  
An exception is thrown if the provider does not exist otherwise it returns a map of key (provider and the query parameter value.

It doesn't do anything with the parameter other than to check that it exists on the query parameter.  It doesn't get the provider object and it doesn't do anything with it.
Then it just returns a map of the key and the string name of the provider that it got from the query parameters. 

If you don't have native id in the url, then you won't get this far in the code.

jceaser (approved):

PR #1640, opened 2022-08-11
CMR-8555 A common directory to store schemas and distribute them to micro services that need them

eriknelson11:
…ent to developers


OtherSchemas/OrderOption/v1.0.0/schema.json
```diff
@@ -0,0 +1,73 @@
+{
```
jceaser, line 1: can you provide (or track down) an example metadata.json file to include?

OtherSchemas/ServiceEntry/v1.0.0/schema.json
```diff
@@ -0,0 +1,72 @@
+{
```
jceaser, line 1: Can you add a metadata.json example file

OtherSchemas/ServiceOption/v1.0.0/schema.json
```diff
@@ -0,0 +1,60 @@
+{
```
jceaser, line 1: can you add a sample metadata.json file?

OtherSchemas/.editorconfig
```diff
@@ -0,0 +1,18 @@
+[*]
```
jceaser, line 1: is this file needed? The one in the parent directory should work

dev-system/dev/user.clj
```diff
@@ -117,6 +117,20 @@
...
 
+(defn- distribute-schemas
+  "A function that will copy the directory OtherSchemas in to all the necessary
+   apps and will be called on reset"
+  []
+  (let [path (System/getProperty "user.dir")
+        apps ["search-app"
+              "ingest-app"
+              "indexer-app"
+              "metadata-db-app"
```
jceaser, line 128: metadata db is needed?

eriknelson11: It was part of the acceptance criteria in the ticket.

jceaser: also these all go inside a directory like indexer-app/resources/schemas/

eriknelson11:
@jceaser Running the repl from the terminal and through an IDE both get the same path being generated from `(System/getProperty "user.dir")`

eudoroolivares2016 (approved):

jceaser:
there may be another issue, we need these schema paths to be lower case, but I guess we put them in the repository with Camel case.  :(

dev-system/dev/user.clj
```diff
@@ -117,6 +117,20 @@
...
+  []
+  (let [path (System/getProperty "user.dir")
+        apps ["search-app/resources"
+              "ingest-app/resources"
+              "indexer-app/resources"
+              "metadata-db-app/resources"
+              "system-int-test/resources"]
+        source "schemas"]
+    (doseq [app apps]
+      (shell/sh "cp" "-r" (clojure.string/replace path #"dev-system" source) (clojure.string/replace path #"dev-system" app)))))
```
jceaser, line 132: the destination path must always be lower case. CMR will use the lowercase value for the generic name (grid vs Grid). In this way we don't have to worry about how the schema was input in the repository and CMR does not have to do a harder search on the file system.

dev-system/dev/user.clj
```diff
@@ -118,18 +118,25 @@
...
               "metadata-db-app/resources"
               "system-int-test/resources"]
-        source "schemas"]
+        source (clojure.string/replace path #"dev-system" "schemas")
+        source-names (-> (:out (shell/sh "ls" source))
+                         (clojure.string/split #"\n"))]
     (doseq [app apps]
-      (shell/sh "cp" "-r" (clojure.string/replace path #"dev-system" source) (clojure.string/replace path #"dev-system" app)))))
+      (doseq [src-name source-names]
+        (if-not (clojure.string/includes? src-name ".md")
```
jceaser, line 135: switch from includes? to ends-with? so that you don't match .md_yada-yada-yada

jceaser (approved):

PR #1643, opened 2022-08-12
CMR-8383: Add token_type to search logging

brianalexander:
Add token_type logging for search queries


isja17 (approved):

johnwteague:
What happens when there is no token present in the request?

brianalexander:
> What happens when there is no token present in the request?

An example log without a token is:

Found 0 collections in 82 ms from client unknown, token_type null in format :xml with params {:result-format :xml}.

TylerHeald1 (approved):

PR #1653, opened 2022-08-18
CMR-8551 "RelatedUrls > MimeType" validation returns a Warning and no…

eriknelson11:
…t an Error

### DIFF
--- a/ingest-app/src/cmr/ingest/validation/validation.clj
+++ b/ingest-app/src/cmr/ingest/validation/validation.clj
@@ -469,7 +469,6 @@
   (let [kms-index (kms-fetcher/get-kms-index context)]
     {:MeasurementIdentifiers (measurement-validation
                               kms-index msg/measurements-not-matches-kms-keywords)
-
      :RelatedURLs (v/every [{:Format (match-kms-keywords-validation-single
                                       kms-index
                                       :granule-data-format
@@ -484,22 +483,37 @@
 
 (defn umm-spec-validate-variable
   "Validate variable through umm-spec validation functions. If warn? flag is
-  true and umm-spec-validation is off, log warnings and return messages, otherwise throw errors."
+   true and umm-spec-validation is off, log warnings and return messages, otherwise throw errors."
   [variable context warn?]
   (when-let [non-ignorable (seq (umm-spec-validation/validate-variable-with-no-defaults
                                  variable
                                  [(variable-keyword-validations-unignorable context)]))]
     (errors/throw-service-errors :invalid-data non-ignorable))
-
-  (when-let [err-messages (seq (umm-spec-validation/validate-variable
-                                variable
-                                [(variable-keyword-validations context)]))]
+  (let [err-messages (seq (umm-spec-validation/validate-variable
+                           variable
+                           [(variable-keyword-validations context)]))
+        kms-index (kms-fetcher/get-kms-index context)
+        warning-messages (seq (umm-spec-validation/validate-variable
+                               variable
+                               [(related-url-validator-warning kms-index)]))]
     (if (or (config/return-umm-spec-validation-errors)
             (not warn?))
-      (errors/throw-service-errors :invalid-data err-messages)
-      (do
-        (warn "UMM-Var UMM Spec Validation Errors: " (pr-str (vec err-messages)))
-        err-messages))))
+       ;;when we are not supposed to return error as warnings
+      (if err-messages
+         ;; throw errors when it exists
+        (errors/throw-service-errors :invalid-data err-messages)
+         ;; throw warnings when error doesn't exist.
+        (when warning-messages
+          (do
+            (warn "UMM-Var UMM Spec Validation Errors: " (pr-str (vec warning-messages)))
+            warning-messages)))
+       ;; when we are supposed to return errors as warnings as well,
+       ;; return both errors and warnings as warnings.
+      (when-let [all-warning-messages (not-empty (remove nil? (merge err-messages warning-messages)))]
+        (do
+          (println "3 " all-warning-messages)
+          (warn "UMM-Var UMM Spec Validation Errors: " (pr-str (vec all-warning-messages)))
+          all-warning-messages)))))
 
 (defn validate-variable-associated-collection
   "Validate the collection being associated to is accessible."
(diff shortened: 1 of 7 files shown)


ingest-app/src/cmr/ingest/validation/validation.clj
```diff
@@ -98,35 +98,35 @@
...
-  {:RelatedUrls 
-   [(match-related-url-kms-keywords-validations kms-index)
-    (v/every {:GetData {:MimeType (match-kms-keywords-validation-single
-                                   kms-index
-                                   :mime-type
-                                   msg/mime-type-not-matches-kms-keywords)}})]})
+   if warn? is set to false, otherwise will check MimeType in GetData for Related URL
+   field. All of which can be inside a ContactInformation or be a standalone field.
+   ContactInformation can themselves be found in DataCenters, ContactGroups, and ContactPersons."
+  [kms-index warn?]
```
jceaser, line 104: I don't know about this, I would think that this function would always return a validator and the issue is if it should be called or not, but I would defer to @sxu123 as she has done the last refactoring of this file. I am leaning in the direction that this is wrong.

ingest-app/src/cmr/ingest/validation/validation.clj
```diff
@@ -156,13 +156,10 @@
...
   [context]
   (let [kms-index (kms-fetcher/get-kms-index context)]
-    (merge (related-url-validator kms-index)
-           (datacenter-url-validators kms-index)
-           (contactpersons-url-validators kms-index)
-           (contactgroups-url-validators kms-index)
-           (useconstraints-onlineresource-validators kms-index)
-           (collectioncitations-onlineresource-validators kms-index)
-           (publicationreferences-onlineresource-validators kms-index))))
+    (merge (related-url-validator kms-index false)
```
jceaser, line 159: if the original function accepts two parameter signatures, one could assume false and then you would only have to change the code that is different.

ingest-app/src/cmr/ingest/validation/validation.clj
```diff
@@ -203,19 +200,26 @@
...
-                        kms-index :spatial-keywords msg/location-keyword-not-matches-kms-keywords)
-     :DataCenters (match-kms-keywords-validation
-                   kms-index :providers msg/data-center-not-matches-kms-keywords)}))
+    (merge (useconstraints-onlineresource-validators kms-index)
+           (collectioncitations-onlineresource-validators kms-index)
+           (publicationreferences-onlineresource-validators kms-index)
+           (datacenter-url-validators kms-index true)
+           (contactpersons-url-validators kms-index true)
+           (contactgroups-url-validators kms-index true)
+           (related-url-validator kms-index true)
```
jceaser, line 209: should we not instead use the conditional to conditionally add these functions here at this level? The function name is "optional" it is already deciding what should be shown for a condition.

ingest-app/src/cmr/ingest/validation/validation.clj
```diff
@@ -98,35 +98,35 @@
...
-   [(match-related-url-kms-keywords-validations kms-index)
-    (v/every {:GetData {:MimeType (match-kms-keywords-validation-single
-                                   kms-index
-                                   :mime-type
-                                   msg/mime-type-not-matches-kms-keywords)}})]})
+   if warn? is set to false, otherwise will check MimeType in GetData for Related URL
+   field. All of which can be inside a ContactInformation or be a standalone field.
+   ContactInformation can themselves be found in DataCenters, ContactGroups, and ContactPersons."
+  [kms-index warn?]
+  (if (not warn?)
```
jceaser, line 105: double negative? so I send false to get a true? Why not rename to something like report-as-error and when it is true you do the thing? There is also an if-not function, but I think you can make it read better.

ingest-app/src/cmr/ingest/validation/validation.clj
eriknelson11: to my understanding the first part of `related-url-validator` should be returned as an error but the second part of the function should be returned as a warning, so would a better way be to create a separate function for the warning portion?

ingest-app/src/cmr/ingest/validation/validation.clj
eriknelson11: `(not warn?)` was used in other parts of the file, I can see how in my case its not as readable though I can change that

jceaser: you can get a second opinion, I will defer if I'm the only one

ingest-app/src/cmr/ingest/validation/validation.clj
jceaser: I think so, but @sxu123 had the most opinions here last time, reach out to her and ask for advice. She reformatted my code and I think it is more readable now.

sxu123 (changes requested):
Sent messages in slack to the developer.

ingest-app/src/cmr/ingest/validation/validation.clj
sxu123: The logic is right. Although I don't like using the flag because it's a little confusing. I messaged Erik about it.

ingest-app/src/cmr/ingest/validation/validation.clj
sxu123: Using another function instead of using flag would be more readable.

ingest-app/src/cmr/ingest/validation/validation.clj
sxu123: Why would you remove           (useconstraints-onlineresource-validators kms-index)
           (collectioncitations-onlineresource-validators kms-index)
           (publicationreferences-onlineresource-validators kms-index))))?
These are not RelatedUrl/MimeType validations. The mime types belong to other fields.

ingest-app/src/cmr/ingest/validation/validation.clj
sxu123: 1. (useconstraints-onlineresource-validators kms-index)
           (collectioncitations-onlineresource-validators kms-index)
           (publicationreferences-onlineresource-validators kms-index) might not belong here since these MimeTypes don't belong to RelatedUrls.

2. DataCenter are validated twice, on different things. After merging, one would be lost. So They need to be combined in the end.

sxu123: A test needs to be added to introduce two errors. One is the DataCenter RelatedUrl/MimeType error, the other is the error from :DataCenters (match-kms-keywords-validation
                           kms-index :providers msg/data-center-not-matches-kms-keywords)})))
Make sure both are returned as warnings, none is lost.

sxu123 (approved):

eereiter (approved):

PR #1656, opened 2022-08-19
CMR-8244-Fix Fixing broken integration tests caused by 8244

dpesall:
Fixed tests: https://ci.earthdata.nasa.gov/chain/result/viewChainResultTestSummary.action?planKey=CMR-DREW30&buildNumber=5

### DIFF
--- a/search-app/src/cmr/search/data/complex_to_simple_converters/spatial.clj
+++ b/search-app/src/cmr/search/data/complex_to_simple_converters/spatial.clj
@@ -214,9 +214,10 @@
           operator (extract-operator (:query-params context))
           spatial-script (shape->script-cond shape operator)
         ; check the MBR first before doing the more expensive check
-          spatial-cond (if (= operator :any)
-                         (gc/and-conds [mbr-cond (gc/or-conds [lr-cond spatial-script])])
-                         spatial-script)]
+          spatial-cond (if (or (= operator :every) (= operator :ignore_br))
+                         spatial-script
+                         (gc/and-conds [mbr-cond (gc/or-conds [lr-cond spatial-script])]))]
+
       (if orbital-cond
         (gc/or-conds [spatial-cond orbital-cond])
         spatial-cond))))
--- a/system-int-test/test/cmr/system_int_test/search/all_granule_search_test.clj
+++ b/system-int-test/test/cmr/system_int_test/search/all_granule_search_test.clj
@@ -77,7 +77,7 @@
         result14 (search/find-refs :granule params8)
         result15 (search/find-refs :granule params9)
         result16 (search/find-refs :granule params10)
-        err-msg (seq '("The CMR does not allow querying across granules in all collections. To help optimize your search, you should limit your query using conditions that identify one or more collections, such as provider, provider_id, concept_id, collection_concept_id, short_name, version or entry_title. For any questions please contact cmr-support@earthdata.nasa.gov."))
+        err-msg "The CMR does not allow querying across granules in all collections. To help optimize your search, you should limit your query using conditions that identify one or more collections, such as provider, provider_id, concept_id, collection_concept_id, short_name, version or entry_title. For any questions please contact cmr-support@earthdata.nasa.gov."
         err-msg-illegal-service-id "Invalid concept_id [S1234-PROV1]. For granule queries concept_id must be either a granule or collection concept ID."
         err-msg-illegal-variable-id "Invalid concept_id [V1234-PROV1]. For granule queries concept_id must be either a granule or collection concept ID."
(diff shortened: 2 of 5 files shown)


search-app/src/cmr/search/data/complex_to_simple_converters/spatial.clj
```diff
@@ -214,9 +214,10 @@
           operator (extract-operator (:query-params context))
           spatial-script (shape->script-cond shape operator)
         ; check the MBR first before doing the more expensive check
-          spatial-cond (if (= operator :any)
-                         (gc/and-conds [mbr-cond (gc/or-conds [lr-cond spatial-script])])
-                         spatial-script)]
+          spatial-cond (if (or (= operator :every) (= operator :ignore_br))
```
ygliuvt, line 217: This code looks wrong. The operator should not have anything to do with we match the MBR first before doing the more expensive exactly spatial shape matching.

system-int-test/test/cmr/system_int_test/search/all_granule_search_test.clj
```diff
@@ -77,26 +77,28 @@
         result14 (search/find-refs :granule params8)
         result15 (search/find-refs :granule params9)
         result16 (search/find-refs :granule params10)
-        err-msg "The CMR does not allow querying across granules in all collections. To help optimize your search, you should limit your query using conditions that identify one or more collections, such as provider, provider_id, concept_id, collection_concept_id, short_name, version or entry_title. For any questions please contact cmr-support@earthdata.nasa.gov."
+        err-msg (seq '("The CMR does not allow querying across granules in all collections. To help optimize your search, you should limit your query using conditions that identify one or more collections, such as provider, provider_id, concept_id, collection_concept_id, short_name, version or entry_title. For any questions please contact cmr-support@earthdata.nasa.gov."))
```
ygliuvt, line 80: You shouldn't have to make this kind of changes in this namespace.

search-app/src/cmr/search/data/complex_to_simple_converters/spatial.clj
dpesall: The reason this is needed is because if they have the `:every` or `:ignore_br` options, because for the granules which have the global bounding rectangles in their geometry data, the match MBR first operation will match with the global bounding rectangle doing this initial operation and return the granule even though the `:ignore_br` option is used.

system-int-test/test/cmr/system_int_test/search/all_granule_search_test.clj
dpesall: I'm at a loss for words as to why this test was failing to begin with, because none of my changes should've affected how error message was returned, but now running it without my change it's no longer failing locally like it was before. Undid the change to this file.

search-app/src/cmr/search/data/complex_to_simple_converters/spatial.clj
ygliuvt: It shouldn't. The MBR condition is to filter out the granules that does not have a chance to match with the real spatial condition and cut down the number of times that we have to run the spatial script.

ygliuvt: This MBR is for the spatial parameter, not the granule's spatial coverage.

system-int-test/test/cmr/system_int_test/search/all_granule_search_test.clj
```diff
@@ -80,14 +80,16 @@
         err-msg "The CMR does not allow querying across granules in all collections. To help optimize your search, you should limit your query using conditions that identify one or more collections, such as provider, provider_id, concept_id, collection_concept_id, short_name, version or entry_title. For any questions please contact cmr-support@earthdata.nasa.gov."
         err-msg-illegal-service-id "Invalid concept_id [S1234-PROV1]. For granule queries concept_id must be either a granule or collection concept ID."
         err-msg-illegal-variable-id "Invalid concept_id [V1234-PROV1]. For granule queries concept_id must be either a granule or collection concept ID."
-        err-msg-illegal-service-id-in-array "Invalid concept_id [[\"\" \"S1234-PROV1\"]]. For granule queries concept_id must be either a granule or collection concept ID."
+        err-msg-illegal-service-id-in-array (seq '("Invalid concept_id [[\"\" \"S1234-PROV1\"]]. For granule queries concept_id must be either a granule or collection concept ID."))
```
ygliuvt, line 83:
```suggestion
        err-msg-illegal-service-id-in-array ["Invalid concept_id [[\"\" \"S1234-PROV1\"]]. For granule queries concept_id must be either a granule or collection concept ID."]
```

PR #1665, opened 2022-08-23
CMR-8578: Removing 'bearer' prefix string; removing extra debug conso…

eudoroolivares2016:
When indexing collections using the bootstrap function, the system token was not being parsed due to the bearer prefix. This removes the prefix; further actions will need to be taken under this PR to allow for all collections to be indexed as opposed to only public collections


eereiter (approved):

jaybarra (approved):

ygliuvt (changes requested):
`CMR_TOKEN_KEY` environment variable should be set up in the serverless.yml file

ygliuvt (approved):

eudoroolivares2016:
Currently holding off merging this; linked NGAP ticket has to be completed so that graphql can retrieve data from Neptune, so that we can remove the public endpoint, so that we can index private records

jceaser (approved):

PR #1689, opened 2022-08-30
CMR-8578: Adds check to skip over broken collections; improve console…

eudoroolivares2016:
… output message;

One of the issues when indexing collections into graphDb was that if there were collections that returned a 404 not found error (such as if it is indexed into elastic but, not found in the oracle database) indexing would halt.  These changes skip over the broken collections and add some useful console outputs. To run locally use: npx sls invoke local --function bootstrapGremlinServer  --path local-bootstrap.json | grep FAILED note you will only see any failed collections in the output but, graphDb will be indexing non-broken collections

### DIFF
--- a/graph-db/serverless/src/indexCmrCollection/__test__/handler.test.js
+++ b/graph-db/serverless/src/indexCmrCollection/__test__/handler.test.js
@@ -132,9 +132,10 @@ describe('indexCmrCollection handler', () => {
 
     const indexed = await indexCmrCollection(event)
 
-    expect(consoleMock).toBeCalledTimes(2)
+    expect(consoleMock).toBeCalledTimes(3)
     expect(consoleMock).toBeCalledWith('Could not complete request due to error: Error: Error Collection not found')
     expect(consoleMock).toBeCalledWith('Collection FAILED during indexing process there may be an issue with the collection verify that the collection for the given env: ', 'C123755555-TESTPROV')
+    expect(consoleMock).toBeCalledWith('Error indexing collection, Execption was thrown: ', new Error('Cannot read properties of null (reading \'data\')'))
 
     const { body } = indexed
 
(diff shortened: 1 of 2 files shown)


graph-db/serverless/src/indexCmrCollection/handler.js
```diff
@@ -48,26 +48,31 @@ const indexCmrCollections = async (event) => {
...
+          console.log(`Start indexing concept [${conceptId}], revision-id [${revisionId}]`)
 
-        await indexCmrCollection(firstCollection, groupList, gremlinConnection)
+          await indexCmrCollection(firstCollection, groupList, gremlinConnection)
 
-        recordCount += 1
+          recordCount += 1
+        }
+      } catch (e) {
+        console.log('Collection FAILED during indexing process there may be an issue with the collection verify that the collection for the given env: ', conceptId)
```
ygliuvt, line 74: We should log the exception in the log message. It is wrong to assume that all exceptions that are thrown would be due to collection not found.

eudoroolivares2016: Added logging of the actual exception.

eereiter (approved):

graph-db/serverless/src/indexCmrCollection/__test__/handler.test.js
```diff
@@ -120,6 +120,28 @@ describe('indexCmrCollection handler', () => {
     expect(statusCode).toBe(200)
   })
 
+  test('test indexing of a broken collection', async () => {
+    // Pass an error to the response of fetchCmrCollection
+    nock(/local-cmr/)
```
jceaser, line 125: do you mean to involve regular expressions? Is this a requirement of the function?

jceaser (approved):
approved, my comments are non-blocking questions

PR #1712, opened 2022-09-15
CMR-8617 - moving the fixed list of generics to a dynamic check in the case default

jceaser:
Found a way to remove a fixed list of concept types by moving it to the default section of the case and doing a lookup there.


jceaser:
made a mistake creating this.

PR #1755, opened 2022-11-01
Feature CMR-8774 - move provider parameter back to the URL path for generic ingest CRUD events

jceaser:
move "provider=PROV1" back to the URL for CRUD events. 

NOTE: documentation of these URL changes is dependent on CMR-8649 and will be completed in a following ticket.

### DIFF
--- a/common-lib/src/cmr/common/concepts.clj
+++ b/common-lib/src/cmr/common/concepts.clj
@@ -24,9 +24,9 @@
    Expected usage is to pass in pluralize-concept-type."
   ([] (get-generic-concept-types-array identity))
   ([modifier-func]
-   (reduce (fn [coll, item] (conj coll (modifier-func item)))
+   (reduce (fn [coll, item] (conj coll (modifier-func (key item))))
            []
-           (keys generic-concept-types->concept-prefix))))
+           (common-generic/approved-generic-concept-prefixes))))
 
 (defn generic-concept?
   "Return true if the passed in concept is a generic concept"
--- a/common-lib/src/cmr/common/concepts.clj
+++ b/common-lib/src/cmr/common/concepts.clj
@@ -18,9 +18,15 @@
   (cset/map-invert generic-concept-types->concept-prefix))
 
 (defn get-generic-concept-types-array
-  "Gets the array of generic concept types."
-  []
-  (vec (keys generic-concept-types->concept-prefix)))
+  "Gets the array of generic concept types and optionally modify those values.
+   By default, no parameters will return the list as is, however if a function
+   is passed in that takes one value, this value will be changed in some way.
+   Expected usage is to pass in pluralize-concept-type."
+  ([] (get-generic-concept-types-array identity))
+  ([modifier-func]
+   (reduce (fn [coll, item] (conj coll (modifier-func item)))
+           []
+           (keys generic-concept-types->concept-prefix))))
 
 (defn generic-concept?
   "Return true if the passed in concept is a generic concept"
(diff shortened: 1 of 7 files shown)


eudoroolivares2016 (approved):

common-lib/src/cmr/common/concepts.clj
```diff
@@ -18,9 +18,15 @@
...
 
 (defn get-generic-concept-types-array
-  "Gets the array of generic concept types."
-  []
-  (vec (keys generic-concept-types->concept-prefix)))
+  "Gets the array of generic concept types and optionally modify those values.
+   By default, no parameters will return the list as is, however if a function
+   is passed in that takes one value, this value will be changed in some way.
+   Expected usage is to pass in pluralize-concept-type."
+  ([] (get-generic-concept-types-array identity))
```
eereiter, line 25: I don't understand this line.  What is identity set to?  Is it defined somewhere else or does nil get passed in?

ingest-app/src/cmr/ingest/api/generic_documents.clj
```diff
@@ -28,37 +28,18 @@
...
-        msg (second required-param-to-check)
-        value (get-in request [:params param-key])]
-    (when-not value
-      (errors/throw-service-error :invalid-data (msg)))))
-
-(defn validate-any-required-query-parameters
-  [request required-parameters]
-  (doseq [param required-parameters]
-    (validate-required-parameter param request)))
+(defn- concept-type->singular
```
eereiter, line 38: Shouldn't this function be moved to cmr/common/concepts?

ingest-app/src/cmr/ingest/api/generic_documents.clj
```diff
@@ -157,10 +138,10 @@
 (defn read-generic-document
   "Read a document from the Native ID and return that document"
   [request]
-  (let [{:keys [route-params request-context params]} request
-        provider-id (:provider params)
+  (let [{:keys [route-params request-context _params]} request
```
eereiter, line 141: If we don't need params here anymore, I think it should be deleted.

ingest-app/src/cmr/ingest/api/generic_documents.clj
```diff
@@ -170,7 +151,7 @@
   [request]
   (let [res (prepare-generic-document request)
         headers (:headers request)
-        {:keys [spec-key spec-version provider-id native-id request-context concept]} res
+        {:keys [spec-key spec-version _provider-id _native-id request-context concept]} res
```
eereiter, line 154: If we don't need provider-id or native-id anymore as keys, they should be deleted.

ingest-app/src/cmr/ingest/api/generic_documents.clj
```diff
@@ -180,18 +161,17 @@
    if successful with the concept id and revision number. A 404 status is returned if the concept has
    already been deleted."
   [request]
-  (let [{:keys [route-params request-context headers params]} request
+  (let [{:keys [route-params _request-context _headers params]} request
```
eereiter, line 164: If we don't need the keys anymore, I think they should be deleted.

common-lib/src/cmr/common/concepts.clj
eudoroolivares2016: **identity** is a single-arg function that simply returns its argument so is that shorthand for something like the default function https://clojuredocs.org/clojure.core/identity

eudoroolivares2016, line 25:
```suggestion
  ;; identity is a single-arg function that simply returns its argument
  ([] (get-generic-concept-types-array identity))
```

eereiter: It seems like extra cpu cycles in most cases and by looking at the docs, it doesn't look like identity is being used in a standard way.  While identity is a function, what is the function with no arguments? is it a no-op?

I still think it would be more readable if the function were 
([] (keys generic-concept-types->concept-prefix))
([modifier-func]
 (reduce (fn [coll, item] (conj coll (modifier-func item)))
           []
           (keys generic-concept-types->concept-prefix))))

jceaser: I can cut out the keys:

> (reduce (fn [coll item] (conj coll (modifier-func (first item))))
          []
          generic-concept-types->concept-prefix))

eereiter (approved):

PR #1758, opened 2022-11-03
CMR-8694 - changing input from URL Form Parameters to JSON

jceaser:
curl was not always being nice when trying to mix Variable metadata with new lines and "payload" data as different named URL parameters "content=<json-of-variable>&data=<payload-json>" so this change instead accepts values as one JSON block: '{"content": <content here>, "data": <data here>"}'

### DIFF
--- a/ingest-app/src/cmr/ingest/api/core.clj
+++ b/ingest-app/src/cmr/ingest/api/core.clj
@@ -231,10 +231,10 @@
    has the side effect of emptying the request body. Don't try to read the body
    again after calling this function.
 
-   For some calls, specificly variables, two documents are sent in. These
+   For some calls, specifically variables, two documents are sent in. These
    documents are Metadata Content and Extended Data. Metadata Content is a
    metadata JSON record, like UMM-V, where Extended Data is any valid JSON which
-   is inteended to be used as data for a Relationship.
+   is intended to be used as data for a Relationship.
 
    Note: this function assumes JSON and not XML as the input."
   [body]
--- a/ingest-app/src/cmr/ingest/api/core.clj
+++ b/ingest-app/src/cmr/ingest/api/core.clj
@@ -226,35 +226,31 @@
   [body]
   (string/trim (slurp body)))
 
-
-(defn tee [any] (println "tee:" any) any)
-
 (defn read-multiple-body!
   "Returns the body content string by slurping the request body. This function
    has the side effect of emptying the request body. Don't try to read the body
    again after calling this function.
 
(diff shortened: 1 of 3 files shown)


ingest-app/src/cmr/ingest/api/core.clj
```diff
@@ -271,7 +273,7 @@
   Don't try to read the body again after calling this function."
   ([concept-type body content-type headers]
    (let [metadata (read-multiple-body! body)]
-     (metadata->concept concept-type metadata content-type headers)))
+     (tee (metadata->concept concept-type metadata content-type headers))))
```
eereiter, line 276: Is tee to be deleted.  It looks like debugging code since the function above uses print.

ingest-app/src/cmr/ingest/api/core.clj
```diff
@@ -225,6 +226,9 @@
   [body]
   (string/trim (slurp body)))
 
+
+(defn tee [any] (println "tee:" any) any)
```
eereiter, line 230: I think this is for debugging and should be deleted.

dpesall (approved):

eereiter (approved):
once tee is removed.

ingest-app/src/cmr/ingest/api/core.clj
```diff
@@ -230,29 +231,26 @@
    has the side effect of emptying the request body. Don't try to read the body
    again after calling this function.
 
-   Note, for testing/reuse use the three parameter function which takes a string
-   and makes fewer assupmtions about the content of the body.
+   For some calls, specificly variables, two documents are sent in. These
+   documents are Metadata Content and Extended Data. Metadata Content is a
+   metadata JSON record, like UMM-V, where Extended Data is any valid JSON which
+   is inteended to be used as data for a Relationship.
```
eudoroolivares2016, line 237:
```suggestion
   is intended to be used as data for a Relationship.
```

ingest-app/src/cmr/ingest/api/core.clj
```diff
@@ -230,29 +231,26 @@
    has the side effect of emptying the request body. Don't try to read the body
    again after calling this function.
 
-   Note, for testing/reuse use the three parameter function which takes a string
-   and makes fewer assupmtions about the content of the body.
+   For some calls, specificly variables, two documents are sent in. These
```
zimzoom, line 234:
```suggestion
   For some calls, specifically variables, two documents are sent in. These
```

zimzoom (approved):

eudoroolivares2016 (approved):

PR #1759, opened 2022-11-03
CMR-8225: Updated service-bridge to build and deploy via uberjar.

ygliuvt:
This changed service-bridge deployment to use uberjar rather than the tedious and messy clojar update approach that is currently applied. This will greatly reduce the deployment complexity of service bridge and speed up service bridge related dev tasks.


jaybarra (approved):

TylerHeald1 (approved):
What is the purpose of moving cmr-exchange to other/cmr-exchange?

PR #1763, opened 2022-11-04
CMR-8717-2: Add search-after to OUS API with new uberjar dir structure

TylerHeald1:
Just the changes from the CMR-8717 branch (with some comments addressed) applied to the new directory structure used for running service-bridge with uberjar

### DIFF
--- a/other/cmr-exchange/ous-plugin/src/cmr/ous/impl/v2_1.clj
+++ b/other/cmr-exchange/ous-plugin/src/cmr/ous/impl/v2_1.clj
@@ -9,18 +9,18 @@
   have the results previously expected when making calls against metadata that
   has not been updated to use UMM-Var 1.2."
   (:require
-    [cmr.exchange.common.results.core :as results]
-    [cmr.exchange.common.results.errors :as errors]
-    [cmr.exchange.common.results.warnings :as warnings]
-    [cmr.exchange.common.util :as util]
-    [cmr.metadata.proxy.concepts.collection :as collection]
-    [cmr.metadata.proxy.concepts.granule :as granule]
-    [cmr.metadata.proxy.concepts.service :as service]
-    [cmr.metadata.proxy.results.errors :as metadata-errors]
-    [cmr.ous.common :as common]
-    [cmr.ous.components.config :as config]
-    [cmr.ous.util.geog :as geog]
-    [taoensso.timbre :as log]))
+   [cmr.exchange.common.results.core :as results]
+   [cmr.exchange.common.results.errors :as errors]
+   [cmr.exchange.common.results.warnings :as warnings]
+   [cmr.exchange.common.util :as util]
+   [cmr.metadata.proxy.concepts.collection :as collection]
+   [cmr.metadata.proxy.concepts.granule :as granule]
+   [cmr.metadata.proxy.concepts.service :as service]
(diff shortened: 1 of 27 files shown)


eudoroolivares2016 (approved):

isja17 (approved):

dpesall (approved):

other/cmr-exchange/ous-plugin/src/cmr/ous/impl/v2_1.clj
```diff
@@ -30,21 +30,24 @@
...
-        coll (collection/extract-metadata coll-promise)
-        tag-data (get-in coll [:tags (keyword collection/opendap-regex-tag) :data])
+        gran-headers (granule/extract-header-data grans-promise)
+        coll-body (collection/extract-body-metadata coll-promise)
+        tag-data (get-in coll-body [:tags (keyword collection/opendap-regex-tag) :data])
         granule-links (map granule/extract-granule-links granules)
-        service-ids (collection/extract-service-ids coll)
-        vars (common/apply-bounding-conditions endpoint token coll params)
+        sa-header (get gran-headers :cmr-search-after)
+        hits-header (get gran-headers :cmr-hits)
```
ygliuvt, line 38: Thanks for fixing `cmr-hits` header as well here.

ygliuvt (changes requested):
Need to update OUS API doc to document the `CMR-Search-After` feature.

The `CMR-Search-After` tests can be added together with `cmr-hits` header as part of CMR-8716.

other/cmr-exchange/service-bridge/resources/docs/rest-api/index.md
```diff
@@ -179,12 +179,19 @@ and, the time taken to process the request.
...
+  "sa-header": "[...]",
   "items": [...],
   "request-id": ...,
   "warnings:" null
 }
 ```
 
+## Search-After
+
+CMR Service-Bridge allows the use of CMR-Search-After. When searching a collection
```
ygliuvt, line 191: Maybe change to something like:

"CMR Service-Bridge allows the use of CMR-Search-After header to take advantage of the CMR Search-After support for paging. Any search against CMR Service Bridge that has results not fully returned in the current request will return a search-after value in the CMR-Search-After header of the search response. User can then pass this returned value in the CMR-Search-After header of the following request to retrieve the next page of result based on the specified page_size. See CMR API document on Search-After for a detailed description."

In your description, it says `the response will include the Search-After header as a header
and in the response body.`. Will the search-after header be returned in the response body also? Why do we want to do that?

TylerHeald1: Unfortunately, with how some of the service-bridge/cmr-exchange stuff is set up, the data where the response headers are set passed through an area of code that adds values to the response body. I tried passing the data through without adding it to the response, but I don't think the code is currently set up in a good way for handling that. the headers CMR-Hits, CMR-Took, and CMR-Request-ID all go through the same thing (ending up in header and response body)

TylerHeald1: With Liu's help, it is now removed from the response body

ygliuvt (approved):

PR #1764, opened 2022-11-07
CMR-8789 Fixed associated-item in generic association response.

sxu123:


search-app/src/cmr/search/services/generic_association_service.clj
```diff
@@ -130,7 +130,12 @@
...
   [mdb-context association operation]
   ;; save each association if there is no errors on it, otherwise returns the errors.
-  (let [association (sort-association association)
+  (let [associated-item-concept-id (:concept-id association)
+        revision-id (:revision-id association)
+        associated-item-revision-id (if (string? revision-id)
+                                      (read-string revision-id)
+                                      revision-id)
+        association (sort-association association)
         {source-concept-id :source-concept-id
```
eereiter, line 139: I don't understand what 
        {source-concept-id :source-concept-id
         source-revision-id :source-revision-id
         concept-id :concept-id
         revision-id :revision-id
         errors :errors} association
is doing in the let and how it is being used in the code after it:  Can you please explain it to me?

eereiter (approved):

system-int-test/test/cmr/system_int_test/search/generic_association_test.clj
```diff
@@ -828,6 +828,13 @@
         (is (= (get-in (first (:body response2)) [:generic-association :revision-id])
                (+ 1 (get-in (first (:body response1)) [:generic-association :revision-id]))))
 
+        ;; The first and second associations contain different associated-item because associated-item
+        ;; should always be related to of the items in the list that's associated to a particular concept.
```
dpesall, line 832: I think `of` in `related to of the items` is an extra word.

dpesall (approved):

search-app/src/cmr/search/services/generic_association_service.clj
jceaser, line 139: my question is what is important about the sort order here, this logic is done before the sorting, should a comment point out any details about why the order is significant?

jceaser (approved):

system-int-test/test/cmr/system_int_test/search/generic_association_test.clj
sxu123: It's should be  ```related to one of the items```

search-app/src/cmr/search/services/generic_association_service.clj
sxu123: native-id is unique in an association concept, just like any other concepts. it's constructed as <source-id>/<source-revision-id>/<destination-id>/<destination-revision-id>. If we don't "sort", then depending on how the request is constructed, the same two concepts/revisions could end up creating two different association concepts, which is what we try to avoid. Either you associate A with B, or B with A, it should end up being the same association. After we sort the two concepts, they will be inserted into the metadata-db.

sxu123: The side-effect with "sorting" is, if in the request I want to associate a service entry with a collection and a variable, in the two associations, the source is not always service entry. and the associated-item is not always collection and variable, the source and associated-item are switched in the service entry and collection association. The client still wants to know if the association with the collection is successful, but if your response says associated-item being service entry, then it doesn't give client any information.  This is why I need to get the original associated-item before the sorting.

sxu123: sort-function does explain why sorting is needed.

PR #1767, opened 2022-11-07
CMR-8755: Change generic concept types code, including API endpoints, to use and expect 'spaces' (hyphens)

zimzoom:
This PR changes the generic concept types code to have 'spaces' in the concept types, including expecting it in the base source array found in common config, and trickling through to the API endpoints. Necessarily this includes:
- removing remaining hard-coded references to concept types
- adjusting tests
- modifying the contents of the schemas directory including sub-directory names.

Also included:
- moving the generic concept types regex function to a common namespace, since ingest, search, and bootstrap APIs were all repeating it.

----------
**"Why use hyphens/dashes instead of underscores?"**
A short google search revealed that the modern state of web development has moved to strongly recommend hyphens/dashes over underscores as a word delimiter. (As discussed, for example, in the O'Reilly REST API Design Rulebook) Reasons include: dashes are easier to type, dashes tend to render better with client-side link styling, dashes are always recognized as word delimiters in operating systems, editors, and programs where many times underscores are not, google search and most other search algorithms also recognize dashes as delimiters but underscores as not.

In our case the obvious additional consideration is that dashes are idiomatic clojure. Therefore, the code should almost certainly retain the use of dashes within the clojure codebase, but with the API endpoints regex function now obeying DRY, it would be trivial to convert the API endpoints to use underscores instead, if the team/reviewers so desire. One industry standard acceptable (but not always necessary) use of underscores in REST API URLs is specifically for query parameters (such as page_size) because these are often desired to be parsed as "one entity". This developer leaves this question to the team.

----------
**Edit - Regarding Index Names:** the generic concept type elasticsearch indices will be different, since index names are generated directly from the base source config list. Indexer code was already expecting clojure-style names with hyphens and converting the hyphens to underscores, so the new index name is for example "1_generic_data_quality_summary". This does require string replacement when sending search-app's query to elasticsearch (`search-app/src/cmr/search/data/elastic_search_index.clj`).


common-lib/src/cmr/common/generics.clj
```diff
@@ -118,3 +118,15 @@
                        (get (json/parse-string index-raw true) :SubConceptType "X")))))
           {}
           (latest-approved-documents)))
+
+(def generic-concept-types-reg-ex
```
eudoroolivares2016, line 122: These really simply things!

eudoroolivares2016 (approved):

Generics.md
```diff
@@ -9,23 +9,23 @@ If adding a new document, you will need to update the defconf variable by either
...
 	(defconfig approved-pipeline-documents
   		{:default {:grid ["0.0.1"]
-             :dataqualitysummary ["1.0.0"]
-             :orderoption ["1.0.0"]
-             :serviceentry ["1.0.0"]
-             :serviceoption ["1.0.0"]}
+             :data-quality-summary ["1.0.0"]
+             :order-option ["1.0.0"]
+             :service-entry ["1.0.0"]
+             :service-option ["1.0.0"]}
```
eereiter, line 15: Can you look in Kibana and tell me what the index names are?  Are they 1_generic_dataqualitysummary
 or 1_generic_data-quality-summary....
If it is the latter I want to talk with either Jay or Leo about deleting the existing indexes before adding these and if we do that then we can use the old toggle and not need the new one, although i think the new one would be good anyway.

schemas/resources/schemas/data-quality-summary/v1.0.0/index.json
```diff
@@ -5,7 +5,7 @@
     "Version": "0.0.1"
   },
   "Generic": {
-    "Name": "DataQualitySummary",
```
jceaser, line 8: so where is this value used, I don't remember? Can we just set this to "Data Quality Summary"? if it is used, don't we have a function in one of the util.clj files to convert names with spaces to - or is that for _?

schemas/resources/schemas/data-quality-summary/v1.0.0/metadata.json
```diff
@@ -3,9 +3,8 @@
   "Name":"CER-BDS_Terra",
   "Summary":"Summary",
   "MetadataSpecification": {
-    "Name": "DataQualitySummary",
+    "Name": "Data-Quality-Summary",
     "Version": "1.0.0",
-    "URL": "https://cdn.earthdata.nasa.gov/generics/dataqualitysummary/v1.0.0"
+    "URL": "https://cdn.earthdata.nasa.gov/generics/data-quality-summary/v1.0.0"
```
jceaser, line 8: I agree with dashes here, but I pushback slightly on the name.

schemas/resources/schemas/data-quality-summary/v1.0.0/schema.json
```diff
@@ -33,12 +33,12 @@
...
           "description": "This element represents the URL where the schema lives. The schema can be downloaded.",
           "type": "string",
-          "enum": ["https://cdn.earthdata.nasa.gov/generics/dataqualitysummary/v1.0.0"]
+          "enum": ["https://cdn.earthdata.nasa.gov/generics/data-quality-summary/v1.0.0"]
         },
         "Name": {
           "description": "This element represents the name of the schema.",
           "type": "string",
-          "enum": ["DataQualitySummary"]
+          "enum": ["Data-Quality-Summary"]
```
jceaser, line 41: no - and space is more human readable, which I think is what name is for

schemas/resources/schemas/data-quality-summary/v1.0.0/metadata.json
eudoroolivares2016: As in the name of generic itself, or that it should be something like "dataquality-summary" etc?

schemas/resources/schemas/data-quality-summary/v1.0.0/index.json
jceaser: found it: (camel-snake-kebab.core/->kebab-case "this is my text")

search-app/src/cmr/search/services/parameters/conversion.clj
```diff
@@ -197,47 +197,14 @@
...
-   :concept-id :string})
-
-(defmethod common-params/param-mappings :grid
-  [_]
-  {:name :string
-   :id :string
-   :provider :string
-   :native-id :string
-   :concept-id :string})
+(doseq [concept-type (cc/get-generic-concept-types-array)]
```
jceaser, line 200: so in the future I assume we will append a function here which will read the config file and add new params? Has the TODO been resolved?

zimzoom: That's a good question. I deleted the original TODO because it read "use doseq OR get from config file" and so it was technically resolve ;) But I think the remaining question of, "should we move this into a config file?" is a good one.

Generics.md
zimzoom: so to answer your question more directly, yes the index names will be different. (just with all underscores)

schemas/resources/schemas/data-quality-summary/v1.0.0/index.json
zimzoom: Done! The 2 files that were changed in order to implement a human-readable "Name" field in the schema (that uses real spaces) are:
`indexer-app/src/cmr/indexer/data/concepts/generic.clj` (line 78)
`ingest-app/src/cmr/ingest/api/generic_documents.clj` (line 60)

schemas/resources/schemas/data-quality-summary/v1.0.0/metadata.json
zimzoom: Done, see comment above

Generics.md
```diff
@@ -73,21 +73,21 @@ Commit and distribute to CMR as done in the addition steps.
...
-native ID of "orderoption-1".
+ingest and plural in search. The examples below use the "order-option"/"order-options" type, with a
+native ID of "order-option-1".
 
 #### ingest
-	curl -v -XPOST -H "$TOKEN" -H "Content-Type:application/vnd.nasa.cmr.umm+json" "https://cmr.earthdata.nasa.gov/ingest/orderoption/orderoption-1?provider=PROV1" -d @orderoption-1.json
-	curl -v -XPUT -H "$TOKEN" -H "Content-Type:application/vnd.nasa.cmr.umm+json" "https://cmr.earthdata.nasa.gov/ingest/orderoption/orderoption-1?provider=PROV1" -d @orderoption-1.json
-	curl -v -XGET -H "$TOKEN" -H "Content-Type:application/vnd.nasa.cmr.umm+json" "https://cmr.earthdata.nasa.gov/ingest/orderoption/orderoption-1?provider=PROV1"
-	curl -v -XDELETE -H "$TOKEN" "https://cmr.earthdata.nasa.gov/ingest/orderoption/dataqualitysummary-1?provider=PROV1"
+	curl -v -XPOST -H "$TOKEN" -H "Content-Type:application/vnd.nasa.cmr.umm+json" "https://cmr.earthdata.nasa.gov/ingest/order-option/order-option-1?provider=PROV1" -d @order-option-1.json
```
jceaser, line 80: reading this today, I think the native id should look more different then the type

common-lib/src/cmr/common/config.clj
```diff
@@ -222,15 +222,15 @@
    are passed through the parser and returned as if it was the default.
    This string should contain a clojure map that looks something like:
    {:grid [\"0.0.1\"],
-    :dataqualitysummary [\"1.0.0\"]
-    :orderoption [\"1.0.0\"]
-    :serviceentry [\"1.0.0\"]
-    :serviceoption [\"1.0.0\"]}"
+    :data-quality-summary [\"1.0.0\"]
```
jceaser, line 225: maybe we no longer need the example as the actual is the same

indexer-app/test/cmr/indexer/test/data/index_set_generics.clj
```diff
@@ -5,8 +5,8 @@
 
 (def index-definition-def
   {:MetadataSpecification
-   {:URL "https://cdn.earthdata.nasa.gov/generics/dataqualitysummary/v1.0.0"
-    :Name "DataQuailtySummary",
+   {:URL "https://cdn.earthdata.nasa.gov/generics/data-quality-summary/v1.0.0"
+    :Name "DataQualitySummary",
```
jceaser, line 9: Did Erich weigh in on adding spaces here?

schemas/resources/schemas/data-quality-summary/v1.0.0/index.json
jceaser: so resolve this then?

jceaser: With or without spaces?

schemas/resources/schemas/data-quality-summary/v1.0.0/search.md
```diff
@@ -134,28 +134,28 @@ Content-Length: 512
...
-                "concept-type": "dataqualitysummary"
+                "concept-type": "data-quality-summary"
             },
             "umm": {
                 "Id": "8EA5CA1F-E339-8065-26D7-53B64074D7CC",
                 "Name": "CER-BDS_Terra",
                 "Summary": "Summary",
                 "MetadataSpecification": {
-                    "Name": "dataqualitysummary1234",
+                    "Name": "data-quality-summary1234",
```
jceaser, line 144: is this suppose to be bad data?

Generics.md
```diff
@@ -73,21 +73,21 @@ Commit and distribute to CMR as done in the addition steps.
...
+	curl -v -XPUT -H "$TOKEN" -H "Content-Type:application/vnd.nasa.cmr.umm+json" "https://cmr.earthdata.nasa.gov/ingest/order-option/order-option-1?provider=PROV1" -d @order-option-1.json
+	curl -v -XGET -H "$TOKEN" -H "Content-Type:application/vnd.nasa.cmr.umm+json" "https://cmr.earthdata.nasa.gov/ingest/order-option/order-option-1?provider=PROV1"
+	curl -v -XDELETE -H "$TOKEN" "https://cmr.earthdata.nasa.gov/ingest/order-option/order-option-1?provider=PROV1"
```
sxu123, line 83: Shouldn't these interface be changed now? I know the following works: 
(curl+ -v -XPOST -H "Echo-Token: mock-echo-system-token"
         -H "Content-Type:application/vnd.nasa.cmr.umm+json"
         "http://localhost:3002/providers/PROV1/serviceentries/serviceentry-8671"
         -d (slurp (io/resource "serviceentry_8671.json")))

eudoroolivares2016: Yes. There is a ticket to update that documentation seperate from this ticket

ingest-app/src/cmr/ingest/api/routes.clj
```diff
@@ -239,7 +232,7 @@
             provider-id native-id request)))
 
        ;; Generic documents are by pattern: /providers/{prov_id}/{concept-type}/{native_id}
-       (context ["/:concept-type" :concept-type generate-generic-concept-types-reg-ex] [concept-type]
+       (context ["/:concept-type" :concept-type (re-pattern common-generic/generic-concept-types-reg-ex)] [concept-type]
```
eudoroolivares2016, line 235: Should this actually be calling the `plural-generic-concept-types-reg-ex` function. The `generate-generic-concept-types-reg-ex` function we are removing here is pluralizing the regex list?

zimzoom: My understanding was that the ingest routes use the singular form, and the the search and bootstrap routes use the plural form.

search-app/src/cmr/search/services/parameters/conversion.clj
zimzoom: We discussed this and decided to put the TODO back in for adding additional params through a config file, and create a ticket for it.

schemas/resources/schemas/data-quality-summary/v1.0.0/search.md
zimzoom: good catch!

indexer-app/src/cmr/indexer/data/concepts/generic.clj
```diff
@@ -74,7 +75,7 @@
         generic-associations (esearch/parse-non-tombstone-associations
                               context
                               (meta-db/get-generic-associations-for-concept context concept))
-        gen-name (util/safe-lowercase (get-in parsed-concept [:MetadataSpecification :Name]))
+        gen-name (csk/->kebab-case (get-in parsed-concept [:MetadataSpecification :Name] ""))
```
sxu123, line 78: safe-lowercase is to guard the nil case. (csk/->kebab-case nil) will throw exception.

indexer-app/test/cmr/indexer/test/data/index_set_generics.clj
sxu123: Actually, I think only the public interface is required to add the dashes...

ingest-app/src/cmr/ingest/api/generic_documents.clj
```diff
@@ -57,7 +57,7 @@
         raw-document (slurp (:body request))
         document (json/parse-string raw-document true)
         specification (:MetadataSpecification document)
-        spec-key (keyword (string/lower-case (:Name specification)))
+        spec-key (csk/->kebab-case-keyword (:Name specification))
```
sxu123, line 60: same comment about nil case.

indexer-app/src/cmr/indexer/data/concepts/generic.clj
zimzoom: Yes, that is what we discussed this morning at standup, and the suggestion proposed (which is shown here) was to add a default case to csk/->kebab-case, which is what the empty string is. In other words, `(csk/->kebab-case nil)` throws exception, but `(csk/->kebab-case nil "")` does not.

Generics.md
eereiter: Ok, I am happy, now.  Just talked with @ygliuvt and since no one is using them yet, we can use the new names and delete the old indices.  Then we can use bulk update to update anything that has been created.

indexer-app/test/cmr/indexer/test/data/index_set_generics.clj
zimzoom: I think (someone correct me if I'm wrong) that while the dashes were only required of the public one, we opted to change both because they will be copies of each other.

ingest-app/src/cmr/ingest/api/generic_documents.clj
zimzoom: good catch, adding the default empty string to the implicit get here, for safe-guarding against nils

eudoroolivares2016 (approved):

indexer-app/test/cmr/indexer/test/data/index_set_generics.clj
sxu123: It's safe to change it all as you don't know when the two parts are interacted.

zimzoom: Done - fixed :Name to match

ingest-app/src/cmr/ingest/api/routes.clj
zimzoom: Done, ingest is plural

Generics.md
zimzoom: I agree that it would be more readable if we made our example native-id's something that does not look like concept-type or Name at first glance.

sxu123 (approved):

PR #1768, opened 2022-11-07
CMR-8732: First cut of the legacy service migration script to support DataQualitySummary migration.

ygliuvt:
In this first cut of the script, it is only coded to work with SIT. We will update the code to work with other environments as we add more data migrations.

### DIFF
--- a/legacy-migration/migrate.py
+++ b/legacy-migration/migrate.py
@@ -40,7 +40,7 @@ def migrate_row(id, name, summary, provider_guid):
               "Authorization": access_token,
               "User-id": "legacy_migration"}
 
-    resp = requests.put(f"{url_root}/ingest/dataqualitysummary/{id}?provider={providers_map[provider_guid]}",
+    resp = requests.put(f"{url_root}/ingest/providers/{providers_map[provider_guid]}/dataqualitysummaries/{id}",
                         data=json.dumps(umm),
                         headers=header)
     if resp.status_code > 300:
--- a/legacy-migration/migrate.py
+++ b/legacy-migration/migrate.py
@@ -1,0 +1,66 @@
+import json
+import oracledb
+import os
+import requests
+import time
+
+# returns strings or bytes instead of a locator
+oracledb.defaults.fetch_lobs = False
+
+user = os.environ.get('MIGRATE_USER')
+pwd = os.environ.get('MIGRATE_PWD')
+access_token = os.environ.get('ACCESS_TOKEN')
+
+url_root = "https://cmr.sit.earthdata.nasa.gov"
+providers_map = {}
+success_count: int = 0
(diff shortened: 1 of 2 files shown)


legacy-migration/migrate.py
```diff
@@ -0,0 +1,66 @@
...
+           "MetadataSpecification": {
+               "Name": "DataQualitySummary",
+               "Version": "1.0.0",
+               "URL": "https://cdn.earthdata.nasa.gov/generics/dataqualitysummary/v1.0.0"}}
+
+    header = {"Content-Type": "application/json",
+              "Authorization": access_token,
+              "User-id": "legacy_migration"}
+
+    resp = requests.put(f"{url_root}/ingest/dataqualitysummary/{id}?provider={providers_map[provider_guid]}",
```
eudoroolivares2016, line 43: With the change of the provider route no longer being a parameter we will need to change the script to
`(f"{url_root}/ingest/providers/providers_map[provider_guid]/dataqualitysummaries/{id}}`

eudoroolivares2016: ingest/providers/<Provier-id>/<plural-generic-concept>/<native-id>

daniel-zamora (approved):

legacy-migration/migrate.py
ygliuvt: Cool. updated.

legacy-migration/migrate.py
```diff
@@ -0,0 +1,66 @@
...
+               "URL": "https://cdn.earthdata.nasa.gov/generics/dataqualitysummary/v1.0.0"}}
+
+    header = {"Content-Type": "application/json",
+              "Authorization": access_token,
+              "User-id": "legacy_migration"}
+
+    resp = requests.put(f"{url_root}/ingest/providers/{providers_map[provider_guid]}/dataqualitysummaries/{id}",
+                        data=json.dumps(umm),
+                        headers=header)
+    if resp.status_code > 300:
```
jceaser, line 46: I saw this and thought >= would be better, I gripped around and did find 301 and no 300 so your probably safe, but if you happen to change for someone else, consider adding the equals.

legacy-migration/migrate.py
```diff
@@ -0,0 +1,66 @@
...
+            for r in cursor.execute(sql):
+                providers_map[r[0]] = r[1]
+
+def migrate_row(id, name, summary, provider_guid):
+    global success_count, failure_count
+    umm = {"Id": id,
+           "Name": name,
+           "Summary": summary,
+           "MetadataSpecification": {
+               "Name": "DataQualitySummary",
```
jceaser, line 35: heads up, this value _may_ change to Data Quality Summary

jceaser (approved):

PR #1772, opened 2022-11-10
CMR-8775: Control access to Generic API ingest interfaces with a new CMR configuration

eudoroolivares2016:
* Adds a configuration to block specified concepts from being ingested into CMR

### DIFF
--- a/common-lib/src/cmr/common/config.clj
+++ b/common-lib/src/cmr/common/config.clj
@@ -240,6 +240,14 @@
   {:default (approved-pipeline-documents)
    :parser #(json/parse-string % true)})
 
+(defconfig generic-ingest-disabled-list
+  "This is a toggle to disallow specified generic concepts from being ingested into CMR,
+   Should be an array of keys with the name of the schema to be disallowed from ingest
+   example [:grid :order-option], would disallow grids and order options from being ingested into CMR
+   if no concepts should be disalllowed set to an empty list"
+  {:default []
+   :parser #(json/parse-string % true)})
+
 (defn check-env-vars
   "Checks any environment variables starting with CMR_ are recognized as known environment variables.
   If any are unrecognized a warning message is logged. Usually this should be called at the start
--- a/ingest-app/src/cmr/ingest/api/generic_documents.clj
+++ b/ingest-app/src/cmr/ingest/api/generic_documents.clj
@@ -21,6 +21,7 @@
 
 (defn disabled-for-ingest?
   "Determine if a generic schema is disallowed for ingest
+   Parameters:
    * schema, the keyword name of an approved generic"
   [schema]
   (some #{schema} (cfg/generic-ingest-disabled-list)))
--- a/ingest-app/src/cmr/ingest/api/generic_documents.clj
+++ b/ingest-app/src/cmr/ingest/api/generic_documents.clj
@@ -9,6 +9,7 @@
    [cmr.common-app.api.launchpad-token-validation :as lt-validation]
    [cmr.common.generics :as gconfig]
    [cmr.common.concepts :as common-concepts]
+   [cmr.common.config :as cfg]
    [cmr.common.log :refer [debug info warn error]]
    [cmr.common.services.errors :as errors]
    [cmr.common.util :refer [defn-timed]]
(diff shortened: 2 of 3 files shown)


common-lib/src/cmr/common/config.clj
```diff
@@ -240,6 +240,14 @@
   {:default (approved-pipeline-documents)
    :parser #(json/parse-string % true)})
 
+(defconfig generic-ingest-disabled-list
+  "This is a toggle to disallow specified generic concepts from being ingested into CMR,
+   Should be an array of keys with the name of the schema to be disallowed from ingest
+   example [:grid :order-option], would disallow grids and order options from being ingested into CMR
+   if no concepts should be disalllowed set to an empty list"
```
dpesall, line 247: Maybe use the terms `prevent` and `prevented` rather than `disallow` and `disallowed` for clarity. Also I'm a bit unclear on `Should be an array`. Is this saying the generic concepts being ingest should be an array of keys?

ingest-app/src/cmr/ingest/api/generic_documents.clj
```diff
@@ -27,13 +35,17 @@
   (if-not (gconfig/approved-generic? schema version)
     (errors/throw-service-error
      :invalid-data
-     (format "The [%s] schema on version [%s] is not an approved schema. This record cannot be ingested." schema version))
-    (if-some [schema-file (gconfig/read-schema-specification schema version)]
-      (let [schema-obj (js-validater/json-string->json-schema schema-file)]
-        (js-validater/validate-json schema-obj raw-json true))
+     (format "The [%s] schema on version [%s] is not an approved schema. This record cannot be ingested" schema version))
```
dpesall, line 38: Period should be at the end of the sentence of the error message.

dpesall (changes requested):

common-lib/src/cmr/common/config.clj
dpesall: That makes sense! I would probably add in `The config's value` before the `should be an array`.

eudoroolivares2016: Good point with "disallow" that does evoke that this has something to do with Access control which it does not, will change to 'prevent'

dpesall (approved):

ingest-app/src/cmr/ingest/api/generic_documents.clj
```diff
@@ -27,13 +35,17 @@
...
-     (format "The [%s] schema on version [%s] is not an approved schema. This record cannot be ingested." schema version))
-    (if-some [schema-file (gconfig/read-schema-specification schema version)]
-      (let [schema-obj (js-validater/json-string->json-schema schema-file)]
-        (js-validater/validate-json schema-obj raw-json true))
+     (format "The [%s] schema on version [%s] is not an approved schema, this record cannot be ingested." schema version))
+    (if (disabled-for-ingest? schema)
       (errors/throw-service-error
        :invalid-data
-       (format "While the [%s] schema with version [%s] is approved, it cannot be found." schema version)))))
+       (format "The %s schema is currently disabled and cannot be ingested" schema))
```
eereiter, line 42: period after ingested.

system-int-test/test/cmr/system_int_test/ingest/generics_test.clj
```diff
@@ -26,17 +26,24 @@
...
-    (with-redefs [config/approved-pipeline-documents (fn [] {:grids ["1.0.0"]})]
+    (with-redefs [config/approved-pipeline-documents (fn [] {:grids ["1.0.0"]}) config/generic-ingest-disabled-list (fn [] [])]
       (is (true? (gcfg/approved-generic? :grids "1.0.0")) "Grids should be an approved format")
       (is (not (gcfg/approved-generic? :grid "0.0.1")) "CMR is using default configuration")
       (is (not (gcfg/approved-generic? :fake "a.b.c")) "A fake type was incorectly approved")))
-
   (testing
    "Verify that a document can be validated as approved schema."
-    (with-redefs [config/approved-pipeline-documents (fn [] {:grid ["0.0.1"]})]
+    (with-redefs [config/approved-pipeline-documents (fn [] {:grid ["0.0.1"]}) config/generic-ingest-disabled-list (fn [] [])]
```
eereiter, line 46: Can you remove the config/generic-ingest-disabled-list (fn [] [])].  Will the code still work and just use the default?

eudoroolivares2016: Yea It will. I just figured it was better to put it into that test since it is being read in the actual function. Will remove from test

eereiter: you have it in other tests, so I thought that one without would be OK.

eereiter (approved):

PR #1775, opened 2022-11-10
CMR-8804: Increasing the max length for the id field in order-option

eudoroolivares2016:
* Increasing the max length for the id field in order-option


eereiter (approved):

TylerHeald1 (approved):

ygliuvt (changes requested):
Although the JIRA ticket only mentioned `id` field of not being long enough, there are other fields (e.g. `Name` and `Description`) in the JSON schema that have maxLength that is shorter than those defined in Legacy Service `EJB_OPTION_DEF` table. These fields need to be changed as well.

ygliuvt (approved):

PR #1787, opened 2022-11-23
CMR-8712: cascading index issue

sxu123:
Association delete events need to be published when collection is deleted.  Association delete events also need to be published when generic document is deleted.

### DIFF
--- a/metadata-db-app/src/cmr/metadata_db/services/concept_service.clj
+++ b/metadata-db-app/src/cmr/metadata_db/services/concept_service.clj
@@ -893,8 +893,8 @@
                              :latest true})]
         ;;For generic associations, conceptn could appear as both source-concept
         ;;and associated-concept so it needs to tombstone additional associations.
-      (tombstone-associations context assoc-type search-params1 true concept-type)
-      (tombstone-associations context assoc-type search-params2 true concept-type)))))
+      (tombstone-associations context assoc-type search-params1 false concept-type)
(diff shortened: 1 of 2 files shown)


metadata-db-app/src/cmr/metadata_db/services/concept_service.clj
```diff
@@ -950,7 +950,17 @@
             ;; collection revision is deleted. Indexer will index the variables associated with
             ;; the collection through the collection delete event. Not the variable association
             ;; delete event.
-            (when-not skip-publication
+
+            ;; CMR-8712: After deleting a collection, searching for services, tools, and generic
+            ;; documents that are associated with the collection, still show the collection association.
+            ;; This is because skip-publication is set to true for all associations when a collection is
+            ;; deleted. Given the above comments regarding the need for the tag and variable associations
```
eudoroolivares2016, line 957:
```suggestion
            ;; deleted. The associations for tag and variables to collections should remain see above
```

eudoroolivares2016 (approved):

metadata-db-app/src/cmr/metadata_db/services/concept_service.clj
sxu123: Thanks. But "should remain" is too vague, in my opinion. I would rather summarize the huge comment block above, i.e. variable and tag association delete event should skip the publication.

zimzoom (approved):

PR #1788, opened 2022-11-29
CMR-8824 combining order options and service option schemas. so that …

eereiter:
…both can be ingested using this 1 schema.

### DIFF
--- a/schemas/resources/schemas/order-option/v1.0.0/schema.json
+++ b/schemas/resources/schemas/order-option/v1.0.0/schema.json
@@ -7,18 +7,17 @@
   "properties": {
     "Id": {
       "description": "The identifier that was assigned to this record by an older system. It is in the schema to preserve search API functionality in order to store this identifier.",
-      "type": "string",
-      "maxLength": 200
+      "type": "string"
     },
     "Name": {
       "description": "The name is a shortened name used to distinguish between other option definitions. This field must be unique per provider and is restricted to 30 characters.",
       "type": "string",
-      "maxLength": 30
+      "maxLength": 255
     },
     "Description": {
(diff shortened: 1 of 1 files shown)


ygliuvt:
Now that order-option and service-option are combined, shouldn't we delete the service-option from the generic concept schemas and configurations?

schemas/resources/schemas/order-option/v1.0.0/schema.json
```diff
@@ -7,18 +7,17 @@
   "properties": {
     "Id": {
       "description": "The identifier that was assigned to this record by an older system. It is in the schema to preserve search API functionality in order to store this identifier.",
-      "type": "string",
-      "maxLength": 200
+      "type": "string"
```
ygliuvt, line 10: Why do we remove the maxLength from the id field? I think it would be good to have a maxLength.

eereiter: Service option schema doesn't have a length associated with the ID field in the rest API.  Do you know if service option ids are stored in the same DB table as order options?  I can add the maxLength back in, I just don't want to eliminate existing data.

eereiter:
I will create another future PR to remove Service Entries and Service Options concepts from the CMR.

schemas/resources/schemas/order-option/v1.0.0/schema.json
ygliuvt: The oracle db schema has maxLength of 200 on the guid field of both tables, so maxLength of 200 will be safe.

eereiter: maxLength added back in.

sxu123 (approved):

eudoroolivares2016 (approved):

PR #1789, opened 2022-11-30
CMR-8826: Fixed searching for generic concepts by native-id.

ygliuvt:
This is only a quick fix of the searching issue to unblock the Legacy migration work. It does not address the missing tests of searching generic concepts by other search parameters or attempt to refactor the existing tests which will be covered in CMR-8828.

### DIFF
--- a/search-app/docs/api.md
+++ b/search-app/docs/api.md
@@ -4506,11 +4506,11 @@ Access to service and service association is granted through the provider via th
 
 #### <a name="service-association"></a> Service Association
 
-A service identified by its concept id can be associated with collections through a list of collection concept revisions. The service association request normally returns status code 200 with a response that consists of a list of individual service association responses, one for each service association attempted to create. Each individual service association response has an `associated_item` field and either a `service_association` field with the service association concept id and revision id when the service association succeeded or an `errors` field with detailed error message when the service association failed. The `associated_item` field value has the collection concept id and the optional revision id that is used to identify the collection during service association. Service association requires that user has update permission on INGEST_MANAGEMENT_ACL for the collection's provider. Here is a sample service association request and its response:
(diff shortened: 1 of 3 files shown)


system-int-test/test/cmr/system_int_test/search/generics_test.clj
```diff
@@ -114,4 +114,11 @@
           (testing "Test that generics will work with concept and revision searches."
             (let [results (search-request (format "concepts/%s/%s" concept-id revision-id) "")
                   status (:status results)]
+              (is (= 200 status) "wrong http status")))
+          
+          (testing "Search generic concept by native-id"
+            (let [results (search-request plural-concept-type-name (str "native-id=" native-id))
+                  status (:status results)
+                  body (:body results)]
+              (is (string/includes? body concept-id) "record not found")
```
ygliuvt, line 123: Don't like this kind of verifications, but didn't refactor the existing tests to do it differently in this PR. I hope this refactoring will happen in CMR-8828.

eereiter (approved):

search-app/docs/api.md
```diff
@@ -4506,11 +4506,11 @@ Access to service and service association is granted through the provider via th
...
 
 ```
 curl -XPOST -i -H "Content-Type: application/json" -H "Authorization: Bearer XXXXX" %CMR-ENDPOINT%/services/S1200000008-PROV1/associations -d \
-'[{"concept_id": "C1200000005-PROV1"},
+'[{"concept_id": "C1200000005-PROV1", "data": {"order_option": "OO1200445588-PROV1"}},
```
eudoroolivares2016, line 4513: In this example, the data payload is a reference to an order-option concept. Is there a use-case for that? I have no problems with it but, I was advised to keep examples of generics outside of the static documentation since they can be dynamically added or removed such as what we intend to do with service-options.

ygliuvt: Yes, this is what Legacy Migration will put on the association. Given this will be how ordering is done going forward, I think it is OK to put this real example data here.

PR #1824, opened 2023-01-03
CMR-8727: Reducing access control Errors when indexing collections into graphDb

eudoroolivares2016:
* Adds a retry policy for indexCmrCollections that reduces instances of concurrent modification errors
* Adds a retry policy using axios-retry for requests to the access-control app
* This does not seem to completely fix the concurrency issues however, there are now scripts to index missing collections into graphDb to compensate
* Update to package.json
* Minor updates to unrelated tests and SC
* Improves logs so that cloudwatch interpretation is easier

### DIFF
--- a/graph-db/serverless/src/indexCmrCollection/__test__/handler.test.js
+++ b/graph-db/serverless/src/indexCmrCollection/__test__/handler.test.js
@@ -120,7 +120,7 @@ describe('indexCmrCollection handler', () => {
     expect(statusCode).toBe(200)
   })
 
-  test('test collection that has an index on CMR but, it is on actually stored in db', async () => {
+  test('test collection that has an index on CMR but, it is Not stored in db', async () => {
     const conceptId = 'C123755555-TESTPROV'
     nock(/local-cmr/)
       .get(/collections/)
--- a/graph-db/serverless/src/indexCmrCollection/__test__/handler.test.js
+++ b/graph-db/serverless/src/indexCmrCollection/__test__/handler.test.js
@@ -120,90 +120,29 @@ describe('indexCmrCollection handler', () => {
     expect(statusCode).toBe(200)
   })
 
-  // test('test collection that has an index but, is not in the oracle db', async () => {
-  //   // Pass an error to the response of fetchCmrCollection
-  //   // Emits an error event on the request object, not the reply
-  //   const conceptId = 'C123755555-TESTPROV'
-  //   nock(/local-cmr/)
-  //     .get(/collections/)
-  //     .replyWithError('Collection not found', { statusCode: 404 })
-  //     // commenting out for now TODO may use for exponential back-off strategy
-  //     // .get(/collections/)
-  //     // .replyWithError('Error Collection not found', { statusCode: 404 })
-  //     // .get(/collections/)
-  //     // .replyWithError('Error Collection not found', { statusCode: 404 })
+  test('test collection that has an index on CMR but, it is on actually stored in db', async () => {
+    const conceptId = 'C123755555-TESTPROV'
+    nock(/local-cmr/)
+      .get(/collections/)
+      .reply(200, {
+        hits: 0,
+        took: 6,
+        items: ['Bad Data in here']
+      })
 
-  //   const event = getEvent(conceptId, 'concept-update')
-
-  //   const consoleMock = jest.spyOn(console, 'log')
-
-  //   const indexed = await indexCmrCollection(event)
-  //   // It will try to reindex TODO won't reindex right now. I'm not sure yet why this is the value 2 since the error is only printed once
-  //   expect(consoleMock).toBeCalledTimes(2)
-  //   // expect(consoleMock).toBeCalledWith('Could not complete request due to error: Error: Error Collection not found')
-  //   // expect(consoleMock).toBeCalledWith(`Could not fetch collection ${conceptId} due to error: Error: Collection not found`)
-  //   // expect(consoleMock).toBeCalledWith('Error indexing collection, Exception was thrown in lambda: Error: Collection not found')
-  //   // expect(consoleMock).toBeCalledWith('Error indexing collection, Exception was thrown in index collection handler: ', new Error('Cannot read properties of null (reading \'data\')'))
-  //   expect(consoleMock).toBeCalledWith('Could not fetch collection C123755555-TESTPROV due to error:')
-  //   // expect(consoleMock).toBeCalledWith('Error indexing collection, Exception was thrown in lambda: Error Collection not found')
-
-  //   // expect(consoleMock).toBeCalledWith('Exception raised, retrying to index collection # ', 0)
-
-  //   const { body } = indexed
-
-  //   expect(body).toBe('Successfully indexed 0 collection(s).')
-  // })
-  // TODO Need more logging information from AWS before using retry keep commented out for now
-  // test.only('testing the retry policy for failed attempts', async () => {
-  //   // Pass errors until the third attempt does not return an error
-  //   // Emits an error event on the request object, not the reply
-  //   nock(/local-cmr/)
-  //     .get(/collections/)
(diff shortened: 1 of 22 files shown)


ygliuvt (approved):

graph-db/serverless/src/indexCmrCollection/__test__/handler.test.js
```diff
@@ -120,53 +120,40 @@ describe('indexCmrCollection handler', () => {
     expect(statusCode).toBe(200)
   })
 
-  test('test indexing of a broken collection', async () => {
-    // Pass an error to the response of fetchCmrCollection
+  test('test collection that has an index on CMR but, it is Not stored in db', async () => {
+    const conceptId = 'C123755555-TESTPROV'
     nock(/local-cmr/)
       .get(/collections/)
-      .replyWithError('Error Collection not found', { statusCode: 404 })
```
jaybarra, line 127: For the case of a 404, is that still handled?

eudoroolivares2016: So this test case was intended to be if we were able to fetch the collection from CMR but, the data that it returned was invalid. I actually think this should have always been a 200 to test this functionality not a 404. We saw this happen only once against SIT but, that is why we wrapped this around a try-catch. The function fetchCmrCollection should 
` if (getConceptType(conceptId) === 'collection' && action === updateActionType) {
      try {
        const collection = await fetchCmrCollection(conceptId, token)

        const { data = {} } = collection
        const { items = [] } = data

        if (items.length === 0) {
          console.log(`Skip indexing of collection [${conceptId}] as it is not found in CMR`)

          skipCount += 1
        } else {
          const [firstCollection] = items

          // Fetch the permitted groups for this collection from access-control
          const groupList = await fetchCollectionPermittedGroups(conceptId, token)

          console.log(`Start indexing concept [${conceptId}], revision-id [${revisionId}]`)

          await indexCmrCollection(firstCollection, groupList, gremlinConnection)

          recordCount += 1
        }`
If it gets data back from CMR then it is in the else block and to test if the data in the response was not good from this function:
 https://github.com/nasa/Common-Metadata-Repository/blob/master/graph-db/serverless/src/utils/cmr/__test__/fetchCmrCollection.test.js
 
 There is a different test in that file called:
'test index of not found collection' which is should be checking against a collection which is not found from CMR in that case it would hit the `if` block. While that test does have 200 response being sent perhaps it should be 404? either way I'm going to add a more detailed comment for the test we were originally discussing. Here is the other test I mentioned: https://github.com/nasa/Common-Metadata-Repository/blob/master/graph-db/serverless/src/indexCmrCollection/__test__/handler.test.js it is called `test index of not found collection`

graph-db/serverless/src/indexCmrCollection/__test__/handler.test.js
```diff
@@ -376,3 +363,59 @@ describe('indexCmrCollection handler', () => {
     await verifyRelatedUrlExistInGraphDb(anothercollectionTitle, newDocUrl)
   })
 })
+
+test('test indexing a collection where the event types is invalid', async () => {
```
eudoroolivares2016, line 367: * event type

dpesall (approved):

jceaser (approved):

eudoroolivares2016:
For those watching this PR, we are currently changing the graphDb engine in the envs so that may affect this PR which is what is holding it up currently

zimzoom (approved):

graph-db/serverless/src/utils/cmr/__test__/indexCollection.test.js
```diff
@@ -58,20 +58,26 @@ describe('utils#indexCmrCollection', () => {
...
+        .reply(200, mockAclResponse)
 
       // Provide `null` for the gremlin connection to throw an error
+      // await expect(indexCmrCollection(collectionObj, [], null))
+
       const result = await indexCmrCollection(collectionObj, [], null)
 
       expect(result).toBeFalsy()
 
+      // Retry policy in place. Called 5 times (maxDepth) uses two different logs for this error the project and the collection itself. One additional log for max attempts being reached
```
eudoroolivares2016, line 70: Remove this no longer the case

graph-db/serverless/src/utils/cmr/__test__/indexCollection.test.js
```diff
@@ -58,20 +58,26 @@ describe('utils#indexCmrCollection', () => {
...
 
       // Error message logged because deleteCmrCollection failed because of null gremlinConnection
       expect(consoleMock.mock.calls[0][0]).toEqual(`Error deleting project vertices only linked to collection [${conceptId}]: Cannot read properties of null (reading 'V')`)
 
-      // Error message logged because addV failed because of null gremlinConnection
-      expect(consoleMock.mock.calls[1][0]).toEqual(`Error indexing collection [${conceptId}]: Cannot read properties of null (reading 'addV')`)
+      // Error message logged because addV failed, first attempt being printed to console
+      expect(consoleMock.mock.calls[1][0]).toEqual(`Error indexing collection into graph database [${conceptId}]: Cannot read properties of null (reading 'addV')`)
+
+      // // Error message that max attempts have been reached according to the retry policy
```
eudoroolivares2016, line 79: Remove these comments as well

graph-db/serverless/src/utils/cmr/fetchCollectionPermittedGroups.js
```diff
@@ -1,11 +1,28 @@
...
-
+import axiosRetry from 'axios-retry'
 /**
  * Fetch a the permitted groups of a collection from CMR access control
  * @param {String} conceptId Collection concept id from CMR
  * @param {String} token An optional Authorization Token
  * @returns [] An array containing the permitted groups of a collection
  */
+
+// Compensate for any misses to the endpoint max retries is going to be 4 using
```
eudoroolivares2016, line 10: Remove these comments since we are no longer using the exponential retry instead we are just multiplying the wait ourselves by the amount of times we have tried

graph-db/serverless/src/utils/cmr/fetchPageFromCMR.js
```diff
@@ -79,21 +82,25 @@ export const fetchPageFromCMR = async ({
...
-
+          // Locally we will call indexCmrCollections directly, otherwise send the job to SQS queue
           await indexCmrCollections({ Records: queueBody })
         } else {
           const sqsEntries = []
-
           chunk.forEach((collection) => {
             const { id: conceptId } = collection
 
+            // TODO: Since we already made a call to CMR we may want to pass that information into the event
```
eudoroolivares2016, line 92: Useful because we can make the decision if this is a good change in the future

graph-db/serverless/src/utils/cmr/fetchPageFromCMR.js
```diff
@@ -106,17 +113,21 @@ export const fetchPageFromCMR = async ({
       })
     }
 
-    // If we have an active searchAfter and there are more results
+    // If we have an active searchAfter so we aren't on the last page and there are more results
+    // Second argument is to parse integers in base 10
     if (cmrsearchAfter && entry.length === parseInt(process.env.PAGE_SIZE, 10)) {
+      // TODO: Investigate putting in a wait function between calls for each page to allow it to get picked
```
eudoroolivares2016, line 119: Useful since we can investigate this for further optimizations to graphDb

graph-db/serverless/src/utils/cmr/fetchCollectionPermittedGroups.js
```diff
@@ -16,6 +33,9 @@ export const fetchCollectionPermittedGroups = async (conceptId, token) => {
 
   let response
   try {
+    // TODO: We can put a wait function here to ensure that this thing is waiting a long time between calls so that we can
```
eudoroolivares2016, line 36: Remove, putting a wait before the call does not make sense, better to put waits between failures to the access-control app if there are any

PR #1826, opened 2023-01-03
CMR-8853: Add associations and association details for variable .json endpoint

eudoroolivares2016:
* Adds associations and association-details to the variable's json response
* Adds documentation for .json response for search app's api.md

### DIFF
--- a/search-app/docs/api.md
+++ b/search-app/docs/api.md
@@ -3996,6 +3996,7 @@ These parameters will match fields within a variable. They are case insensitive 
   * options: ignore_case, or
 measurement_identifiers parameter is a nested parameter with subfields: contextmedium, object and quantity. Multiple measurement_identifiers can be specified via different indexes to search variables. The following example searches for variables that have at least one measurement_identifier with contextmedium of Med1, object of Object1 and quantity of Q1, and another measurement_identifier with contextmedium of Med2 and object of Obj2.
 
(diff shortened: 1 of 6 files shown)


isja17 (approved):

dpesall (approved):

ygliuvt:
We should update the API doc to document the associations and association details fields in the variable search JSON result format.

search-app/docs/api.md
```diff
@@ -4099,8 +4110,38 @@ Content-Length: 292
...
+  "hits" : 1,
+  "took" : 41,
+  "items" : [ {
+    "concept_id" : "V1200000022-PROV1",
+    "revision_id" : 2,
+    "provider_id" : "PROV1",
+    "native_id" : "sample-variable",
+    "name" : "methaneVar",
+    "long_name" : "Total Methane",
+    "associations" : {
```
ygliuvt, line 4130: With the documentation, it makes this observation clearer: For JSON result format that does not originally have the `associations` field, do we want to make the result more compact with `details` in the associations field, or do we want to make the `associations` and `association_details` fields separate just to be the same as old JSON result format which is not clean? I am OK with your choice here, but I just want to point this out that you made this choice (perhaps unintentionally).

ygliuvt (approved):

PR #1830, opened 2023-01-06
CMR-8828: Update the generic tests to add all of the parameters to make sure they work

eudoroolivares2016:
* Adds more tests for search parameters for generics

### DIFF
--- a/system-int-test/test/cmr/system_int_test/search/generics_test.clj
+++ b/system-int-test/test/cmr/system_int_test/search/generics_test.clj
@@ -286,7 +286,7 @@
                   body (:body results)]
               (is (string/includes? body concept-id) "record not found")
               (is (= 200 status) "wrong http status")))
-              
+
           (testing "Search generic concept by provider"
             (let [results (search-request plural-concept-type-name (str "provider=" provider))
                   status (:status results)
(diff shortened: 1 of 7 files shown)


jaybarra (approved):

dpesall (approved):

system-int-test/test/cmr/system_int_test/search/generics_test.clj
```diff
@@ -71,8 +72,164 @@
...
+            (string/upper-case native-id)
+            "options[native-id][ignore-case]=true"
+
+            "Search using native-id pattern parameter search"
+            (inf/plural concept-type-string)
+            "native-id"
+            (str (string/join "" (drop-last native-id)) "*")
+            "options[native-id][pattern]=true"))
+            ;; test for legacy documents but, some generics do not have a guid-id
+            (if guid-id
```
ygliuvt, line 219: Neither the API doc nor this test is very clear on what the `id` parameter is actually matching against. From reading the code, I can see the `id` is actually optional for generic concepts, this is not intuitive. I think we need to make that clear in the API doc and the test here.

ygliuvt (approved):

PR #1831, opened 2023-01-09
CMR-8857: Risk of Unredacted Invalid Token String Returned to User

TylerHeald1:


dpesall (approved):

isja17 (approved):

system-int-test/test/cmr/system_int_test/search/tagging/tag_crud_test.clj
```diff
@@ -273,7 +273,7 @@
           token (echo-util/login (system/context) "user1")
           {:keys [status errors]} (tags/create-tag token tag)]
       (is (= 400 status))
-      (is (= ["Launchpad token is required. Token [ABC-1] is not a launchpad token."] errors))))
+      (is (= ["Launchpad token is required. Token [XXX] is not a launchpad token."] errors))))
```
jaybarra, line 276: If the token is entirely scrubbed, we should just remove the "XXX", if it scrubs parts of it we should make the test token longer.

PR #1835, opened 2023-01-19
CMR-8925 Stepping down snapshot version

dpesall:
The CMR Search app has a dependency on `org.geotools/gt-shapefile` which was pulling version 29 snapshot which is no longer on the Maven Repository with a Java 8 version. Snapshot version reverted to 28.


eudoroolivares2016 (approved):

jceaser (approved):
it was explained to me that the snapshots were the releases that have the improved security releases, this is why we are not on a hard version.

PR #1846, opened 2023-02-06
Cmr 8939

eereiter:


umm-spec-lib/src/cmr/umm_spec/models/umm_variable_models.clj
```diff
@@ -257,7 +257,6 @@
    ;; The name of the dimension of the variable represented in the data field. For example, 'XDim.
    Name
 
-   ;; The size of the dimension of the variable represented in the data field. For example, '1200'.
```
TylerHeald1, line 260: What's the reason to remove this comment?

isja17 (approved):

umm-spec-lib/src/cmr/umm_spec/models/umm_variable_models.clj
eereiter: When a UMM schema changes, the developer needs to run the lein generate-umm-records clojure code as documented in the umm-spec-lib/README.md file. Developers should not hand edit this file since it is program generated.  The schema changed and the overall comment moved in the schema from the top level to the new sub-elements.  Because of this change the clojure code took it out.  I did not hand edit this file.

TylerHeald1 (approved):

PR #1848, opened 2023-02-13
CMR-8685: In the generics search API the version number should be allowed to be included

eudoroolivares2016:
In the generics search API the version number should be allowed to be included

curl -v -H "$TOKEN" -H "Accept: application/vnd.nasa.cmr.umm_results+json;version=1.0.0" http://localhost:3003/dataqualitysummaries?name="CER-BDS_Terra"

right now ;version=1.0.0 doesn't work.


common-app-lib/src/cmr/common_app/services/search/query_validation.clj
```diff
@@ -24,9 +25,11 @@
                         result-format)]
     (when-not (get (supported-result-formats concept-type) result-format)
       (if-let [version (mt/version-of mime-type)]
-        [(format "The mime type [%s] with version [%s] is not supported for %ss."
-                 (mt/base-mime-type-of mime-type) version (name concept-type))]
-        [(format "The mime type [%s] is not supported for %ss." mime-type (name concept-type))]))))
+        ;; Some concepts aren't plural with just an 's'
+        [(format "The mime type [%s] with version [%s] is not supported for %s."
+                 (mt/base-mime-type-of mime-type) version (inf/plural (name concept-type)))]
```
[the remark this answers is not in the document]
eudoroolivares2016: That is fair, I would maybe push back that it is already being used in the Common App project: `[inflections "0.13.0"]` I didn't really introduce it, I can look for an alternative way to pluralize though.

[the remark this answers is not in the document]
eereiter: What I would suggest in common-lib/src/cmr/common/concepts.clj on line 80 there is a function called pluralize-concept-type which takes a singular concept-key (such as :collection) and returns the plural key (:collections).  Since you need the name and not the key, could you please add 2 new functions into concept.clj

(defn pluralize-concept-type-name
  "Pluralizes the passed in concept keyword/string and returns the string. The compliment
   function is singularize-concept-type-name"
  [concept-key]
  (inf/plural (name concept-key)))

(defn singularize-concept-type-name
  "Singularize the passed in concept keyword/string and returns the string. The compliment
   function is pluralize-concept-type-name"
  [concept-key]
  (inf/singular (name concept-key)))

Then change starting on line 80 (the old number) the function to be:
(defn pluralize-concept-type
  "Pluralizes the passed in concept keyword/string and returns it. The compliment
   function is singularize-concept-type"
  [concept-key]
  (keyword (pluralize-concept-type-name concept-key)))

(defn singularize-concept-type
  "Singularize the passed in concept keyword/string and returns it. The compliment
   function is pluralize-concept-type"
  [concept-key]
  (keyword (singularize-concept-type-name concept-key)))

Then in your code use the pluralize-concept-type-name function.  I agree with Lucia that should use the wrapper when we can.  

Doing a search through the code, I do see it being used in many places where we should use the wrapper.  I will write a ticket to change these to use the common-lib functions.

system-int-test/test/cmr/system_int_test/search/generics_test.clj
```diff
@@ -37,6 +38,27 @@
...
+   (-> {:method :get
+        :url (format "%s%s.%s" (url-helper/search-root) concept-type-ext url-extension)
+        :connection-manager (system/conn-mgr)
+        :throw-exceptions false}
+       (client/request))))
+
+
+(defn search-request-version-accept-header
+  "This function will make a request to one of the generic URLs passing the accept header to specify version"
+  ([concept-type-ext accept]
```
daniel-zamora, line 54: multiarity syntax isn't needed if you aren't using it.

system-int-test/test/cmr/system_int_test/search/generics_test.clj
```diff
@@ -37,6 +38,27 @@
         :throw-exceptions false}
        (client/request))))
 
+
+(defn search-request-version-url-extension
+  "This function will make a request to one of the generic URLs using and extended url to include version"
+  ([concept-type-ext url-extension]
```
daniel-zamora, line 44: multiarity syntax isn't needed if you aren't using it.

search-app/src/cmr/search/validators/validation.clj
```diff
@@ -18,7 +19,7 @@
   "Returns the umm versioned result formats for the given concept-type"
   [concept-type]
   (for [format-key [:umm-json :umm-json-results]
-        version (umm-version/versions concept-type)]
+        version ((merge umm-version/versions generic/documents-all-versions) concept-type)]
```
daniel-zamora, line 22: There is something weird going on here that is resulting in two `((` to start,  I think it would be better to use `(get`

daniel-zamora (approved):
Just a few style comments

common-app-lib/src/cmr/common_app/services/search/query_validation.clj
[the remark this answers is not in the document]
eudoroolivares2016: Added the functions and swapped according to @eereiter comment

common-app-lib/src/cmr/common_app/services/search/query_validation.clj
```diff
@@ -24,9 +25,11 @@
                         result-format)]
     (when-not (get (supported-result-formats concept-type) result-format)
       (if-let [version (mt/version-of mime-type)]
-        [(format "The mime type [%s] with version [%s] is not supported for %ss."
-                 (mt/base-mime-type-of mime-type) version (name concept-type))]
-        [(format "The mime type [%s] is not supported for %ss." mime-type (name concept-type))]))))
+        ;; Some concepts aren't plural with just an 's'
```
jceaser, line 28: this comment is to remind us that other extensions may exist, or that is why we use a library?

schemas/resources/schemas/service-entry/v1.0.0/search.md
```diff
@@ -93,7 +93,7 @@ The JSON response includes the following fields.
 __Example__
 
 ```
-    curl -H "Cmr-Pretty: true" "%CMR-ENDPOINT%/service-entries.json?name="service-entry-name"
+    curl -H "Cmr-Pretty: true" "%CMR-ENDPOINT%/service-entries.json?name="Serviceentry-name-v1"
```
jceaser, line 96: why is this example without -?

eudoroolivares2016: To make sure it matches the response but, I can set them all to use the "-" characters instead if that is preferred, that way would be congruent with the actual concept name

common-app-lib/src/cmr/common_app/services/search/query_validation.clj
```diff
@@ -24,9 +25,11 @@
                         result-format)]
     (when-not (get (supported-result-formats concept-type) result-format)
       (if-let [version (mt/version-of mime-type)]
```
jceaser, line 27: is version still optional, I can not tell from the tests

common-app-lib/src/cmr/common_app/services/search/query_validation.clj
jceaser: I don't think it is needed

schemas/resources/schemas/service-entry/v1.0.0/search.md
jceaser: I see, it's just data

common-app-lib/src/cmr/common_app/services/search/query_validation.clj
eudoroolivares2016: Do you mean for generics? If so yes, the rest of the search tests don't pass version so they were my "control" would comments in `generics_test` be sufficient?

schemas/resources/schemas/service-entry/v1.0.0/search.md
eudoroolivares2016: Right, perhaps it's even better so not as to imply that there is some kind of required link between the name and the concept-type when it's really just a string.

dpesall (approved):

PR #1854, opened 2023-02-23
Cmr 8784 - Adding Bulk updates capability to add/update/remove RelatedURLs in UMM-G.

eereiter:


ingest-app/src/cmr/ingest/services/granule_bulk_update/online_resource_url/echo10.clj
```diff
@@ -3,29 +3,30 @@
    and OnlineAccessURL in bulk update."
   (:require
    [clojure.data.xml :as xml]
-   [clojure.set :as set]
+   [clojure.set :as cset]
```
jaybarra, line 6: This shouldn't be necessary, importing it as `set` is inline with [the style guide](https://guide.clojure.style/#use-idiomatic-namespace-aliases) (despite us not doing the same for `str`)

jaybarra (approved):

TylerHeald1 (approved):

PR #1859, opened 2023-02-28
CMR-8907 Define EFS protocol for use in Metadata-DB

TylerHeald1:
This ticket has undergone some change since it's creation, but the idea remains the same: create some basic interface for metadata-db to interact with EFS and add a feature toggle to wrap EFS functionality in. This is just to declare some basic functions and library content


johnwteague (commented):
I guess, even after the pivot, I was expecting that you would have functional place-holders with the feature toggle in place.  So now I'm confused where the toggles will go.

TylerHeald1:
@johnwteague The issue with placing functional place-holders at this stage having pivoted is that I was very much getting into the weeds of where and how to implement those for specific CRUD items. Create and update I think will be relatively easy because the save-concept function is the primary location for actually inserting concepts into the DB, but reads from the database happen in multiple places and in different ways, and the data read from the DB is propagated up to be used in calling functions. There are some functions that get singular concepts and some that pull a selection of concepts at a time. It was slowing me down quite a bit and I didn't see an easy or quick solution to location for a toggle and then parsing data from oracle or from efs for calling functions so I wanted to move the addition of feature toggles for CRUD operations to the ticket for those CRUD items so that I could move on to work on create and hopefully update this sprint.

I certainly can spend more time looking into it if you want some mocked feature toggles in this PR.

johnwteague:
I don't want to expand the scope of this ticket, if we don't need to identify/create the CRUD methods for EFS in order to meet the spirit of the work item (given that you are no longer creating a protocol to handle this switch) but the A/C states:

    _A basic set of required functions for interacting with EFS are declared
    EFS functionality can be toggled with an environment variable_
   
And I'm not sure this defines or delivers that. Which is fine as long as we know what we will do and where, in the individual CRUD tickets, but the purpose here was to build out an implementation skeleton with functional implementation to come. Do you think this PR fulfills that? 

And we don't need to identify the READ points now; I think we can declare that out of scope for this (even though the ticket doesn't explicitly state CRUD only, that's our PI objective)

TylerHeald1:
Good points. I would say that it doesn't really meet that second AC. I'll put some more work into this and put together more of a framework.

efs-lib/src/cmr/efs/connection.clj
```diff
@@ -0,0 +1,39 @@
...
+  "Returns the health status of the EFS instance."
+  []
+  )
+
+(defn health
+  "Returns the efs health with timeout handling."
+  []
+  ())
+
+(defn save-concept
```
eereiter, line 20: If a concept already exists does the save overwrite or leaves what is there?

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
```diff
@@ -345,21 +347,35 @@
...
-  ([db concept-type provider concept-id revision-id]
-   (if revision-id
-     (let [table (tables/get-table-name provider concept-type)]
-       (j/with-db-transaction
-         [conn db]
+       (when (or
+              (= "efs-on" efs-config/efs-toggle)
+              (= "efs-only" efs-config/efs-toggle))
+         (efs/get-concept provider concept-type concept-id))
+       (when (or
```
eereiter, line 354: Can you just check to see if (not (= efs-only efs-config/efs-toggle)).

efs-lib/src/cmr/efs/connection.clj
TylerHeald1: Saving will be based on current rules. If different revisions of a concept are kept, then saving to EFS will have to as well

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
TylerHeald1: I suppose I could do that for both. check for (not (= efs-only...)) and (not (= efs-off...))
Functionally would make no real difference, would just mean the third toggle value could be any nonsense value

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
```diff
@@ -380,11 +388,16 @@
                                      (where `(and (= :c.concept-id :t.concept-id)
                                                   (= :c.revision-id :t.revision-id)))))
 
-              result (doall (map (partial db-result->concept-map concept-type conn provider-id)
-                                 (su/query conn stmt)))
+              result-oracle (if (not (= "efs-only" efs-config/efs-toggle))
```
eereiter, line 391: You should use (when instead of (if in this case.  When returns nil if not true.

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
```diff
@@ -380,11 +388,16 @@
...
+                              (doall (map (partial db-result->concept-map concept-type conn provider-id)
+                                          (su/query conn stmt)))
+                              nil)
+              result-efs (if (not (= "efs-off" efs-config/efs-toggle))
```
eereiter, line 395: You should use (when instead of (if in this case.  When returns nil if not true.

efs-lib/src/cmr/efs/connection.clj
```diff
@@ -0,0 +1,39 @@
+(ns cmr.efs.connection
```
eereiter, line 1: When are the functions going to be written for EFS?  Is that another ticket that will be worked later?

efs-lib/src/cmr/efs/config.clj
```diff
@@ -0,0 +1,11 @@
...
+  "Contains functions for retrieving import EFS details from environment variables"
+  (:require [cmr.common.config :as cfg :refer [defconfig]]))
+ 
+ (defconfig efs-directory
+   "Directory EFS is mounted to"
+   {:default "/efs"})
+ 
+ (defconfig efs-toggle
+   "Three-way toggle for EFS functionality. 'efs-off' uses only Oracle, 'efs-on' uses both Oracle and EFS, and 'efs-only' uses only EFS"
+   {:default "efs-off"})
```
eereiter, line 11: don't forget the empty line at the bottom of this file.

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
```diff
@@ -410,28 +423,31 @@
...
+                           table
+                           (string/join "," cols)
+                           seq-name
+                           (string/join "," (repeat (count values) "?")))]
+          (trace "Executing" stmt "with values" (pr-str values))
+          (when (not (= "efs-only" efs-config/efs-toggle))
+            (j/db-do-prepared db stmt values)
+            (after-save conn provider concept))
+          (when (not (= "efs-off" efs-config/efs-toggle))
+            (efs/save-concept provider concept-type concept))
```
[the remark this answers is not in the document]
TylerHeald1: All that EFS is being used for is storing the metadata element. Some things, like concept-id and revision-id might be in EFS as a result of the file path, but other metadata about the metadata is going to be stored in a different system(an as of yet undetermined system. DynamoDB is on the list to consider). For the save function, I have the entire concept being passed in because there are elements that will need to be pulled from it to determine where to store it in EFS

eereiter (approved):

johnwteague (approved):

PR #1872, opened 2023-03-16
CMR-8961: change Launchpad token authentication call from legacy services to URS

zimzoom:
Before, the control flow in transmit-lib's `tokens.clj` was: 
if this is a JWT token and local EDL verification is turned on, then verify the token locally; 
else send the token to legacy services. 
This meant that legacy services was being used for BOTH echo tokens and Launchpad tokens.

Because we are still supporting echo tokens today, the legacy services call is still there. One more `if` has been added to check if the token is a Launchpad token, if so send it to URS. 

Adding that `if` check, means that the token checking code had to be moved from common-app-lib to common-lib in order to avoid a circular dependency, because transmit-lib needs to be loaded before common-app-lib. (As common-app-lib requires transmit-lib)

### DIFF
--- a/common-app-lib/src/cmr/common_app/api/launchpad_token_validation.clj
+++ b/common-app-lib/src/cmr/common_app/api/launchpad_token_validation.clj
@@ -9,31 +9,7 @@
    [cmr.common.util :as common-util]
    [cmr.transmit.config :as transmit-config]))
 
-;; Note: Similar code exists at gov.nasa.echo.kernel.service.authentication
-(def URS_TOKEN_MAX_LENGTH 100)
-(def BEARER "Bearer ")
-
-(defn- is-legacy-token?
-  "There are two uses cases captured by this test, the Legacy token and the
-   new style legacy token made to behave like a legacy token but are prefixed
-   with EDL- to aid in indentification. This function will not match very short
-   JWT tokens.
-   Note: Similar code exists at gov.nasa.echo.kernel.service.authentication."
-  [token]
-  (or (<= (count token) URS_TOKEN_MAX_LENGTH)
-      (string/starts-with? token "EDL-")
-      (string/starts-with? token (str BEARER "EDL-"))))
-
-(defn is-launchpad-token?
-  "Returns true if the given token is a launchpad token.
-   If the token is not a Legacy (ECHO), Heritage (EDL+), or JWT (newest) token,
-   then it must be a Launchpad token.
-   Note: Similar code exists at gov.nasa.echo.kernel.service.authentication."
-  [token]
-  ;; note: ordered from least expensive to most
-  (not (or (is-legacy-token? token)
-           (common-util/is-jwt-token? token))))
-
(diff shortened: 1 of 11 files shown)


common-app-lib/src/cmr/common_app/api/launchpad_token_validation.clj
```diff
@@ -53,7 +29,7 @@
   [request-context]
```
eereiter, line 29: on line 28 of this file the validate-launchpad-token description still states 
Ingest will only be allowed if the user is in the NAMS CMR Ingest group
   (done in legacy services) and
Is this statement still true?  If not, could you please change it so that it documents what is true?

common-lib/src/cmr/common/util.clj
```diff
@@ -1022,6 +1022,32 @@
...
+;; TODO - remove legacy token check after legacy token retirement
+(defn is-legacy-token?
+  "There are two uses cases captured by this test, the Legacy token and the
+   new style legacy token made to behave like a legacy token but are prefixed
+   with EDL- to aid in indentification. This function will not match very short
+   JWT tokens.
+   Note: Similar code exists at gov.nasa.echo.kernel.service.authentication."
+  [token]
+  (or (<= (count token) URS_TOKEN_MAX_LENGTH)
+      (string/starts-with? token "EDL-")
```
eereiter, line 1038: I think (string/starts-with? token "EDL-"). is dead code. Check with Jay he would know for sure.

common-lib/src/cmr/common/util.clj
```diff
@@ -1022,6 +1022,32 @@
...
+      (string/starts-with? token (str BEARER "EDL-"))))
```
eereiter, line 1039: I think this is dead code as well: (string/starts-with? token (str BEARER "EDL-")))).  Please check with Jay first though.

transmit-lib/src/cmr/transmit/urs.clj
```diff
@@ -30,6 +31,10 @@
   [conn username]
   (format "%s/api/user_groups/groups_for_user/%s" (conn/root-url conn) username))
 
+(defn- launchpad-validation-url
+  [conn username]
+  (format "%s/api/nams/edl_user?token=%s" (conn/root-url conn) username))
```
eereiter, line 36: Do you need to full user profile here?  or just the user id?

jaybarra (approved):

common-lib/src/cmr/common/util.clj
```diff
@@ -1022,6 +1022,29 @@
           false))
       false)))
 
+;; Note: Similar code exists at gov.nasa.echo.kernel.service.authentication
+(def URS_TOKEN_MAX_LENGTH 100)
+(def BEARER "Bearer ")
```
eereiter, line 1027: You don't need this anymore. Please remove it.

eereiter (approved):

PR #1874, opened 2023-03-22
CMR-8912 READ concept from EFS

TylerHeald1:


isja17 (approved):

efs-lib/src/cmr/efs/connection.clj
```diff
@@ -2,16 +2,39 @@
...
+  "Returns a listing of revisions of the concept stored on EFS"
+  [provider concept-type concept-id]
+  (let [concept-dir-path (format "%s/%s/%s/%s" (efs-config/efs-directory) provider concept-type concept-id)]
+    (map (fn [file]
+           (.getName file)) (.listFiles (File. concept-dir-path)))))
+
+(defn get-latest-revision
+  [provider concept-type concept-id]
+  (first (sort > (map (fn [file-name]
+                        (Integer/parseInt (subs (second (str/split file-name #"\.")) 1))) (get-revision-file-names provider concept-type concept-id)))))
```
eereiter, line 30: Can you please put the (get-revision-file-name ... on the next line, that lines up with (fn ...) since this is another parameter to the map function.

efs-lib/src/cmr/efs/connection.clj
```diff
@@ -2,16 +2,39 @@
...
+
+(defn concept-revision-exists
+  [provider concept-type concept-id revision-id]
+  (let [concept-path (format "%s/%s/%s/%s/%s.r%d.zip" (efs-config/efs-directory) provider concept-type concept-id concept-id revision-id)]
+    (.exists (File. concept-path))))
```
eereiter, line 35: While the probability is low, this is a race condition waiting to happen.  The AWS EFS is built on a multitasking OS, so it is possible that after this function is called, but before you read the specific file, the file can get deleted.  Its better to just read the file and deal with the fact if it exists or not.

efs-lib/src/cmr/efs/connection.clj
```diff
@@ -26,19 +49,27 @@
...
 
 (defn get-concept
   "Gets a concept from EFS"
-  ([provider concept-type concept-id])
-  ([provider concept-type concept-id revision-id]))
+  ([provider concept-type concept-id]
+   (get-concept provider concept-type concept-id (get-latest-revision (:provider-id provider) (name concept-type) concept-id)))
+  ([provider concept-type concept-id revision-id]
+   (if (and revision-id
+            (concept-revision-exists (:provider-id provider) (name concept-type) concept-id revision-id))
```
eereiter, line 63: 1) Any time you have and if and you want to return a nil, you should use (when instead of (if
2) While the probability is low, you are creating a race condition when checking to see if a file exists and then later you try and read it.  The file could get deleted in the time between the successful exists and the time the file gets read.  You should just read the file and create logic to handle the case where it doesn't exist.

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
```diff
@@ -364,58 +364,76 @@
...
-       (when (not (= "efs-only" (efs-config/efs-toggle)))
-         (db-result->concept-map concept-type conn (:provider-id provider)
-                                 (su/find-one conn (select '[*]
-                                                           (from table)
-                                                           (where `(= :concept-id ~concept-id))
-                                                           (order-by (desc :revision-id)))))))))
+   (let [efs-concept-get (when (not (= "efs-off" (efs-config/efs-toggle)))
+                           (util/time-execution (efs/get-concept provider concept-type concept-id)))
+         oracle-concept-get (when (not (= "efs-only" (efs-config/efs-toggle)))
+                             (util/time-execution (j/with-db-transaction
```
eereiter, line 370: please move the (j/with-db-transaction to the next line.

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
```diff
@@ -364,58 +364,81 @@
...
-          (if result-oracle result-oracle result-efs))))
+    (let [efs-concepts-get (when (not (= "efs-off" (efs-config/efs-toggle)))
+                             (util/time-execution
+                              (efs/get-concepts provider concept-type concept-id-revision-id-tuples)))
+          oracle-concepts-get (when (not (= "efs-only" (efs-config/efs-toggle)))
+                                (util/time-execution
+                                 (j/with-db-transaction
+                                   [conn db]
+                                                       ;; use a temporary table to insert our values so we can use a join to
+                                                       ;; pull everything in one select
```
eereiter, line 422: pull these comments over to the right.

eereiter (approved):

efs-lib/src/cmr/efs/connection.clj
```diff
@@ -26,19 +46,27 @@
...
-  ([provider concept-type concept-id revision-id]))
+  ([provider concept-type concept-id]
+   (get-concept provider concept-type concept-id (get-latest-revision (:provider-id provider) (name concept-type) concept-id)))
+  ([provider concept-type concept-id revision-id]
+   (when revision-id
+     (let [concept-path (format "%s/%s/%s/%s/%s.r%d.zip" (efs-config/efs-directory) (:provider-id provider) (name concept-type) concept-id concept-id revision-id)]
+       (try
+         (info "Getting concept from EFS at path " concept-path)
+         {:revision-id revision-id :metadata (util/gzip-bytes->string (Files/readAllBytes (Paths/get concept-path (into-array String [])))) :deleted false}
+         (catch Exception _e nil))))))
```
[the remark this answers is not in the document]
TylerHeald1: I think I'll go ahead and log it for now. My expectation right now is that most errors would be a result of the revision not existing and causing a file doesn't exist error, but I could see a situation where some other error gets returned. And if we decide after doing EFS testing that it's just noise, it would be simple to remove

PR #1877, opened 2023-03-30
Cmr 9013 ingest 500 error

sxu123:


eereiter (approved):

system-int-test/test/cmr/system_int_test/ingest/granule_ingest_test.clj
```diff
@@ -474,15 +474,30 @@
                                (ingest/concept :granule "PROV1" "foo" :iso-smap valid-gran-metadata))]
          (is (= 201 status))))))
 
-(deftest CMR-5226-invalid-iso-smap-geographic-description-test
+(deftest CMR-9013-invalid-orbit-data
```
jaybarra, line 477: Would prefer not having specific tickets as part of test names in general.

sxu123: It seems that when we fix a specific bug, we add the ticket number in the test. I see other tests like that so I just did the same.

PR #1879, opened 2023-03-31
Cmr 8975 - Update UMM-C so that the tiling Identification Systems numeric/string coordinate values is managed from the schema.

eereiter:


TylerHeald1 (approved):

common-lib/src/cmr/common/util.clj
```diff
@@ -1079,14 +1080,20 @@
...
     s))
 
 (defn str->num
   "If the string can be converted to a number, return that number, otherwise return nil."
   [s]
-  (if (numeric? s)
-    (safe-read-string s)
+  (if (and (string? s)
+           (numeric? s))
+      (safe-read-string s)
```
sxu123, line 1096: not lined up right.  Also, where is numeric? from?  I couldn't run it.

eereiter: Fixed.  

numeric? is in the same file (common-lib/src/cmr/common/util.clj) on line 474.

PR #1880, opened 2023-03-31
CMR-8965: Create script to update legacy providers to new system (CMR)

eudoroolivares2016:
Write a script to move data from the old system to the new schema by reading the old tables and generating json files which are then loaded into the provider system using the new command line program.

This ticket does not need to talk directly to the database, the restful legacy service api may be enough. Also it does not need to write directly to CMR, it could just create metadata which is uploaded latter.

### DIFF
--- a/legacy-migration/providers/migrate.py
+++ b/legacy-migration/providers/migrate.py
@@ -1,4 +1,5 @@
 """Migrates providers from legacy services REST API into new providers interface"""
+import argparse
 import json
 import os
 import xml.etree.ElementTree as ET
@@ -6,11 +7,14 @@ import re
 import time
 import sys
 import requests
+import logging
+import validators
 
 URL_ROOT = "https://cmr.sit.earthdata.nasa.gov"
 
 access_token = os.environ.get('SIT_SYSTEM_TOKEN')
 CMR_INGEST_ENDPOINT = "https://cmr.sit.earthdata.nasa.gov/ingest"
+CMR_LOCAL_ingest_ENDPOINT = 'http://localhost:3002'
 
 # To test local ingesting into CMR
 local_cmr_access_token = os.environ.get('localHeader')
@@ -18,7 +22,7 @@ local_cmr_access_token = os.environ.get('localHeader')
 # TODO I want more clarity on how to handle these organization and contact roles since there is not an enum accessible substitute in the old format
 DEFAULT_ORGANIZATION_ROLE = "SERVICE PROVIDER"
 DEFAULT_CONTACT_ROLE = "TECHNICAL CONTACT"
-
+ 
 # Metadata specification for ingesting providers
 # It should not be expected this version or schema directories will change
 # within the lifecycle of this migration
@@ -27,9 +31,22 @@ metadata_specification = {
             "Version": "1.0.0",
             "URL": "https://cdn.earthdata.nasa.gov/schemas/provider/v1.0.0"}
 
+def retrieve_legacy_providers():
+    """Requests all of the providers on legacy-services and their metadata which we will be
+    re-ingesting into the new providers interface"""
+    try:
+        response = requests.get(f"{URL_ROOT}/legacy-services/rest/providers.json",
+            headers={"Authorization": access_token, "User-id": "legacy_migration"},
+            timeout=80)
+        providers = response.json()
+    except requests.exceptions.ConnectionError:
+        print('Failed to retrieve legacy service providers')
+        return None
+    return providers
+
 def ingest_provider(provider_metadata, provider_id):
     """Attempts to ingest metadata into CMR (locally)"""
-    response = requests.post("http://localhost:3002/providers/",
+    response = requests.post(f"{CMR_LOCAL_ingest_ENDPOINT}/providers/",
                         data=provider_metadata,
                         headers={"Authorization": local_cmr_access_token, "User-id": "legacy_migration",
                         "Content-Type": "application/json"},
@@ -49,12 +66,25 @@ def is_admin_group(group):
     # if in the name of the text admin comes up anywhere then it is an admin group
     return bool(re.search('admin', group.lower()))
 
+def convert_to_uri_value(url, provider_id):
+    """Where applicable convert the url in to URI format. For example:
+    www.example.com becomes https://example.com"""
+    # startsWith requires a tuple of strings if you want multiple ORed together
+    if validators.domain(url):
+        # if this is a valid domain ending but, front of URI is missing
+        if not url.startswith(('http://', 'https://')):
+            url = 'https://' + url
+        return url
+    else:
+        # We will not be able to migrate this provider since we'll want to manually check it's url
+        logging.warning('This url could not be coerced into a valid URI format ' + url + ' for provider ' + provider_id)
+        return url
+
 def get_admin_usernames(member_guid_arr):
     """Parses the list of member_guids and retrieves the edl usernames though an API call"""
     username_arr = []
-    if len(member_guid_arr) < 1:
+    if not member_guid_arr:
         return username_arr
-    #TODO: can we do this in one request instead of iterating over each username
     for user_id in member_guid_arr:
         print("User_id being passed in request", user_id)
         try:
@@ -70,19 +100,6 @@ def get_admin_usernames(member_guid_arr):
             print("Failed to Retrieve edl-username for guid" + str(user_id))
             return None
     return username_arr
-
-def retrieve_legacy_providers():
-    """Requests all of the providers on legacy-services and their metadata which we will be
-    re-ingesting into the new providers interface"""
-    try:
-        response = requests.get(f"{URL_ROOT}/legacy-services/rest/providers.json",
-            headers={"Authorization": access_token, "User-id": "legacy_migration"},
-            timeout=80)
-        providers = response.json()
-    except requests.exceptions.ConnectionError:
-        print('Failed to retrieve legacy service providers')
-        return None
-    return providers
 
 def get_groups_by_provider(owner_id):
     """Requests all groups that belong to a specific provider by passing the owner_id"""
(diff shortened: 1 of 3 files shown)


legacy-migration/providers/migrate.py
```diff
@@ -0,0 +1,312 @@
...
+            edl_username = user_element.find('username')
+            print("Retrieved edl-username for guid", user_id)
+            print('The edl username', edl_username.text)
+            username_arr.append(edl_username.text)
+        except requests.exceptions.ConnectionError:
+            print("Failed to Retrieve edl-username for guid" + str(user_id))
+            return None
+    return username_arr
+
+def retrieve_legacy_providers():
```
eudoroolivares2016, line 74: move to the top

legacy-migration/providers/migrate.py
```diff
@@ -0,0 +1,312 @@
...
+def get_admin_usernames(member_guid_arr):
+    """Parses the list of member_guids and retrieves the edl usernames though an API call"""
+    username_arr = []
+    if len(member_guid_arr) < 1:
+        return username_arr
+    #TODO: can we do this in one request instead of iterating over each username
+    for user_id in member_guid_arr:
+        print("User_id being passed in request", user_id)
+        try:
+            response = requests.get(f"{URL_ROOT}/legacy-services/rest/users/{user_id}",
```
jceaser, line 61: URL_ROOT is just for legacy services or should this also be for all CMR?

jceaser: CMR_INGEST_ENDPOINT is a whole URL, but here your just a partial, I like the partial approach, if the prefix is the same for both, then both should use the same var

legacy-migration/providers/migrate.py
```diff
@@ -0,0 +1,312 @@
...
+        organizations.append(organization)
+    return organizations
+
+def migrate_administrators(provider_owner_guid, provider_id):
+    """Parses admin's fields between old-style format into new schema"""
+    all_the_groups_specific_provider = get_groups_by_provider(provider_owner_guid)
+    member_guids_specific_provider = get_all_provider_guids(all_the_groups_specific_provider)
+    # TODO: Iron out the pieces
+    if member_guids_specific_provider is None:
+        print('🚀 Could not retrieve the member guids for ', provider_id)
```
jceaser, line 237: if this went to a log, we could find them all easy

legacy-migration/providers/migrate.py
```diff
@@ -0,0 +1,312 @@
...
+# It should not be expected this version or schema directories will change
+# within the lifecycle of this migration
+metadata_specification = {
+            "Name": "provider",
+            "Version": "1.0.0",
+            "URL": "https://cdn.earthdata.nasa.gov/schemas/provider/v1.0.0"}
+
+def ingest_provider(provider_metadata, provider_id):
+    """Attempts to ingest metadata into CMR (locally)"""
+    response = requests.post("http://localhost:3002/providers/",
```
jceaser, line 32: why is localhost hard coded?

legacy-migration/providers/migrate.py
```diff
@@ -0,0 +1,312 @@
...
+        ingest_provider(json_data, provider_id)
+
+        #TODO will remove below useful for debugging
+        # provider_id_list.append(json_data)
+        provider_migration_count += 1
+
+    # Migration Complete
+    print(f'The total number of providers migrated {provider_migration_count}')
+
+# Main:
```
jceaser, line 307: this is not the "standard" way a main method is called:

```
if __name__ == "__main__":
    main()
```

legacy-migration/providers/migrate.py
```diff
@@ -0,0 +1,312 @@
...
+
+    for provider in json_response:
+        provider_id = provider['provider']['provider_id']
+        print(f"Migrating {provider_id} providers")
+        description_of_holdings = provider['provider']['description_of_holdings']
+        
+        # Remove any "\" characters in the description of holdings string
+        # TODO item of discussion is the length that we will allow for DOH
+        # description_of_holdings=description_of_holdings.replace("\\","")
+        re.sub('[^A-Za-z0-9]+', '', description_of_holdings)
```
jceaser, line 262: \w is all of these plus _

legacy-migration/providers/migrate.py
```diff
@@ -0,0 +1,312 @@
...
+        organization["Roles"] = [DEFAULT_ORGANIZATION_ROLE]
+        organizations.append(organization)
+    return organizations
+
+def migrate_administrators(provider_owner_guid, provider_id):
+    """Parses admin's fields between old-style format into new schema"""
+    all_the_groups_specific_provider = get_groups_by_provider(provider_owner_guid)
+    member_guids_specific_provider = get_all_provider_guids(all_the_groups_specific_provider)
+    # TODO: Iron out the pieces
+    if member_guids_specific_provider is None:
```
jceaser, line 236: so maybe an else statement to print out all the found admins to a log file, I'm just now thinking it would be nice to have an output of all the people who have been selected to be an admin across the entire holdings. This would then be used to verify that we actually know all these people and the accounts are active.

legacy-migration/providers/migrate.py
jceaser, line 32: Set both CMR client and browser agent in all scripts you write, they can be the same value.

jceaser (changes requested):

legacy-migration/providers/migrate.py
```diff
@@ -0,0 +1,312 @@
...
+        data = {}
+        data["MetadataSpecification"]= metadata_specification
+        data["ProviderId"] = provider_id
+        data["DescriptionOfHolding"] = description_of_holdings
+        data["Organizations"] = organizations
+        data["Administrators"] = admin_usernames
+        data["ContactPersons"] = contact_persons
+
+        # Dump the new metadata schema
+        json_data = json.dumps(data)
```
DanielJDufour, line 291: could consider dumping with some indentation to improve readability
```python
json.dumps(data, indent=4)
```
https://stackoverflow.com/questions/12943819/how-to-prettyprint-a-json-file

DanielJDufour:
left some comments, just wanted to add that I really appreciate the use of `dict.get('key')` vs `dict['key']`.  I really like the if statements and defensive safeguards to prevent dict lookup and other errors.

also, I never knew how easy you could time things using the time module.  (I've previously done a much more cumbersome approach of using datetime).  Thanks for teaching me that :-)

legacy-migration/providers/migrate.py
eudoroolivares2016: They're the same that is a good point I'll reuse it

legacy-migration/providers/migrate.py
```diff
@@ -0,0 +1,312 @@
...
+        address_metadata["StateProvince"] = state
+        address_metadata["PostalCode"] = postal_code
+
+    # Look for street *
+    provider_street_keys = {key: val for key, val in address.items()
+       if key.startswith('street')}
+    # print('🚀 ~ file: migrate.py:115 ~ street_list:', str(provider_street_keys.values()))
+
+    # If street addresses were in the record add them to the new metadata
+    if len(provider_street_keys.values()) > 0:
```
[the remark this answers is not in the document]
eudoroolivares2016: I think I need the .values() call but, I should remove the unnecessary `len` logic

legacy-migration/providers/migrate.py
```diff
@@ -0,0 +1,312 @@
...
+            contact_mechanisms.append(email_contact)
+
+        if contact.get("phones"):
+            phone_numbers = parse_phones(contact)
+            for phone_number in phone_numbers:
+                phone_contact = {"Type": "Telephone", "Value": phone_number}
+                # Use the same contact mechanism object as emails
+                contact_mechanisms.append(phone_contact)
+
+        if len(addresses) > 0 or len(contact_mechanisms) > 0 or len(phone_contacts) > 0 :
```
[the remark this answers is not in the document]
eudoroolivares2016: No because if ANY of them exist it will create that `ContactInformation` field for the new schmea

legacy-migration/providers/migrate.py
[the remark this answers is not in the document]
eudoroolivares2016: Yea I did change what I was doing with the doh. What I was really trying to do was get rid of non-ascii chars and found a better way to do so

jceaser (approved):

dpesall (approved):

legacy-migration/providers/migrate.py
```diff
@@ -0,0 +1,312 @@
...
+import os
+import xml.etree.ElementTree as ET
+import re
+import time
+import sys
+import requests
+
+URL_ROOT = "https://cmr.sit.earthdata.nasa.gov"
+
+access_token = os.environ.get('SIT_SYSTEM_TOKEN')
```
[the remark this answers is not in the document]
eudoroolivares2016: Added these as flags to the script now

legacy-migration/providers/migrate.py
```diff
@@ -0,0 +1,418 @@
...
+import xml.etree.ElementTree as ET
+import re
+import time
+import string
+import sys
+import requests
+import logging
+import validators
+
+# URL_ROOT = "https://cmr.sit.earthdata.nasa.gov"
```
eudoroolivares2016, line 14: remove these

legacy-migration/providers/migrate.py
```diff
@@ -0,0 +1,418 @@
...
+
+url_map = { "sit": ".sit", "uat":".uat", "ops":"", "prod": ""}
+token_map = { "sit": "SIT_SYSTEM_TOKEN", "uat":"UAT_SYSTEM_TOKEN", "ops":"OPS_SYSTEM_TOKEN", "prod": "PROD_SYSTEM_TOKEN"}
+
+# access_token = os.environ.get('SIT_SYSTEM_TOKEN')
```
eudoroolivares2016, line 19: and this

PR #1885, opened 2023-04-10
CMR-8910 (3) Adding DELETE to EFS

TylerHeald1:
Fixing an issue present with cascading deletes (as found when a collection is 'deleted' and granules associated have a cascading delete done on them)

### DIFF
--- a/metadata-db-app/src/cmr/metadata_db/data/oracle/sql_helper.clj
+++ b/metadata-db-app/src/cmr/metadata_db/data/oracle/sql_helper.clj
@@ -13,6 +13,14 @@
    [cmr.common.util :as util])
   (:import cmr.oracle.connection.OracleStore))
 
+;;----------- HELPER FOR EFS DELETE --------------
+(defn efs-concept-delete-helper
+  [revision-concept-tuple]
+  (let
+   [{:keys [concept_id
+            revision_id]} revision-concept-tuple]
+    [concept_id revision_id]))
+
 (defn find-params->sql-clause
   "Converts a parameter map for finding concept types into a sql clause for inclusion in a query. The type
   of value determines the nature of the clause. If the value for a parameter is sequential then a clause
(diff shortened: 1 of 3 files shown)


metadata-db-app/src/cmr/metadata_db/data/oracle/sql_helper.clj
```diff
@@ -53,6 +65,14 @@
...
         table (ct/get-table-name provider concept-type)
+        get-stmt (su/build (select [:concept-id :revision-id]
+                                   (from table)
+                                   (where (find-params->sql-clause params))))
+        get-values (when (not (= "efs-off" (efs-config/efs-toggle)))
+                     (j/query db get-stmt))
         stmt (su/build (delete table
-                         (where (find-params->sql-clause params))))]
+                               (where (find-params->sql-clause params))))]
+    (when (not (= "efs-off" (efs-config/efs-toggle)))
```
eudoroolivares2016, line 75: nit; maybe this could be `(when = "efs-on" (efs-config/efs-toggle))` but, as long as it is consistent throughout the other toggles that is more important

metadata-db-app/src/cmr/metadata_db/data/oracle/sql_helper.clj
```diff
@@ -5,10 +5,22 @@
...
+   [cmr.common.log :refer [debug error info trace warn]]
    [cmr.metadata-db.data.oracle.concept-tables :as ct]
-   [cmr.oracle.sql-utils :as su :refer [insert values select from where with order-by desc delete as]])
+   [cmr.oracle.sql-utils :as su :refer [insert values select from where with order-by desc delete as]]
+   [cmr.efs.config :as efs-config]
+   [cmr.efs.connection :as efs]
+   [cmr.common.util :as util])
   (:import cmr.oracle.connection.OracleStore))
 
+;;----------- HELPER FOR EFS DELETE --------------
```
eudoroolivares2016, line 16: nit; maybe this function could just have a doc-string that explains why it's needed instead of a comment up here

metadata-db-app/src/cmr/metadata_db/data/oracle/sql_helper.clj
TylerHeald1: It's like this because there are two options for efs begin on, 'efs-on' and 'efs-only'. Originally I checked for multiple values, but this not check was more concise without really changing function

isja17 (approved):

PR #1896, opened 2023-05-05
CMR-9145: Updating Jetty, Ring, and JSON Validator library versions

dpesall:


common-lib/project.clj
```diff
@@ -41,17 +41,17 @@
...
+                 [org.eclipse.jetty/jetty-servlets "9.4.48.v20220622"]
+                 [org.eclipse.jetty/jetty-util "9.4.48.v20220622"]
                  ;; load jts core lib first to make sure it is available for shapefile integration,
                  ;; otherwise ES referenced 1.15.0 version will be mistakenly picked for shapefile
                  [org.locationtech.jts/jts-core "1.18.2"]
                  [org.ow2.asm/asm "7.0"]
                  [potemkin "0.4.5"]
-                 [ring/ring-core "1.9.2"]
-                 [ring/ring-jetty-adapter "1.9.2"]
+                 [ring/ring-core "1.9.6"]
```
eudoroolivares2016, line 53: I think we'll want to synch all of the ring upgrades. Ring is also being imported by `system-int-test/project.clj`, 'search-app/project.clj', 'metadata-db-app/project.clj'

common-lib/project.clj
```diff
@@ -41,17 +41,17 @@
                  [org.clojure/tools.nrepl "0.2.13"]
                  [org.clojure/tools.reader "1.3.2"]
                  ;;these dependencies should be updated in tandem with the ring dependiences below
```
eudoroolivares2016, line 43: I actually think that we want to have the ` 9.4.51.v20230217` version on here. I think this from going to https://clojars.org/ring/ring-jetty-adapter/versions/1.10.0 and then the dependency on the right side is that value. Either way we should add a comment here that explains how to find the matching dependency that existing comment is not very helpful.

eudoroolivares2016, line 43: nit; know it's not your comment but, change dependiences to dependencies

dpesall: The version I upgraded to is `1.9.6`, not `1.10.0`, which depends on `9.4.48.v20220622`

eudoroolivares2016: You are correct should have looked more closely I just went to latest ty!

jaybarra (approved):

daniel-zamora (approved):

PR #1901, opened 2023-05-15
CMR-9109 adds caching for urs client token

daniel-zamora:


transmit-lib/src/cmr/transmit/urs.clj
```diff
@@ -36,9 +38,22 @@
...
-;; Request functions
+;; URS cache functions
+
+(def urs-cache-key
+  "The cache key for a urs cache."
+  :urs)
+
+(def URS_CACHE_TIME
+  "The number of milliseconds CMR client bearer token information will be cached."
+  (* 24 60 1000))
```
dpesall, line 49: Should this be `(* 24 60 60 1000))` instead? Currently it is 24 minutes instead of 24 hours.

zimzoom: i second drew's comment. 1 hour = 60 minutes = 60 × 60 seconds = 3600 seconds = 3600 × 1000 milliseconds = 3,600,000 ms.
so milliseconds to hours should be `60 * 60 * 1000`

dpesall (approved):

PR #1903, opened 2023-05-19
CMR-8872: Fixed a in memory db issue for variable collection association

sxu123:
The related tests already exist. It's just that they are run with real db, instead of in-memory db in bamboo and have always run successfully.  With this fix,  I could run the tests successfully locally using in-memory db.

### DIFF
--- a/indexer-app/src/cmr/indexer/services/index_service.clj
+++ b/indexer-app/src/cmr/indexer/services/index_service.clj
@@ -439,11 +439,6 @@
    so that it aligns with what index-concept :default expects."
   [context concept]
   (let [combined-assocs (meta-db/find-associations context concept)
-        ;; in-memory db returns variable-collection association in an additional [].
-        combined-assocs (map #(if (and (sequential? %) (= 1 (count %)))
-                                (first %)
-                                %)
-                             combined-assocs)
         grouped-assocs (group-by :concept-type combined-assocs)
         new-key-map (into {} (map #(create-association-map-keys %) (keys grouped-assocs)))]
     (cset/rename-keys grouped-assocs new-key-map)))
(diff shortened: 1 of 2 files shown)


indexer-app/src/cmr/indexer/services/index_service.clj
```diff
@@ -439,6 +439,11 @@
    so that it aligns with what index-concept :default expects."
   [context concept]
   (let [combined-assocs (meta-db/find-associations context concept)
+        ;; in-memory db returns variable-collection association in an additional [].
+        combined-assocs (map #(if (and (sequential? %) (= 1 (count %)))
```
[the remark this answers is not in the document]
eereiter: Keeping the data structures the same going to/from a storage device might be harder to implement, but I think it is a much better solution, then to sprinkle fixes in other classes.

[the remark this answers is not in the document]
sxu123: Absolutely agreed.  It's just that I encountered some weird behavior.  Will continue working on this. Thanks.

metadata-db-app/src/cmr/metadata_db/data/memory_db.clj
```diff
@@ -302,7 +302,7 @@
   involve the passed in concept id."
   [concepts concept-id]
   (util/remove-nils-empty-maps-seqs
-   (merge
```
sxu123, line 305: There are two issues with using merge: 
1. (merge [] [1 2 3]) returns [[1 2 3]], rather than [1 2 3].
2. (merge [1 2 3] [4 5 6]) returns [1 2 3 [4 5 6]], rather than [1 2 3 4 5 6].
We need both situations to work correctly.

eereiter: 1. (reduce conj [][1 2 3])
2.(reduce conj [1 2 3] [4 5 6])

sxu123: concat works fine too. Thanks.

eereiter (approved):

jaybarra (approved):

PR #1905, opened 2023-05-24
CMR-8973 Updated docs to include Launchpad tokens

isja17:


tbs1979 (approved):

eudoroolivares2016:
For consistency we should probably remove the "Bearer" prefix in the ingest.md files for the generic documents. Search `Ingest.md` in the project to find. You can choose to ignore `service-option` and `service-entries` documentation because we plan to deprecate those CMR concepts as they were not needed for migration out of legacy-services

isja17:
> 

We want

PR #1906, opened 2023-05-24
CMR-9160: Create a library for DynamoDB

TylerHeald1:
Create a library that has basic configuration info and connection info and stubbed functions that will be needed for interacting with DynamoDB


dynamo-lib/src/cmr/dynamo/config.clj
```diff
@@ -0,0 +1,11 @@
+(ns cmr.dynamo.config
+  "Contains functions for retrieving important DynamoDB details from environment variables"
+  (:require [cmr.common.config :as cfg :refer [defconfig]]))
+
+(defconfig dynamo-table
+   "DynamoDB table for meta-metadata items"
+   {:default "meta-metadata"})
+
+(defconfig dynamo-toggle
+  "Three-way toggle for DynamoDB functionality. 'dynamo-off' uses only Oracle, 'dynamo-on' uses both Oracle and EFS, and 'dynamo-only' uses only EFS"
```
eereiter, line 10: Is this comment correct?  dynamo-off means that only oracle is used and not EFS and/or Oracle?
dynamo-on - shouldn't dynamo be used and not oracle? or are all three being used?
dynamo-only uses only EFS?  not Dynamo?

dynamo-lib/src/cmr/dynamo/connection.clj
```diff
@@ -0,0 +1,35 @@
+(ns cmr.dynamo.connection
+  "Contains functions for interacting with the DynamoDB storage instance"
+  (:require [cmr.common.config :as cfg :refer [defconfig]]
+            [taoensso.faraday :as far]))
```
eereiter, line 4: why is taoensso.faraday in here?  it isn't being used.

dynamo-lib/src/cmr/dynamo/config.clj
TylerHeald1: Oh, yeah that's wrong. I'd copied some stuff from elsewhere and hadn't realized that part hadn't been updated.

dynamo-lib/project.clj
```diff
@@ -0,0 +1,64 @@
...
+
+            ;; Linting aliases
+            "kibit" ["do"
+                     ["with-profile" "lint" "shell" "echo" "== Kibit =="]
+                     ["with-profile" "lint" "kibit"]]
+            ;; Eastwood needs special handling with libs that include oracle
+            ;; drivers in the deps, in particulear:
+            ;;   java.lang.ClassNotFoundException: oracle.dms.console.DMSConsole
+            "eastwood" ["with-profile" "lint" "eastwood"
+                        "{:namespaces [:source-paths] :exclude-namespaces [cmr.oracle.connection]}"]
```
eereiter, line 58: Since this isn't an oracle project do we need the special handling?

eereiter (changes requested):
The biggest issue is the dynamo description.

dynamo-lib/project.clj
TylerHeald1: Nope. Project file was largely copied from elsewhere because it was quick to do

dynamo-lib/src/cmr/dynamo/connection.clj
```diff
@@ -0,0 +1,34 @@
...
+
+(def connection-options
```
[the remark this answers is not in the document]
eereiter: I am not quite understanding your comment. Is the endpoint adjustable without having to change the software locally?

zimzoom (approved):

dynamo-lib/src/cmr/dynamo/connection.clj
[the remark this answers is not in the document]
TylerHeald1: I put the endpoint itself as a defconfig in the config.clj so that it can be adjusted by using an environment variable. It would still need a restart/redeployment to pick up the change though

eereiter (approved):

PR #1907, opened 2023-05-24
CMR-9188: Refactoring MDB calls for Access Control Permissions endpoint

dpesall:
Access Control Permissions was not very efficient in how it processes large lists of records it gets from Metadata DB. Requests to the permissions endpoint was making redundant calls to metadata-db, resulting in slow response times for requests with a large number of granules.

### DIFF
--- a/acl-lib/src/cmr/acl/core.clj
+++ b/acl-lib/src/cmr/acl/core.clj
@@ -170,8 +170,10 @@
 
 (defn get-acl-enforcement-collection-fields-fn
   [concept]
-  {:AccessConstraints (umm-spec-core/parse-concept-access-value concept)
-   :TemporalExtents (umm-spec-core/parse-concept-temporal concept)})
+  {:AccessConstraints (when-not (= "" (:metadata concept))
+                        (umm-spec-core/parse-concept-access-value concept))
+   :TemporalExtents (when-not (= "" (:metadata concept))
+                        (umm-spec-core/parse-concept-temporal concept))})
 
 (defn get-acl-enforcement-collection-fields
   [context concept]
--- a/acl-lib/src/cmr/acl/core.clj
+++ b/acl-lib/src/cmr/acl/core.clj
@@ -13,20 +13,19 @@
    [cmr.common.services.errors :as errors]
    [cmr.common.util :as util]
    [cmr.transmit.access-control :as access-control]
-   [cmr.transmit.config :as transmit-config]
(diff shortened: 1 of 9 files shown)


acl-lib/src/cmr/acl/core.clj
```diff
@@ -23,6 +23,10 @@
   "Flag that indicates if we accept the 'Echo-Token' header."
   {:default true :type Boolean})
 
+(def collection-field-constraints-cache-key
+  "The cache key for a urs cache."
```
jaybarra, line 27:
```suggestion
  "The cache key for an URS cache."
```

acl-lib/src/cmr/acl/core.clj
```diff
@@ -154,6 +162,11 @@
   []
   (mem-cache/create-in-memory-cache :ttl {} {:ttl TOKEN_IMP_CACHE_TIME}))
 
+(defn create-access-constraints-cache
+  "Creates a cache for access constraint mapping."
```
jaybarra, line 166:
```suggestion
  "Create a cache for access constraint mapping."
```

acl-lib/src/cmr/acl/core.clj
```diff
@@ -154,6 +162,11 @@
...
+  []
+  (mem-cache/create-in-memory-cache :ttl {} {:ttl CONCEPT_MAP_CACHE_TIME}))
```
jaybarra, line 168: Since this is an in-memory cache, are they are concerns with multiple requests coming in to different instances of access-control where only some have cached data?

acl-lib/src/cmr/acl/core.clj
```diff
@@ -140,6 +144,10 @@
   "The cache key for the token to subscription management permission cache."
   :token-smp)
 
+(def CONCEPT_MAP_CACHE_TIME
+  "The number of milliseconds token information will be cached."
```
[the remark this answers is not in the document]
dpesall: I initially made a separate cache time for this since I wasn't sure how long I wanted to cache for, but I ended up leaving it at the same duration as the other time. I can remove this one and use the other cache time value with the same 5-minute duration, but in the event where we want to change how long this is cached for, we'd need to separate them out again.

acl-lib/src/cmr/acl/core.clj
dpesall: Fixed

jaybarra (approved):

daniel-zamora (approved):

PR #1910, opened 2023-05-25
CMR-9200 CMR-9125

daniel-zamora:
This PR covers two issues.  First is geotools snyk issues cropping up in service-bridge can be avoided by removing the dependency altogether as it is not used.  Second, we have some extraneous calls to legacy services for token-info that needed to be rerouted to EDL.   We first do a local verification using the jwt public key from EDL, if that fails then we try launchpad through EDL's api.


other/cmr-exchange/authz/src/cmr/authz/token.clj
```diff
@@ -58,27 +71,73 @@
...
+        (if (:errors bearer-token)
+          (deliver (promise) bearer-token)
+          (do
+            (log/trace "retrieved cmr client token= " bearer-token)
+            (log/trace "user info not found for jwt token, attempting to authenticate user via launchpad")
+            (request/async-post
+             url
+             (-> {:body (str "token=" token)}
+                 (http/add-options)
+                 (request/add-token-header (str "Bearer " (:body bearer-token))))
```
jaybarra, line 131: Will this ever possibly receive launchpad tokens that don't use `Bearer `?

jaybarra (approved):

other/cmr-exchange/authz/src/cmr/authz/token.clj
daniel-zamora: That is the cmr client token, so no.

PR #1911, opened 2023-05-31
CMR-9159: Implement Feature Toggle for DynamoDB

TylerHeald1:
Part of this work was a realization that the toggle and return values from EFS are really subsumed by toggle and return values from the meta metadata store.


metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
```diff
@@ -376,46 +377,58 @@
...
+         dynamo-concept-get (when (not (= "dynamo-off" (dynamo-config/dynamo-toggle)))
+                              (util/time-execution
+                               ()))]
+     (when (not (= "dynamo-off" (dynamo-config/dynamo-toggle)))
        (info "Runtime of EFS get-concept: " (first efs-concept-get) " ms.\nValue gotten from EFS get-concept: " (second efs-concept-get)))
-     (when (not (= "efs-only" (efs-config/efs-toggle)))
+     (when (not (= "dynamo-only" (dynamo-config/dynamo-toggle)))
        (info "Runtime of Oracle get-concept: " (first oracle-concept-get) " ms."))
+     (when (not (= "dynamo-off" (dynamo-config/dynamo-toggle)))
+       (info "Runtime of DynamoDB get-concept: " (first dynamo-concept-get) " ms.\nValue gotten from DynamoDB get-concept: " (second dynamo-concept-get)))
```
jaybarra, line 389: Nice use of `gotten`, don't see it used correctly often.

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
```diff
@@ -376,46 +377,58 @@
...
                                                                                      (where `(and (= :concept-id ~concept-id)
-                                                                                                  (= :revision-id ~revision-id)))))))))]
-       (when (not (= "efs-off" (efs-config/efs-toggle)))
-         (info "Runtime of EFS get-concept: " (first efs-concept-get) " ms.\nValue gotten from EFS get-concept for concept-id " concept-id " and revision-id " revision-id ": " (second efs-concept-get)))
-       (when (not (= "efs-only" (efs-config/efs-toggle)))
+                                                                                                  (= :revision-id ~revision-id)))))))))
+           dynamo-concept-get (when (not (= "dynamo-off" (dynamo-config/dynamo-toggle)))
+                                (util/time-execution
+                                 ()))]
+       (when (not (= "dynamo-off" (dynamo-config/dynamo-toggle)))
```
jaybarra, line 412: All of the `(not (=` can be replaced with `(not=`

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
```diff
@@ -364,10 +365,10 @@
 
 (defn get-concept
```
jaybarra, line 366: This function is getting a bit long, and is a candidate for being converted to a multi-method based on the state of the dynamo-config. To be functional we should properly pass in the config, as opposed to having a global state object change the behavior of the function.

jaybarra (changes requested):

jaybarra:
I'm generally concerned by EFS logic being done in the oracle namespace. If possible we should make the storage calls through an interface, and not care if it's being done under the hood as Oracle or EFS.
How much of the calls could be made the oracle part was duplicated into a `efs` namespace that mirrored the oracle calls? Or could it be generalized even further to `db` and depend on what jar was being loaded?

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
```diff
@@ -376,46 +377,58 @@
...
                                                                                      (from table)
                                                                                      (where `(and (= :concept-id ~concept-id)
-                                                                                                  (= :revision-id ~revision-id)))))))))]
-       (when (not (= "efs-off" (efs-config/efs-toggle)))
-         (info "Runtime of EFS get-concept: " (first efs-concept-get) " ms.\nValue gotten from EFS get-concept for concept-id " concept-id " and revision-id " revision-id ": " (second efs-concept-get)))
-       (when (not (= "efs-only" (efs-config/efs-toggle)))
+                                                                                                  (= :revision-id ~revision-id)))))))))
+           dynamo-concept-get (when (not (= "dynamo-off" (dynamo-config/dynamo-toggle)))
+                                (util/time-execution
+                                 ()))]
```
jaybarra, line 411: Is this an extra internal `()`?

TylerHeald1: Thats for the next ticket of actually implementing the functionality.

TylerHeald1:
@jaybarra To clarify some of the design considerations in the prototype: long-term the functionality is going to be in a new DataStore record (Oracle functionality is currently in a OracleStore record that extends DataStore). That was originally the intention here, but a further layer of abstraction is needed to make that an effective solution (there are multiple services that make use of, and directly reference OracleStore, so it would need to be abstracted such that there is one specified data store type that other services access and also implementing a DataStore for the service would essentially require making the fully finished data store interface), and Metadata-DB isn't set up in a way that multiple DataStore records could be used at the same time (which is desired here for prototype testing). As there are some elements of further abstracting for the data store that aren't currently understood and we are looking to test and decide on a meta-metadata store by the end of the PI, the decision was made to currently run the functionality alongside Oracle functionality. Let me know if there are still concerns or questions about that.

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
eereiter: instead of a multi-method could we use cond or condp or case to call other functions?  Or are you being more generic?  I keep reading that multi-methods are much slower.

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
```diff
@@ -376,46 +377,58 @@
...
-       (when (not (= "efs-only" (efs-config/efs-toggle)))
+                                                                                                  (= :revision-id ~revision-id)))))))))
+           dynamo-concept-get (when (not= "dynamo-off" (dynamo-config/dynamo-toggle))
+                                (util/time-execution
+                                 ()))]
+       (when (not= "dynamo-off" (dynamo-config/dynamo-toggle))
+         (info "Runtime of EFS get-concept: " (first efs-concept-get) " ms."))
+       (when (not= "dynamo-only" (dynamo-config/dynamo-toggle))
          (info "Runtime of Oracle get-concept: " (first oracle-concept-get) " ms."))
+       (when (not= "dynamo-off" (dynamo-config/dynamo-toggle))
```
eereiter, line 416: This when is identical to the when on line 412.  can you put the info's together or is one of the when clauses wrong?

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
```diff
@@ -376,46 +377,58 @@
...
+                                                                                     (order-by (desc :revision-id)))))))))
+         dynamo-concept-get (when (not= "dynamo-off" (dynamo-config/dynamo-toggle))
+                              (util/time-execution
+                               ()))]
+     (when (not= "dynamo-off" (dynamo-config/dynamo-toggle))
        (info "Runtime of EFS get-concept: " (first efs-concept-get) " ms.\nValue gotten from EFS get-concept: " (second efs-concept-get)))
-     (when (not (= "efs-only" (efs-config/efs-toggle)))
+     (when (not= "dynamo-only" (dynamo-config/dynamo-toggle))
        (info "Runtime of Oracle get-concept: " (first oracle-concept-get) " ms."))
+     (when (not= "dynamo-off" (dynamo-config/dynamo-toggle))
```
eereiter, line 388: The when on line 388 is the same as the when on line 384. Can you put the two info's together or is one of the whens wrong?

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
```diff
@@ -431,18 +444,23 @@
...
+                                                 (su/query conn stmt)))))))
+          dynamo-concepts-get (when (not= "dynamo-off" (dynamo-config/dynamo-toggle))
+                                (util/time-execution
+                                 ()))]
+      (when (not= "dynamo-off" (dynamo-config/dynamo-toggle))
+        (info "Runtime of EFS get-concepts: " (first efs-concepts-get) " ms."))
+      (when (not= "dynamo-only" (dynamo-config/dynamo-toggle))
         (info "Runtime of Oracle get-concepts: " (first oracle-concepts-get) " ms."))
+      (when (not= "dynamo-off" (dynamo-config/dynamo-toggle))
+        (info "Runtime of DynamoDB get-concepts: " (first dynamo-concepts-get) " ms."))
```
eereiter, line 456: same comment here - the when clauses on 456 and 451 are the same, should they be or can we combine the messages into 1 when?

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
TylerHeald1: During testing, all parts of this function are actually going to be called at the same time. I'm not aware if multi-methods can make that work. Also, I would like to clarify that the current structure of these functions is for prototype testing purposes, this is not the state it will be in when merged into master

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
TylerHeald1: Good catch, I can absolutely combine those

eereiter (approved):

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
```diff
@@ -540,19 +563,24 @@
                            tmp.concept_id = t1.concept_id AND
                            tmp.revision_id = t1.revision_id)"
                           table)]
-            efs-delete (when (not (= "efs-off" (efs-config/efs-toggle)))
+            efs-delete (when (not= "dynamo-off" (dynamo-config/dynamo-toggle))
```
jaybarra, line 566: `efs-delete` and `dynamo-delete` are true on the same condition, so they could be merged

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
```diff
@@ -540,19 +563,24 @@
...
         (when efs-delete
           (info "Runtime of EFS force-delete-concepts: " (first efs-delete) " ms."))
         (when oracle-delete
           (info "Runtime of Oracle force-delete-concepts: " (first oracle-delete) " ms."))
+        (when dynamo-delete
+          (info "Runtime of DynamoDB force-delete-concepts: " (first dynamo-delete) " ms."))
         (if oracle-delete
           (second oracle-delete)
-          (second efs-delete))))))
+          (second dynamo-delete))))))
```
jaybarra, line 583: This logic needs to be distinct between backends. The various backends needs to be broken out into multiple functions.

jaybarra (changes requested):

jaybarra (commented):
This very much breaks principles of single responsibility and the control flow is not easy to follow.

I will still recommend moving all of the efs/dynamo logic into functions that runs at the end of any of the existing functions that takes any necessary input needed to complete testing of the underlying data store interactions.

johnwteague:
You collapsed the toggles for EFS and DynamoDB into a single switch because you are always going to interact with EFS if you are interacting with DyDB and vice-versa?

jaybarra (approved):

TylerHeald1:
Correct. EFS will always be used with DynamoDB and vice-versa.

jasongilman:
Hello,

I'm interested in the activity I've seen around adding the EFS capability. Are there any descriptions anywhere of the intent of the switch to EFS and dynamo DB? Is the idea to switch away from Oracle to use EFS for storage of metadata concepts?

I'm also curious about the drivers for that. Is cost of the Oracle RDS the issue? Was any consideration given to switching to Aurora PostgreSQL?

TylerHeald1:
@jasongilman There are a couple places in the Earthdata wiki space that document the reasoning and action around moving towards EFS capability. Primarily, there was an analysis done years ago and more recent documents in the Ops and Infrastructure space about the work that's been done and that will happen.
The intent is indeed to store metadata concepts in EFS. They take up by far the largest chunk of DB space and it was determined in the EFS analysis that it would be cheaper with potentially better CRUD time. That being said, EFS is inherently not queryable, so there is a need for another system that stores the metadata surrounding the metadata files (meta-metadata is what we've been referring to that as)
What we realized was that there had not actually been an analysis done on what that secondary data store for the meta-metadata would be. As such, we have been requested to run some tests integrated with CMR to get some numbers and determine a secondary datastore to move forward with. We are going to test DynamoDB (which this PR is for) and Aurora PostgreSQL.

TylerHeald1:
@jasongilman I checked in with others, and I don't know that I can actually put potentially internal only wiki links out in public unfortunately. However, I can probably answer any specific questions you might have. You can also reach out to me directly at tyler.heald@nasa.gov

PR #1916, opened 2023-06-08
CMR-9218: Updating Cucumber tests to allow ingests against any given provider

dpesall:


system-validation-test/features/ingest_basic.feature
```diff
@@ -11,7 +11,9 @@ Feature: Basic Ingest API calls
     Scenario: Using an invalid Bearer token value
       Given I am ingesting a "Collection"
       And I clear headers
-      And I add url path "providers/CMR_ONLY/collections/TestCollection001"
+      And I use the "providers" endpoint
+      And I add the environment variable "CMR_TEST_PROV" to the url path
+      And I add url path "/collections/TestCollection001"
```
eudoroolivares2016, line 16: If its just one line, I know there was discussion about trying to have internal CMR calls use a client-id or a user-agent.

PR #1924, opened 2023-06-23
cmr-9165: add aurora lib with connection config

zimzoom:
DATASTORE PROTOTYPE -- NOT GOING TO MASTER -- SEEKING 1 APPROVAL ONLY (FROM TYLER H)

Aurora Lib with needed connection stuff, a pinging utility, and protocol stubs


TylerHeald1 (approved):

jasongilman:
As a past contributor to the CMR it's great to see Aurora being considered. I always thought it made sense to switch from Oracle RDS to Aurora. Given the CMR started on-premises it made sense at the time to use the database that was available, Oracle. I believe that there were no Oracle specific features used in Metadata DB so Aurora would be a natural fit.

PR #1934, opened 2023-07-14
CMR-9187 - Sending provider events to ordering

jceaser:
Calling CRM Ordering when a provider is created or updated so that it can resync the copy of providers that it has

### DIFF
--- a/transmit-lib/src/cmr/transmit/ordering.clj
+++ b/transmit-lib/src/cmr/transmit/ordering.clj
@@ -1,0 +1,59 @@
+(ns cmr.transmit.ordering
+  "Handle all communications to the CMR-Ordering application, a graphql app which
+   processes orders for Earthdata Search and CMR. This application needs to know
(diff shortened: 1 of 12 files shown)


transmit-lib/src/cmr/transmit/ordering.clj
```diff
@@ -0,0 +1,59 @@
+(ns cmr.transmit.ordering
+  "Handle all communications to the CMR-Ordering application, a graphql app which
+   processes orders for Earthdata Search and CMR. This application needs to know
+   when CMR has changed a proveder so that it can download the latest provider
```
zimzoom, line 4:
```suggestion
   when CMR has changed a provider so that it can download the latest provider
```

jceaser:
After quite some time, I finally was able to get a green build on my branch using a new build plan.

TylerHeald1 (approved):

zimzoom (approved):

PR #1955, opened 2023-08-14
CMR-9261: When delete a draft, remove it from database, not tombstone it…

sxu123:
… it.

### DIFF
--- a/ingest-app/src/cmr/ingest/api/generic_documents.clj
+++ b/ingest-app/src/cmr/ingest/api/generic_documents.clj
@@ -216,8 +216,10 @@
     (ingest-document request-context concept headers)))
 
 (defn delete-generic-document
-  "Deletes a generic document in elasticsearch and creates a tombstone in the database. Returns a 201
-   if successful with the concept id and revision number. A 404 status is returned if the concept has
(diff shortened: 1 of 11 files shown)


ingest-app/src/cmr/ingest/api/generic_documents.clj
```diff
@@ -230,7 +230,9 @@
              request-context :update :provider-object provider-id)
             (acl/verify-provider-context-permission
              request-context :read :provider-object provider-id))]
-    (api-core/delete-concept concept-type provider-id native-id request)))
+    (if (is-draft-concept? request)
+      (api-core/delete-draft concept-type provider-id native-id request)
```
eudoroolivares2016, line 234: Does the `delete-draft` function in ingest/core leave a tombstone? If it does not we should update the doc-string in this function with something like `Deletes a generic document in elasticsearch and creates a tombstone in the database if the concept is not a draft concept.`

eudoroolivares2016 (approved):

ingest-app/src/cmr/ingest/api/generic_documents.clj
sxu123: updated the doc string.

eereiter (approved):

PR #1962, opened 2023-08-24
cmr-9193: add option to not use metadata column in aurora postgres db

zimzoom:
this is for the datastore prototype. this is not going to master. 

the intention is to simply not write to the metadata blob column in concept tables (and thus not read it back while doing prototype runs), as a quick option creation rather than complete column removal.


metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
```diff
@@ -594,15 +594,20 @@
                            (string/join "," cols)
                            seq-name
                            (string/join "," (repeat (count values) "?"))) 
-              ;; replace instances of 'false' with 0 and 'true' with 1 for Postgres 'deleted' column
-              pg-values (map #(if (boolean? %) (if % 1 0) %) values)
+              ;; if set to not use metadata column in Aurora database, then remove it and replace with nil
+              pg-values (if (aurora-config/aurora-metadata?)
+                          values
+                          (map #(if (or (string? %) (number? %) (boolean? %) (nil? %)) % nil) values))
```
TylerHeald1, line 600: Just to confirm, the metadata field is a byte array at this point and 'string?' doesn't return true on it?

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
```diff
@@ -594,15 +594,20 @@
                            (string/join "," cols)
                            seq-name
                            (string/join "," (repeat (count values) "?"))) 
-              ;; replace instances of 'false' with 0 and 'true' with 1 for Postgres 'deleted' column
-              pg-values (map #(if (boolean? %) (if % 1 0) %) values)
```
johnwteague, line 598: Do you not need this boolean -> int fix-up?

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
johnwteague, line 600: Sorry for what is probably a dumb question (I don't speak clojure), but can you walk me through the logic of this?

johnwteague, line 600: I guess following on Tyler's question, but why not have a positive check on the data type of the metadata column, rather than a list of "not-metadata-types"?

zimzoom: correct, that's exactly right

zimzoom: sure, we could do that for readability

johnwteague (approved):

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
zimzoom: done!

zimzoom: updated this to explicitly check for bytes array instead

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
```diff
@@ -594,15 +594,20 @@
...
+              ;; if set to not use metadata column in Aurora database, then remove it and replace with nil
+              pg-values (if (aurora-config/aurora-metadata?)
+                          values
+                          (map #(if (bytes? %) nil %) values))
```
jaybarra, line 600:
```suggestion
                          (map #(when-not (bytes? %) %) values))
```

TylerHeald1 (approved):

PR #1964, opened 2023-08-28
CMR-9372 CMR Internal server error when disassociating tags

daniel-zamora:
After legacy services shutdown, invalid tokens can get routed to a 301 returning endpoint for verification.  If we receive a 301 we will now assume the token does not exist as it has failed jwt and launchpad verification.  This is done to preserve the mock-echo testing functionality for now.  I ticket will be written to decouple CMR tests from mock echo.


eudoroolivares2016:
There's a TODO comment on Line 12 in the function `get-token-type` in the namespace `common-app-lib/src/cmr/common_app/api/launchpad_token_validation.clj` does this take care of that? Also do we need to make that call `(common-util/is-legacy-token? token) "Legacy-EDL"` anymore at all? since its no longer supported

eudoroolivares2016 (approved):

eereiter (approved):

PR #1965, opened 2023-08-29
CMR-9270: As CMR DevOps I want CMR services to abstractly create a DB connection

TylerHeald1:


zimzoom:
I like this approach. I do kind of wonder whether this should merge into a feature branch ala `feature/db-abstraction` , since:

* it is a very small piece of what will ultimately be a large refactor, which has not been fully designed yet
* it would make easier to incorporate this change with the other very large pending database code change
* it would make it easier for devs on other teams to review all the pieces in their context
* it would help keep CMR git history tidy (which some have voiced lately has been getting painful)

metadata-db-app/test/cmr/metadata_db/test/data/oracle/concepts.clj
```diff
@@ -38,7 +39,7 @@
 ;; This test is commented out until CMR-1303 is resolved
 #_(deftest db-result->concept-map-test
```
jaybarra, line 40: This test should probably be deleted if it hasn't been re-enabled since CMR-1303 was written.

jceaser (approved):

zimzoom (approved):

PR #1972, opened 2023-09-26
CMR-9410: Missing MetadataSpecification handling

sxu123:


ingest-app/src/cmr/ingest/api/generic_documents.clj
```diff
@@ -129,10 +130,15 @@
...
-             (not= (common-concepts/get-concept-type-of-draft concept-type) spec-key))
-      (throw (UnsupportedOperationException. (format "%s version %s is not supported." spec-key spec-version)))
+    (if (or (not spec-key)
+            (and (not= concept-type spec-key)
+                 (not= (concept-type draft-concept->spec-map) spec-key)
+                 (not= (common-concepts/get-concept-type-of-draft concept-type) spec-key)))
+      (errors/throw-service-error
+       :invalid-data
+       (if-not spec-key
+         "MetadataSpecification is missing"
```
eereiter, line 140: Can you change the string to 

"The MetadataSpecification schema element is missing from the record being ingested."

sxu123: Done

ingest-app/src/cmr/ingest/api/generic_documents.clj
```diff
@@ -129,10 +130,15 @@
...
-      (throw (UnsupportedOperationException. (format "%s version %s is not supported." spec-key spec-version)))
+    (if (or (not spec-key)
+            (and (not= concept-type spec-key)
+                 (not= (concept-type draft-concept->spec-map) spec-key)
+                 (not= (common-concepts/get-concept-type-of-draft concept-type) spec-key)))
+      (errors/throw-service-error
+       :invalid-data
+       (if-not spec-key
+         "The MetadataSpecification schema element is missing from the record being ingested."
+         (str spec-key " version " spec-version " are not supported")))
```
eudoroolivares2016, line 141: Can we add a test for if the version is incorrect

sxu123: If you look at the code I replaced, I was only adding the case when spec-key doesn't exist,  which is when MetadataSpecification doesn't exist. the other part of the functionality is carried over.  That is also only checking the spec key.  version is not being checked. It's just returned in the message, together with the invalid spec-key.

sxu123: added one more test when spec-key is wrong.

eudoroolivares2016 (approved):

ingest-app/src/cmr/ingest/api/generic_documents.clj
eudoroolivares2016: Phrased that wrong you are correct. We do handle the version already, and with this we can now handle the wrong "Name" thanks for adding test.

eereiter (approved):

PR #1985, opened 2023-10-24
CMR-9399 Adding Processing levels as a new keyword set that the CMR s…

eereiter:
…upports from KMS.


TylerHeald1 (approved):

system-int-test/test/cmr/system_int_test/search/keyword_endpoint_test.clj
```diff
@@ -697,181 +697,99 @@
...
+                 {"value" "application/octet-stream",
+                  "uuid" "b77e64ef-ce80-4dab-b552-c6062990a6e0"}
+                 {"value" "text/plain",
+                  "uuid" "fea4e0a7-d794-481d-9915-52f1be226714"}
+                 {"value" "image/jpeg",
+                  "uuid" "3f697f52-6a1c-4e2c-bd4b-13aaaf45f2e6"}
+                 {"value" "application/gzip",
+                  "uuid" "a8ee535a-8bc8-46fd-8b97-917bd7ea7666"}
+                 {"value" "application/x-hdf5",
+                  "uuid" "4e80047b-c50b-4805-ac68-789dbc38803f"}]}
```
sxu123, line 775: Why do we need to change the granule-data-format and mime-type here?  Is it that the kms structure changed for these two?

sxu123 (approved):

jceaser (approved):

system-int-test/test/cmr/system_int_test/search/keyword_endpoint_test.clj
eereiter: Because what they were passing back wasn't correct.  We don't need to pass back a uuid structure that contains a uuid value, just uuid itself is what is needed.

PR #1991, opened 2023-11-02
CMR-9503: Transfer the index name caches from Search to redis

TylerHeald1:


system-int-test/test/cmr/system_int_test/admin/cache_api_test.clj
```diff
@@ -168,7 +168,7 @@
         (url/search-read-caches-url) "acls" ["acls"]
         (url/search-read-caches-url) "collections-for-gran-acls" []
         (url/search-read-caches-url) "has-granules-map" []
-        (url/search-read-caches-url) "index-names" []
+        (url/search-read-caches-url) "index-names" [":acls-hash-code" ":collection-granule-aggregation-cache-hash-code" ":concept-indices-hash-code" ":concept-mapping-types-hash-code" ":providers-hash-code" ":write-enabled-hash-code" "collection-granule-aggregation-cache" "kms"]
```
TylerHeald1, line 171: This is a result of the get-keys function in the RedisCache object simply getting all the keys in redis, which includes keys that are associated with other cache names

eudoroolivares2016: Just verifying so the last three items in that vector `"collection-granule-aggregation-cache" "kms"` aren't keys (they don't have `:` is that intentional thought the test of the items are?

TylerHeald1: They are keys in the redis cache, but certainly they are not in clojure key format. The definition and use of those keys is done elsewhere and I couldn't really say whether that was intentional or not

eudoroolivares2016 (approved):

jmaeng72 (approved):

PR #1994, opened 2023-11-03
CMR-9535: Add InstanceInformation, Definition, and Science Keywords to elastic-search and the variable JSON response

eudoroolivares2016:
Root cause of the slowness seen on EDSC (https://bugs.earthdata.nasa.gov/browse/EDSC-3867) is because the `umm_json` response for variables is causing timeout in cases where the collection has many variables. Analysis was done to determine if there was some obvious inefficiencies in the `umm_json` response for variables. Since none was seen we should add the fields for variables that EDSC requires into elastic-search and then reindex all variables in elastic search. So that EDSC can retrieve all the fields that it needs from `variables` through the `.json` endpoint instead (through `cmr-graphql` which will also require an update after this)

Note that we will need to do a re-indexing job for `variables` on each env as this propagates up

### DIFF
--- a/search-app/docs/api.md
+++ b/search-app/docs/api.md
@@ -3489,7 +3489,7 @@ The keyword endpoint is used to retrieve the full list of keywords for each of t
 
 The keywords are returned in a hierarchical JSON format. The response format is such that the caller does not need to know the hierarchy, but it can be inferred from the results. Keywords are not guaranteed to have values for every subfield in the hierarchy, so the response will indicate the next subfield below the current field in the hierarchy which has a value. It is possible for the keywords to have multiple potential subfields below it for different keywords with the same value for the current field in the hierarchy. When this occurs the subfields property will include each of the subfields.
 
-Supported keywords include `platforms`, `instruments`, `projects`, `temporal_keywords`, `location_keywords`, `science_keywords`, `archive_centers`, `data_centers`, `granule-data-format`, "mime-type and `measurement-name`. The endpoint also supports `providers` which is an alias to `data_centers` and `spatial_keywords` which is an alias to `location_keywords`.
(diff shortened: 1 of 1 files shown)


indexer-app/src/cmr/indexer/data/concepts/variable.clj
```diff
@@ -45,6 +49,7 @@
                                 :set-names (map :Name (:Sets parsed-concept))})
                         schema-keys)
         all-assocs (concat variable-associations generic-associations)]
+
```
eudoroolivares2016, line 52: put back

eudoroolivares2016:
We should also update the search apps `api.md` for the JSON response

dpesall (approved):

search-app/docs/api.md
```diff
@@ -3489,7 +3489,7 @@ The keyword endpoint is used to retrieve the full list of keywords for each of t
...
 
-Supported keywords include `platforms`, `instruments`, `projects`, `temporal_keywords`, `location_keywords`, `science_keywords`, `archive_centers`, `data_centers`, `granule-data-format`, "mime-type`, `measurement-name`, and `processing-levels`. The endpoint also supports `providers` which is an alias to `data_centers` and `spatial_keywords` which is an alias to `location_keywords`.
+Supported keywords include `platforms`, `instruments`, `projects`, `temporal_keywords`, `location_keywords`, `science_keywords`, `archive_centers`, `data_centers`, `granule-data-format`, "mime-type and `measurement-name`. The endpoint also supports `providers` which is an alias to `data_centers` and `spatial_keywords` which is an alias to `location_keywords`.
```
eudoroolivares2016, line 3492: fix

PR #1995, opened 2023-11-07
CMR-9532 Pass a "has_combine" field in the JSON response for collections 

daniel-zamora:
Adds has_combine to umm_json, json, and atom coll search results to facilitate order option presentation on earthadata search.

### DIFF
--- a/indexer-app/src/cmr/indexer/data/concepts/service.clj
+++ b/indexer-app/src/cmr/indexer/data/concepts/service.clj
@@ -106,6 +106,16 @@
   [service]
   (has-subset-type? service :VariableSubset))
 
+(defn- has-combine?
+  "Returns true if the given service has ConcatenateDefault = true, false otherwise."
+  [service]
+  (-> service
+      :ServiceOptions
+      :Aggregation
+      :Concatenate
+      :ConcatenateDefault
+      boolean))
+
 (defn- has-transforms?
   "Returns true if the given service has a defined SubsetTypes or InterpolationTypes,
   or multiple supported projections values."
@@ -125,6 +135,7 @@
   [services]
   {:has-formats (boolean (some has-formats? services))
    :has-transforms (boolean (some has-transforms? services))
+   :has-combine (boolean (some has-combine? services))
    :has-variables (boolean (some has-variables? services))
    :has-spatial-subsetting (boolean (some has-spatial-subsetting? services))
    :has-temporal-subsetting (boolean (some has-temporal-subsetting? services))})
@@ -144,7 +155,7 @@
         harmony-services (filter #(= "Harmony" (:Type %)) services)]
     (util/remove-map-keys
       empty?
-      {:opendap (get-trimmed-has-features opendap-services )
+      {:opendap (get-trimmed-has-features opendap-services)
        :esi (get-trimmed-has-features esi-services)
        :harmony (get-trimmed-has-features harmony-services)})))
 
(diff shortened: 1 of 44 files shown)


system-int-test/test/cmr/system_int_test/search/service/collection_service_search_test.clj
```diff
@@ -499,3 +499,47 @@
...
+    (index/wait-until-indexed)
+    (au/associate-by-concept-ids token serv1-concept-id [{:concept-id (:concept-id coll1)}])
+    (au/associate-by-concept-ids token serv2-concept-id [{:concept-id (:concept-id coll2)}])
+    (au/associate-by-concept-ids token serv3-concept-id [{:concept-id (:concept-id coll3)}])
+    (index/wait-until-indexed)
+    (testing "when Aggregation is not present"
+      (service-util/assert-collection-search-result coll1 {:has-combine false :service-features {:harmony {:has-combine false}}} [serv1-concept-id]))
+    (testing "when concatenateDefault is true"
+      (service-util/assert-collection-search-result coll2 {:has-combine true :service-features {:harmony {:has-combine true}}} [serv2-concept-id]))
+    (testing "when concatenateDefault is false"
```
eudoroolivares2016, line 544: Thinking about this more closely in this case, even if the value for `ConcatenateDefault` is `false` don't we still have the option to do `combines` because the `serviceOption` obj of `:Concatenate` does exist.
Should `has_combine` be true in that case?

eudoroolivares2016: So on EDSC I think we would still want to show the Badge even if the `concatenateDefault` is false

indexer-app/src/cmr/indexer/data/concepts/service.clj
```diff
@@ -125,6 +135,7 @@
   [services]
   {:has-formats (boolean (some has-formats? services))
    :has-transforms (boolean (some has-transforms? services))
+   :has-combine (boolean (some has-combine? services))
    :has-variables (boolean (some has-variables? services))
```
eereiter, line 139: Can we change this to has-concatenate ?  Concatenate has a specific meaning and it also aligns better to the actual UMM-S metadata field.

system-int-test/test/cmr/system_int_test/search/service/collection_service_search_test.clj
eereiter: Yes, the existence of the UMM-S field in the metadata means that the data set can be concatenated.  It is irrelevant if the value is true or false.  Also can we change the name from has-combined to has-concatenate as concatenate has a specific meaning to the subsetting people and it aligns better to the UMM-S field that holds the data.

daniel-zamora: The ticket, which you wrote, says "On the collection's JSON response, we should add a field called has_combine which is a boolean that is determined whether or not the collection is associated to a UMM-S record that contains an InstanceInformation field where the sub field is true such as this:"
```   
 "aggregation": {"concatenate": {"concatenateDefault": true} 
```

Until the ticket changes the implementation will remain this way.

daniel-zamora: Ah I didn't see erich's update, will adjust then.

dpesall (approved):

system-int-test/test/cmr/system_int_test/search/service/collection_service_search_test.clj
eereiter: I understand what happen.  Leaving the name is good.

eereiter (approved):

eudoroolivares2016 (approved):

PR #1998, opened 2023-11-09
CMR-9546: Updating GraphDb to Node-18

eudoroolivares2016:
Update `graphDb` to use node18 this is important as node-16 has already reached EOL


eudoroolivares2016:
I believe the reason there are so many changes in the `package-lock` file is just because of minor patches  picked up by NPM on the dependencies

graph-db/scripts/deploy-bamboo.sh
```diff
@@ -12,7 +12,7 @@ node_modules
 EOF
 
 cat <<EOF > Dockerfile
-FROM node:16
+FROM node:18.16-bullseye
```
eudoroolivares2016, line 15: This is what `cmr-graphl` is currently set to and I saw a chat in `#ci/cd` about needing this, the dependencies may be updated enough at this point that we could use `node:18-bullseye`

jceaser (approved):

daniel-zamora (approved):

eudoroolivares2016:
The build for this is not passing due to the `snyk` issue however the build was successful I was told that should be done as a separate ticket 
https://ci.earthdata.nasa.gov/browse/CN2-CG230-1

PR #2005, opened 2023-11-15
CMR-9554: update temporal search documentation.

isja17:
Update CMR temporal search documentation to include "between Julian days x,y"


zimzoom (approved):

jmaeng72 (approved):

johnwteague:
Do we know why/how this is supported? Is it part of the ISO 8601 datetime spec or is it our own addition?

isja17:
> Do we know why/how this is supported? Is it part of the ISO 8601 datetime spec or is it our own addition?

I believe it is our own addition that we made a long time ago specifically for ASF, but we never documented it. At least, that is my interpretation of what Gregory Short was saying in the #cmr thread. ISO has a different notation for doing the same thing.

PR #2012, opened 2023-11-29
CMR-9555: Incorporate integration test coverage results into code coverage step

TylerHeald1:
Adding to test configuration so a codecov file is generated by the integration tests


eereiter (approved):

jceaser (approved):

jaybarra:
Code coverage does not provide useful metrics in the integration test suite. It is only adding coverage of the test suite itself and nothing of the production code. Coverage metrics are only useful when done run as unit tests.

PR #2014, opened 2023-11-29
CMR-9558 Fixing indexer kms cache.

eereiter:


common-app-lib/src/cmr/common_app/services/kms_fetcher.clj
```diff
@@ -82,7 +86,8 @@
   (let [cache (cache/context->cache context kms-cache-key)
         gcmd-keywords-map (fetch-gcmd-keywords-map context)]
     (info "Refreshed KMS cache.")
-    (cache/set-value cache kms-cache-key gcmd-keywords-map)))
+    (cache/set-value cache kms-cache-key gcmd-keywords-map)
+    (info (format "*** the KMS cache size is %d" (cache/cache-size cache)))))
```
jceaser, line 90: two things, is this a number or a string, line 77 uses %s ; second thing, something should be different in these two lines so we can tell which is which. We may be confused as to which function wrote this?

indexer-app/src/cmr/indexer/data/concepts/collection.clj
```diff
@@ -211,6 +211,29 @@
       (distinct (conj consortiums "GEOSS"))
       consortiums)))
 
+(defn- get-memory-statistics
```
jceaser, line 214: I think you should use the same function in cmr.common-app.services.jvm-info by making it public.

metadata-db-app/src/cmr/metadata_db/services/concept_service.clj
```diff
@@ -684,46 +684,52 @@
...
-                                                (set (keys concepts-by-concept-id)))]
-        (errors/throw-service-errors
-          :not-found
-          (map msg/concept-does-not-exist
-               missing-concept-ids))))))
+  (if (<= (count concept-ids) 0)
+    (do
+      (info "Calling get-latest-concepts with 0 concept-ids.")
+      '())
+    (let [_ (info (format "Getting [%d] latest concepts by concept-id %s" (count concept-ids) (doall (reduce #(str %1 " " %2)
```
jceaser, line 691: if you make other changes, could you make this one slightly less long, with those parameters on a new line way over to the edge. Actually as the first statement in the let, does it even need to be in the let and not above?
Change if you are already making other changes.

jceaser (changes requested):

common-app-lib/src/cmr/common_app/services/kms_fetcher.clj
eereiter: First off cache/cache-size returns a number, so they both should be %d.  
Secondly, I don't think we need these anymore as we now report the redis cache sizes as part of the job that reports on all of the cache sizes.  I initially used this to test why the indexer kept running out of memory. But I now know why.

eereiter: I have removed these log statements, as I think they are no longer needed.

indexer-app/src/cmr/indexer/data/concepts/collection.clj
eereiter: Better yet I removed this function and the log-jvm-statstics function and added to the common-app/services/jvm-info.clj namespace the ability to add in an application name or any other string before the Maximum Memory message.

metadata-db-app/src/cmr/metadata_db/services/concept_service.clj
eereiter: I either need it in the let, because it is part of the if statement or I need to wrap it all into a do function. I left it in the let statement.
So far I didn't need to make other changes in this namespace, but I mored the parameters to the format function over to the right.

isja17 (approved):

PR #2015, opened 2023-11-30
Feature CMR-9539 - Update the coll-cache from standard redis to new Hash Map Redis 

jceaser:
Moving the code over to the new hash map redis cache and breaking up the data into two different hashes within that cache to limit the request data being transmitted over the "wire".

### DIFF
--- a/search-app/src/cmr/search/services/acls/collections_cache.clj
+++ b/search-app/src/cmr/search/services/acls/collections_cache.clj
@@ -104,27 +104,32 @@
   (let [cache (hash-cache/context->cache context cache-key)
         collections-map (fetch-collections-map context)]
     (doseq [[coll-key coll-value] collections-map]
-      (debug (format "refresh-cache %s" coll-key))
+      (debug (format "collections-cache refresh-cache working on %s" coll-key))
       (hash-cache/set-value cache cache-key coll-key (clj-times->time-strs coll-value)))))
 
 (defn- get-collections-map
-  "Gets the cached value."
-  [context key-type]
+  "Gets the cached value. By default an error is thrown if there is no value in
+   the cache. Pass in anything other then :return-errors for the third parameter
+   to have null returned when no data is in the cache. :return-errors is for the
+   default behavior."
+  ([context key-type]
+   (get-collections-map context key-type :return-errors))
+  ([context key-type throw-errors]
   (let [coll-cache (hash-cache/context->cache context cache-key)
         collection-map (hash-cache/get-value
                         coll-cache
                         cache-key
                         key-type)]
-    (if (empty? collection-map)
+    (if (and (= :return-errors throw-errors) (empty? collection-map))
       (errors/internal-error! (coll-msg/collections-not-in-cache key-type))
-      (time-strs->clj-times collection-map))))
+      (time-strs->clj-times collection-map)))))
 
 (defn get-collection
   "Gets a single collection from the cache by concept id. Handles refreshing the cache if it is not found in it.
   Also allows provider-id and entry-title to be used."
   ([context concept-id]
-   (let [by-concept-id (get-collections-map context :by-concept-id)]
-     (when-not (by-concept-id concept-id)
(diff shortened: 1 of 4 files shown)


search-app/src/cmr/search/services/acls/collections_cache.clj
```diff
@@ -88,6 +90,7 @@
                                          (for [{:keys [provider-id EntryTitle] :as coll}
                                                collections]
                                            [[provider-id EntryTitle] coll]))]
+    (debug "fetch-collections-map count" (count collections))
     ;; We could reduce the amount of memory here if needed by only fetching the collections that
     ;; have granules.
     {:by-concept-id by-concept-id
```
eereiter, line 96: Should we write a ticket to reduce the amount of memory by only fetching the collections that have granules?  Do we need the collections in this cache that don't have granules?

search-app/test/cmr/search/test/unit/services/acls/granule_acls.clj
```diff
@@ -91,7 +91,7 @@
         old-mem-cache (get-in context [:system :caches :collections-for-gran-acls])
         context (assoc-in context [:system :caches :collections-for-gran-acls] rcache)
         data (.get-value old-mem-cache :collections)]
-    (.set-value rcache :collections-for-gran-acls data)
+    (.set-values rcache :collections-for-gran-acls data)
```
eereiter, line 94: Is .set-values a short cut way of writing hash-cache/set-values ?

eereiter, line 94: Why did you leave old-mem-cache in the test?  Can you remove this and just set up the redis cache as the old-mem-cache (replace old-mem-cache)?  All you do with it is setup the memory cache and then take its contents and put it into the redis cache and then not use it anymore.

eereiter (approved):
It looks like the search app is loading in the cache instead of bootstrap.  I thought that at the beginning of the PI we were going to put the loading of the cache into bootstrap so that only 1 task loads the cache.  I saw this with a couple of other cache changes as well.  I think we should write a ticket to put the job scheduler of loading this cache into bootstrap.

search-app/src/cmr/search/services/acls/collections_cache.clj
jceaser: that comment was already there, I did not change the logic of what was going to be cached

search-app/test/cmr/search/test/unit/services/acls/granule_acls.clj
jceaser: I was trying to keep the code as similar as possible. I can switch over to the other syntax.

search-app/src/cmr/search/services/acls/collections_cache.clj
```diff
@@ -98,34 +101,40 @@
   to keep the cache fresh. This will throw an exception if there is a problem fetching collections. The
   caller is responsible for catching and logging the exception."
   [context]
-  (let [cache (cache/context->cache context cache-key)
+  (let [cache (hash-cache/context->cache context cache-key)
         collections-map (fetch-collections-map context)]
-    (cache/set-value cache cache-key (clj-times->time-strs collections-map))))
+    (doseq [[coll-key coll-value] collections-map]
+      (debug (format "collections-cache refresh-cache working on %s" coll-key))
```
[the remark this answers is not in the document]
jceaser: removing comment

search-app/test/cmr/search/test/unit/services/acls/granule_acls.clj
eereiter: you don't need to change over, I was just wondering.

jceaser: I don't think I even need this code, I removed it and tests pass locally, testing on bamboo now.

jceaser: switched to the clojure syntax as suggested.

PR #2020, opened 2023-12-11
CMR-9582 Making the values a def config so that they can be set in AW…

eereiter:
…S, setting the maximum search result value to 200K.

### DIFF
--- a/common-app-lib/src/cmr/common_app/config.clj
+++ b/common-app-lib/src/cmr/common_app/config.clj
@@ -28,3 +28,16 @@
   "CMR support email address"
   {:default "cmr-support@earthdata.nasa.gov"})
 
+(defconfig es-unlimited-page-size
+  "This is the number of items we will request from elastic search at a time when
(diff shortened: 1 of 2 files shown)


zimzoom (approved):

common-app-lib/src/cmr/common_app/config.clj
johnwteague, line 1: The current issue is a result of exceeding the max-unlimited-hits (100K) correct? Can you explain the need to increase unlimited-page-size from 10K to 100K?

jceaser (approved):

common-app-lib/src/cmr/common_app/config.clj
eereiter: I didn't mean to increase it.  I will change it back to 10,000.

eereiter: Fixed.

johnwteague (approved):

PR #2036, opened 2024-01-17
CMR-9625 

sxu123:
Added functionality to allow MMT to add private data in the metadata for variable draft. It's removed at publishing.
No new tests are necessary.  If it's not removed correctly, the existing tests should fail during schema validation at publishing.

### DIFF
--- a/system-int-test/src/cmr/system_int_test/utils/generic_util.clj
+++ b/system-int-test/src/cmr/system_int_test/utils/generic_util.clj
@@ -37,7 +37,7 @@
                             (slurp)
                             (json/parse-string true)))
 
-(def variable-draft (-> "schemas/variable-draft/v1.0.0/metadata.json"
+(def variable-draft (-> "schemas/variable-draft/v1.0.0/metadata_with_private_data.json"
                         (jio/resource)
                         (slurp)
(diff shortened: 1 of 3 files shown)


system-int-test/src/cmr/system_int_test/utils/generic_util.clj
```diff
@@ -37,7 +37,7 @@
                             (slurp)
                             (json/parse-string true)))
 
-(def variable-draft (-> "schemas/variable-draft/v1.0.0/metadata.json"
+(def variable-draft (-> "schemas/variable-draft/v1.0.0/metadata_with_private_data.json"
```
cgokey, line 40: I can't read Clojure, but do you need a test to verify that the _private doesn't show up in the resulting metadata?

sxu123: No. I mentioned it at the very beginning that tests are not needed. The existing tests covered it.  If the _private is not removed, the schema validation would fail when publishing the variable-draft because the element doesn't exist in the schema.

cgokey (approved):

jceaser (approved):

PR #2040, opened 2024-01-24
CMR-9591: Write a memory dump on exist if requested 

jceaser:
1. Write a memory dump to the file system if sent a `kill -15 <pid>`. This memory dump can then be loaded into visual vm for analyses.
2. RefreshKmsCacheJob now uses a try/catch and writes a short error message if an exception is thrown to shorten the logs. This normally happens locally when starting CMR
3. Adding a new way to make local configurations through refresh_persistent_settings.clj which will not trigger a file change in git status

### DIFF
--- a/common-app-lib/src/cmr/common_app/app_events.clj
+++ b/common-app-lib/src/cmr/common_app/app_events.clj
@@ -1,0 +1,74 @@
+(ns cmr.common-app.app-events
+  "Common events which applications can use when starting and stopping."
+  (:require
+   [clojure.java.io :as io]
+   [cmr.common-app.config :as config]
+   [cmr.common.log :refer [debugf infof warnf errorf]]))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(import 'com.sun.management.HotSpotDiagnosticMXBean)
+(import 'java.lang.management.ManagementFactory)
+
+;;
+;; The internet suggests the following as how one would write the dump file in java:
+;;
+;;public static void dumpHeap (String filePath, boolean live) throws IOException {
+;;  MBeanServer server = ManagementFactory.getPlatformMBeanServer ();
+;;  HotSpotDiagnosticMXBean mxBean = ManagementFactory.newPlatformMXBeanProxy (server,
+;;      "com.sun.management:type=HotSpotDiagnostic", HotSpotDiagnosticMXBean.class);
+;;   mxBean.dumpHeap (filePath, live);
+;;}
+
+(defn- dump-heap
+  "Write a diagnostics file to the file system for analysis in Visual VM"
+  [^String app-name ^String file-path ^java.lang.Boolean live]
+  (let [bean (-> (ManagementFactory/getPlatformMBeanServer)
+                 (ManagementFactory/newPlatformMXBeanProxy "com.sun.management:type=HotSpotDiagnostic"
+                                                           HotSpotDiagnosticMXBean))]
+    (try
+      (when (.exists (io/file file-path))
+        (warnf (format "dump-heap: %s deleting diagnostic file %s." app-name file-path))
+        (io/delete-file file-path))
+      (.dumpHeap bean file-path live)
+      (infof "dump-heap: %s finished creating diagnostic file %s." app-name file-path)
+      (catch Exception err
+        (errorf "dump-heap: %s because %s" app-name (.getMessage err))))))
+
+(defn- writeDump
+  "A call back function which writes a diagnostics file if the application config
+   dump-diagnostics-on-exist-to has been set to a file path. To test this out,
+   run the following commands:
+   >CMR_DUMP_DIAGNOSTICS_ON_EXIT_TO='/tmp/dump-test.hprof'
+       java -cp cmr-search-app-0.1.0-SNAPSHOT-standalone.jar clojure.main
+       -m cmr.search.runner
+   >kill -15 $(jps | grep main | cut -f 1 -d ' ')"
+  [app-name]
+  (let [path-to-file (config/dump-diagnostics-on-exit-to)]
+    (infof "dump-heap: %s is writing a diagnostic file to %s..." app-name path-to-file)
+    (dump-heap app-name path-to-file false)))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(defn stop-on-exit-hook
+  "Add a shutdown hook to call the CMR app stop command found in most if not all
+   CMR microservices. Caller must supply a function that calls the application's
+   stop function such as: #(cmr.search.system/stop system)"
+  [app-name stop-thread]
+  (.addShutdownHook (Runtime/getRuntime) (new Thread stop-thread)))
+
+(defn dump-on-exit-hook
+  "Add a shutdown hook to the current run time which will write a memory dump file
+   if the app has been configured to do so. Set the config dump-diagnostics-on-exist-to
+   a path to the file the log should be saved to."
+  [app-name]
+  (let [path-to-file (config/dump-diagnostics-on-exit-to)]
+    (when-not (empty? path-to-file)
+      (debugf "dump-heap: %s Will write a diagnostic file to %s on exit." app-name path-to-file)
+      (.addShutdownHook (Runtime/getRuntime) (new Thread #(writeDump app-name))))))
+
+(comment
+  ;; use these to create a memory file locally for use in Visual VM
+  (config/set-dump-diagnostics-on-exit-to! "/tmp/dump-auto.txt")
+  (dump-heap "/tmp/dump.txt" true)
+  (dump-heap "/tmp/dump.hprof" false))
--- a/common-app-lib/src/cmr/common_app/services/kms_fetcher.clj
+++ b/common-app-lib/src/cmr/common_app/services/kms_fetcher.clj
@@ -91,7 +91,13 @@
 ;; Job for refreshing the KMS keywords cache. Only one node needs to refresh the cache.
 (def-stateful-job RefreshKmsCacheJob
   [_ system]
-  (refresh-kms-cache {:system system}))
+  (try
+    ;; Locally KMS may not be configured or ready to go, in this case a bad URL
+    ;; May exist. Trap the error and print out a shorter message to shorten the
+    ;; logs
+    (refresh-kms-cache {:system system})
+    (catch java.net.MalformedURLException mue
+      (error "Malformed URL"  (.getMessage  mue)))))
 
 (defn refresh-kms-cache-job
   "The singleton job that refreshes the KMS cache. The keywords are infrequently updated by the
--- a/common-lib/src/cmr/common/system.clj
+++ b/common-lib/src/cmr/common/system.clj
@@ -2,7 +2,7 @@
   "Contains helper functions for application systems."
   (:require
    [cmr.common.lifecycle :as lifecycle]
-   [cmr.common.log :as log :refer (debug info warn error)]))
+   [cmr.common.log :as log :refer (debug info warn error reportf)]))
 
 (defn stop
   "Stops the system using the given component order."
@@ -46,19 +46,19 @@
   starting and started"
   [system-name component-order]
   (fn [s]
-    (info (str system-name " System starting"))
+    (reportf "%s System starting" system-name)
     (try
       (start s component-order)
       (finally
-        (info (str system-name " System started"))))))
+        (reportf "%s System started" system-name)))))
 
 (defn stop-fn
   "Creates a generic system stop function that logs when the system is stopping
   and stopped"
   [system-name component-order]
   (fn [s]
-    (info (str system-name " System stopping"))
+    (reportf "%s System stopping" system-name)
     (try
       (stop s component-order)
       (finally
-        (info (str system-name " System stopped"))))))
+        (reportf "%s System stopped" system-name)))))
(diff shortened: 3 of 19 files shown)


common-app-lib/src/cmr/common_app/services/kms_fetcher.clj
johnwteague, line 1: This seems unrelated to the ticket, is it?

common-lib/src/cmr/common/system.clj
```diff
@@ -46,19 +46,19 @@
   starting and started"
   [system-name component-order]
   (fn [s]
-    (info (str system-name " System starting"))
+    (reportf "%s System starting" system-name)
```
johnwteague, line 49: Are you changing the level categorization from INFO to REPORT here? If so, why?

dev-system/dev/refresh_persistent_settings.clj
```diff
@@ -1,14 +1,39 @@
...
+   This file is only read at initial launch of CMR, (user/reset) will not pickup
+   changes.
+   "
+  [which default]
+  (let [home (System/getProperty "user.dir")
+        file-path (format "%s/local.edn" home)
+        exists (.exists (io/file file-path))
+        local-settings (if exists (read-string (slurp file-path)) {})
+        settings (get local-settings which default)]
+    ;; Aggresivly display that settings are being set from this process
```
johnwteague, line 30: typo "aggressively"

common-app-lib/src/cmr/common_app/app_events.clj
```diff
@@ -0,0 +1,74 @@
...
+;;public static void dumpHeap (String filePath, boolean live) throws IOException {
+;;  MBeanServer server = ManagementFactory.getPlatformMBeanServer ();
+;;  HotSpotDiagnosticMXBean mxBean = ManagementFactory.newPlatformMXBeanProxy (server,
+;;      "com.sun.management:type=HotSpotDiagnostic", HotSpotDiagnosticMXBean.class);
+;;   mxBean.dumpHeap (filePath, live);
+;;}
+
+(defn- dump-heap
+  "Write a diagnostics file to the file system for analysis in Visual VM"
+  [^String app-name ^String file-path ^java.lang.Boolean live]
```
johnwteague, line 25: Am I correct that the file-path is configuration driven only and so the name of the resulting dump file is not built from the app-name, or am I overlooking that here? It looks like the app-name (set as instance-name below I think?) is only used for the message events?
This might be a question for the follow-up ticket (where we write the dump file to some external and accessible location) but thinking that we probably want the dump file name to be unique and identify the source task; you already create the unique (ish) instance-name.

search-app/src/cmr/search/system.clj
```diff
@@ -164,20 +165,16 @@
   "Performs side effects to initialize the system, acquire resources,
   and start it running. Returns an updated instance of the system."
   [this]
-  (info "search System starting")
```
johnwteague, line 167: Why are you removing these start/stop info log events?

search-app/src/cmr/search/runner.clj
```diff
@@ -1,12 +1,13 @@
...
+  "Entry point for the application. Defines a main method that accepts arguments."
+  (:require [cmr.common-app.app-events :as events]
+            [cmr.common.config :as cfg]
+            [cmr.search.system :as system]))
 
 (defn -main
   "Starts the App."
   [& args]
   (let [system (system/start (system/create-system))]
-    (info "Running...")
```
johnwteague, line 11: Why remove this log event?

ingest-app/src/cmr/ingest/runner.clj
```diff
@@ -1,16 +1,14 @@
...
-            [cmr.ingest.api.routes :as routes]
+            [cmr.common-app.app-events :as events]
             [cmr.common.config :as cfg])
   (:gen-class))
 
 (defn -main
   "Starts the App."
   [& args]
   (let [system (system/start (system/create-system))]
-    (info "Running...")
```
johnwteague, line 15: Why remove this log event? (same question in multiple places).  I assume you are cleaning up, guess I'll just ask you to confirm that you are removing redundant messaging then.

ingest-app/src/cmr/ingest/runner.clj
```diff
@@ -1,16 +1,14 @@
...
-    (cfg/check-env-vars)))
+    (cfg/check-env-vars)
+    (events/stop-on-exit-hook (:instance-name system) #(system/stop system))
+    (events/dump-on-exit-hook (:instance-name system))))
```
johnwteague, line 14: So does dump-on-exit get invoked (and the dump file created if configured to do so) for _all_  shutdowns? Hm. Maybe I should take this question up outside of the PR.

common-app-lib/src/cmr/common_app/app_events.clj
```diff
@@ -0,0 +1,74 @@
...
+                 (ManagementFactory/newPlatformMXBeanProxy "com.sun.management:type=HotSpotDiagnostic"
+                                                           HotSpotDiagnosticMXBean))]
+    (try
+      (when (.exists (io/file file-path))
+        (warnf (format "dump-heap: %s deleting diagnostic file %s." app-name file-path))
+        (io/delete-file file-path))
+      (.dumpHeap bean file-path live)
+      (infof "dump-heap: %s finished creating diagnostic file %s." app-name file-path)
+      (catch Exception err
+        (errorf "dump-heap: %s because %s" app-name (.getMessage err))))))
```
johnwteague, line 36: Does the dump-heap take long to complete? Any chance of it getting "stuck" so that the shutdown hangs?

common-app-lib/src/cmr/common_app/app_events.clj
jceaser: app-name is used for the logs to report which app (search vs ingest) is doing this. I have to have this as a parameter because I put this code in common-app-lib instead of having a copy in each app.

file-path is passed in and decided by the caller.

common-app-lib/src/cmr/common_app/app_events.clj
jceaser: it has not taken long for me, but it could hang sure. I think that AWS would just kill more aggressively in that case. I could time it if you like and write the time out.

common-app-lib/src/cmr/common_app/services/kms_fetcher.clj
jceaser: your right, it is, but I have been passing this along for some time and it gets in the way of debugging and development when I have to wade through these false errors.

ingest-app/src/cmr/ingest/runner.clj
jceaser: a kill -9 never gets here, that is done at the OS level, a kill -15 and control-c (local) show up here.

ingest-app/src/cmr/ingest/runner.clj
jceaser: it is redundant with the alerts in common-lib/src/cmr/common/system.cl

search-app/src/cmr/search/runner.clj
jceaser: I found it is redundant with another log, common-lib/src/cmr/common/system.clj

search-app/src/cmr/search/system.clj
jceaser: common-lib/src/cmr/common/system.clj should handle this

common-lib/src/cmr/common/system.clj
jceaser: I'm condensing two logs into one, these fire for the start and stop events and look redundant with the others. As for why the level, I think it is important to always know if a server is going up and down. I don't think this adds that much to logs as it happens just once. I also think it is needed to debug which sever has failed to come up locally. (we run in info on AWS from what I understand, so no change to production, only locally where I think it is important to see that the boot process has failed to start properly).

jceaser: one problem I have with a bunch of these logs is that there is no way of connecting which search node is going down. I added :instance-name to the system object and would like to look into using that to be able to better report which node is shutting down, or more important, being able to count exactly how many match start ups and shutdowns there are so we can tell how AWS killed the software (forcibly or gracefully).

common-app-lib/src/cmr/common_app/app_events.clj
```diff
@@ -0,0 +1,74 @@
...
+       -m cmr.search.runner
+   >kill -15 $(jps | grep main | cut -f 1 -d ' ')"
+  [app-name]
+  (let [path-to-file (config/dump-diagnostics-on-exit-to)]
+    (infof "dump-heap: %s is writing a diagnostic file to %s..." app-name path-to-file)
+    (dump-heap app-name path-to-file false)))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(defn stop-on-exit-hook
```
TylerHeald1, line 53: Will using this hook function make sure that the stop command is actually run in the microservices? I know that until now searching for "system stopped" in logs doesn't really show anything

common-lib/src/cmr/common/system.clj
johnwteague: OK. Currently the task IP address is emitted with the log message. Any improvement you add to help identify the task is fine by me.
2024-01-23T21:56:11.949Z ip-10-6-65-205.ec2.internal INFO [cmr.search.system:?] - search System starting

common-app-lib/src/cmr/common_app/services/kms_fetcher.clj
johnwteague: Ok

common-app-lib/src/cmr/common_app/app_events.clj
johnwteague: Yeah, I think my point was that we are writing the same file name, which is defined by a configuration item rather than creating a coredump file name, dynamically, which serves to identify the app source. So every dump file, created from each of the tasks on every shutdown will have the same name.

search-app/src/cmr/search/runner.clj
johnwteague: You mean there are multiple logs with the same info? What's the text of the common-lib/src/cmr/common/system.clj log?

common-app-lib/src/cmr/common_app/app_events.clj
jceaser: yes, that is a problem locally, I can change that

ingest-app/src/cmr/ingest/runner.clj
johnwteague: Yeah, I understand. My question is one of how to manage the all of the dump files we don't care about (nominal shutdown) vs the ones we do care about (AWS Fargate sends the SIGTERM prior to the SIGKILL). The real value of this ticket and functionality is that we capture the stack trace at the time of task failure, during Fargate initiated termination. We need to be able to separate the wheat from the chaff, since the task (and we have 7 of them for Search and Ingest) will create this file everytime we deploy.
This needs to be taken up as part of the next ticket, which notionally is simply "writing the file to some accessible place" but really also needs to be identification and management so that we don't overwrite the files we want to preserve for investigation with those which are just generated as part of the normal shutdown process.

common-app-lib/src/cmr/common_app/app_events.clj
jceaser: yes, it should, unless something uses either kill -9 or the Task is blown away at the top. AWS should try to shutdown nicely, if that is the case we will see our shutdown messages. Locally I test this with `kill -15`.

common-app-lib/src/cmr/common_app/app_events.clj
johnwteague: Capturing the time might be interesting yeah, thanks. And yeah, I'd expect that AWS would kill the task if shutdown hung for any reason, including this.

common-app-lib/src/cmr/common_app/app_events.clj
johnwteague: Hm, well, ok, that was kinda my question. Are you using the app specific name as part of the target file? Bc it doesn't look like path-to-file (dump-heap app-name path-to-file false))) includes the app-name? Sorry for not seeing it, but where is that used in the file name? (obv you have generated these files so know what the name is)

common-app-lib/src/cmr/common_app/app_events.clj
johnwteague: We will know when we deploy and then re-deploy, initiating shutdown. We may not know what happens in "bad" situations until one occurs naturally however.

johnwteague (approved):

common-app-lib/src/cmr/common_app/app_events.clj
jceaser: I reworked some of this (the logging) to be more clear, now all apps will report in the same way, not just the two I originally added the SIGTERM response to. The file name should be unique to application if your running multiple applications.

ingest-app/src/cmr/ingest/runner.clj
jceaser: I don't have a way to do that. The process to turn on the dump is an ENV variable that is only settable at deploy time, however, I could check some other resource or provide an end point you could hit to toggle the behavior on. In fact, I don't see why that would be a problem, to write an end point (like clear cache) that you hit when you want to start capturing dumps.

common-app-lib/src/cmr/common_app/app_events.clj
jceaser: right. I'm very interested in seeing how this behaves in the real world.

TylerHeald1 (approved):

common-app-lib/src/cmr/common_app/app_events.clj
johnwteague: Ok, thanks. But I think we will revisit this in the next ticket. Also, picking a random number between 0,1000 doesn't guarantee uniqueness so it may or may not actually generate 1000 discrete filenames. Timestamp may be the better way to go. 
>even if you're running multiple Tasks in AWS 
Note, it's very uncommon for us to only have a single task instance for any deployed CMR service in any environment. Bootstrap is the only one I can think of bc that's kinda treated as a singleton.

PR #2048, opened 2024-01-31
EDSC-3935 adds redis cache serverless deployment for browse-scaler

daniel-zamora:


browse-scaler/serverless.yml
```diff
@@ -0,0 +1,53 @@
+service: browse-scaler-${self:provider.stage}
```
eudoroolivares2016, line 1: Don't' we need this as a PR on EDSC since this inf is meant to go on that account?

PR #2057, opened 2024-02-12
Add PR Template

jmaeng72:
I believe that having a standard PR template will help clarify PRs for reviewers and help developers go through testing processes

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -21,8 +21,8 @@ Please put in details of performance results from testing in WL
 # Checklist
 
 - [ ] I have updated/added unit and integration tests that prove my fix is effective or that my feature works
-- [ ] New and existing unit tests pass locally with my changes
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,29 @@
...
+
+### What areas of the application does this impact?
+
+List impacted areas.
+
+# Testing
+
+### Performance Test Results
+
+Please put in details of performance results from testing in WL
```
jmaeng72, line 19: If we are uncomfortable putting performance results, then we can add this as a checklist instead. Please LMK what you think

isja17, line 19: we shouldn't require WL runs to be ran pre-pr. Most tickets won't even need WL runs.

jmaeng72: That makes sense, I can remove this

isja17 (approved):

eudoroolivares2016 (approved):
Great change

PR #2059, opened 2024-02-15
CMR-9684: Part 2 - Switching humanizer report from 24 hr interval to daily schedule

jmaeng72:
Updating the humanizer-report to trigger an hour after the collection metadata cache does

### DIFF
--- a/search-app/src/cmr/search/services/humanizers/humanizer_report_service.clj
+++ b/search-app/src/cmr/search/services/humanizers/humanizer_report_service.clj
@@ -197,8 +197,10 @@
 
 (defn refresh-humanizer-report-cache-job
   [job-key]
-  "The job definition used by the system job scheduler."
+  "The job definition used by the system job scheduler to refresh the humanizer report cache.
+  This cache is directly reliant on the collection-metadata-cache being refreshed and present (refresh-collections-metadata-cache-job).
+  Currently the collection metadata cache is refreshed daily at 6 AM UTC. So we will start this job an hour after that."
   {:job-type HumanizerReportGeneratorJob
    :job-key job-key
-   :interval (* 60 60 24) ; every 24 hours
-   :start-delay (+ (default-job-start-delay) (humanizer-report-generator-job-delay))})
+   :start-delay (* 10 60) ;; 10 minutes delay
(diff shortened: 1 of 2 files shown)


search-app/src/cmr/search/services/humanizers/humanizer_report_service.clj
```diff
@@ -197,8 +197,10 @@
...
-  "The job definition used by the system job scheduler."
+  "The job definition used by the system job scheduler to refresh the humanizer report cache.
+  This cache is directly reliant on the collection-metadata-cache being refreshed and present (refresh-collections-metadata-cache-job).
+  Currently the collection metadata cache is refreshed daily at 6 AM UTC. So we will start this job an hour after that."
   {:job-type HumanizerReportGeneratorJob
    :job-key job-key
-   :interval (* 60 60 24) ; every 24 hours
-   :start-delay (+ (default-job-start-delay) (humanizer-report-generator-job-delay))})
+   :start-delay (* 10 60) ;; 10 minutes delay
+   :daily-at-hour-and-minute [07 00]})
```
johnwteague, line 206: Would it be better to use a modified humanizer-report-generator-job-delay, which is still at 400s I think? Maybe I don't get the point of the start-delay here. You're scheduling at 7:00 UTC, plus 10 min delay, so 7:10 UTC start time?

johnwteague: This ticket is already Verified Internal, how are we tracking this change?

jmaeng72: I put the delay in there, for the current bootstrap problem where all our caches start at deployment and cause errors. However, it will not be necessary once the bootstrap problem is fixed in the future.

I will create a new ticket for this and update the PR name to reflect that

johnwteague: Oh, the 'start-delay' defines a delay after application start? haha, I didn't know what it meant. Thought it was delay after the scheduled time or something, that's why I was confused.
Thanks!

isja17 (approved):

PR #2069, opened 2024-02-29
CMR-9748 Moving collection metadata cache to its own and separating r…

eereiter:
…eaders and writers.

# Overview

### What is the feature/fix?
Moving collection metadata cache to its own and separating readers and writers.

### What is the Solution?

Added some new configuration parameters for the new cache and updated the redis code to use them based on the passed in cache key.

redis

# Checklist

- [N/A] I have updated/added unit and integration tests that prove my fix is effective or that my feature works
- [ X] New and existing unit and int tests pass locally and remotely with my changes
- [X] I have made corresponding changes to the documentation


redis-utils-lib/src/cmr/redis_utils/config.clj
```diff
@@ -35,13 +63,40 @@
 
 (defconfig redis-retry-interval
   "The number of milliseconds to wait between Redis retry command."
-  {:default 100 :type Long})
+  {:default 400 :type Long})
 
 (defn redis-conn-opts
```
TylerHeald1, line 68: All the new option lists are ops, so either this one needs to change or the others should end with 'opts'

redis-utils-lib/src/cmr/redis_utils/redis.clj
```diff
@@ -6,39 +6,52 @@
...
+                      (error (format "Redis %s %s failed with exception %s"  wr-string# ~key e#)))))))]
+        (with-retry# (config/redis-num-retries))))
 
 (defn healthy?
   "Returns true if able to reach Redis."
   [& args]
   {:ok?
    (try
-     (= "PONG" (wcar* (carmine/ping)))
+     (= "PONG" (wcar* "healthy?" true (carmine/ping)))
```
TylerHeald1, line 47: Do we not need to use this function on the redis-collection-metadata-cache? With new code for wcar* it seems like 'healthy?' 'reset' and 'get-keys' will never be able to be used on the new redis cache

redis-utils-lib/src/cmr/redis_utils/redis.clj
```diff
@@ -6,39 +6,52 @@
...
+    The passed in key allows for specific redis caches as well as more logging information.
+    The read? is a true/false flag that signifies that a read-only replica redis connection
+    should be used. This flag is also used in more logging information.
+    For available commands refer to Redis docuemntation or:
+    https://github.com/ptaoussanis/carmine/blob/master/src/taoensso/carmine/commands.edn."
+    [key read? & body]
+    `(let [with-retry#
+            (fn with-retry# [num-retries#]
+              (let [wr-string# (if ~read? "read" "write")
+                    config# (if (= :collection-metadata-cache ~key)
```
TylerHeald1, line 20: The only usages of wcar* I saw outside of the defrecord RedisCache and defrecord RedisHashCache was for tests (as wcar* and the get-keys function in the same file) and in dev-system/src/cmr/dev_system/control.clj for the /reset and /clear-cache endpoints (with the reset function from this file).
Would we be able to add the connection options as a part of the redis cache record details and then just pass the connection options into wcar* as an argument so we can avoid this logic?

redis-utils-lib/src/cmr/redis_utils/config.clj
eereiter: fixed.

common-app-lib/src/cmr/common_app/services/search.clj
```diff
@@ -41,14 +42,18 @@
   "Returns a cache backed by Redis. This cache is used to store a map of cmr scroll-ids to ES scroll-ids 
    in a consistent way acrosss all instances of search."
   []
-  (redis-cache/create-redis-cache {:ttl (/ (scroll-id-cache-ttl) 1000)}))
+  (redis-cache/create-redis-cache {:ttl (/ (scroll-id-cache-ttl) 1000)
```
jceaser, line 45: I know you did not do this, but should 1000 be a def?

redis-utils-lib/src/cmr/redis_utils/config.clj
```diff
@@ -20,6 +44,10 @@
   "Timeout for connection in ms."
   {:default 3000 :type Long})
 
+(defconfig redis-collection-metadata-timeout-ms
+  "Timeout for connection in ms."
+  {:default 300000 :type Long})
```
jceaser, line 49: consider using math to show that this is 5 min, the zeros all run together with my eyes.  :)

redis-utils-lib/src/cmr/redis_utils/config.clj
```diff
@@ -35,13 +63,40 @@
 
 (defconfig redis-retry-interval
   "The number of milliseconds to wait between Redis retry command."
-  {:default 100 :type Long})
+  {:default 400 :type Long})
```
jceaser, line 66: why 4 times longer?

common-app-lib/src/cmr/common_app/services/search.clj
eereiter: Even better, since no other cashes use these time to live values I was able to just remove dividing by 1000 and change the original value from milliseconds to seconds.

redis-utils-lib/src/cmr/redis_utils/config.clj
eereiter: fixed.

redis-utils-lib/src/cmr/redis_utils/config.clj
eereiter: I logged how long it takes these connections.  While most of these transactions are under 100 milliseconds, I found that some times these transactions take 100-300 milliseconds, so just by increasing the timeout, I reduce the number of requests to redis during peak load.  When we re-try redis still has to finish the original task, but then we ask it to do it again, and if that one takes longer than 100 milliseconds we ask it to do it again, so under peak load it really adds un-necessarily more work.

jceaser: can you summarize this into a 1 sentence comment to warn a future developer from randomly making the value shorter then 400?

jceaser (approved):
consider my suggestion on a comment for "400" value.

PR #2090, opened 2024-04-15
CMR-9774: Updating the CMR 

eereiter:
# Overview

### What is the feature/fix?
Changing the acl software to mitigate linter findings.

### What is the Solution?
Removing not needed required statements and putting meaningless declarations in front of every macro so the linter doesn't complain about them. .

Acl library and all other dependencies.

# Checklist

- [x ] I have updated/added unit and integration tests that prove my fix is effective or that my feature works
- [x ] New and existing unit and int tests pass locally and remotely with my changes
- [x ] I have performed a self-review of my own code
- [x ] I have made corresponding changes to the documentation


umm-lib/resources/schema/iso_mends/resources/Codelist/eosCodelists.xml
```diff
@@ -187,4 +187,53 @@
          </gmx:codeEntry>
       </gmx:CodeListDictionary>
    </gmx:codelistItem>
+   <!--=== DS_AssociationTypeCode ===-->
```
TylerHeald1, line 190: For clarification, what do the XML changes have to do with linter issues?

acl-lib/dev/user.clj
```diff
@@ -1,12 +1,3 @@
-(ns user
```
TylerHeald1, line 1: Does removing these :require, :use, and reset affect local dev at all?

eereiter: Nothing.  Stuff from a previous merge got it.  I will need to redo this branch.

eereiter: No.  My guess is that someone who was developing the acl lib used things in the user namespace, then got rid of the code in the user namespace, but never got rid of the require statements.

umm-lib/resources/schema/iso_mends/resources/Codelist/eosCodelists.xml
eereiter: Nothing.  It looks like the code from my previous issue that I was working got into this PR.  I will need to redo my changes and issue a new PR.

PR #2098, opened 2024-04-22
Cmr 9840 Access Control search endpoints do not properly handle the search-after header

sxu123:
# Overview

### What is the feature/fix?

Pass the search-after header to context.

### What is the Solution?

Fixed the ACL search so that the search-after header could be passed to the code that handles it.

ACL search.

# Checklist

- [ x] I have updated/added unit and integration tests that prove my fix is effective or that my feature works
- [ x] New and existing unit and int tests pass locally and remotely with my changes
- [ x] I have performed a self-review of my own code
- [ x] My changes generate no new warnings


transmit-lib/src/cmr/transmit/http_helper.clj
```diff
@@ -105,6 +121,35 @@
...
+        response (http-response->raw-response-with-headers
+                   (client/request
+                     (merge connection-params
+                            {:url (url-fn conn)
+                             :method method
+                             :throw-exceptions false}
+                            (when use-system-token?
+                              {:headers {config/token-header (config/echo-system-token)}})
+                            http-options)))]
+    response))
```
eereiter, line 151: why can't we use the original request function on line 79?  Its the exact same function, except that it calls the raw-response-with-headers that I am questioning why we need it as well as the return of response, which I think is incorrect.  On line 151 I think it should be (response-handler request response).

transmit-lib/src/cmr/transmit/http_helper.clj
```diff
@@ -45,6 +45,22 @@
...
+  (let [content-type (mt/mime-type->format
+                       (mt/content-type-mime-type headers)
+                       ;; don't return a default
+                       nil)]
+    {:status status
+     :body (if (= content-type :json)
+             (safe-parse-json body)
+             body)
+     :content-type content-type
+     :headers headers}))
```
eereiter, line 62: Why do we need this function?  Can't we just add the headers to http-response->raw-response function on line 33?

transmit-lib/src/cmr/transmit/access_control.clj
```diff
@@ -201,6 +202,13 @@
   ([context params options]
    (search-for-acls* context params (merge options {:method :post}))))
 
+(defn search-for-acls-with-returned-headers
```
eereiter, line 205: Why can't we just use search-for-acls and put the header information in the downstream functions?  I don't think we need to duplicate all of the functions 1 to handle the exact same requests with headers and one without. We should always handle the headers.

eereiter (changes requested):
-

transmit-lib/src/cmr/transmit/access_control.clj
sxu123: because 1. it causes a lot of test failures, 2. if you look at the detail of the function, there is some differences other than returning headers.

transmit-lib/src/cmr/transmit/http_helper.clj
sxu123: I did that first. and it causes a lot of test failures.

sxu123:
> you will need to fix your merge conflict.

Done.

eereiter (approved):
I don't think we need 2 separate set of functions for search:  1 set to handle requests with headers and 1 set to handle requests without headers.  We should just handle headers for all search requests.

eudoroolivares2016:
The AC mentioned that we needed to update the Documentation for this I didn't see usage of paging on the existing `https://cmr.earthdata.nasa.gov/search/site/docs/access-control/api.html`

eudoroolivares2016 (approved):
Approving assuming that the documentation is added or that the AC is removed and put into a new ticket since this task is a blocker to other team

eudoroolivares2016 (changes requested):
I think we should just do the documentation in this ticket its in the AC

PR #2107, opened 2024-04-29
CMR-9850: Update usage of Testcontainers

TylerHeald1:
# Overview
Updating usage of Testcontainers
### What is the feature/fix?

updating our usage of testcontainers a bit

### What is the Solution?

Changed what we look for for container startup, but also the ordering of withStartupTimeout. Set startup timeout gets overwritten by setting a waitingFor WaitStrategy.
Also added a docker compose file that can potentially be used for local dev and addressed some linter output

Integration tests, local dev

# Checklist

- [ ] New and existing unit and int tests pass locally and remotely with my changes


dev-system/src/cmr/dev_system/system.clj
```diff
@@ -117,7 +117,7 @@
 
 (defmethod create-elastic :external
   [_]
-  (elastic-config/set-elastic-port! 9209))
```
jmaeng72, line 120: so we are no longer supporting external elastic? If so, then should we just get rid of this method entirely?

dev-system/compose.yaml
```diff
@@ -0,0 +1,41 @@
+services:
```
jmaeng72, line 1: When you say this docker compose can be 'potentially' used for local dev. Are you going to add how to do this in README?

dev-system/src/cmr/dev_system/system.clj
TylerHeald1: We can use external elastic just fine, it just doesn't seem there's a real need to set the port to 9209 when you can just use 9210 which is the default that CMR checks for elasticsearch on

dev-system/compose.yaml
TylerHeald1: Sure I can add a bit of documentation

jmaeng72 (approved):

eereiter (approved):

PR #2119, opened 2024-05-20
CMR-9891: parallel unit tests

jceaser:
# Overview

This is an effort to speed up unit test phase in bamboo by running unit tests in parallel.

### What is the feature/fix?
Run modules in parallel to take advantage of multi-core processors.

### What is the Solution?

To run the unit tests in parallel a python script is used to create a pool of jobs and a set of "work-bees" to process them. The script requests from lein the number of modules to be processed and the script then filters out known modules with no tests, and also a small set of modules that can not be run in parallel. All remaining modules have their unit tests run. When all these modules are done, the remaining modules run one at a time. 

the top level lein responds to `lein ci-utest` which will start the python script and run the tests.

In order for this to work, two things must happen:

1. a new cmr builder docker image is needed. The needed docker image has been created
2. the bamboo build scripts must be changed to take advantage of this new call

# Checklist

- [-] I have removed unnecessary/dead code and imports in files I have changed
- [x] I have made corresponding changes to the documentation


README.md
```diff
@@ -76,7 +76,7 @@ When running CMR locally, the values in this file do not need to be change, CMR
 This process will take quite a long time, be patient and monitor progress.
 
 #### Running CMR
-You can run CMR two different ways, everything in one or as separte. 
+You can run CMR two different ways, everything in one or as separte.
```
TylerHeald1, line 79: > separte
separate

bin/unit_test_script/color.py
```diff
@@ -0,0 +1,112 @@
...
+#mark - Imports
+
+from enum import Enum   #Creating Enums
+#import argparse         #command line parsing
+#import code             #interactive shell
+#import os               #file handling
+#import re               #filter in man()
+#import readline         #interactive shell
+#import subprocess       #calling unix commands
+#import sys              #exiting
```
TylerHeald1, line 22: remove commented out imports if not needed

jceaser: that is indeed legacy code

bin/unit_test_script/run_unit_tests.py
```diff
@@ -0,0 +1,159 @@
...
+from threading import Lock
+import datetime
+import multiprocessing
+import os
+import time
+
+# Originally written using https://github.com/jceaser/shellwrap, the two files
+# imported have instead been included so as to not require any pip actions when
+# running on bamboo ; trying to be as lights as posible
+#from shellwrap import color, unix
```
TylerHeald1, line 23: If shellwrap is not used, remove commented out import

bin/unit_test_script/unix.py
```diff
@@ -0,0 +1,59 @@
...
+""" template script for future scripts """
+
+#mark - Imports
+
+#from enum import Enum   #Creating Enums
+#import argparse         #command line parsing
+#import code             #interactive shell
+#import os               #file handling
+#import re               #filter in man()
+#import readline         #interactive shell
```
TylerHeald1, line 20: remove commented out imports

bin/unit_test_script/run_unit_tests.py
```diff
@@ -0,0 +1,159 @@
...
+
+        color.cprint(color.tcode.white, f"+task {id_number} working on {task}.",
+            environment=env)
+
+        st = time.time()
+        # Run the external command in the task directory inside a try/except
+        # block, to ensure thread never dies
+        try:
+            os.chdir(base+"/"+task)
+            unix.pipe(["lein", "ci-utest"])
```
TylerHeald1, line 101: Why is this a pipe?

bin/unit_test_script/unix.py
```diff
@@ -0,0 +1,59 @@
...
+    for cmd in cmd_lists:
+        if pipe_result is None:
+            pipe_result = subprocess.run(cmd, check=True, capture_output=True)
+        else:
+            pipe_result = subprocess.run(cmd, input=pipe_result.stdout,
+                capture_output=True, check=True) # use check?
+            #print (f"return:{pipe_result.returncode}")
+    return pipe_result.stdout.decode('utf-8')
+
+def curl(*flags):
```
TylerHeald1, line 50: This doesn't seem to be used, remove if unneeded

bin/unit_test_script/unix.py
```diff
@@ -0,0 +1,59 @@
...
+    """ Build a curl command list which can be passed to pipe() or call() """
+    cmd = ["curl", "-s", '-A', 'tcherry-script']
+    cmd.extend(flags)
+    return cmd
+
+def ccurl(*flags):
```
TylerHeald1, line 56: This doesn't seem to be used, remove if unneeded

bin/unit_test_script/run_unit_tests.py
```diff
@@ -0,0 +1,159 @@
...
+import color # add color and verbose printing
+import unix # unix command calling
+
+# for a more descriptive run locally, remove these settings
+env = {"color": False, "verbose": color.VMode.ERROR}
+
+# There is nothing wrong with using a couple of globals vars to manage a bunch
+# of threads because this is way cleaner then doing it all in bash
+# pylint: disable=global-statement
+# pylint: disable=global-variable-not-assigned
```
TylerHeald1, line 33: remove unneeded commented code

bin/unit_test_script/run_unit_tests.py
TylerHeald1: Since there is only one command being used, and the output of it isn't getting used, can we just use the subprocess.run directly?

jmaeng72 (approved):

bin/unit_test_script/run_unit_tests.py
jceaser: so this comes from a library function, I included the file so I would not need to do any pip actions, do you really want me to strip out all the functions because this tool does not use it?

TylerHeald1 (approved):

PR #2124, opened 2024-05-30
CMR-9777: lint common-lib

jceaser:
# Overview

NOTE: bamboo is still building this

### What is the feature/fix?

Ran clj-kondo, bikeshed, and kibit against the common lib

Please summarize the feature or fix.

### What is the Solution?

* General cleanups
* Moved a Regex API into the spatial lib which is the only user of this code
* Moved some validation code out of common and into cmr.umm-spec.validation.util which is shared between the client libraries

List impacted areas.
* common-lib
* spatial-lib
* metadata-db
* search-app
* umm-spec

# Checklist

- [x] I have updated/added unit and integration tests that prove my fix is effective or that my feature works


common-lib/src/cmr/common/api/context.clj
```diff
@@ -61,6 +61,6 @@
   ([context]
    (context->user-id context "Valid user token required."))
   ([context msg]
-   (if-let [token (:token context)]
+   (if-let [_token (:token context)]
```
eereiter, line 64: I changed this function - in order not to get a merge conflict you should change this to
(if (:token context) 

you don't need the let statement because the token variable is not used.

common-lib/src/cmr/common/api/errors.clj
```diff
@@ -46,12 +49,7 @@
            (csk/->PascalCaseString path-item)))
        field-path))
 
-(defmulti errors->body-string
-  "Converts a set of errors into a string to return in the response body
-  formatted according to the requested response format."
-
-  (fn [response-format errors]
-    response-format))
+;; *************************************
```
eereiter, line 52: Why has ;; *******... been added?

common-lib/src/cmr/common/api/errors.clj
```diff
@@ -66,33 +64,46 @@
   [error]
   (update-in error [:path] keyword-path->string-path))
 
-(defmethod errors->body-string mt/json
-  [_ errors]
-  (json/generate-string {:errors (map error->json-element errors)}))
+;; *************************************
```
eereiter, line 67: why has ;; **** ... been added?

common-lib/src/cmr/common/api/errors.clj
jceaser: I want a more physical separation between sets of defmulti calls. error->xml-element and errors->body-string are right on top of each other and I want it clear where they each are.

eereiter: Looking at the full file I now see why.
Can you just convert the three functions into 1 and use a case statement? It will execute much faster that way.  defmulti/defmethods are not needed here.

common-lib/src/cmr/common/api/errors.clj
eereiter: Looking at the full file I now see why.
Can you just convert the three functions into 1 and use a case statement? It will execute much faster that way.  defmulti/defmethods are not needed here.

jceaser: I"m only doing this when I find there are more then one defmulti in a file and it seams hard to tell them apart.

common-lib/src/cmr/common/api/context.clj
jceaser: will show up in my next update.

common-lib/src/cmr/common/api/errors.clj
```diff
@@ -66,33 +64,46 @@
...
 
 (defmulti error->xml-element
   "Converts an individual error element to the equivalent XML structure."
   type)
 
 (defmethod error->xml-element String
   [error]
-  (x/element :error {} (mask-token-error error)))
+  (xml/element :error {} (mask-token-error error)))
```
eereiter, line 75: Can you just convert the three functions into 1 and use a case statement? It will execute much faster that way.  defmulti/defmethods are not needed here.

common-lib/src/cmr/common/api/errors.clj
jceaser: You will see that I did this in other files as well. I also found that in some cases no care was taken to keep the order of things clear, defmethod of different types would be mixed with functions in between them too. I cleaned these up moving the functions outside of the "blocks" and ordering defmulti and defmethods so as to not break load order but also collect them together. The comment line I think may help enforce the seperation with future developers.

common-lib/src/cmr/common/api/errors.clj
```diff
@@ -66,33 +64,46 @@
...
         :path {} (string/join "/" (keyword-path->string-path path)))
-      (x/element
-        :errors {} (for [error errors] (x/element :error {} error))))))
+      (xml/element
+        :errors {} (for [error errors] (xml/element :error {} error))))))
+
+;; *************************************
+
+(defmulti errors->body-string
+  "Converts a set of errors into a string to return in the response body
```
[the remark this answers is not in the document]
jceaser: yea, I probably could do that. Let me search to see if these are only in this file. I tried that in another place and had to roll it back but that was where class was used and clojure apparently handles class differently.

common-lib/src/cmr/common/cache/cache_spec.clj
```diff
@@ -1,13 +1,14 @@
 (ns cmr.common.cache.cache-spec
   "Defines a common set of tests for checking cache implementations"
-  (:require [clojure.test :refer :all]
-            [cmr.common.cache :as c]
+  (:require [clojure.test :refer [are is testing]]
```
eereiter, line 3: our standards have the [clojure.test ... on the next line and lines up with : (colon)   Look at the next file as an example.

common-lib/src/cmr/common/util.clj
```diff
@@ -198,40 +198,6 @@
                        "Timed function %s/%s took %d ms."
                        ~ns-str ~fn-name-str elapsed#)))))))))))
 
-(defn ^{:deprecated true} build-validator
-  "Creates a function that will call f with it's arguments. If f returns any
-  errors then it will throw a service error of the type given.
-
-  DEPRECATED: we should use the validations namespace."
```
TylerHeald1, line 205: I believe this is intended for the cmr.common.validations namespace, not the cmr.umm-spec.validation

umm-spec-lib/src/cmr/umm_spec/validation/util.clj
TylerHeald1, line 1: Would recommend moving to cmr.common.validations instead

isja17 (approved):

common-lib/src/cmr/common/util.clj
```diff
@@ -1039,12 +1009,12 @@
...
                    (= "Earthdata Login" (:origin header-data)))
               false)
-            (catch com.fasterxml.jackson.core.JsonParseException e false))
+            (catch com.fasterxml.jackson.core.JsonParseException _e false))
           false))
       false)))
 
 ;; Note: Similar code exists at gov.nasa.echo.kernel.service.authentication
-(def URS_TOKEN_MAX_LENGTH 100)
+(def URS_TOKEN_MAX_LENGTH "Back in the day, this is how you knew it was URS" 100)
```
eereiter, line 1017: funny!

common-lib/src/cmr/common/xml.clj
```diff
@@ -2,9 +2,8 @@
   "Contains XML helpers for extracting data from XML structs created using clojure.data.xml.
   See the test file for examples."
   (:require [cmr.common.date-time-parser :as p]
-            [cmr.common.services.errors :as errors]
-            [clojure.string :as string]
-            [clojure.java.io :as io])
+            [clojure.string :as string])
```
eereiter, line 5: This could be moved over to below the colon.

common-lib/src/cmr/common/xml.clj
```diff
@@ -2,9 +2,8 @@
...
+  #_{:clj-kondo/ignore [:unused-import]}
   (:import javax.xml.validation.SchemaFactory
```
eereiter, line 7: same here.

common-lib/test/cmr/common/test/mime_types.clj
```diff
@@ -1,46 +1,47 @@
 (ns cmr.common.test.mime-types
   "Tests for mime-type functions."
   (:require
-   [clojure.string :as str]
-   [clojure.test :refer :all]
+   [clojure.test :refer [are deftest is testing]]
    [cmr.common.mime-types :as mt]
-   [cmr.common.util :refer [are2]]))
+   [cmr.common.util :refer [are3]]))
```
eereiter, line 6: Nice!

eereiter (approved):

common-lib/src/cmr/common/api/errors.clj
jceaser: I converted the last one, the other two are based on java classes and I did not have much luck with that.

common-lib/src/cmr/common/cache/cache_spec.clj
jceaser: I have been converting those when I notice them, I'll change this one.

common-lib/src/cmr/common/cache/in_memory_cache.clj
```diff
@@ -120,15 +122,15 @@
 
 (defmulti create-core-cache
   "Create a cache using cmr.core-cache of the given type."
-  (fn [type value opts]
+  (fn [type _value _opts]
```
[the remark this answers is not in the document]
jceaser: if you can figure out how to do size-in-bytes in that same file, then I'll be impressed.

common-lib/src/cmr/common/util.clj
jceaser: I removed the need for that and moved this into what library I could that was shared between metadata-db and search. I'm not sure why it was deprecated.

umm-spec-lib/src/cmr/umm_spec/validation/util.clj
jceaser: It is mostly used by metadata-db. Why was the code marked for deprecation?

common-lib/src/cmr/common/util.clj
jceaser: it's really hard coming up with doc strings after the fact

common-lib/src/cmr/common/xml.clj
jceaser: also alphabatized now.

common-lib/test/cmr/common/test/mime_types.clj
jceaser: I got a few of these!

common-lib/src/cmr/common/cache/in_memory_cache.clj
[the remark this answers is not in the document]
jceaser: specifically I could not get it to handle clojure.lang.IPersistentMap correctly. In my testing maps went to the default and I'm not sure why. Defmethod is the preferred way, I read, to handle java class branching code.

zimzoom (approved):

TylerHeald1 (approved):

PR #2128, opened 2024-06-05
CMR-9185: Require clj-kondo in GitHub

jceaser:
# Overview

Require that code passed clj-kondo before allowing a merge into master

### What is the feature/

This change will add clj-kondo as a step to a GitHub action which will be successful if there are no errors from clj-kondo. Warnings will be allowed to pass.

### What is the Solution?

The solution is to call clj-kondo in a docker container against the 7 sub-projects that have been cleaned up. The list of sub projects is expected to increase to all. To do this, a new script has been added to dev-system/support/run-kondo.sh to be run by the clj-kondo docker container. 

This should impact changes made to access-control-app/src, acl-lib/src, common-app-lib/src, common-lib/src, elastic-utils-lib/src, es-spatial-plugin/src, message-queue-lib/src

# Checklist

- [-] New and existing unit and int tests pass locally and remotely with my changes
- [-] I have removed unnecessary/dead code and imports in files I have changed
- [-] My changes generate no new warnings

### DIFF
--- a/dev-system/support/run-kondo.sh
+++ b/dev-system/support/run-kondo.sh
@@ -11,7 +11,7 @@
 # error or warning
 fail_level=error
 
-# true or anything but true
+# set to the string true to always pass, anything else will be considered a false
 always_pass=false
 
 # A list of directories to run kondo on
--- a/dev-system/support/run-kondo.sh
+++ b/dev-system/support/run-kondo.sh
@@ -1,0 +1,52 @@
+#!/bin/bash
+
+# **************************************************************************************************
+# This script is meant to be run from the CMR project root by Github.com as a workflow action step.
+# The purpouse of this script is to run all the linter checks using clj-kondo in a docker image and
+# to return a failed error code if there are any errors.
+
+fail_level=error
+always_pass=false
+projects_to_check='access-control-app/src \
+            acl-lib/src \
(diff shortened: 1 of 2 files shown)


jceaser:
if you want to see an example of this code running, you can take a look at my fork:

https://github.com/jceaser/Common-Metadata-Repository/actions/runs/9386228406

TylerHeald1 (approved):

dev-system/support/run-kondo.sh
```diff
@@ -0,0 +1,63 @@
...
+# The purpouse of this script is to run all the linter checks using clj-kondo in a docker image and
+# to return a failed error code if there are any errors.
+#
+# This script is designed to be called from inside a docker image by github action step like this:
+#     run: ./dev-system/support/run-kondo.sh >> report.log
+
+# error or warning
+fail_level=error
+
+# true or anything but true
```
jmaeng72, line 14: unsure what this comment means

jmaeng72 (approved):

dev-system/support/run-kondo.sh
jceaser: bash does not have true and false, just strings, so "true" is the only thing it checks for, but I'll clear that up.

PR #2145, opened 2024-07-03
CMR-10009: Fix job-router load balancer referencing issues

TylerHeald1:
# Overview

### What is the feature/fix?

If some infrastructure change causes the LB to be remade with the same name, it might not have the same DNS name, which requires job-router to need it's variables updated

### What is the Solution?

Changed job-router to use the name of the LB to query AWS for the DNS name, so it isn't reliant on a more-likely-to-change value

job-router

# Checklist

- [x ] I have updated/added unit and integration tests that prove my fix is effective or that my feature works
- [x ] I have made corresponding changes to the documentation
- [x ] I have cleaned up integration tests by doing one or more of the following:

### DIFF
--- a/job-utilities/job_router/lambda_function.py
+++ b/job-utilities/job_router/lambda_function.py
@@ -30,8 +30,8 @@ def handler(event, _):
     if environment is None:
         print("ERROR: Environment variable not set!")
         error_state = True
-    if cmr_url is None:
-        print("ERROR: CMR_URL variable not set!")
+    if cmr_lb_name is None:
+        print("ERROR: CMR_LB_NAME variable not set!")
         error_state = True
     if error_state:
         sys.exit(1)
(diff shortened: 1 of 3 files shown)


job-utilities/job_router/lambda_function.py
```diff
@@ -19,36 +19,40 @@ def handler(event, _):
...
-    cmr_url = os.getenv('CMR_LB_URL')
+    cmr_lb_name = os.getenv('CMR_LB_NAME')
     service = event.get('service', 'bootstrap')
     endpoint = event.get('endpoint')
     single_target = event.get('single-target', True)
-
-    error_state = False
+    request_type = event.get('request-type', "GET")
 
     if environment is None:
```
johnwteague, line 28: Just curious: you are trying to avoid having multiple error exit points so that's why it's deferred? Otherwise you could just exit as soon as you tried and failed to get the environment

johnwteague (approved):

job-utilities/job_router/lambda_function.py
TylerHeald1: Figured it was fine to do it this way with only a couple variables, but anything larger probably would've had it done one by one

jmaeng72 (approved):

eereiter (approved):

PR #2146, opened 2024-07-17
CMR-10020 - Adding indexes to granule tables to speed up DB queries for cleaning up old concepts.

eereiter:
# Overview

### What is the feature/fix?

Adding indexes to granule tables to speed up DB queries for cleaning up old concepts.

### What is the Solution?

Adding indexes to granule tables to speed up DB queries for cleaning up old concepts.

Granule Tables

# Checklist

- [X] New and existing unit and int tests pass locally and remotely with my changes

### DIFF
--- a/metadata-db-app/src/cmr/metadata_db/migrations/090_add_granule_delete_revision_index.clj
+++ b/metadata-db-app/src/cmr/metadata_db/migrations/090_add_granule_delete_revision_index.clj
@@ -11,7 +11,7 @@
     (try
       (if (string/includes? table-name "SMALL_PROV")
         (mh/sql (format "create index %s_DR ON %s (provider_id, deleted, revision_date)" table-name table-name))
-        (mh/sql (format "create index %s_DD ON %s (deleted, revision_date)" table-name table-name)))
+        (mh/sql (format "create index %s_DR ON %s (deleted, revision_date)" table-name table-name)))
       (catch java.sql.BatchUpdateException e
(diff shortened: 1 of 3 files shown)


metadata-db-app/src/cmr/metadata_db/migrations/090_add_granule_delete_revision_index.clj
```diff
@@ -0,0 +1,25 @@
...
+      (catch java.sql.BatchUpdateException e
+        ;; If the index already exists we can ignore the error. This could
+        ;; happen if a migration was moved down, then back up.'
+        (when-not (.contains (.getMessage e) "ORA-00955")
+          (throw e))))))
+
+(defn down
+  "Migrates the database down from version 90."
+  []
+  ;; do nothing as we want to keep the new indexes
```
jmaeng72, line 24: So if we roll back to a previous version of the db, we still want these indices to exist?

eereiter: This is what I did on the 089 as well.  I think we want to keep these indexes on the tables until the tables get deleted. Even if we rolled back the database 20 versions back for several years, I think we still want these indexes on the database tables because it speeds up finding the data to be deleted.

jmaeng72 (approved):

TylerHeald1 (approved):

PR #2148, opened 2024-07-23
Feature: CMR-10013 - Removal of sha/md5 by default

jceaser:
# Overview

The inclusion of SHA1 and MD5 output in the request logs really have no practical use for every day logs and are of minor interest outside of debugging.

This change will push these values to debug only status, meaning they will only display when logs are set to debug or more verbose.

This impacts the custom access log

### DIFF
--- a/common-app-lib/src/cmr/common_app/api/request_logger.clj
+++ b/common-app-lib/src/cmr/common_app/api/request_logger.clj
@@ -141,7 +141,7 @@
              ;; that reporting scripts can try to protect against change. Humans
              ;; may be creating Splunk or ELK reports based on this content.
              note (-> {"log-type" "action-log" "log-version" "1.0.0"}
-                      ;(assoc-hashes response)
+                      (assoc-hashes response)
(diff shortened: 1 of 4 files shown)


common-app-lib/src/cmr/common_app/api/request_logger.clj
```diff
@@ -82,6 +82,18 @@
                                (assoc-in [:headers "Content-SHA1"] body-sha1))]
       updated-response)))
 
+(defn assoc-hashes
```
jmaeng72, line 85: can we rename this func to be more clear of what it does? something like 'add-hash-to-data'?

TylerHeald1 (approved):

common-app-lib/src/cmr/common_app/api/request_logger.clj
jceaser: glad you said something, I also don't like the name.

zimzoom (approved):

PR #2150, opened 2024-07-24
Feature: CMR-10025 - Condensed Index Event logs

jceaser:
# Overview

### What is the feature/fix?

Please summarize the feature or fix.

### What is the Solution?

There are three logs which generate logs in the millions per day. These three logs are for the same indexing event and represent: Start, finish, time to visibility.

The change here is to introduce a new configuration allowing Indexer to issue a single log entry when reporting time to visibility with additional information found in the start/stop log entries. This new log is also in json and is as short as possible.

There are also Kondo updates in this change.

# Checklist

- [x] I have updated/added unit and integration tests that prove my fix is effective or that my feature works

### DIFF
--- a/indexer-app/src/cmr/indexer/api/routes.clj
+++ b/indexer-app/src/cmr/indexer/api/routes.clj
@@ -1,18 +1,14 @@
 (ns cmr.indexer.api.routes
   "Defines the HTTP URL routes for the application."
   (:require
-   [cheshire.core :as json]
-   [clojure.stacktrace :refer [print-stack-trace]]
-   [clojure.string :as string]
    [clojure.walk :as walk]
-   [compojure.core :refer :all]
+   [compojure.core :refer [DELETE GET POST PUT context routes]]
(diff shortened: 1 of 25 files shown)


indexer-app/src/cmr/indexer/api/routes.clj
```diff
@@ -35,34 +31,34 @@
 (def ^:private index-set-routes
   "Routes providing index-set operations"
   (context "/index-sets" []
-    (POST "/" {body :body request-context :request-context params :params headers :headers}
```
jmaeng72, line 38: why did we get rid of the params and headers for this endpoint?

zimzoom (approved):

indexer-app/src/cmr/indexer/api/routes.clj
jceaser: kondo told me to

jmaeng72 (approved):

PR #2151, opened 2024-07-24
CMR-10023: Part 1

jmaeng72:
# Overview

### What is the feature/fix?

Change logs from info to debug for search api and data folders

### What is the Solution?

Switch logs from info to debug for logs that are already repeated in the REPORT log

Search App logs

# Checklist

- [ ] New and existing unit and int tests pass locally and remotely with my changes

### DIFF
--- a/search-app/src/cmr/search/api/concepts_lookup.clj
+++ b/search-app/src/cmr/search/api/concepts_lookup.clj
@@ -3,7 +3,7 @@
   (:require
    [cmr.common-app.api.routes :as common-routes]
    [cmr.common.concepts :as concepts]
-   [cmr.common.log :refer (debug info warn error)]
+   [cmr.common.log :refer (debug)]
    [cmr.common.mime-types :as mt]
    [cmr.common.services.errors :as svc-errors]
    [cmr.search.api.core :as core-api]
@@ -73,13 +73,13 @@
 (defn- find-concept-by-concept-id*
   "Perfrom the retrieval of concept by concept id and revision id"
   ([ctx result-format concept-id]
-   (info (format "Search for concept with cmr-concept-id [%s]" concept-id))
+   (debug (format "Search for concept with cmr-concept-id [%s]" concept-id))
    (core-api/search-response
     ctx
     (query-svc/find-concept-by-id ctx result-format concept-id)))
 
   ([ctx result-format concept-id revision-id]
-   (info (format "Search for concept with cmr-concept-id [%s] and revision-id [%s]"
+   (debug (format "Search for concept with cmr-concept-id [%s] and revision-id [%s]"
                  concept-id
                  revision-id))
    (core-api/search-response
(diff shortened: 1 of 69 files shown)


search-app/src/cmr/search/api/concepts_lookup.clj
```diff
@@ -73,13 +73,13 @@
 (defn- find-concept-by-concept-id*
   "Perfrom the retrieval of concept by concept id and revision id"
   ([ctx result-format concept-id]
-   (info (format "Search for concept with cmr-concept-id [%s]" concept-id))
+   (debug (format "Search for concept with cmr-concept-id [%s]" concept-id))
```
isja17, line 76: Putting this comment here but really it applies to most of these changes.

Removing the start log is an interesting approach. I _think_ that we don't use them in dashboards, reports, etc - a quick look through some dashboards showed that we use the "found x <concept> in y ms" logs. We need to double, triple check that that's true before merging this.  I am a little bit concerned about losing end-to-end visibility of a searches lifecycle, but that might be fine.

search-app/src/cmr/search/api/tags.clj
```diff
@@ -101,7 +99,7 @@
   (verify-tag-modification-permission context :update)
   (common-enabled/validate-write-enabled context "search")
   (validate-tag-content-type headers)
-  (info (format "Tagging [%s] on collections by query: %s by client: %s."
+  (debug (format "Tagging [%s] on collections by query: %s by client: %s."
```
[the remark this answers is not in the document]
isja17: that's a reason to cut it imo. we might be surprised by how much we gain by cutting out all of these small logs that only show up a handful of times a day. i would bet we have a lot of them throughout all of apps, and they are mostly noise.

search-app/src/cmr/search/api/concepts_lookup.clj
zimzoom: IIRC the "found x in y ms" logs only get created when results are found. So searches that don't find anything or error out won't get captured

search-app/src/cmr/search/api/tags.clj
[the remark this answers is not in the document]
zimzoom: seconded; logs should be cut across the board based on usefulness alone and not predicted impact to target. Minimalism for the long game and quality of life.

search-app/src/cmr/search/api/concepts_lookup.clj
zimzoom: But I think the access logs should cover it

zimzoom (approved):

jceaser (approved):

search-app/src/cmr/search/api/concepts_lookup.clj
jmaeng72: I went through the Splunk dashbaords and alerts and did not find any query that uses 'Search for concept...' I only found the one using 'Searching for...'

Ad for the actual info, this is already captured in the REPORT log

isja17: > IIRC the "found x in y ms" logs only get created when results are found. So searches that don't find anything or error out won't get captured

Yeah, that's what my comment was saying - we don't use these for reports. IIRC, john had some reason to not get rid of this particular log though, right?

jmaeng72: I will put these logs back

PR #2156, opened 2024-07-30
CMR-10023: Part 2 Reduce Search Logs

jmaeng72:
# Overview

### What is the feature/fix?

Reducing search logs in the src/cmr/search/services folder

### What is the Solution?

Remove some associations logs

Search

# Checklist

### DIFF
--- a/search-app/src/cmr/search/services/query_service.clj
+++ b/search-app/src/cmr/search/services/query_service.clj
@@ -211,7 +211,7 @@
         result-format (rfh/printable-result-format (:result-format query))
         log-message (log-search-result-metadata (:hits results) (name concept-type)
                                                 total-took (:client-id context) (:token context) result-format
-                                                "" nil)]
+                                                "with params %s" (pr-str params))]
     (info (cond
             scroll-id (format "%s, scroll-id: %s." log-message (str (hash scroll-id)))
             search-after (format "%s, search-after: %s, new search-after: %s."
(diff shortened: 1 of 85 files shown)


search-app/src/cmr/search/services/query_service.clj
```diff
@@ -485,7 +485,7 @@
         pr-result-format (rfh/printable-result-format result-format)
         log-message (log-search-result-metadata hits "deleted collections"
                                                 total-took (:client-id context) (:token context) pr-result-format
-                                                "with params %s" (pr-str params))]
+                                                "" nil)]
     (info log-message)
```
jmaeng72, line 489: So for the "Found X granules/collections in Y ms from client Z logs", we also print out the params here.

Example)
Found 12 granules in 144 ms from client asf_search_v7.0.7, token_type null in format :umm-json-results 1.6.6 with params {...}

which can get quite large.

Since we already get the same params values further up the log chain from the "Searching for ...." logs that means these params logs in the "Found x granules..." logs are repeats and can be removed.

Open to thoughts

jmaeng72: I kept the params parameter in the log-search-result-metadata func intact, because another func calls it with "with JSON Query %s and query params...". I was close to removing this log altogether, but decided to keep the json query one.

jceaser: did someone not have a comment about passing 'nil' as a parameter before and not like the practice? Can't the function be changed to accept two fewer parameters?

jmaeng72: We're not going this route, so reverted the changes

isja17 (approved):

eereiter (approved):

PR #2157, opened 2024-08-02
Feature CMR-10024 : Less Timed Logs from Ingest

jceaser:
# Overview

Remove large quantities of logs which are not helpful.

### What is the feature/fix?

1. Removing defn-timed calls for functions which are either fast enough to not worry about, infrequent, or fast and frequent but inside another defn-timed.

2. Added a new macro, defn-timed-level which allows code to specify the log level to send timed info to. Using this for one log.

### What is the Solution?

removing defn-timed macro from function definition.

Most changes are in ingest, one change is in common.

I have also run clj-kondo on ingest and made significant changes.

# Checklist

### DIFF
--- a/common-lib/src/cmr/common/util.clj
+++ b/common-lib/src/cmr/common/util.clj
@@ -163,11 +163,48 @@
          result# (do ~@body)]
      [(- (System/currentTimeMillis) start#) result#]))
 
+(defmacro defn-timed-level
+  "Creates a function that logs how long it took to execute the body. It
+   supports multiarity functions but only times how long the last listed arity
+   version takes. This means it should be used with multiarity functions where
+   it calls itself with the extra arguments.
+   This function is different from defn-timed by allowing the log function to be
+   passed in so as to allow the calling function to pick which level to log to."
+  [fn-name level & fn-tail]
+  (let [fn-name-str (name fn-name)
+        ns-str (str *ns*)
+        ;; Extract the doc string from the function if present
+        [doc-string fn-tail] (if (string? (first fn-tail))
+                               [(first fn-tail) (next fn-tail)]
+                               [nil fn-tail])
+        ;; Wrap single arity functions in a list
+        fn-tail (if (vector? (first fn-tail))
+                  (list fn-tail)
+                  fn-tail)
+        ;; extract other arities defined in the function which will not be
+        ;; timed.
+        other-arities (drop-last fn-tail)
+        ;; extract the last arity definitions bindings and body
+        [timed-arity-bindings & timed-arity-body] (last fn-tail)]
+    `(defn ~fn-name
+       ~@(when doc-string [doc-string])
+       ~@other-arities
+       (~timed-arity-bindings
+        (let [start# (System/currentTimeMillis)]
+          (try
+            ~@timed-arity-body
+            (finally
+              (let [elapsed# (- (System/currentTimeMillis) start#)]
+                (when (true? (cfg/defn-timed-debug))
+                  (~level
+                   (format
+                    "Timed function %s/%s took %d ms."
+                    ~ns-str ~fn-name-str elapsed#)))))))))))
 (defmacro defn-timed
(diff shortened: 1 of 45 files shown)


jceaser:
I do have an open question about the macro defn-timed-level. I wonder if there is a more graceful way of doing this macro by maybe reusing the code in defn-timed instead of just making a copy. Or I could deprecate defn-timed and migrate everything oven to the new function (in the future). Any thoughts?

common-lib/src/cmr/common/util.clj
```diff
@@ -163,11 +163,48 @@
...
 
+(defmacro defn-timed-level
+  "Creates a function that logs how long it took to execute the body. It
+   supports multiarity functions but only times how long the last listed arity
+   version takes. This means it should be used with multiarity functions where
+   it calls itself with the extra arguments.
+   This function is different from defn-timed by allowing the log function to be
+   passed in so as to allow the calling function to pick which level to log to."
+  [fn-name level & fn-tail]
+  (let [fn-name-str (name fn-name)
```
eereiter, line 174: Can you first check the logging level before doing all of this work and if the level is below the current setting then don't do the code.?

eereiter: Or is the level being passed in what to set this particular log level to - which then my previous comment doesn't make sense.

ingest-app/src/cmr/ingest/api/translation.clj
```diff
@@ -167,12 +167,11 @@
 
 (def random-metadata-routes
```
eereiter, line 168: This isn't for your pull request, but I am wondering why is this function in the code and not in test code.

ingest-app/src/cmr/ingest/data/bulk_update.clj
```diff
@@ -187,25 +184,33 @@
...
 
+(declare get-bulk-update-collection-statuses-for-task)
+#_{:clj-kondo/ignore [:unresolved-symbol]}
 (defn-timed get-bulk-update-collection-statuses-for-task
   [context task-id]
   (get-bulk-update-task-collection-status (context->db context) task-id))
 
+(declare create-bulk-update-task)
+#_{:clj-kondo/ignore [:unresolved-symbol]}
 (defn-timed create-bulk-update-task
```
eereiter, line 206: Do we need get-bulk-update-task-status-for-provider, get-bulk-update-collection-statuses-for-task, create-bulk-update-task, update-bulk-update-task-collection-status all timed - can these be moved to debug?

ingest-app/src/cmr/ingest/data/granule_bulk_update.clj
```diff
@@ -318,6 +324,8 @@
               (util/html-escape task-id))]))
   task-id)
 
+(declare task-completed?)
+#_{:clj-kondo/ignore [:unresolved-symbol]}
 (defn-timed task-completed?
```
eereiter, line 329: Do we need most of these functions are timed?

eereiter (approved):

common-lib/src/cmr/common/util.clj
jceaser: the function still has to be executed, so I can't skip all of it, but I was thinking of adding a defconf to allow the level to be changed.

ingest-app/src/cmr/ingest/data/bulk_update.clj
jceaser: I would love to, I was trying to be very conservative. These must have had a middle of the road usage and time consumption.

jceaser: we talked on line, this is good as it is.

ingest-app/src/cmr/ingest/data/granule_bulk_update.clj
jceaser: there is a report out there.

TylerHeald1 (approved):

PR #2162, opened 2024-08-12
CMR-10071: Fixing the Time To Visibility Log

jceaser:
Removing fv field and adding a concept type to ease parsing. Test also added.

# Overview

### What is the feature/fix?

1. This change will correct the JSON output in the time to visibility log so that it is parsable by Splunk.
2. Added Concept Type so that splunk queries do not have to do a regex to discover it
3. Removed spaces for even more space savings

This should impact anything that relies on the new time to visibility log.

# Checklist

- [x] I have updated/added unit and int tests that prove my fix is effective or that my feature works
- [x] clj-kondo has been run locally and all errors corrected

### DIFF
--- a/indexer-app/src/cmr/indexer/services/index_service.clj
+++ b/indexer-app/src/cmr/indexer/services/index_service.clj
@@ -5,6 +5,7 @@
    [clj-time.core :as t]
    [clj-time.format :as f]
    [clojure.set :as set]
+   [clojure.string :as string]
    [cmr.acl.acl-fetcher :as acl-fetcher]
    [cmr.common.cache :as cache]
    [cmr.common.concepts :as cs]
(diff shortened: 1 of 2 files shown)


TylerHeald1 (approved):

indexer-app/src/cmr/indexer/services/index_service.clj
```diff
@@ -307,22 +308,27 @@
...
-  (let [all-value (if all-revisions-index? 1 0)]
-    (format (str "{\"fv\": 1 \"mg\": \"index-vis\", "
-                 "\"ci\": \"%s\", "
-                 "\"ri\": \"%s\", "
-                 "\"vm\": \"%d\", "
-                 "\"ar\": %d}")
+  (let [all-value (if all-revisions-index? 1 0)
+        concept-type (re-find (re-pattern "^[A-Z]+") concept-id)]
+    (format (str "{\"fv\":1,"
+                 "\"mg\":\"index-vis\","
```
isja17, line 322: fv and mg are hard coded, is there any reason to keep these? I don't know what value they provide. 

also, how many bytes do we gain by making this a bit more human-readable, i.e. not using abbreviations?

jceaser: yes, as I write new logs, I will use a different mg value so that you can find one or the other. As for fv, this is the "format version". If I want to change the format radically I can change this value and then maybe you can query that and know what to do?

isja17 (approved):

zimzoom (approved):

PR #2176, opened 2024-09-25
CMR-10134 Adding in the new subscription schema and setting the subscription cache when the ingest subscriptions are added or deleted.

eereiter:
# Overview

### What is the feature/fix?

Adding in the new subscription schema and setting the subscription cache when the ingest subscriptions are added or deleted.

### What is the Solution?

Adding in the new subscription schema and setting the subscription cache when the ingest subscriptions are added or deleted.

Ingest, subscriptions, metadata_db

# Checklist

### DIFF
--- a/ingest-app/src/cmr/ingest/api/subscriptions.clj
+++ b/ingest-app/src/cmr/ingest/api/subscriptions.clj
@@ -311,7 +311,6 @@
   (let [parsed (json/parse-string (:metadata sub-concept) true)
         provider-id (get-provider-id parsed)
         subscriber-id (get-subscriber-id context parsed)
-        _ (println "The parsed subscription is:" parsed)
         parsed (if (and (= nil (:Query parsed))
                         (= "granule" (:Type parsed)))
                  (assoc parsed :Query (str "collection-concept-id=" (:CollectionConceptId parsed)))
--- a/ingest-app/src/cmr/ingest/api/subscriptions.clj
+++ b/ingest-app/src/cmr/ingest/api/subscriptions.clj
@@ -310,7 +310,12 @@
   (perform-basic-validations sub-concept)
   (let [parsed (json/parse-string (:metadata sub-concept) true)
         provider-id (get-provider-id parsed)
-        subscriber-id (get-subscriber-id context parsed)]
+        subscriber-id (get-subscriber-id context parsed)
+        _ (println "The parsed subscription is:" parsed)
+        parsed (if (and (= nil (:Query parsed))
+                        (= "granule" (:Type parsed)))
+                 (assoc parsed :Query (str "collection-concept-id=" (:CollectionConceptId parsed)))
+                 parsed)]
     (when-not (= CMR_PROVIDER provider-id)
       (api-core/verify-provider-exists context provider-id))
     (validate-user-id context subscriber-id)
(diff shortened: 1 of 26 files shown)


metadata-db-app/src/cmr/metadata_db/services/subscription_cache.clj
```diff
@@ -0,0 +1,49 @@
+(ns cmr.metadata-db.services.subscription-cache
+  "Defines common functions and defs for the subscription cache.
+	Structure of the hash-cache is as follows:
+	<collection-concept-id> --> <ingest subscription map>
+
+	Example:
+	'C12344554-PROV1' --> { 'enabled' true|false
+                          'mode'    New|Update|Delete|All} "
```
TylerHeald1, line 8: Is this choosing a single mode, or is this more like flags? Can a mode be set for New AND Update, or would that need to be set to All to get both events?

eereiter: Right now it is single mode. You can only choose 1.  I didn't think about choosing multiple.  In the API I also have it set to 1. You could create 2 subscriptions, but it would be better to allow multiple.

TylerHeald1 (approved):

ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -310,7 +310,12 @@
   (perform-basic-validations sub-concept)
   (let [parsed (json/parse-string (:metadata sub-concept) true)
         provider-id (get-provider-id parsed)
-        subscriber-id (get-subscriber-id context parsed)]
+        subscriber-id (get-subscriber-id context parsed)
+        _ (println "The parsed subscription is:" parsed)
```
jmaeng72, line 314: do you want to keep this println?

metadata-db-app/src/cmr/metadata_db/config.clj
```diff
@@ -101,3 +101,7 @@
                      (ingest-exchange-name)
                      (access-control-exchange-name)
                      (deleted-granule-exchange-name)]))
+
+(defconfig ingest-subscription-enabled
```
jmaeng72, line 105: are there other types of subscriptions? If so, we should renamed this to be more specific

metadata-db-app/src/cmr/metadata_db/services/subscription_cache.clj
```diff
@@ -0,0 +1,49 @@
+(ns cmr.metadata-db.services.subscription-cache
+  "Defines common functions and defs for the subscription cache.
```
jmaeng72, line 2: this is a much appreciated explanation

jmaeng72 (approved):

metadata-db-app/src/cmr/metadata_db/config.clj
eereiter: right now we only have subscriptions that are driven by search and the results get emailed out.

ingest-app/src/cmr/ingest/api/subscriptions.clj
eereiter: No, thanks.

PR #2185, opened 2024-10-30
CMR-10140-1: Adding subscribing to the local and AWS topics and publishing messages.

eereiter:
# Overview

### What is the feature/fix?

Adding the capability to subscribe/unsubscribe to topics either AWS or local
Adding the capability to publish notification messages when it meets subscription criteria

### What is the Solution?

Added the capability to subscribe/unsubscribe an AWS queue to either an AWS or local topic.
Added the capability to metadata_db to send notification messages if the concepts meet the subscription criteria.

Ingest Subscriptions and Metadata_DB.

# Checklist

### DIFF
--- a/message-queue-lib/src/cmr/message_queue/topic/aws_topic.clj
+++ b/message-queue-lib/src/cmr/message_queue/topic/aws_topic.clj
@@ -1,8 +1,12 @@
 (ns cmr.message-queue.topic.aws-topic
   "Defines an AWS implementation of the topic protocol."
   (:require
+   [cheshire.core :as json]
    [cmr.common.dev.record-pretty-printer :as record-pretty-printer]
    [cmr.common.log :refer [error]]
+   [cmr.common.util :as util]
+   [cmr.message-queue.config :as config]
+   [cmr.message-queue.queue.aws-queue :as aws-queue]
    [cmr.message-queue.topic.topic-protocol :as topic-protocol])
   (:import
    (software.amazon.awssdk.regions Region)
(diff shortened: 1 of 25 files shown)


message-queue-lib/src/cmr/message_queue/topic/aws_topic.clj
```diff
@@ -47,26 +54,93 @@
...
+
+(defn set-filter-policy
+  "For a given subscription set the filter policy so that the queue
+    only gets the notificiation messages that it wants. The passed in
+    filter policy is a hash map - for example:
+    {\"collection-concept-id\": \"C12345-PROV1\"
+     \"mode\": [\"New\", \"Update\"]}"
+  [sns-client subscription-arn subscription]
+  ;; Turn the clojure filter policy to json
+  (when (or (:CollectionConceptId subscription)
```
TylerHeald1, line 77: If both :CollectionConceptId and :Mode are required, would it better to do an and?
Or does the presence of one guarantee the other?

message-queue-lib/src/cmr/message_queue/topic/local_topic.clj
```diff
@@ -45,11 +63,44 @@
...
   (subscribe
-    [_this subscription]
-    (swap! subscription-atom conj subscription))
+   [this subscription]
+   ;; to speed up development startup, the setup call is here and setup checks first to see if it is already setup.
+   ;; Otherwise on startup the system would have to wait for the elasticmq to start before it could continue with setting
+   ;; up the database slowing down all the tests.
+   (setup-infrastructure this)
+   (let [metadata (:metadata subscription)
+         sqs-client (queue/create-sqs-client (config/sqs-server-url))
```
TylerHeald1, line 72: Does an SQS client need to be made for each subscription? Is it like a different interface to the SQS service for each subscription?

jmaeng72 (approved):

TylerHeald1 (approved):

message-queue-lib/src/cmr/message_queue/topic/local_topic.clj
eereiter: I never figured out where the underlying http connection (assuming its http) lived in the AWS libraries.  I thought that it would live in the client, so I wanted to limit those and keep them open for quicker communication, but if that is not the case, then we don't need it in the subscription. It is possible that a queue could live in a different place from other queues, but when the subscription is set up all of the clients are set to use the local queue manager.

PR #2204, opened 2025-01-03
CMR-10194: Exception During Collection Draft Publish (Delete)

jmaeng72:
# Overview

### What is the feature/fix?

High rate of errors where the Collection Draft instance has been deleted from Oracle but is still indexed in Elasticsearch. 

### What is the Solution?

Error was caused by race condition between two queues with same concept-delete msg that indexer picks up and uses to find concept in db and delete from 2 es indices. But if one index instance deletes from the db first, the other will not delete from second es index.
Current solution is to delete from both indexes at same time from 1 queue and ignoring the 'concept not found in db' when read again from 2nd queue.

Delete any draft concept

# Checklist


TylerHeald1:
If the solution is to allow the first item off the queue to handle both deletes, then I think you should only push onto one queue from the publish action. The way it is set up currently feels like just repackaging the race condition

TylerHeald1 (changes requested):
I don't believe this addresses the race condition

indexer-app/src/cmr/indexer/services/index_service.clj
```diff
@@ -697,50 +697,70 @@
...
     (cs/concept-id->type concept-id)))
 
-(defmethod delete-concept :default
-  [context concept-id revision-id options]
-  ;; Assuming ingest will pass enough info for deletion
-  ;; We should avoid making calls to metadata db to get the necessary info if possible
+(defn- delete-concept-default-helper
+  [context concept concept-id revision-id options]
+  (if (nil? concept)
+    (errors/throw-service-error
```
jceaser, line 703: will this code ever run given that (when) is used below?

indexer-app/src/cmr/indexer/services/index_service.clj
```diff
@@ -697,50 +697,70 @@
   (fn [_context concept-id _revision-id _options]
     (cs/concept-id->type concept-id)))
 
-(defmethod delete-concept :default
-  [context concept-id revision-id options]
-  ;; Assuming ingest will pass enough info for deletion
-  ;; We should avoid making calls to metadata db to get the necessary info if possible
+(defn- delete-concept-default-helper
```
jceaser, line 700: None of these functions touched in this change have documentation. I don't remember if the linter requires private functions to be documented, but the functions intent is good to capture now while you remember.

indexer-app/src/cmr/indexer/services/index_service.clj
```diff
@@ -697,50 +697,70 @@
...
+
+(defn- delete-draft-concept
+  [context concept concept-id revision-id options]
+  (when concept
+    (delete-concept-default-helper context concept concept-id revision-id (assoc options :all-revisions-index? true))
+    (delete-concept-default-helper context concept concept-id revision-id (assoc options :all-revisions-index? false))
+    (try
+      (meta-db/delete-draft context concept)
+      (catch Exception e
+        (info (format "Force delete draft ran into exception for concept-id %s due to race condition between indexers
```
jceaser, line 752: why is this not a warning? Does this mean that a deletion is not successful meaning the system is not correct? Would you ever want to filter these out by being in info?

jmaeng72:
> If the solution is to allow the first item off the queue to handle both deletes, then I think you should only push onto one queue from the publish action. The way it is set up currently feels like just repackaging the race condition

Even if we publish to one queue from the publish side, we'll still need this separation of funcs with a delete-draft-concept func. The only difference would be that it would no longer need to catch an exception and ignore it.

I can make the publish API write to one queue instead, but that will require a lot of changes to the funcs it calls. Currently, the publish endpoint calls the metadata /concepts API that eventually writes to the two queues.

I can create new funcs within metadata that only writes to one queue instead and then redo all the other funcs that call it. It's about 6 func calls.

I originally decided against it because we might do a full re-design of this entire flow anyway. But I am open to changing the publish func to write to one queue and then writing new funcs for the new draft flow.

indexer-app/src/cmr/indexer/services/index_service.clj
jmaeng72: The "delete-concept :default" func can be called when other concepts like granules, collections, etc are deleted and if they aren't found in the db we'll want to throw this error here

indexer-app/src/cmr/indexer/services/index_service.clj
jmaeng72: This is a message to let us know the race condition occurred with the draft index. It's technically not a warning because we're expecting this race condition and choosing to ignore this error

jmaeng72:
> I don't believe this addresses the race condition

Do you mean that you don't believe this solves the race condition? If so, I agree. I think the only solution is to do a re-design, rather than a hot-fix solution like this

TylerHeald1 (approved):

TylerHeald1:
Discussed offline, turns out I had misread part of the PR

PR #2209, opened 2025-01-27
CMR-10310: Allow the subscription endpoint to take an https URL in the EndPoint element

jmaeng72:
# Overview

### What is the feature/fix?

Ingest subscription could only take in SQS ARN as endpoint to send sub messages to

### What is the Solution?

Add the ability to send sub message to any HTTP/S endpoint

Ingest

# Checklist

- [x] I have updated/added unit and int tests that prove my fix is effective or that my feature works
- [x] I have made corresponding changes to the documentation

### DIFF
--- a/ingest-app/test/cmr/ingest/api/subscriptions_test.clj
+++ b/ingest-app/test/cmr/ingest/api/subscriptions_test.clj
@@ -1,8 +1,9 @@
 (ns cmr.ingest.api.subscriptions-test
   (:require
-   [clojure.string :as string]
-   [clojure.test :refer :all]
-   [cmr.ingest.api.subscriptions :as subscriptions]))
+    [clojure.string :as string]
+    [clojure.test :refer :all]
+    [cmr.common.util :as util]
+    [cmr.ingest.api.subscriptions :as subscriptions]))
 
 (deftest generate-native-id-test
   (let [parsed {:Name "the_beginning"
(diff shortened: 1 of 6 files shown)


metadata-db-app/test/cmr/metadata_db/test/services/subscriptions_test.clj
```diff
@@ -682,3 +682,64 @@
...
+          "http://www.google.com"
+
+          "invalid url - non-existent domain"
+          false
+          "hello.blach"
+
+          "invalid url - some string"
+          false
+          "this is just some string"
+          )))
```
TylerHeald1, line 745: add newline at end of file

ingest-app/test/cmr/ingest/api/subscriptions_test.clj
```diff
@@ -14,3 +15,32 @@
...
+  (testing "validate subscription endpoint str -- expected valid"
+    (util/are3 [subscription-concept]
+               (let [fun #'cmr.ingest.api.subscriptions/validate-subscription-endpoint]
+                 (is (= nil (fun subscription-concept))))
+
+               "given method is search -- endpoint ignored"
+               {:EndPoint "blahblah", :Method "search"}
+
+               "given method is search and endpoint not given -- endpoint ignored"
+               {:EndPoint "blahblah", :Method "search"}
```
TylerHeald1, line 29: test looks the same as the one above?

metadata-db-app/src/cmr/metadata_db/services/subscriptions.clj
```diff
@@ -110,31 +114,79 @@
...
+
+(defn- is-local-test-queue
+  "Checks if subscription endpoint is a local url that point to the local queue -- this is for local tests.
+  Returns true or false."
+  [endpoint]
+  (some? (re-matches #"http://localhost:9324.*" endpoint)))
+
+(defn- send-sub-to-url-dest
+  "Sends subscription details to url given. Throws error if subscription is not successful."
+  [subscription-concept dest-endpoint]
```
eereiter, line 136: Metadata DB shouldn't send messages out. 

All endpoints should subscribe to the external SNS topic. The lambda should be a CMR lambda that listens to the SNS topic that was subscribed. When an URL endpoint arrives it should send messages to the endpoint.  

All metadata db does is check to see if a granule is part of a collection where a subscription exists and then creates enough information to be placed onto an internal subscription queue.  This is where the subscription workers will pick it up where they can do ACL checks based on the data that is in the message. If the message can be sent on, then the workers will create the proper notification message and place it on the external SNS Topic, that is where the CMR Lambda will pick up the message and forward it to the URL in the message.

eereiter (changes requested):
Metadata DB should not be sending out notifications to subscribers, It needs to put enough information into the notification message so that the subscription workers can do more processing and then it needs to place the messages onto the internal subscription notification topic.  The sending of the notification to the end point URL needs to happen in the lambda that is listening to the external CMR topic.

PR #2213, opened 2025-02-11
CMR-10311: Adding a lambda that can send notification messages to scubscription URLs.

eereiter:
# Overview

### What is the feature/fix?

Adding a lambda that can send notification messages to scubscription URLs.

### What is the Solution?

Adding a lambda that can send notification messages to scubscription URLs.

subscription notifications through URLs.

# Checklist


subscription/notify_lambda/notification-lambda.py
```diff
@@ -0,0 +1,41 @@
...
+        logger.debug("Message: ", message)
+        response = requests.post(url, headers=headers, json=message)
+
+        # Check if the request was successful
+        if response.status_code == 200:
+            logger.debug("Message sent successfully!")
+            logger.debug(f"Response: {response.text}")
+        else:
+            logger.error(f"Failed to send message. Status code: {response.status_code}")
+            logger.debug(f"Response: {response.text}")
```
TylerHeald1, line 37: do you need a sys.exit(1) to ensure that the lambda marks itself as having failed? Though I guess it might depend on what kind of error occurred, since it could simply be an error on the other end of the request

TylerHeald1: Or perhaps the fact that it sent the request regardless of return value is success as far as the lambda is concerned. Generally I'm trying to think if there's a way we could know how many lambdas are getting non 2xx responses from sending the data

subscription/notify_lambda/test.sh
```diff
@@ -0,0 +1,8 @@
...
...d '{"Records": [{"EventSource": "aws:sns","EventVersion": "1.0","EventSubscriptionArn": "arn:aws:sns:us-east-1:832706493240:cmr-subscriptions-sit:1f071817-9e7b-450e-b5a5-606bef7c5b71","Sns": {"Type": "Notification","MessageId": "ed8c7ee0-c70a-5050-8ef9-1effe57d3dde","TopicArn": "arn:aws:sns:us-east-1:832706493240:cmr-subscriptions-sit","Subject": "testing again","Message": "testing again","Timestamp": "2025-02-06T20:48:55.564Z","SignatureVersion": "1","Signature": "iNvFhB7SvB2Um4xkJ4czhaDtQ9WMr3VWz91j9aBJclHQOSR1vznwdT6pJLJOj1fUA/C9JatNEnouUiPut+8DACZ8pVXfMw5bvHoyci63Y7z5gcmVIHpEMetn1Ms7SpgnuddrIoKpCN/tAQGrEfmzDlMbTXjRWrXpEs8xXsJ+Xq2Mp17FoMU00/JV/eAP8ktHX7TtS+dBGLsPd/QPY3eN5QYcmYAcoFd99Va4Qt6dMqOoH/fjIs1aCwnmD02MSnbjCWOgNSteZChnVAT/+l26krKoidHSop53guPt5KWNBGXTSy4Ouhq630qourN+AiVbKqzUtVypNTXas7iBEoT5Nw==","SigningCertUrl": "https://sns.us-east-1.amazonaws.com/SimpleNotificationService-9c6465fa7f48f5cacd23014631ec1136.pem","UnsubscribeUrl": "https://sns.us-east-1.amazonaws.com/?Action=Unsubscribe&SubscriptionArn=arn:aws:sns:us-east-1:832706493240:cmr-subscriptions-sit:1f071817-9e7b-450e-b5a5-606bef7c5b71","MessageAttributes": {"endpoint":"http://host.docker.internal:5001/store"}}}]}'
```
TylerHeald1, line 6: What's assumed to be running on localhost:5001 for this line?

subscription/notify_lambda/notification-lambda.py
eereiter: I could put the data that fails to send the notifications back onto a queue with a logged message to be retried later, and after a few tries, it just gets abandoned?  Do you think we should track success counts vs failures? Right now in terraform I don't have these logs forwarded to splunk, but we could.

TylerHeald1: I was thinking back and forth on this quite a bit, and I've decided I think it's actually fine as is, but it would be great if we could get something set up in cloudwatch or something to track what status codes get logged in the error output (from "Failed to send message. Status code: {response.status_code}"). So that way we can have metrics available to see if/when someone brings an issue with subscriptions to us we can check the chart and see if there was a sudden surge in http error codes.
My concern is that if we just let the lambda succeed and don't have any note anywhere about what error got logged, we'll have a much harder time investigating any major issues someone might bring to us

TylerHeald1 (approved):

subscription/notify_lambda/logger.py
```diff
@@ -0,0 +1,29 @@
+import os
+import logging
+import sys
+
+LOG_LEVEL = os.getenv("LOG_LEVEL")
+if not LOG_LEVEL:
+    LOG_LEVEL = logging.INFO
+
+def setup_logger(name, log_file=None, level=logging.INFO):
```
jceaser, line 9: in general I think more type hints should be used.

subscription/notify_lambda/notification-lambda.py
jceaser: I don't think exit() is the best way, just doing a google and an AI request

jceaser (commented):
add some type hints, and I think we need to decided on the exit() issue

subscription/notify_lambda/notification-lambda.py
TylerHeald1: No, I think letting the lambda just print to the error output and have the lambda itself succeed is fine, but think it would be great if there's some way we could capture the metrics of how many error codes we get (though if the team doesn't think we need that, that's also fine)

subscription/notify_lambda/logger.py
eereiter: I have added Type Hints.

jceaser (approved):

PR #2222, opened 2025-02-26
CMR-10388: Add access control checks to the subscription worker to take out any notifications when the subscriber no longer has access.

eereiter:
# Overview

### What is the feature/fix?

Add access control checks to the subscription worker to take out any notifications when the subscriber no longer has access.

### What is the Solution?

Add access control checks to the subscription worker to take out any notifications when the subscriber no longer has access.

Subscription worker check permissions with access control.

# Checklist


subscription/src/access_control.py
```diff
@@ -1,26 +1,27 @@
...
+import json
 import requests
 from env_vars import Env_Vars
 from sys import stdout
 from logger import logger
 
 class AccessControl:
     """Encapsulates Access Control API.
-    This class needs the following environment variables set:
+    This class needs the following environment variables set with an exxample value:
```
TylerHeald1, line 10: exxample -> example

eereiter: Fixed.

metadata-db-app/src/cmr/metadata_db/services/subscriptions.clj
```diff
@@ -63,19 +63,20 @@
   "Creates a mode-to-endpoints map given a list of subscriptions all for the same one collection.
   Returns a map in this structure:
   {
-    new: ['sqs:arn:111', 'https://www.url1.com', 'https://www.url2.com'],
```
jmaeng72, line 66: I do not think that putting the subscriber info under the 'Mode' parent key is correct. Why not create a different parent key 'Subscriber' and put the user info in there?
So that the map ends up looking like:

{Mode: {new: [sqs1, url1]},
Subscriber: {new: user1}}

Why do the users have to be tied to the mode? Are subscriber permissions mode level based?

jmaeng72: Also, the the mode-types were supposed to point to a set of sqs or url, to prevent duplicates will this new format support that previous function?

What if multiple users have different permissions to the same sqs, mode, collection combo? What happens then?

eereiter: Its not that the subscriber has to be tied to the mode its that the subscriber needs to be tied to the URL. Oh, you know what I am not sure that I am sending out the message the correct way though.

eereiter: Yeah, I forgot to update the http lambda.  I need to know the subscriber and the URL together for the http notifications.  Otherwise user1 can see the notification, and user 2 can't and I don't have the subscriber information tied with the URL, then I won't know to what URL the message can't or can be sent.

subscription/src/subscription_worker.py
```diff
@@ -39,18 +40,24 @@ def delete_messages(sqs_client, queue_url, messages):
...
+            message_attributes = message_body["MessageAttributes"]
+            logger.debug(f"Subscription worker: Received message including attributes: {message_body}")
+
+            subscriber = message_attributes['subscriber']['Value']
+            collection_concept_id = message_attributes['collection-concept-id']['Value']
+
+            if(access_control.has_read_permission(subscriber, collection_concept_id)):
+                logger.debug(f"Subscription worker: {subscriber} has permission to receive granule notifications for {collection_concept_id}")
+                sns_client.publish_message(topic, message)
+            else:
```
TylerHeald1, line 56: Looks like the message is just being ignored, do we need to remove the subscription so it doesn't take up future resources?

eereiter: Yes, the lack of permission is being logged, but that it is.  We could, but sometimes providers mess up their ACLs and fix them later.  I have the same question if the endpoint we are sending it to goes away, SQS is deleted or URL no longer works, do we remove the subscription?  May be after a length of time for the same error?  I can write a ticket for us to discuss this, but I think for this ticket, its too much work.

subscription/test/access_control_test.py
```diff
@@ -8,84 +8,104 @@
...
-        mock_get_url.return_value = "https://test.com"
+        mock_env_vars.return_value = mock_env_vars_instance
+
+        self.access_control.get_url_from_parameter_store()
+        expected_url = "https://cmr.sit.earthdata.nasa.gov:3011/access-control"
+        self.assertEqual(self.access_control.url, expected_url)
+
+    def test_get_url(self):
+        with patch.object(AccessControl, 'get_url_from_parameter_store') as mock_method:
+            self.access_control.url = "http://test-url.com"
```
jceaser, line 33: that url is not a dead URL but it does not return either. Switch to example.gov or example.com

jceaser (changes requested):
update that URL

subscription/test/access_control_test.py
eereiter: I have updated the URL

metadata-db-app/test/cmr/metadata_db/test/services/subscriptions_test.clj
```diff
@@ -602,12 +621,13 @@
...
-                 subscription-arn (subscriptions/attach-subscription-to-topic test-context sub-concept)
-                 sub-concept (assoc-in sub-concept [:extra-fields :aws-arn] subscription-arn)]
+                 sub-concept (subscriptions/attach-subscription-to-topic test-context sub-concept)
+                 subscription-arn (get-in sub-concept [:extra-fields :aws-arn])]
 
              (is (some? subscription-arn))
-             (when subscription-arn
-               (is (some? (subscriptions/delete-ingest-subscription test-context sub-concept))))
+            ; (when subscription-arn
+            ;   (println "Subscription ARN:" subscription-arn)
```
jmaeng72, line 629: why did you remove this?

metadata-db-app/test/cmr/metadata_db/test/services/subscriptions_test.clj
```diff
@@ -602,12 +621,13 @@
...
+            ;   (is (some? (subscriptions/delete-ingest-subscription test-context sub-concept))))
```
jmaeng72, line 630: why is this commented out?

subscription/src/access_control.py
```diff
@@ -91,8 +96,36 @@ def get_permissions(self, subscriber_id, concept_id):
...
+
+        try:
+            # Call the get_permissions function
+            permissions_str = self.get_permissions(subscriber_id, collection_concept_id)
+            
+            if permissions_str:
+                permissions = json.loads(permissions_str)
+
+                # Check if the permissions is a dictionary
+                if isinstance(permissions, dict):
```
jmaeng72, line 119: these two if statements can become one if statement with an 'AND'

subscription/src/access_control.py
```diff
@@ -91,8 +96,36 @@ def get_permissions(self, subscriber_id, concept_id):
...
+                    # Check if the collection_concept_id is in the permissions dictionary
+                    if collection_concept_id in permissions:
+                        # Check if "read" is in the list of permissions for the collection
+                        return "read" in permissions[collection_concept_id]
+                    else: return False
```
jmaeng72, line 124: for python you don't need to do a return false for each else. You can just put a default return False in the try block level, so that if it doesn't catch the 'return "read" in permissions[collection_concept_id], it will automatically go to the outer 'return False' statement.

subscription/src/env_vars.py
```diff
@@ -12,7 +12,15 @@ def __init__(self):
         self.ssm_client = boto3.client('ssm', region_name=os.getenv("AWS_REGION"))
     
     def get_var(self, name, decryption=False):
```
jmaeng72, line 14: can you rename to something more specific? Like: get_env_var_from_parameter_name. Or strip_env_var_from_parameter_name. And the variable name be changed to parameter_name or something like that

subscription/src/env_vars.py
```diff
@@ -12,7 +12,15 @@ def __init__(self):
...
-        value = os.getenv(name)
+        """The name parameter looks like /sit/ingest/ENVIRONMENT_VAR. To check if the environment
+           variable exists strip off everything except for the actual variable name. Otherwise
+           go to the AWS ParameterStore and get the values."""
+
+        logger.debug(f"Subscription worker: Getting the environment variable called {name}")
+        parts = name.split('/')
+        os_name = next(part for part in reversed(parts) if part)
```
jmaeng72, line 21: are you trying to get the ENVIRONMENT_VAR? If so, why not just get the last element in the split string?

name.split('/')[-1]

subscription/test/access_control_test.py
```diff
@@ -8,84 +8,104 @@
...
-            params={"user_id": "sub123", "concept_id": "concept456"}
+
+    @patch.object(AccessControl, 'get_permissions')
+    def test_has_read_permission(self, mock_get_permissions):
+        # Test when user has read permission
+        mock_get_permissions.return_value = "{\"C1200484253-CMR_ONLY\": [\"read\", \"update\"]}"
+        result = self.access_control.has_read_permission("user1", "C1200484253-CMR_ONLY")
+        self.assertTrue(result)
+
+        # Test when user doesn't have read permission
```
jmaeng72, line 71: i like these comments very helpful

metadata-db-app/test/cmr/metadata_db/test/services/subscriptions_test.clj
eereiter: I duplicated the lines when I was testing and forgot to delete it.

subscription/src/access_control.py
eereiter: I have combined the ifs and removed the elses.

subscription/src/env_vars.py
eereiter: Yes. That is a better way.  I have change the code as you suggest.

subscription/test/access_control_test.py
eereiter: Thank you.

subscription/src/subscription_worker.py
eereiter: I created issue https://bugs.earthdata.nasa.gov/browse/CMR-10415

metadata-db-app/src/cmr/metadata_db/services/subscriptions.clj
```diff
@@ -63,19 +63,20 @@
   "Creates a mode-to-endpoints map given a list of subscriptions all for the same one collection.
```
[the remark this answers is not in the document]
eereiter: I have updated the documentation.

jceaser (approved):

TylerHeald1 (approved):

jmaeng72 (approved):

PR #2236, opened 2025-04-04
CMR-10455: Update subscription doc on subscription edge case limit

jmaeng72:
# Overview

### What is the feature/fix?

To update doc on limitation of subscription queue create and update.
Currently, the ingest (NRT) subscriptions has a situation where if User1 creates a subscription to Coll1 and then User2 creates a different subscription to the same Coll1 for the same queue, it will overwrite the first subscription in the backend SNS subscription filter for that queue.

### What is the Solution?

Add the limit to the CMR Ingest documentation page so that users understand what to expect

Ingest Doc

# Checklist

### DIFF
--- a/ingest-app/docs/api.md
+++ b/ingest-app/docs/api.md
@@ -1169,9 +1169,10 @@ For NRT Notification subscriptions to be used there are three new fields that ar
         </ul>
 </ul>
 
-**IMPORTANT: There is limit to creating and updating subscriptions related to the specific scenario highlighted below:
-Currently, if User1 creates a subscription to Coll1 and then User2 creates a different subscription to the same Coll1 for the same queue, it will overwrite the first subscription for that specific queue.
-So please have each of your queues to have only one agreed upon subscription.
+IMPORTANT: There is limit to creating and updating subscriptions related to the specific scenario highlighted below:
+Currently, if <User 1> creates a subscription to <Collection 1> and then <User 2> creates a different subscription 
+to the same <Collection 1> for the same queue, it will overwrite the first subscription for that specific queue.
+So please, have each queue have only one agreed upon subscription.
 
 ##### NRT Notification Subscription POST Request
 
@@ -1232,9 +1233,10 @@ If a native-id is provided in a POST, and a subscription already exists for that
 
 PUT requests should be used for updating subscriptions. Creating subscriptions with PUT may be deprecated in the future. All PUT requests require a native-id to be part of the request URL.
 
-**IMPORTANT: There is limit to creating and updating subscriptions related to the specific scenario highlighted below:
-Currently, if User1 creates a subscription to Coll1 and then User2 creates a different subscription to the same Coll1 for the same queue, it will overwrite the first subscription for that specific queue.
-So please have each of your queues to have only one agreed upon subscription.
+IMPORTANT: There is limit to creating and updating subscriptions related to the specific scenario highlighted below:
+Currently, if <User 1> creates a subscription to <Collection 1> and then <User 2> creates a different subscription
+to the same <Collection 1> for the same queue, it will overwrite the first subscription for that specific queue.
+So please, have each queue have only one agreed upon subscription.
 
 ##### Create Batch Notification subscription with %CMR-ENDPOINT%/subscriptions/<native-id> URL
 ```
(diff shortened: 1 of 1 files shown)


ingest-app/docs/api.md
```diff
@@ -1169,6 +1169,10 @@ For NRT Notification subscriptions to be used there are three new fields that ar
         </ul>
 </ul>
 
+**IMPORTANT: There is limit to creating and updating subscriptions related to the specific scenario highlighted below:
+Currently, if User1 creates a subscription to Coll1 and then User2 creates a different subscription to the same Coll1 for the same queue, it will overwrite the first subscription for that specific queue.
```
jceaser, line 1173: User1 rubs me the wrong way, only because people complain about our API docs a lot. While I am probably responsible for a good number of documentation complaints (having created a lot of them) I feel like we could do something better here. What about `"<User One>"` and `"<Collection One>"`?

TylerHeald1 (approved):

ingest-app/docs/api.md
```diff
@@ -1169,6 +1169,10 @@ For NRT Notification subscriptions to be used there are three new fields that ar
...
+So please have each of your queues to have only one agreed upon subscription.
```
jceaser, line 1174: is this the correct "voice" for documentation, the use of "your"?

ingest-app/docs/api.md
jceaser: I also do not like the use of ** for markdown when it has no markdown meaning. This could confuse bad parsers.

ingest-app/docs/api.md
```diff
@@ -1228,6 +1232,10 @@ If a native-id is provided in a POST, and a subscription already exists for that
 
 PUT requests should be used for updating subscriptions. Creating subscriptions with PUT may be deprecated in the future. All PUT requests require a native-id to be part of the request URL.
 
+**IMPORTANT: There is limit to creating and updating subscriptions related to the specific scenario highlighted below:
```
jceaser, line 1235: all the same comments from above.

jceaser (changes requested):

ingest-app/docs/api.md
```diff
@@ -1169,6 +1169,10 @@ For NRT Notification subscriptions to be used there are three new fields that ar
         </ul>
 </ul>
 
+**IMPORTANT: There is limit to creating and updating subscriptions related to the specific scenario highlighted below:
```
zimzoom, line 1172:
```suggestion
**IMPORTANT:** There is limit to creating and updating subscriptions related to the specific scenario highlighted below:
```

ingest-app/docs/api.md
zimzoom, line 1173:
```suggestion
Currently, if <User One> creates a subscription to <Collection 1> and then <User 2> creates a different subscription to the same <Collection 1> for the same queue, it will overwrite the first subscription for that specific queue.
```

ingest-app/docs/api.md
zimzoom, line 1174:
```suggestion
So please, have each queue have only one agreed upon subscription.
```

ingest-app/docs/api.md
zimzoom, line 1235:
```suggestion
**IMPORTANT:** There is a limit to creating and updating subscriptions related to the specific scenario highlighted below:
```

ingest-app/docs/api.md
```diff
@@ -1228,6 +1232,10 @@ If a native-id is provided in a POST, and a subscription already exists for that
 
 PUT requests should be used for updating subscriptions. Creating subscriptions with PUT may be deprecated in the future. All PUT requests require a native-id to be part of the request URL.
 
+**IMPORTANT: There is limit to creating and updating subscriptions related to the specific scenario highlighted below:
+Currently, if User1 creates a subscription to Coll1 and then User2 creates a different subscription to the same Coll1 for the same queue, it will overwrite the first subscription for that specific queue.
```
zimzoom, line 1236:
```suggestion
Currently, if <User 1> creates a subscription to <Collection 1> and then <User 2> creates a different subscription to the same <Collection 1> for the same queue, it will overwrite the first subscription for that specific queue.
```

zimzoom (approved):
Coded and batched Thomas' suggestions for your convenience

jceaser (approved):

PR #2257, opened 2025-05-13
CMR-10213 new jetty - Attempt 2

jceaser:
# Overview

The first attempt at pushing out the Jetty change failed in SIT as that environment ran the code in a different way (using different settings) which exposed an error. This version removes the use of libraries which caused the problem. 

It seams that the web_server.clj file was not always being compiled unless you called `cmr setup dev` between changes. After doing this I was able to get runtime errors.

NOTE: this pull request is into the first branch so that I can show just what was changed here. I will then merge that branch (which was approved already) into master.

### What is the feature/fix?

1. Removed library calls which no longer exist.
2. Exposed dev-system configurations globally so that all apps can choose to run without compression and logs instead of assuming.

1. All web servers now have two new options (which should not be needed for deployment).
2. Dev-system will use logs now conforming to the global default.

# Checklist

- [-] I have updated/added unit and int tests that prove my fix is effective or that my feature works
- [-] New and existing unit and int tests pass locally and remotely
- [-] clj-kondo has been run locally and all errors corrected

### DIFF
--- a/common-lib/src/cmr/common/api/web_server.clj
+++ b/common-lib/src/cmr/common/api/web_server.clj
@@ -19,6 +19,19 @@
   ring jetty adapter default of 8."
   8)
 
+(declare use-web-compression?)
+(defconfig use-web-compression?
+  "Indicates whether the servers will use gzip compression. Disable this to
+  make tcpmon usable"
+  {:default true
+   :type Boolean})
+
+(declare use-access-log)
+(defconfig use-access-log
+  "Indicates whether the servers will use the access log."
+  {:default true
+   :type Boolean})
+
 (declare MAX_THREADS)
 (defconfig MAX_THREADS
   "The maximum number of threads for Jetty to use to process requests. This was originally set to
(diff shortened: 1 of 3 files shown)


common-lib/src/cmr/common/api/web_server.clj
```diff
@@ -19,6 +19,19 @@
...
 
+(declare use-web-compression?)
+(defconfig use-web-compression?
+  "Indicates whether the servers will use gzip compression. Disable this to
+  make tcpmon usable"
+  {:default true
+   :type Boolean})
+
+(declare use-access-log)
+(defconfig use-access-log
```
jceaser, line 30: NOTE: I'm turning this to true which was the default in web_server.clj. This means that dev-system will log now unless you set an env, which I don't think is a big deal.

dev-system/src/cmr/dev_system/config.clj
```diff
@@ -29,17 +29,6 @@
   {:default 5601
    :type Long})
 
-(defconfig use-web-compression?
-  "Indicates whether the servers will use gzip compression. Disable this to
-  make tcpmon usable"
-  {:default true
-   :type Boolean})
-
-(defconfig use-access-log
```
jceaser, line 38: was false

dev-system/src/cmr/dev_system/system.clj
```diff
@@ -78,8 +79,8 @@
   "Update the web configuration options for the passed app system."
   [app-system]
   (-> app-system
-      (assoc-in [:web :use-compression?] (dev-config/use-web-compression?))
-      (assoc-in [:web :use-access-log?] (dev-config/use-access-log))))
+      (assoc-in [:web :use-compression?] (web-serv/use-web-compression?))
+      (assoc-in [:web :use-access-log?] (web-serv/use-access-log))))
```
jceaser, line 83: this will now be true, not false

eereiter (approved):

TylerHeald1 (approved):

PR #2277, opened 2025-07-25
CMR-10693: Fixing Bug when granules do not have a DataGranule element an exception was thrown and when the location doesn't include the search context 

eereiter:
# Overview

### What is the feature/fix?

Fixing Bug when granules do not have a DataGranule element an exception was thrown and when the location doesn't include the search context 

### What is the Solution?

Fixing Bug when granules do not have a DataGranule element an exception was thrown and when the location doesn't include the search context 

Subscription worker

# Checklist


subscription/src/search.py
```diff
@@ -147,7 +147,10 @@ def get_concept(self, concept_id, revision_id):
 
     def get_producer_granule_id(self, metadata):
         """Get the granule producer id from the metadata."""
-        identifiers = metadata.get('DataGranule').get('Identifiers')
+        data_granule = metadata.get('DataGranule')
+        identifiers = None
+        if data_granule:
```
jmaeng72: instead of setting identifier and pgi as none, you can put all these in the if statement. Since you're not using the vars outside of the condition.

subscription/src/search.py
```diff
@@ -160,6 +163,17 @@ def get_producer_granule_id(self, metadata):
                 return None
         else:
             return None
+        
+    def fix_location(self, location):
+        """This function fixes the location bug that is in
```
jmaeng72: not sure if we need this type of comment in here, saying it's fixing a bug. Also would consider renaming this func to be more specific, like 'add_search_context_to_location_url. What is the location type? A string representing a url? If so, can rename to location_url for clearer understanding

subscription/src/search.py
```diff
@@ -160,6 +163,17 @@ def get_producer_granule_id(self, metadata):
...
+        metadata-db-app/src/cmr/metadata_db/services/subscriptions.clj
+        on line 260 where the location is missing the context value of search."""
+        
+        if "search" in location:
```
jmaeng72: according to the func that calls this func, location could be nil, might want to add a catch for that

subscription/src/search.py
```diff
@@ -177,10 +191,12 @@ def process_message(self, message):
          \"location\": \"http://localhost:3003/concepts/G1200484356-ERICH_PROV/39\"}
         """
         # Get the granule concept-id from the message.
-        print(f"Search process_message message type {type(message)}")
+        logger.debug(f"Search process_message message type {type(message)}")
         message_dict = json.loads(message)
         concept_id = message_dict["concept-id"]
         revision_id = message_dict.get("revision-id", None)
+        location = message_dict.get("location")
+        fixed_location = self.fix_location(location)
```
jmaeng72: can we rename to something more clear? 'fixed' means what?

subscription/src/search.py
eereiter: I have re-written the function.

subscription/src/search.py
eereiter: I have changed the function name to add_search_context_to_location_url and changed location to location_url.  Yes, the location type is a string which represents a URL.

subscription/src/search.py
eereiter: I have added code to catch the None condition.

subscription/src/search.py
eereiter: I changed the variable name from fixed_location to location_url

PR #2286, opened 2025-08-29
CMR-10784: Create aliases for indices created during migration and rebalancing.

ygliuvt:
# Overview

### What is the feature/fix?

Create aliases for indices created during migration and rebalancing.

### What is the Solution?

Create default aliases (in the form of <index-name>_alias) for CMR indices in the index-set during access-control and indexer application's db-migrate operations. If there are new indices that need to be added, the new indices and their default aliases with also be created as a result of db-migrate.

New granule indices created as a result of rebalancing will also have their default aliases created automatically.

Note: Separate granule indices moved back to small_collections index during rebalancing will not have their aliases removed because the finalize operation in rebalancing will not delete the separate granule index automatically  (to avoid issue with search cache refreshing). As a result, we will rely on the cascading deletion of aliases when the separate granule index gets deleted eventually.

I have tested locally and in WL4 to verify:
1) New index added to index-set will be created together with its default alias as a result of db-migrate during deployment for both the Access-Control and Indexer apps.
2) Aliases for existing index in Access-Control app (acls, groups) will be added automatically as a result of db-migrate during deployment.
3) Aliases for existing index in Indexer app will NOT be added automatically (if the rest of the indices stay the same) as a result of db-migrate during deployment. We can manually create them by invoking the `db-migrate` endpoint on Indexer app with `force=true` parameter.

Access-control app and Indexer app db-migrate endpoints.

# Checklist

### DIFF
--- a/indexer-app/src/cmr/indexer/data/index_set_elasticsearch.clj
+++ b/indexer-app/src/cmr/indexer/data/index_set_elasticsearch.clj
@@ -13,6 +13,10 @@
    [cmr.indexer.services.messages :as m]
    [cmr.transmit.config :as t-config]))
 
+(def resharding-index-pattern
+  "Regex pattern for index names created by resharding"
+  #"^([0-9]+_c[0-9]+_[A-Za-z0-9_]+)_([0-9]+)_shards$")
+
 (defn- decode-field
   "Attempt to decode a field using gzip, b64. Return the original field json decoded
   if there are any exceptions. This function is here to deal with any legacy fields that are not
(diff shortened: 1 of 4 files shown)


eereiter (approved):

indiejames (approved):

indexer-app/src/cmr/indexer/data/index_set_elasticsearch.clj
```diff
@@ -73,6 +77,12 @@
         (do
           (info "Index" index-name "does not exist so it will be created")
           (esi-helper/create conn index-name {:settings settings :mappings mapping})))
+
+      ;; if the index is not a resharding index and alias does not exist, add it
```
chris-durbin, line 81: Probably worth adding a test around the logic for when and when not to create an alias.

ygliuvt: Too much work to set up mocks and subs for unit testing this. The plan is to do an integration test when we implement resharding.

PR #2287, opened 2025-09-02
CMR-10788: Improve performance of nested aggregation queries performed to support facets.

chris-durbin:
# Overview

### What is the feature/fix?

Granule searches that include facets are significantly slower than all other types of granule queries. For the past week the median performance of queries that include facets is 3900 ms against 300 ms for any granule queries that do not include facets. This PR looks to improve that performance.

### What is the Solution?

The main cause of the slow performance is reverse nested aggregations which use a sub aggregation to get counts by concept-id by performing a terms search and grouping by concept-id which is slow. I changed a few things.

1. The cycle query was performing this reverse nested aggregation by concept-id unnecessarily since cycle is not a nested field. I dropped it from the query.
2. For queries that include a cycle or pass and need to get reverse nested aggregation counts I switched from using a terms aggregation which is expensive to a more efficient cardinality query. Since we are only getting counts and don't need the concept-ids in the aggregation changing it does not cause any difference in behavior.
3. I found collection facet queries were doing something similar with a terms aggregation when we only wanted counts, not buckets by collection id, so I switched that one to use cardinality as well.

Note that terms queries are guaranteed to be exact, but cardinality queries are approximate. The default precision value of 3000 means it should be an exact count for most of our queries and otherwise may be slightly off. Since this doesn't affect the hits returned - only the count displayed within the facet bucket - this should not be a problem. If anyone needs an exact count for really large values they can perform a search and use the hits returned. See https://www.elastic.co/docs/reference/aggregations/search-aggregations-metrics-cardinality-aggregation#_counts_are_approximate for more details.

It impacts both granule and collection searches that include facets. I expect a much larger impact to granule queries mainly by number 1 in the solution above and due to the number of granules being vastly larger than the number of collections.

Note that I found a couple of other places in the collection facets and the provider holdings queries that also use nested terms aggregations and might not need it. I'll take a look at those places separately from this ticket and write new tickets if they should be changed.

# Checklist

- [X] I have updated/added unit and int tests that prove my fix is effective or that my feature works (tested existing tests)


search-app/src/cmr/search/services/query_execution/facets/cycle_facets.clj
```diff
@@ -1,17 +1,16 @@
...
-  facets. Creates aggregations for cycles, further aggregated by 
+  "Functions for generating the spatial facets within v2 granule
+  facets. Creates aggregations for cycles, further aggregated by
   pass.")
 
-(def aggregate-by-concept-id {:concept-id
-                              {:terms {:field :concept-id
-                                       :size 1}}})
 
 (def aggregate-by-cycle {:histogram
```
chris-durbin, line 7: Solution change number 1 - don't aggregate by concept-id because it isn't needed in this case.

search-app/src/cmr/search/services/query_execution/facets/facets_results_feature.clj
```diff
@@ -22,7 +22,7 @@
   "Used to build an aggregation to get a count of unique concepts included in the current nested
   aggregation."
   {:reverse_nested {}
-   :aggs {:concept-id {:terms {:field :concept-id :size 1}}}})
+   :aggs {:concept-id {:cardinality {:field :concept-id}}}})
```
chris-durbin, line 25: Solution change number 3 - switch from terms aggregation to cardinality for nested fields in collection facets searches.

ygliuvt (approved):

search-app/src/cmr/search/services/query_execution/facets/granule_v2_facets.clj
```diff
@@ -104,10 +104,6 @@
   "The titles of temporal facet group nodes in order."
   ["Year" "Month" "Day"])
 
-(def group-nodes-in-order-spatial
```
chris-durbin, line 107: The changes in this file were to drop definitions that were not used anywhere making debugging confusing.

search-app/src/cmr/search/services/query_execution/facets/cycle_facets.clj
```diff
@@ -1,17 +1,16 @@
...
                          {:field :cycle
                           :min_doc_count 1
-                          :interval 1}
-                         :aggs aggregate-by-concept-id})
+                          :interval 1}})
+
+(def aggregate-by-concept-id {:concept-id
```
chris-durbin, line 12: Solution change number 2 - switch from terms aggregation to cardinality for cycle and pass nested aggregations.

PR #2293, opened 2025-09-08
CMR-10768: As a CMR operator, I want to CMR to use aliases when searching against Elasticsearch so that we can switch between underlying ES indices

ygliuvt:
# Overview

### What is the objective?

As a CMR operator, I want to CMR to use aliases when searching against Elasticsearch so that we can switch between underlying ES indices

### What are the changes?

Switch CMR to use aliases when searching against Elasticsearch so that we can switch between underlying ES indices

access-control, indexer and bootstrap apps.

# Required Checklist

# Additional Checklist

### DIFF
--- a/access-control-app/src/cmr/access_control/services/acl_service.clj
+++ b/access-control-app/src/cmr/access_control/services/acl_service.clj
@@ -20,6 +20,7 @@
    [cmr.common.services.errors :as errors]
    [cmr.common.services.search.query-model :as qm]
    [cmr.common.util :as util :refer [defn-timed]]
+   [cmr.elastic-utils.es-index-helper :as esi-helper]
    [cmr.elastic-utils.search.es-group-query-conditions :as gc]
    [cmr.elastic-utils.search.es-index :as common-esi]
    [cmr.elastic-utils.search.es-params-converter :as cp]
@@ -470,7 +471,7 @@
   ;; Search is not a dependency of access-control and this must be
   ;; defined for collection search to work
   {:index-name (if (:all-revisions? query)
-                 "1_all_collection_revisions"
+                 (esi-helper/index-alias "1_all_collection_revisions")
                  "collection_search_alias")
    :type-name "collection"})
 
--- a/elastic-utils-lib/src/cmr/elastic_utils/search/ac_elasticsearch.clj
+++ b/elastic-utils-lib/src/cmr/elastic_utils/search/ac_elasticsearch.clj
@@ -41,7 +41,7 @@
     (let [{:keys [concept-id revision-id]} concept
           type (concept->type concept)
           elastic-version revision-id
-          index-name (access-control-index/concept-type->index-name type)
+          index-alias (access-control-index/concept-type->index-alias type)
           elastic-doc (parsed-concept->elastic-doc context concept concept)
           ;; "only index the document if the given version is equal or higher than
           ;; the version of the stored document."
@@ -50,7 +50,7 @@
                              {:_id concept-id
                               :version elastic-version
                               :version_type version-type})]
-      (assoc elastic-doc :_index index-name))
+      (assoc elastic-doc :_index index-alias))
 
     (catch Throwable e
       (error e (str "Skipping failed catalog item. Exception trying to convert "
--- a/indexer-app/src/cmr/indexer/data/collection_granule_aggregation_cache.clj
+++ b/indexer-app/src/cmr/indexer/data/collection_granule_aggregation_cache.clj
@@ -104,7 +104,7 @@
    concept id to collection information. The collection will only be in the map if it has granules."
   [context]
   (-> (es-helper/search (indexer-util/context->conn context)
-                        "1_small_collections,1_c*" ;; Searching all granule indexes
+                        "1_small_collections_alias,1_c*_alias" ;; Searching all granule indexes
                         ["granule"] ;; With the granule type.
                         {:query (esq/match-all)
                          :size 0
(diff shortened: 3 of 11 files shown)


access-control-app/src/cmr/access_control/services/acl_service.clj
```diff
@@ -470,7 +471,7 @@
   ;; Search is not a dependency of access-control and this must be
   ;; defined for collection search to work
   {:index-name (if (:all-revisions? query)
-                 "1_all_collection_revisions"
+                 (esi-helper/index-alias "1_all_collection_revisions")
```
jceaser, line 474: is there a const somewhere that defines the 1_ part of collections?

jmaeng72: on this note, there needs be a refactor ticket that substitutes all these hardcoded index names, because right now, the "1" is based on index-set-id, which could be changed in the future and would require us to go into the entire code and change all these strings to "2_" which is... not great.

jmaeng72: meanwhile cmr.indexer.data.index-set has some defs of index names already in there

elastic-utils-lib/src/cmr/elastic_utils/search/ac_elasticsearch.clj
```diff
@@ -41,7 +41,7 @@
     (let [{:keys [concept-id revision-id]} concept
           type (concept->type concept)
           elastic-version revision-id
-          index-name (access-control-index/concept-type->index-name type)
+          index-alias (access-control-index/concept-type->index-alias type)
```
jmaeng72, line 44: Just to be clear, we only care about using aliases to search? We don't want to use aliases to delete, right?

I'm asking because our delete funcs also use direct index names and I was wondering if we planned to use aliases to delete the index

indexer-app/src/cmr/indexer/data/collection_granule_aggregation_cache.clj
```diff
@@ -119,7 +119,7 @@
   (let [revision-date (t/minus (tk/now) (t/seconds granules-updated-in-last-n))
         revision-date-str (datetime-helper/utc-time->elastic-time revision-date)]
     (-> (es-helper/search (indexer-util/context->conn context)
-                          "1_small_collections,1_c*" ;; Searching all granule indexes
+                          "1_small_collections_alias,1_c*_alias" ;; Searching all granule indexes
```
jmaeng72, line 122: wondering if we should have a universal 'convert-to-alias-name' func of some type and use it everywhere, instead of hardcoding in the '_alias' suffix

I believe you already have a func like this called index-alias

indexer-app/src/cmr/indexer/data/elasticsearch.clj
```diff
@@ -246,8 +246,7 @@
                       (update :tool-associations #(parse-non-tombstone-associations context %)))
           elastic-id (get-elastic-id concept-id revision-id all-revisions-index?)
           index-names (idx-set/get-concept-index-names
-                       context concept-id revision-id options
-                       concept)
+                       context concept-id revision-id options concept)
```
jmaeng72, line 249: There is a func called (defn get-deleted-granules) inside `cmr.search.services.query-service` that might need to be converted into an alias form

elastic-utils-lib/src/cmr/elastic_utils/search/ac_elasticsearch.clj
ygliuvt: We should use alias when deleting content from an index, but not when deleting the index itself.

indexer-app/src/cmr/indexer/data/collection_granule_aggregation_cache.clj
ygliuvt: The helper function doesn't apply on a list of indices as a string. This is just a string literal. I didn't think it is worth the effort to make the string literal based on some other string literal. It feels cleaner this way.

access-control-app/src/cmr/access_control/services/acl_service.clj
ygliuvt: Agreed. We have index-set-id defined in some namespaces, but hardcoded in most. Basically we don't support dynamic index-set-id as originally planned. We should probably just remove this index-set id feature from CMR in the future, but not in this ticket.

indexer-app/src/cmr/indexer/data/elasticsearch.clj
ygliuvt: Good point. I will change it to use alias.

elastic-utils-lib/src/cmr/elastic_utils/search/ac_elasticsearch.clj
jmaeng72: So when we delete a concept we use the (defn- delete-concept-default-helper) func that uses the direct list of indexes to delete documents from indices in ES. Do we want to keep that same behavior as is, or use aliases?

This (defn- delete-concept-default-helper) func is called in the indexer app during the process of deleting a granule from ingest app

I think, overall, I would like to know the plan with aliases when it comes to deleting concepts in ES

indexer-app/src/cmr/indexer/data/collection_granule_aggregation_cache.clj
jmaeng72: This is fine for now and can be cleaned up when we allow flexible index naming in the future

elastic-utils-lib/src/cmr/elastic_utils/search/ac_elasticsearch.clj
ygliuvt: The rule is: if we know the exact index name for a given concept/collection (for example. when we get the index name from the index-set directly as in the `delete-concept-default-helper` function), we should use the index name directly. There is no need to go from the exact index name to alias then down to the index again. If we don't know the exact index (such as when searching in general, we don't get the index name directly from index-set), we should use alias and elasticsearch will follow the alias to the correct index.

To be specific to the function you mentioned above, since the function gets the exact index name directly from index-set prior to make the ES call, we should just use the index name. Had the function went with implying the index name based on the concept-id, we should use alias. I hope this clears things up.

jmaeng72: Ah I see, yes it does clear it up. Thank you. I am good with this approach

PR #2299, opened 2025-09-16
CMR-10697: Moved ingest (NRT) subscriptions from metadata-db to ingest to make it more performant.

eereiter:
# Overview

### What is the objective?

Moved ingest (NRT) subscriptions from metadata-db to ingest to make it more performant.

### What are the changes?

Moved ingest (NRT) subscriptions from metadata-db to ingest to make it more performant.

Ingest subscriptions

# Required Checklist

# Additional Checklist

### DIFF
--- a/access-control-app/src/cmr/access_control/services/acl_service.clj
+++ b/access-control-app/src/cmr/access_control/services/acl_service.clj
@@ -20,6 +20,7 @@
    [cmr.common.services.errors :as errors]
    [cmr.common.services.search.query-model :as qm]
    [cmr.common.util :as util :refer [defn-timed]]
+   [cmr.elastic-utils.es-index-helper :as esi-helper]
    [cmr.elastic-utils.search.es-group-query-conditions :as gc]
    [cmr.elastic-utils.search.es-index :as common-esi]
    [cmr.elastic-utils.search.es-params-converter :as cp]
@@ -470,7 +471,7 @@
   ;; Search is not a dependency of access-control and this must be
   ;; defined for collection search to work
   {:index-name (if (:all-revisions? query)
-                 "1_all_collection_revisions"
+                 (esi-helper/index-alias "1_all_collection_revisions")
                  "collection_search_alias")
    :type-name "collection"})
 
--- a/elastic-utils-lib/src/cmr/elastic_utils/search/ac_elasticsearch.clj
+++ b/elastic-utils-lib/src/cmr/elastic_utils/search/ac_elasticsearch.clj
@@ -41,7 +41,7 @@
     (let [{:keys [concept-id revision-id]} concept
           type (concept->type concept)
           elastic-version revision-id
-          index-name (access-control-index/concept-type->index-name type)
+          index-alias (access-control-index/concept-type->index-alias type)
           elastic-doc (parsed-concept->elastic-doc context concept concept)
           ;; "only index the document if the given version is equal or higher than
           ;; the version of the stored document."
@@ -50,7 +50,7 @@
                              {:_id concept-id
                               :version elastic-version
                               :version_type version-type})]
-      (assoc elastic-doc :_index index-name))
+      (assoc elastic-doc :_index index-alias))
 
     (catch Throwable e
       (error e (str "Skipping failed catalog item. Exception trying to convert "
--- a/elastic-utils-lib/src/cmr/elastic_utils/search/access_control_index.clj
+++ b/elastic-utils-lib/src/cmr/elastic_utils/search/access_control_index.clj
@@ -8,6 +8,7 @@
    [cmr.common.services.errors :as errors]
    [cmr.common.util :as util :refer [defn-timed]]
    [cmr.elastic-utils.es-helper :as es-helper]
+   [cmr.elastic-utils.es-index-helper :as esi-helper]
    [cmr.elastic-utils.index-util :as m :refer [defmapping defnestedmapping]]
    [cmr.elastic-utils.search.es-index :as esi]
    [cmr.elastic-utils.search.es-query-to-elastic :as q2e]
@@ -132,7 +133,7 @@
 
 (defmethod esi/concept-type->index-info :access-group
   [_context _ _]
-  {:index-name group-index-name
+  {:index-name (esi-helper/index-alias group-index-name)
    :type-name group-type-name})
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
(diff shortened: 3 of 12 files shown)


ingest-app/src/cmr/ingest/api/subscriptions.clj
```diff
@@ -85,8 +85,8 @@
         endpoint (:EndPoint subscription-concept)
         default-url-validator (UrlValidator. UrlValidator/ALLOW_LOCAL_URLS)]
 
-    (if (= method "ingest")
-      (if-not (or (some? (re-matches #"arn:aws:sqs:.*" endpoint)) (.isValid default-url-validator endpoint))
+    (when (= method "ingest")
```
indiejames: nit: I think you can collapse this to a single `when`

ingest-app/src/cmr/ingest/services/nrt_subscription/subscriptions.clj
```diff
@@ -36,13 +37,29 @@
...
-                                      :concept-type :subscription
-                                      :collection-concept-id coll-concept-id})))
+   (let [response (mdb2/find-concepts context
+                                      {}
+                                      :subscription
+                                      {:raw? true
+                                       :http-options {:query-params
+                                                      {:collection-concept-id coll-concept-id
+                                                       :latest true
+                                                       :concept-type :subscription}}})]
```
indiejames: Should this have `:subscription-type` set as well?

ingest-app/src/cmr/ingest/services/nrt_subscription/subscriptions.clj
```diff
@@ -36,13 +37,29 @@
   "Get the subscriptions from the database. This function primarily exists so that
   it can be stubbed out for unit tests."
   ([context]
-   (mdb-search/find-concepts context {:latest true
-                                      :concept-type :subscription
-                                      :subscription-type "granule"}))
+   (let [response (mdb2/find-concepts context
```
indiejames: there is a lot of duplicate code in these two methods. I think you could just have the first one invoke the second one with `nil` for `coll-concept-id`, then in the second one set `:collection-concept-id` on the `query-params` map if it is not `nil`

ingest-app/src/cmr/ingest/services/nrt_subscription/subscriptions.clj
```diff
@@ -36,13 +37,29 @@
...
-                                      :collection-concept-id coll-concept-id})))
+   (let [response (mdb2/find-concepts context
+                                      {}
+                                      :subscription
+                                      {:raw? true
+                                       :http-options {:query-params
+                                                      {:collection-concept-id coll-concept-id
+                                                       :latest true
+                                                       :concept-type :subscription}}})]
+     (if (= (:status response) 200)
```
indiejames: I think you can just use a `when` here instead of `if` with an explicit `nil`. `when` evaluates to `nil` if its argument is `false`.

ingest-app/src/cmr/ingest/services/nrt_subscription/subscriptions.clj
eereiter: I doesn't need it as the collection_concept_id is only filled out for granule subscription types, but it doesn't hurt to add it because it does make the code a little easier to read, once its refactored.

ingest-app/src/cmr/ingest/services/nrt_subscription/subscriptions.clj
eereiter: I have updated the code and combined the two methods.

ingest-app/src/cmr/ingest/api/subscriptions.clj
eereiter: I have combined them into a single when clause.

ingest-app/src/cmr/ingest/services/nrt_subscription/subscriptions.clj
eereiter: I have updated the code.

ingest-app/src/cmr/ingest/services/ingest_service/subscription.clj
```diff
@@ -25,11 +29,48 @@
...
         {:keys [concept-id revision-id]} (mdb2/save-concept
                                           context
                                           (assoc concept :provider-id (:provider-id concept)
                                                  :native-id (:native-id concept)))]
+    (subscriptions/add-or-delete-ingest-subscription-in-cache context concept)
     {:concept-id concept-id
      :native-id (:native-id concept)
      :revision-id revision-id}))
+
+(declare delete-subscription)
```
chris-durbin, line 44: We should only need to declare something if it's used in the file prior to being defined, but I see it is defined immediately afterwards - is this some issue with the `defn-timed` macro?

ingest-app/src/cmr/ingest/services/ingest_service/subscription.clj
```diff
@@ -25,11 +29,48 @@
...
+#_{:clj-kondo/ignore [:unresolved-symbol]}
```
chris-durbin, line 45: I think there's a way to make this work for the `defn-timed` macro without needing to put this in every time it is used. I'll see if I can submit a simple PR for it.

subscription/src/subscription_worker.py
```diff
@@ -53,10 +52,13 @@ def process_messages(sns_client, topic, messages, access_control, search):
             subscriber = message_attributes['subscriber']['Value']
             collection_concept_id = message_attributes['collection-concept-id']['Value']
 
-            if(access_control.has_read_permission(subscriber, collection_concept_id)):
-                logger.debug(f"Subscription worker: {subscriber} has permission to receive granule notifications for {collection_concept_id}")
-                message_msg = search.process_message(message_body['Message'])
-                message_body['Message'] = message_msg
+            #sets the time in milliseconds
```
chris-durbin, line 56: Commented out code - I'm assuming you want to keep these log messages in?

ingest-app/src/cmr/ingest/services/ingest_service/subscription.clj
eereiter: There were thousands of linter errors and the linters don't recognize the macros, so we had to conduct a campaign to remove them sometime last year. This is one of the outcomes of that effort.

ingest-app/src/cmr/ingest/services/ingest_service/subscription.clj
eereiter: That would be great.  Its not because of the code or clojure, its because the linter doesn't understand the macro.

subscription/src/subscription_worker.py
eereiter: Yes, we will need to add acls back in, but we have to cache the values, otherwise we will bring CMR down.

chris-durbin: Oh I missed line 57 - for commented out code like this and in the test file I think we should have a ticket number in the comments - e.g. ## Re-enable ACL check with CMR-xxxxx

eereiter: I have added the comment: 
 Re-enable ACL check with CMR-10855 
to the subscription worker and subscription worker test in the commented out ACL code.

PR #2300, opened 2025-09-17
Cmr 10805 - Add information about number of ES searches conducted and number of shards read during searches to log message

indiejames:
# Overview

### What is the objective?
Add logging to show number of ES searches performed and number of shards read during search.
We need this information to identify possible performance improvements.

### What are the changes?
Added logging

search

# Required Checklist

# Additional Checklist

### DIFF
--- a/elastic-utils-lib/src/cmr/elastic_utils/search/es_index.clj
+++ b/elastic-utils-lib/src/cmr/elastic_utils/search/es_index.clj
@@ -231,7 +231,7 @@
    (do-send context index-info query)))
 
 (defn- add-shard-count-to-context
-  "Add information about the indexes used to the `meta` field of the context"
+  "Add information about the shards read during search to the metadata of the context"
   [context shard-count]
   (when-let [metadata (meta context)]
     (let [shard-counts-atom (:shard-counts metadata)]
--- a/elastic-utils-lib/src/cmr/elastic_utils/search/es_index.clj
+++ b/elastic-utils-lib/src/cmr/elastic_utils/search/es_index.clj
@@ -229,6 +229,12 @@
   (transmit-conn/handle-socket-exception-retries
    (do-send context index-info query)))
 
+(defn- add-shard-count-to-context
+  "Add information about the indexes used to the `meta` field of the context"
+  [context shard-count]
+  (when-let [meta (:meta context)]
+    (swap! meta update-in [:shard-counts] conj shard-count)))
+
 (defmulti send-query-to-elastic
   "Created to trace only the sending of the query off to elastic search."
   (fn [_context query]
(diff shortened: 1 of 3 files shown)


search-app/src/cmr/search/services/query_service.clj
```diff
@@ -194,23 +194,32 @@
                 context concept-type params)]
     (common-params/generate-param-query-conditions context concept-type params)))
 
+(defn- add-meta-to-context
+  "Add metadata to the context to hold an atom that will contain shard counts"
+  [context]
+  (let [shard-counts (atom [])]
+    (with-meta context {:shard-counts shard-counts})))
+
```
indiejames, line 202: I'm using an `atom` here because I want to be able to modify it at the end of the call stack. By adding it as metadata to the `context`  it can be passed back up to the top-level without modifying the return values of all the intermediate functions.

jmaeng72 (dismissed):

elastic-utils-lib/src/cmr/elastic_utils/search/es_index.clj
```diff
@@ -249,15 +256,17 @@
...
+           "with sort" (pr-str sort-params)
+           "with aggregations" (pr-str aggregations)
+           "and highlights" (pr-str highlights))
     (when-let [scroll-id (:scroll-id query-map)]
       (debug "Using scroll-id" scroll-id))
     (when-let [search-after (:search_after query-map)]
       (debug "Using search-after" (pr-str search-after)))
-    (let [response (send-query context index-info query-map)]
+    (let [response (send-query context index-info query-map)
+          shard-count (get-in response [:_shards :total])]
```
jmaeng72: so ES gives back this total shards value automatically in a search query response?

chris-durbin: I'm wondering if we need to worry about this possibly not being returned for some queries. Do you think it makes sense to add in a default value of 0, or logging an error if it isn't present? The code will throw an exception when it tries to  add nil in `(apply + @shard-counts-atom)`.

indiejames: yes

indiejames: I thought it was always returned, but maybe not. I can either use a default of 0 or just not add that info to the logging when it's not available
. I don't think an error message makes sense.

chris-durbin: default to 0 sounds best to me

indiejames: Changed the code to default to 0

PR #2301, opened 2025-09-19
CMR-10831: Add Refresh granule count cache to event bridge.

eereiter:
# Overview

### What is the objective?

Adding Refresh of granule counts cache to event bridge.

### What are the changes?

Adding Refresh of granule counts cache to event bridge.

job router and a new function in bootstrap.

# Required Checklist

# Additional Checklist


job-utilities/eventbridge_schedule/deploy_schedule.py
```diff
@@ -48,81 +49,169 @@ def make_schedule_expression(job):
...
+    """
+    lambda_client = boto3.client('lambda')
+    try:
+        lambda_details = lambda_client.get_function(
+            FunctionName="job-router-"+environment
+        )
+    except ClientError as e:
+        print("Error getting lambda function: " + e.response['Error']['Code'])
+        sys.exit(1)
+    return [lambda_client, lambda_details]
```
TylerHeald1: If we need the lambda client later (and other boto3 clients), can we just declare it outside of functions below the imports so we aren't having to handle passing it around

eereiter: I have updated the code to use global variables.

job-utilities/eventbridge_schedule/deploy_schedule.py
```diff
@@ -48,81 +54,174 @@ def make_schedule_expression(job):
...
         except ClientError as e:
             print("Error adding permissions to lambda: " + e.response["Error"]["Code"])
             print("This does not mean the deployment will fail, it could just indicate the permission already exists")
 
+def create_scheduler_rules(jobs_map):
+    """
+    Creates AWS EventBridge scheduler rules to schedule all the refresh cache jobs.
+    """
+    rules = {}
+    global events_client
```
TylerHeald1: Doesn't need to be explicitly declared as a global in this way and assigned in a function, the events_client = boto3.client('events') can just be assigned at the top of the module

eereiter: I have updated the code.

PR #2309, opened 2025-09-26
CMR-10826: Create granule counts cache for all collections

DuJuan:
# Overview

### What is the objective?

Implement a new caching mechanism for granule counts in the CMR search application. This will improve query performance and reduce load on the Elasticsearch cluster.  

### What are the changes?

Implemented a new granule-counts-cache namespace in the search application
Added Redis-based caching for granule counts
Modified the existing granule count retrieval logic to use the new cache
Added configuration options for cache TTL and Redis connection details
Implemented cache refresh mechanisms

CMR Search Application
CMR Ingest Application (for cache invalidation)

# Required Checklist

- [] clj-kondo has been run locally and all errors in changed files are corrected

# Additional Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a Redis-backed granule counts cache and an authenticated admin POST endpoint to refresh it.
* **Performance**
  * Granule counts are cached to reduce repeated queries and improve response times.
* **Reliability**
  * Cache refresh, clear, and validation reduce anomalous aggregation results; refresh errors are logged.
* **Tests**
  * Added unit and integration tests covering cache lifecycle, refresh/clear behavior, and admin API exposure.

### DIFF
--- a/metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
+++ b/metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
@@ -653,7 +653,6 @@
     :else
     (let [table (tables/get-table-name provider :granule)
           ;; Oracle caps IN clause at 1000, so chunk the IDs
-          ;; This is only theoretical, it is unlikely we will get close to even 100.
           batches (partition-all 900 granule-concept-ids)]
       (->> batches
            (mapcat
--- a/search-app/docs/api.md
+++ b/search-app/docs/api.md
@@ -2969,7 +2969,7 @@ It supports all normal granule parameters. It requires the following parameters.
     * `entry_id` - Collection entry id (concatenation of short_name and version).
     * `entry_title` - Collection entry title.
     * `short_name` and `version` - Both parameters must be provided together to identify a specific collection.
-  
+
   It is recommended that the timeline search be limited to a few collections for good performance.
 
 The response format is in JSON. Intervals are returned as tuples containing three numbers like `[949363200,965088000,4]`. The two numbers are the start and stop date of the interval represented by the number of seconds since the epoch. The third number is the number of granules within that interval.
--- a/search-app/src/cmr/search/services/parameters/parameter_validation.clj
+++ b/search-app/src/cmr/search/services/parameters/parameter_validation.clj
@@ -795,7 +795,7 @@
                 (:entry-title params)
                 (:echo-collection-id params)
                 (and (:short-name params) (:version params)))
-    ["Timeline searches must include a collection identifier parameter (concept-id, collection-concept-id, echo-collection-id, entry-id, entry-title, or both short-name and version) to limit the search scope."]))
+    ["Timeline searches must include a collection identifier parameter (concept_id, collection_concept_id, entry_id, entry_title, or both short_name and version) to limit the search scope."]))
 
 (defn- timeline-start-date-validation
   "Validates the timeline start date parameter"
--- a/system-int-test/test/cmr/system_int_test/search/granule/granule_timeline_test.clj
+++ b/system-int-test/test/cmr/system_int_test/search/granule/granule_timeline_test.clj
@@ -92,14 +92,14 @@
 
         (testing "missing collection identifier"
           (is (= {:status 400
-                  :errors ["Timeline searches must include a collection identifier parameter (concept-id, collection-concept-id, echo-collection-id, entry-id, entry-title, or both short-name and version) to limit the search scope."]}
+                  :errors ["Timeline searches must include a collection identifier parameter (concept_id, collection_concept_id, entry_id, entry_title, or both short_name and version) to limit the search scope."]}
                  (search/get-granule-timeline {:start-date "2000-01-01T00:00:00Z"
                                                :end-date "2001-01-01T00:00:00Z"
                                                :interval :year}))))
 
         (testing "short-name without version"
           (is (= {:status 400
-                  :errors ["Timeline searches must include a collection identifier parameter (concept-id, collection-concept-id, echo-collection-id, entry-id, entry-title, or both short-name and version) to limit the search scope."]}
+                  :errors ["Timeline searches must include a collection identifier parameter (concept_id, collection_concept_id, entry_id, entry_title, or both short_name and version) to limit the search scope."]}
                  (search/get-granule-timeline {:start-date "2000-01-01T00:00:00Z"
                                                :end-date "2001-01-01T00:00:00Z"
                                                :interval :year
@@ -107,7 +107,7 @@
 
         (testing "version without short-name"
           (is (= {:status 400
-                  :errors ["Timeline searches must include a collection identifier parameter (concept-id, collection-concept-id, echo-collection-id, entry-id, entry-title, or both short-name and version) to limit the search scope."]}
+                  :errors ["Timeline searches must include a collection identifier parameter (concept_id, collection_concept_id, entry_id, entry_title, or both short_name and version) to limit the search scope."]}
                  (search/get-granule-timeline {:start-date "2000-01-01T00:00:00Z"
                                                :end-date "2001-01-01T00:00:00Z"
                                                :interval :year
@@ -154,7 +154,7 @@
 
         (testing "missing parameters with post"
           (is (= {:status 400 :errors ["Parameter [foo] was not recognized."
-                                       "Timeline searches must include a collection identifier parameter (concept-id, collection-concept-id, echo-collection-id, entry-id, entry-title, or both short-name and version) to limit the search scope."
+                                       "Timeline searches must include a collection identifier parameter (concept_id, collection_concept_id, entry_id, entry_title, or both short_name and version) to limit the search scope."
                                        "start_date is a required parameter for timeline searches"
                                        "end_date is a required parameter for timeline searches"
                                        "interval is a required parameter for timeline searches"]}
@@ -355,7 +355,7 @@
           :start-date "1992-01-01T00:00:00Z"
           :end-date "2002-02-01T00:00:00Z"
           :interval :month}))
-      
+
       (testing "correct headers are returned"
         (let [response (search/get-granule-timeline {:concept-id (:concept-id coll1)
                                                      :start-date "2000-01-01T00:00:00Z"


search-app/src/cmr/search/data/elastic_search_index.clj
```diff
@@ -182,7 +182,7 @@
 
 (defn get-collection-granule-counts
   "Returns the collection granule count by searching elasticsearch by aggregation"
-  [context provider-ids]
+  [context provider-ids]
```
eereiter: remove space after provider-ids]

search-app/src/cmr/search/data/granule_counts_cache.clj
```diff
@@ -0,0 +1,95 @@
...
+   [cmr.common.services.errors :as e]
+   [cmr.common.services.search.query-model :as qm]
+   [cmr.elastic-utils.search.es-query-to-elastic :as q2e]
+   [cmr.elastic-utils.search.es-index :as common-esi]
+   [cmr.redis-utils.config :as redis-config]
+   [cmr.redis-utils.redis-cache :as redis-cache]))
+
+(defconfig granule-counts-cache-time-to-live
+  "Time to live for granule counts cache in seconds."
+  {:default 3600
```
eereiter: We are going to make the schedule to refresh the cache 1 hour for now, but this value should be bigger.  We don't want to empty the cache before it can load - Change 3600 to (* 4 3600). This is 4 hours and it is easier for developers to know how long it is.

search-app/src/cmr/search/data/granule_counts_cache.clj
```diff
@@ -0,0 +1,95 @@
...
+(defn refresh-granule-counts-cache
+  "Refreshes the granule counts cache with the latest data."
+  ([context]
+   (refresh-granule-counts-cache context #(get-collection-granule-counts context nil)))
+  ([context func]
+   (let [granule-counts (func)
+         cache (cache/context->cache context granule-counts-cache-key)]
+     (log/info "Refreshing granule counts cache with" (count granule-counts) "entries")
+     (when cache
+       (cache/set-value cache granule-counts-cache-key granule-counts)))))
```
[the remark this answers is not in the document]
eereiter: This is a good suggestion.

search-app/src/cmr/search/routes.clj
```diff
@@ -21,7 +21,8 @@
    [cmr.search.middleware.shapefile-simplification :as shapefile-simplifier]
    [cmr.search.services.messages.common-messages :as msg]
    [cmr.search.site.routes :as site-routes]
-   [compojure.core :refer [GET context routes]]
+   [cmr.search.data.granule-counts-cache :as granule-counts-cache]
```
eereiter: This line 24 should be up a little so that the list is alphabetical.

search-app/src/cmr/search/system.clj
```diff
@@ -32,7 +32,8 @@
    [cmr.search.services.query-execution.has-granules-results-feature :as hgrf]
    [cmr.transmit.config :as transmit-config]
    [cmr.transmit.launchpad-user-cache :as launchpad-user-cache]
-   [cmr.transmit.urs :as urs]))
+   [cmr.transmit.urs :as urs]
+   [cmr.search.data.granule-counts-cache :as granule-counts-cache]))
```
eereiter: This line should be up a bit so the requires are alphabetical.

search-app/src/cmr/search/routes.clj
DuJuan: Good call, done.

search-app/src/cmr/search/system.clj
DuJuan: Good call, done.

search-app/src/cmr/search/data/granule_counts_cache.clj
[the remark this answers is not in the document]
DuJuan: Indeed and done.

search-app/src/cmr/search/data/granule_counts_cache.clj
DuJuan: You're absolutely correct. Increasing the cache time-to-live from 1 hour to 4 hours is a good idea to prevent the cache from emptying before it can be fully loaded. Updated accordinly.

search-app/src/cmr/search/data/granule_counts_cache.clj
```diff
@@ -0,0 +1,98 @@
...
+  {:default 3600
+   :type Long})
+
+(def granule-counts-cache-key
+  :granule-counts-cache)
+
+(defn create-granule-counts-cache
+  "Creates and returns a new cache for storing granule counts."
+  []
+  (redis-cache/create-redis-cache {:keys-to-track [granule-counts-cache-key]
```
jmaeng72, line 19: do we want to add a check to see if the cache already exists before creating it?

search-app/src/cmr/search/data/granule_counts_cache.clj
```diff
@@ -0,0 +1,98 @@
...
+  (let [condition (if (seq provider-ids)
+                    (qm/string-conditions :provider-id provider-ids true)
+                    qm/match-all)
+        query (qm/query {:concept-type :granule
+                         :condition condition
+                         :page-size 0
+                         :aggregations {:by-provider
+                                        {:terms {:field (q2e/query-field->elastic-field
+                                                         :provider-id :granule)
+                                                 :size 10000}
```
jmaeng72: will this size never change?

search-app/src/cmr/search/data/granule_counts_cache.clj
eereiter: I saw that you pushed some changes, but this hasn't been changed yet.

search-app/src/cmr/search/data/granule_counts_cache.clj
eereiter: This code doesn't actually create the cache, this code creates the redis client to the redis cache which runs as an external service.  This code follows what we have done with all of our other redis client cache connections.

search-app/src/cmr/search/data/granule_counts_cache.clj
eereiter: This code is copied verbatim from  search-app/src/cmr/search/data/elastic_search_index.clj.  I don't know why the size was hard coded - probably a good idea as a def.  The size parameter controls the number of top term buckets for the aggregation. In this case 10,000 provider buckets.  I think its a large number to get all of the providers without running into not getting them all.  The elastic search default for the size is 10. 
Below the size is up to 10,000 collection buckets per provider.

search-app/src/cmr/search/data/granule_counts_cache.clj
DuJuan: Forgot to include that change in the commit. You should see it now.

search-app/src/cmr/search/data/granule_counts_cache.clj
DuJuan: Great question and response! Would the following provide better clarity?
```suggestion
(defn create-granule-counts-cache
  "Creates and returns a new Redis client for connecting to the granule counts cache.
   This does not create a new cache, but establishes a connection to the existing
   Redis cache service."
  []
  (redis-cache/create-redis-cache {:keys-to-track [granule-counts-cache-key]
                                   :read-connection (redis-config/redis-read-conn-opts)
                                   :primary-connection (redis-config/redis-conn-opts)
                                   :ttl (granule-counts-cache-time-to-live)}))
```

search-app/src/cmr/search/data/granule_counts_cache.clj
DuJuan: I did think about removing the hard coded value initially. Still can if you both feel it is necessary.

search-app/src/cmr/search/data/granule_counts_cache.clj
```diff
@@ -0,0 +1,100 @@
...
+               (do
+                 (when (> extra-collection-count 0)
+                   (e/internal-error!
+                    (format "Provider %s has more collections ([%s]) with granules than we support"
+                            provider-id
+                            extra-collection-count)))
+                 [(concepts/build-concept-id {:sequence-number coll-seq-id
+                                              :provider-id provider-id
+                                              :concept-type :collection})
+                  num-granules])))))
```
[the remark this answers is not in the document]
DuJuan: Good catch, updated accordingly.

search-app/src/cmr/search/data/granule_counts_cache.clj
jmaeng72: feel like we should rename this to 'create-redis-cache-client' then

DuJuan: Great suggestion, done.

search-app/src/cmr/search/routes.clj
```diff
@@ -106,6 +107,15 @@
        (GET "/robots.txt" req (get-robots-txt-response test-environment)))
      (api-routes/build-routes system)
      (site-routes/build-routes system)
+     ;; Endpoints for managing the granule counts cache
+     (POST "/refresh-granule-counts-cache" {:keys [request-context]}
+       (acl/verify-ingest-management-permission request-context :update)
+       (granule-counts-cache/refresh-granule-counts-cache request-context)
+       {:status 200})
+     (POST "/clear-granule-counts-cache" {:keys [request-context]}
```
ygliuvt: No need to have this clear endpoint. Remove.

search-app/src/cmr/search/data/granule_counts_cache.clj
```diff
@@ -21,7 +21,7 @@
 (def granule-counts-cache-key
   :granule-counts-cache)
 
-(defn create-granule-counts-cache
+(defn create-redis-cache-client
```
jmaeng72: Actually, instead of redis, it should be granule count because it's specific to that.


'create-granule-counts-cache-client'

ygliuvt (changes requested):
The granule count cache is set up with a TTL of 4 hours which causes it to expire every 4 hours and Maybe I missed it, but I don't see any code to automatically refresh the cache and the `get-granule-counts` function is defined but not used. This leads me to believe that this ticket is only to create the cache, but not use it? We should not have done that.

Also, the way the cache is set up in this PR will not work properly. The reason is that the cache will only be repopulated after it expires even after we change all the access code to use the `get-granule-counts` function. But that means every 4 hours we will have a search request that is slowed because it needs to repopulate the cache. 

We should change the ticket to have all the code working for setting up the cache and using the cache. The cache should be refreshed by a background job that runs every 4 hours by default and we can remove the TTL setting from the cache.

search-app/src/cmr/search/routes.clj
DuJuan: Good catch, I've updated accordingly.

search-app/src/cmr/search/data/granule_counts_cache.clj
DuJuan: Agreed,'create-granule-counts-cache-client' is a better name for the function. Done!

search-app/src/cmr/search/data/granule_counts_cache.clj
```diff
@@ -0,0 +1,86 @@
...
+
+(def granule-counts-cache-key
+  :granule-counts-cache)
+
+(defn create-redis-cache-client
+  "Creates and returns a new cache for storing granule counts."
+  []
+  (redis-cache/create-redis-cache {:keys-to-track [granule-counts-cache-key]
+                                   :read-connection (redis-config/redis-read-conn-opts)
+                                   :primary-connection (redis-config/redis-conn-opts)}))
```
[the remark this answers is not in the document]
DuJuan: The refresh of the granule count cache will be triggered by a Lambda function, which is activated by an EventBridge scheduled event.

search-app/src/cmr/search/data/granule_counts_cache.clj
```diff
@@ -0,0 +1,86 @@
...
+                                      provider-id
+                                      extra-collection-count)))
+                         coll-buckets (get-in provider-bucket [:by-collection-id :buckets])]]
+               (for [coll-bucket coll-buckets
+                     :let [coll-seq-id (:key coll-bucket)
+                           num-granules (:doc_count coll-bucket)]]
+                 [(concepts/build-concept-id {:sequence-number coll-seq-id
+                                              :provider-id provider-id
+                                              :concept-type :collection})
+                  num-granules])))))
```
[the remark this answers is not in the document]
DuJuan: Good catch, done.

search-app/src/cmr/search/routes.clj
```diff
@@ -106,6 +107,11 @@
        (GET "/robots.txt" req (get-robots-txt-response test-environment)))
      (api-routes/build-routes system)
      (site-routes/build-routes system)
+     ;; Endpoints for managing the granule counts cache
+     (POST "/refresh-granule-counts-cache" {:keys [request-context]}
+       (acl/verify-ingest-management-permission request-context :update)
+       (granule-counts-cache/refresh-granule-counts-cache request-context)
+       {:status 200})
```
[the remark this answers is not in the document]
DuJuan: This is will addressed on a later ticket.

PR #2314, opened 2025-10-01
CMR-10871: Performance optimization for granule search with granule concept id parameter

ygliuvt:
# Overview

### What is the objective?

This PR updates granule search with a granule concept ID parameter to restrict the Elasticsearch query to only the indices of the collections tied to those concept IDs, instead of all collections from the associated providers.

### What are the changes?

This PR updates granule search with a granule concept ID search parameter to retrieve the corresponding collection concept IDs directly from the metadata DB. It then limits the subsequent granule query in Elasticsearch to only the indices of those collections, instead of searching across all collections belonging to the providers derived from the granule concept IDs.

It adds provider cache in the search app to make retrieving the providers info more efficient which is needed to construct the collection concept ids sql query for a list of given granule concept ids.

This PR also added fix to short circuit granule queries with MatchNoneCondition to Elasticsearch to improve the search performance.

This PR improves search performance of the following granule searches:

#### Search by granule concept ids (only goes to the Elasticsearch indices of the collections tied to the granules)
e.g. with granule search-by-day-night data, the following search only goes to the small collections index rather than all collection indices of the providers:
curl -i -g "http://localhost:3003/granules?concept_id[]=G1200000011-PROV1&concept_id[]=G1200000018-PROV2"

#### Search by non-existent granule concept ids (skip granule query to Elasticsearch)
e.g.
curl -i "http://localhost:3003/granules?concept_id=G123456-PROV1"

#### Search without ACL no longer executing all granule searches
e.g. with collection-search-with-temporal-acls-test data, the following search without the granule ACL no longer results in all granule searches in Elasticsearch:
curl -i "http://localhost:3003/granules?collection_concept_id=C1200000012-PROV1"

# Required Checklist

# Additional Checklist

## Summary by CodeRabbit

* **New Features**
  * Granule searches using granule concept IDs now derive related collection IDs for more accurate combined filtering.
  * A provider cache was added and is now listed in admin cache endpoints.

* **Performance**
  * Granule queries that cannot match are short-circuited to return empty results without querying Elasticsearch.

* **Tests**
  * Added and updated tests covering provider lookups and collection ID derivation.


metadata-db-app/src/cmr/metadata_db/services/concept_service.clj
```diff
@@ -1151,3 +1151,10 @@
          tombstone-cut-off-date
          concept-truncation-batch-size)
       concept-truncation-batch-size)))
+
+(defn get-collection-concept-ids
+  "Returns the collection concept ids for the given provider and granule concept id
+   provider and granule concept id has been verified to match before invoking this function."
+  [context provider granule-concept-ids]
+  (let [db (util/context->db context)]
+    (c/get-collection-concept-ids db provider granule-concept-ids)))
```
eereiter: Add a new line here.

search-app/src/cmr/search/services/parameters/conversion.clj
```diff
@@ -330,24 +332,42 @@
 (defmethod common-params/parameter->condition :granule-concept-id
   [context concept-type _param value options]
   (let [values (if (sequential? value) value [value])
-        {granule-concept-ids :granule
+        {granule-concept-ids    :granule
```
eereiter: extra spaces between -ids and    :granule

search-app/test/cmr/search/test/unit/services/query_execution.clj
```diff
@@ -3,116 +3,143 @@
...
   (:require
    [clojure.test :refer :all]
+   [cmr.common.hash-cache :as hc]
    [cmr.elastic-utils.search.es-params-converter :as pc]
    [cmr.elastic-utils.search.query-execution :as qe]
+   [cmr.metadata-db.data.concepts :as c]
    [cmr.search.services.query-execution :as search-qe]))
 
+(def test-context
+  {:system
```
eereiter, line 13: Why did you create this?  Were you getting a hashcache protocal for nil class error?  Or is there some other reason?  Or is this so you don't need to spin up an embedded redis server fixture like we do in a lot of other tests?

search-app/test/cmr/search/test/unit/services/query_execution.clj
```diff
@@ -3,116 +3,143 @@
...
-
-      ;; Facets require elastic strategy
-      {:concept-id "G1-PROV1" :result-format :echo10 :include-facets "true"} :elasticsearch
-
-      ;; All revisions uses elastic strategy
-      {:concept-id "C1-PROV1" :result-format :opendata :all-revisions "true"} :elasticsearch
-      {:concept-id "C1-PROV1" :result-format :echo10 :all-revisions "true"} :elasticsearch
+  ;; stub get-collection-concept-ids to always return ["C1-PROV1"]
+  (with-redefs [c/get-collection-concept-ids
+                (fn [_db _provider _granule-ids]
```
eereiter, line 45: I have tried now twice in a week and I could not get with-redefs to work for me - is that why you setup the test-context above?  I end up writing parameters that I can pass in the function so that I can mock out those functions.

search-app/test/cmr/search/test/unit/services/query_execution.clj
```diff
@@ -3,116 +3,143 @@
...
+        {:concept-id "C1-PROV1" :result-format :dif} :elasticsearch
+        {:concept-id "C1-PROV1" :result-format :dif10} :elasticsearch
+        {:concept-id "C1-PROV1" :result-format :iso19115} :elasticsearch
+        {:concept-id "C1-PROV1" :result-format :iso-smap} :elasticsearch))
 
-  (testing "Granule Specific"
-    (are [params expected-strategy]
-         (is (= expected-strategy (params->query-execution-strategy :granule params)))
+    (testing "Granule Specific"
+      (are [params expected-strategy]
```
eereiter, line 124: I think your editor changed the spacing so not asking you to change this.   Otherwise I would ask you to use are3.

eereiter (approved):

search-app/test/cmr/search/test/unit/services/query_execution.clj
ygliuvt: This is needed to make the provider hash work with the test. Since this is unit test, we don't have any real system component set up in the test.

search-app/test/cmr/search/test/unit/services/query_execution.clj
ygliuvt: This setup is to redefine get-collection-concept-ids to stub out the real db queries to retrieve the collection concept ids. It does not directly depend on the test-context which is needed for stub out the get provider from provider cache.

search-app/test/cmr/search/test/unit/services/query_execution.clj
ygliuvt: Right, I didn't change this part of the test. It is just format change by the editor auto-format.

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
```diff
@@ -644,6 +644,31 @@
                 []
                 concept-id-rev-ids-map))))
 
+(defn get-collection-concept-ids
```
chris-durbin, line 647: Missing docstring

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
```diff
@@ -644,6 +644,31 @@
...
+  [db provider granule-concept-ids]
+  (cond
+    (empty? granule-concept-ids)
+    []
+
+    :else
+    (let [table (tables/get-table-name provider :granule)
+          ;; Oracle caps IN clause at 1000, so chunk the IDs
+          ;; This is only theoretical, it is unlikely we will get close to even 100.
```
chris-durbin, line 656: Just FYI I've seen plenty of EDSC queries that will send 2000 granule concept-ids, so it's good that you're handling it here.

search-app/src/cmr/search/services/parameters/conversion.clj
```diff
@@ -332,22 +334,40 @@
...
-                                         (common-params/parameter->condition
-                                          context :collection :provider provider-ids {}))
-                                        (common-params/string-parameter->condition
-                                         concept-type :concept-id granule-concept-ids options)])))]
-    (if (and collection-cond granule-cond)
-      (gc/and-conds [collection-cond granule-cond])
-      (or collection-cond granule-cond))))
+
+        ;; get collection concept ids for granule-concept-ids from metadata-db
+        granules-by-provider (group-by (comp :provider-id cc/parse-concept-id)
```
chris-durbin: I was thinking we'd only need to do this if collection concept IDs could not be figured out. If a granule concept id is provided for a different collection the set intersection on line 351 is going to remove it.

So I think the logic should be:
```
if-not (seq collection-concept-ids)
...
Figure it out from granule concept-ids
```
That way any time collection concept ids are provided we don't have to query Oracle.

chris-durbin (changes requested):
Looks good - I just have one comment that we should only check Oracle if no collection concept IDs are provided.

search-app/src/cmr/search/services/parameters/conversion.clj
ygliuvt: I thought about it, but there is no guarantee that the client pass in all the collection concept ids or even correct ones for the granule concept ids in the search parameter. So we will have to do the search all the time to be fool proof.

ygliuvt: OK. I see your points now. So even though doing an intersection of the collection concept ids from parameters directly and from the ones derived from granule concept ids will make the searching indices more accurate. It is probably not worth it due to the unlikeliness of user passing in hundreds of collection_concept_ids. I am going to make that change to simplify the logic here. Thanks!

metadata-db-app/src/cmr/metadata_db/data/oracle/concepts.clj
ygliuvt: It's the CMR coding standard that we don't duplicate docstring for implementation functions for a protocol. The docstring is in the protocol function itself which is in `cmr.metadata-db.data.concepts` in this case.

PR #2317, opened 2025-10-08
CMR-10850: Searching for LARC granules without using a token currently results in an internal server error

TylerHeald1:
# Overview
LARC searches have been triggering a condition grouping logic error that presents as a 500 status.
### What is the objective?

Add logging to better diagnose the issue in any future occurrence and resolve root cause.

### What are the changes?

Logging has been updated to better diagnose future occurrences and a root cause that was discovered in SIT is now being handled.

Search and Access Control, though this issue seems to be largely during processing ACLs

# Required Checklist

# Additional Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Correctly handles groups that would previously produce no matches, yielding a clean “no results” outcome without extraneous conditions.

* **Improvements**
  * More precise grouping behavior to avoid spurious match logic and reduce unnecessary nesting.
  * More informative error messages when all conditions are filtered out, including the originating operation and initial inputs.

* **Tests**
  * Added unit tests covering AND/OR group composition for match-all, match-none, and mixed condition sets.

* **Refactor**
  * Internal cleanup of the condition-processing pipeline for clarity and maintainability.

### DIFF
--- a/elastic-utils-lib/src/cmr/elastic_utils/search/es_group_query_conditions.clj
+++ b/elastic-utils-lib/src/cmr/elastic_utils/search/es_group_query_conditions.clj
@@ -72,7 +72,7 @@
 
     (when (empty? processed-conditions)
       (errors/internal-error! (format "Logic error while grouping initial conditions [%s] with operation [%s]. No conditions found"
-                                      conditions, operation)))
+                                      conditions operation)))
 
     (if (= (count processed-conditions) 1)
       (first processed-conditions)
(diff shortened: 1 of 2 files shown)


elastic-utils-lib/src/cmr/elastic_utils/search/es_group_query_conditions.clj
chris-durbin, line 36: The problem is if this method returns empty results - we should make the change here - I think this should address it. 
```
(defmethod filter-group-conds :or
  [_operation conditions]
  (let [filtered (remove #(= % q/match-none) conditions)]
    (if (seq filtered)
      filtered
      [q/match-none])))
```

PR #2320, opened 2025-10-14
CMR-10930: Eliminate need for adding linter hints to any code using defn-timed, defn-timed-level or while-let macros

chris-durbin:
# Overview

### What is the objective?

I found many places in the source code where we were declaring and adding hints for the linter. I removed those declares and linter directives from the Clojure code and adding configuration to the linter to handle these three custom macros. 

### What are the changes?

Linter configuration allowing us to no longer add linter ignores in the source code files when using those three custom macros.

No changes so applications - only linter and code editors impacted.

# Required Checklist

# Additional Checklist

## Summary by CodeRabbit

- Chores
  - Removed many internal lint suppressions and forward declarations to improve code quality and tooling.
- Refactor
  - Simplified internal declarations and metadata; no functional or behavioral changes for users.
- Observability
  - Added an informational log when a bulk-update item status is not "UPDATED" to aid troubleshooting.
- Removal
  - Removed an internal ingest utility for deleting concepts; no changes to public APIs or user workflows.

### DIFF
--- a/.clj-kondo/config.edn
+++ b/.clj-kondo/config.edn
@@ -12,9 +12,9 @@
                 cache-in-mem {:linters {:unresolved-symbol {:exclude [cmr.common.cache.in-memory-cache]}}}}
  ;; CMR custom macros
  :lint-as {cmr.common.config/defconfig clojure.core/def
-           cmr.common.util/defn-timed clj-kondo.lint-as/def-catch-all
-           cmr.common.util/defn-timed-level clj-kondo.lint-as/def-catch-all
-           cmr.common.util/while-let clj-kondo.lint-as/def-catch-all
+           cmr.common.util/defn-timed clojure.core/defn
+           cmr.common.util/defn-timed-level clojure.core/defn
+           cmr.common.util/while-let clojure.core/loop
            cmr.common.jobs/defjob clojure.core/defn
            cmr.common.jobs/def-stateful-job clojure.core/defn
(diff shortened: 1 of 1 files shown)


.clj-kondo/config.edn
```diff
@@ -12,10 +12,13 @@
...
  ;; CMR custom macros
  :lint-as {cmr.common.config/defconfig clojure.core/def
+           cmr.common.util/defn-timed clojure.core/defn
+           cmr.common.util/defn-timed-level clojure.core/defn
+           cmr.common.util/while-let clojure.core/loop
            cmr.common.jobs/defjob clojure.core/defn
            cmr.common.jobs/def-stateful-job clojure.core/defn
            cmr.common.util/are3 clj-kondo.lint-as/def-catch-all
-           cmr.common.util/are2 clj-kondo.lint-as/def-catch-all} 
+           cmr.common.util/are2 clj-kondo.lint-as/def-catch-all}
```
jceaser, line 21: is there a reason to keep are2, I thought we were trying to get rid of them. Will adding this rule mean we will continue to ignore their existence? I don't feel very strong against this, just wondered where everyone else is on this.

chris-durbin: I didn't change anything for that line other than deleting a hidden trailing whitespace. I think we want to keep this linter rule until we've fully removed `are2` from the tests.

PR #2330, opened 2025-10-22
CMR-10941: Update db-migrate to update index-set with granule indexes

jmaeng72:
# Overview

### What is the objective?

To reduce downtime of CMR when transitioning to split cluster CMR logic

### What are the changes?

Update db-migrate to update index-set to get the old granule indexes list

All

# Required Checklist

# Additional Checklist

### DIFF
--- a/system-int-test/tests.group4.edn
+++ b/system-int-test/tests.group4.edn
@@ -1,5 +1,3 @@
 ["test/cmr/system_int_test/ingest/collection"
  "test/cmr/system_int_test/ingest/granule"
- "test/cmr/system_int_test/ingest/misc"
- "test/cmr/system_int_test/search/collection"
- "test/cmr/system_int_test/search/granule"]
+ "test/cmr/system_int_test/ingest/misc"]
--- a/system-int-test/tests.group5.edn
+++ b/system-int-test/tests.group5.edn
@@ -1,0 +1,2 @@
+["test/cmr/system_int_test/search/collection"
+ "test/cmr/system_int_test/search/granule"]


indexer-app/src/cmr/indexer/data/elasticsearch.clj
```diff
@@ -208,7 +210,17 @@
       (do
         (info "Ignoring update gran indexes request because gran index set is unchanged.")
         (info "Existing gran index set:" (pr-str existing-gran-index-set))
-        (info "New gran index set:" (pr-str expected-gran-index-set))))))
+        (info "New gran index set:" (pr-str expected-gran-index-set))))
+
+    ;; this is during our transition
+    (if (nil? existing-gran-index-set)
+      (let [_ (info "Creating gran index set for first time, using the old index-sets value if available.")
+            old-index-set (cmr.indexer.data.index-set-elasticsearch/get-old-index-set context es-config/gran-elastic-name idx-set/index-set-id)]
```
eereiter: change cmr.indexer.data.index-set-elasticsearch to index-set-es

indexer-app/src/cmr/indexer/data/elasticsearch.clj
```diff
@@ -208,7 +210,17 @@
       (do
         (info "Ignoring update gran indexes request because gran index set is unchanged.")
         (info "Existing gran index set:" (pr-str existing-gran-index-set))
-        (info "New gran index set:" (pr-str expected-gran-index-set))))))
+        (info "New gran index set:" (pr-str expected-gran-index-set))))
+
+    ;; this is during our transition
+    (if (nil? existing-gran-index-set)
```
eereiter: don't use if here use while because you don't have a statement for false.

indexer-app/src/cmr/indexer/data/elasticsearch.clj
```diff
@@ -208,7 +210,17 @@
...
+      (let [_ (info "Creating gran index set for first time, using the old index-sets value if available.")
```
eereiter: Move the info up one line out of the let. You can do this since we are changing the if to a while.

indexer-app/src/cmr/indexer/data/elasticsearch.clj
```diff
@@ -208,7 +210,17 @@
...
+            old-index-set (cmr.indexer.data.index-set-elasticsearch/get-old-index-set context es-config/gran-elastic-name idx-set/index-set-id)]
+        ;; put that value into the new index sets
+        (if (not (nil? old-index-set))
+          (do
```
eereiter: Replace (if (not with while-not  because a false condition doesn't exist.  Then you can also remove the do.

linter should have found this.

indexer-app/src/cmr/indexer/data/elasticsearch.clj
```diff
@@ -208,7 +210,17 @@
...
+            (info "old index-sets was found. Creating gran index-sets index with old index-sets value.")
+            (cmr.indexer.services.index-set-service/put-index-set context old-index-set)))))))
```
eereiter: replace cmr.indexer.services.index-set-service with index-set-es

indexer-app/src/cmr/indexer/data/index_set_elasticsearch.clj
```diff
@@ -115,6 +115,20 @@
           (get-in [:_source :index-set-request])
           decode-field))))
 
+(defn get-old-index-set
+  "Fetch index-sets that existing in the pre-split cluster archtecture.
+  This func should only be called one the first time we transition to the split cluster.
```
eereiter: replace one with once,

indexer-app/src/cmr/indexer/data/index_set_elasticsearch.clj
```diff
@@ -115,6 +115,20 @@
...
+  It may be deleted after the transition is determined to be successful and permanent."
```
eereiter: Can we write a ticket to delete this function and anything that calls it and put the ticket number in the comments?

indexer-app/src/cmr/indexer/services/index_set_service.clj
```diff
@@ -280,6 +281,18 @@
 
     (index-requested-index-set context index-set es-cluster-name)))
 
+(defn put-index-set
+  "Upsert the given index-set to the index-sets index in ES."
```
eereiter, line 284: I had to look up Upsert, didn't know that was an actual word.

indexer-app/src/cmr/indexer/services/index_set_service.clj
```diff
@@ -280,6 +281,18 @@
...
+  [context index-set]
+  (let [split-index-set-map (split-index-set-by-cluster index-set)]
+    (acl/verify-ingest-management-permission context :update)
```
eereiter: Just curious as to why you moved the acl check in this function instead of leaving it in the routes namespace?  The other routes that have the check, check the acl in the routes namespace. The rest is a good move.

indexer-app/src/cmr/indexer/data/elasticsearch.clj
jceaser: yea, this is one of those cases where I disagree with the language. They should not allow this to compile.

jmaeng72: I think you mean use 'when'? And yes I can do that

indexer-app/src/cmr/indexer/services/index_set_service.clj
jmaeng72: whoops. Yes it would be better int he routes.clj file. Moving

indexer-app/src/cmr/indexer/data/index_set_elasticsearch.clj
jmaeng72: Yep, created and added to the deploy epic [CMR-10949]

indexer-app/src/cmr/indexer/data/elasticsearch.clj
jmaeng72: I must not be linting properly... will ask others how they are linting to double check my process

indexer-app/src/cmr/indexer/data/elasticsearch.clj
```diff
@@ -208,7 +210,16 @@
       (do
         (info "Ignoring update gran indexes request because gran index set is unchanged.")
         (info "Existing gran index set:" (pr-str existing-gran-index-set))
-        (info "New gran index set:" (pr-str expected-gran-index-set))))))
+        (info "New gran index set:" (pr-str expected-gran-index-set))))
+
+    ;; this is during our transition
+    (when (nil? existing-gran-index-set)
```
jceaser, line 216: does (when (not also work here? Question not critique.

jceaser (approved):

indexer-app/src/cmr/indexer/data/elasticsearch.clj
jmaeng72: I think it would. Although the two methods are not exactly the same, in this scenario, it would work the same. Although I imagine the nil? is more clear of what I expect the value to be (rather than false)

PR #2336, opened 2025-10-24
CMR-10948: Fix rebalancing of resharded individual collection granule index.

ygliuvt:
# Overview

### What is the objective?

Fix rebalancing of resharded individual collection granule index.

### What are the changes?

This PR figures out the correct granule index based on the index-set rather than simple naming conventions during rebalancing to fix the Internal Server Error on baseline.

CMR indexer 

### Test 

The `reshard-individual-granule-index-with-rebalancing-test` integration test is commented out because rebalance collections uses delete-by-query which cannot be force refreshed. As a result, after the granules are moved from small_collections index to separate index, the granule deleted version is still in the small_collections index waiting to be permanently deleted. At this time, if we move back to small collections index immediately as in this test, the rebalancing would fail because of version conflict. It is hard to set up the perfect condition in test.

We can test it locally by enable the test and comment out the part in test that doing the rebalancing back to small collections starting at line 184 and run with logging level at debug. Then run the following test locally and verify that the rebalancing completes successfully and searching for granules on the collection id runs against the small_collections index and found 1 granule.

curl -i -XPOST -H "Authorization: mock-echo-system-token" "http://localhost:3006/rebalancing_collections/C1200000009-PROV1/start?target=small-collections"

curl -i -H "Authorization: mock-echo-system-token" "http://localhost:3006/rebalancing_collections/C1200000009-PROV1/status"

curl -i -XPOST -H "Authorization: mock-echo-system-token" "http://localhost:3006/rebalancing_collections/C1200000009-PROV1/finalize"

curl -i -H "Authorization: mock-echo-system-token" "http://localhost:3003/granules?collection_concept_id=C1200000009-PROV1"

# Required Checklist

# Additional Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved granule index validation and removal by standardizing concept-id handling (supports dash- and underscore-delimited and already-canonical ids).

* **Tests**
  * Added tests covering additional index-name/concept-id edge cases.
  * Disabled an unstable integration test to improve suite reliability.

### DIFF
--- a/indexer-app/src/cmr/indexer/services/index_set_service.clj
+++ b/indexer-app/src/cmr/indexer/services/index_set_service.clj
@@ -262,26 +262,28 @@
           (vals (:index-set index-set))))
 
 (defn index-name->concept-id
-  "Extracts a concept ID from a granule index name. Does not work for non-granule indexes.
+  "Extracts a concept ID from a granule index name. Works for both underscore- and dash-delimited names.
    Examples:
    '1_c2317035855_nsidc_ecs' -> 'C2317035855-NSIDC_ECS'
-   '1_c2317035855_nsidc_ecs_5_shards' -> 'C2317035855-NSIDC_ECS'"
+   '1_c2317035855_nsidc_ecs_5_shards' -> 'C2317035855-NSIDC_ECS'
+   'C2317035855-NSIDC_ECS' -> 'C2317035855-NSIDC_ECS'"
   [index-name]
   (when index-name
     (let [;; Remove shard count suffix if present
           without-shards (string/replace index-name #"_\d+_shards$" "")
           ;; Remove leading number prefix (e.g., "1_")
-          without-prefix (string/replace without-shards #"^\d+_" "")
-          ;; Find the first underscore after the concept type+number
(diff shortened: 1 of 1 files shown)


system-int-test/test/cmr/system_int_test/bootstrap/reshard_index_test.clj
```diff
@@ -131,3 +132,64 @@
      ;; After the cache is cleared the right amount of data is found
      (search/clear-caches)
      (bootstrap/verify-provider-holdings expected-provider-holdings "After finalize after clear cache"))))
+
+;; Rebalance collections uses delete-by-query which cannot be force refreshed.
```
daniel-zamora, line 136: Can we mark this test with ticket numbers, both for when this test was commented out and a ticket for refactoring the test.  Also, is this comment reflected in comments elsewhere in the relevant code, or do we only explain this interaction here?

ygliuvt: We have similar comments on the existing commented-out rebalancing test `rebalance-collection-back-to-small-collections-test`. I think we have decided that we weren't going to file a ticket to address the test back then because of the effort level and importance.

PR #2366, opened 2025-12-23
CMR-11037: Fix middleware for body-copy to work properly with form data

zimzoom:
# Overview

### What is the objective?

Original reported bug noted that collection facets search was returning an inaccurate field; further investigation determined the root cause was that this middleware function for attaching "body-copy" when processing form encoded data was not working correctly, and actually just returning an empty string. This middleware has been fixed in this PR to correctly parse the body copy in the case of form encoded requests while still maintaining the original slurp functionality otherwise.

### What are the changes?

This middleware has been fixed in this PR to correctly parse the body copy in the case of form encoded requests while still maintaining the original slurp functionality otherwise.

search-app

# Required Checklist

- [ x] New and existing unit and int tests pass locally and remotely
- [ x] clj-kondo has been run locally and all errors in changed files are corrected
- [ x] I have commented my code, particularly in hard-to-understand areas
- [ x] I have made changes to the documentation (if necessary)
- [ x] My changes generate no new warnings

# Additional Checklist
- [ x] I have removed unnecessary/dead code and imports in files I have changed
- [ x] I have cleaned up integration tests by doing one or more of the following:

## Summary by CodeRabbit

* **New Features**
  * Added multipart/form-data parsing so multipart POSTs are parsed and reconstructed for reliable processing.

* **Bug Fixes**
  * Improved request-body preservation so POST form bodies can be read multiple times (body now preserved and replayable).

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/search-app/src/cmr/search/routes.clj
+++ b/search-app/src/cmr/search/routes.clj
@@ -16,7 +16,7 @@
    [cmr.common.config :refer [defconfig]]
    [cmr.common.mime-types :as mt]
    [cmr.common.services.errors :as svc-errors]
-   [cmr.search.api.routes :as api-routes] 
+   [cmr.search.api.routes :as api-routes]
    [cmr.search.data.granule-counts-cache :as granule-counts-cache]
    [cmr.search.middleware.shapefile :as shapefile]
    [cmr.search.middleware.shapefile-simplification :as shapefile-simplifier]
@@ -24,6 +24,7 @@
    [cmr.search.site.routes :as site-routes]
    [compojure.core :refer [GET POST context routes]]
    [ring.middleware.keyword-params :as keyword-params]
+   [ring.middleware.multipart-params :as multipart]
    [ring.middleware.nested-params :as nested-params]
    [ring.middleware.params :as params]))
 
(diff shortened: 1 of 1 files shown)


search-app/src/cmr/search/routes.clj
```diff
@@ -55,13 +56,21 @@
        [(msg/mixed-arity-parameter-msg mixed-param)]))
     (handler request)))
 
+(defn parse-multipart-body [request]
+  (let [parsed (multipart/multipart-params-request request)]
+    (string/join "&" (for [[k v] (:multipart-params parsed)]
```
jceaser, line 61: did ring not have anything for you to do this? None of these would work: https://ring-clojure.github.io/ring/ring.middleware.multipart-params.html#var-parse-multipart-params

jmaeng72 (approved):

search-app/src/cmr/search/routes.clj
zimzoom: Hmm ok, the one you suggested did not work, but there is another one that might. Double checking

zimzoom: Update: the other method did not work either. This is how it has to be.

jceaser (approved):

PR #2383, opened 2026-02-11
Reimplement spatial-lib as a Java library

jpolchlo:
# Overview

### What is the objective?

In order to move to Elasticsearch 8, we have to contend with a much stricter security model for plugins—especially for ES 8.18 and up, which move to the Java Entitlements system (the old Java security manager is disabled in ES8.18, and for all ES8 incarnations, it seems, the security is much tighter with regard to class loaders and privileges).  I've struggled with this, and despite my best efforts to wrap as much access to clojure in Java at the plugin level, I could not overcome class loader difficulties.  There might be a way, but I didn't find it.

What I decided to do was try out a pure Java conversion for the spatial plugin.  That isn't possible without a pure Java spatial-lib.  So this is that second part.

Let me be clear that this was implemented by Copilot (with Claude sonnet 4.5 as the acting model), and it's a **massive** change.  Quite risky as far as these things go, since we're relying on the test suite in spatial-lib to really characterize the correctness of the module.  I guided the AI to the best of my ability to stick closely to the Clojure implementation.  I fear there's a good amount of content here that is overwrought, and I have definitely not laid eyes on every line of the solution.

I definitely don't expect that this PR will ever be merged.  My interest was to make sure you could pull this into your tree, because ES7 will eventually go out of support, and maybe you'll need to do what I think needs to be done for the ES plugin to work for you again.

Note: unit tests were the sole arbiter of success.  I have not run the integration tests, which might show impacts.

### What are the changes?

The business logic of spatial-lib has been pulled out into a separate set of Java classes, and the Clojure interface delegates to it to keep compatibility downstream.

Direct effects are on umm-lib, umm-spec-lib, search-app, es-spatial-plugin.  Indirect effects fall out of those.

# Required Checklist

- [x] Claude has commented its code, particularly in hard-to-understand areas

# Additional Checklist

## Summary by CodeRabbit

* **New Features**
  * Added circle geometry and a Java-backed shape intersection API for richer spatial queries.

* **Bug Fixes**
  * Improved correctness for pole cases and antimeridian (date-line) crossings.

* **Refactor**
  * Core spatial predicates and containment checks now delegate to new Java geometry implementations for robustness and performance.

* **Chores**
  * Project configured to include Java sources.

* **Tests**
  * New comprehensive Java and Clojure test suites covering intersections and deserialization.


spatial-lib/src/java/cmr/spatial/geometry/CircleIntersections.java
```diff
@@ -0,0 +1,222 @@
...
+     * @return Sanitized longitude
+     */
+    private static double sanitizeLongitude(double lon) {
+        if (lon > 180.0) {
+            return lon - 360.0;
+        } else if (lon < -180.0) {
+            return lon + 360.0;
+        }
+        return lon;
+    }
```
jaybarra, line 221: This is not guaranteed to return values between -180 and 180,  in the case of values accidentally being greater or less than +- 180 originally. Unlikely but possible

PR #2410, opened 2026-04-13
CMR-11249: Index UUID for Reverse Lookup in Collections on the keyword2 index (projects)

eudoroolivares2016:
# Overview

### What is the objective?

Add search capability to search for collections by the `uuid` keyword under `projects` this is added to the free-text search interface via `keyword2` where it will be appended

### What are the changes?

Adds the `project-uuid` into the collection index holding it as `kms-uuids` which then adds those unique `uuids` to the `keyword2` index and makes it available for search via the  GET request API

Adds some supporting tests for the indexes and the new cache that is being added. The new cache will hold the keyword `shortname` as a key and the `uuid` as a value

I am also changing one integration test so that each test can be run individually instead of being under a single `deftest` 
I am using an `atom` to avoid having to reset fixtures that aren't changing in this search test and thus allowing us to simply `destructure`  the atom and match assertions against it this is from a previous change where the 

CMR Search using free-text search GET API; CMR indexing for collection records

Here is [build](https://ci.earthdata.nasa.gov/browse/CN2-CSN2947-3)

image

# Required Checklist

- [NA] I have made changes to the documentation (if necessary)

# Additional Checklist

## Summary by CodeRabbit

* **New Features**
  * Project-based collection search now supports hierarchical project fields (short name, long name, UUID) with pattern and ignore-case options; index now includes project identifiers.

* **Bug Fixes**
  * Fixed minor documentation typos.

* **Documentation**
  * README formatting cleaned and examples clarified.

* **Tests**
  * Collection campaign tests split into focused scenarios; new project keyword tests added.

* **Data Updates**
  * KMS keyword catalogs refreshed (projects, formats, locations, measurements, mimetypes, etc.)

* **Chores**
  * KMS example tooling: configurable output directory and promote option added.


indexer-app/test/cmr/indexer/test/data/concepts/collection/resolution.clj
```diff
@@ -78,7 +78,7 @@
       :Unit "Kilometers"}]))
 
 (deftest get-horizontal-data-resolutions-test
-  "Test the extraction and conversion of horizontal data resolutions."
+  ;; Test the extraction and conversion of horizontal data resolutions.
```
eudoroolivares2016: The linter was complaining about these thats why I am updating it here maybe they should just be `(testing )` blocks actually

indexer-app/src/cmr/indexer/data/concepts/collection.clj
```diff
@@ -281,10 +282,17 @@
...
+        projects-nested (map #(project/project-short-name->elastic-doc context %)
+                             (->> (map :ShortName (:Projects collection))
+                                  (map string/trim)))
+        project-short-names (map :short-name projects-nested)
+        project-long-names (->> projects-nested
+                                (keep :long-name)
+                                distinct
+                                (map string/trim)
+                                (remove string/blank?))
+        project-uuids (keep :uuid projects-nested)
```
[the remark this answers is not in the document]
eudoroolivares2016: Huh? That is how instruments and platforms work

zimzoom:
Looks good except we can't introduce the breaking change to the JSON API, since we are not currently versioning it

common-app-lib/src/cmr/common_app/services/kms_lookup.clj
```diff
@@ -148,7 +148,7 @@
 
 (def keywords-to-lookup-by-short-name
   "Set of KMS keywords that we need to be able to lookup by short name."
-  #{:providers :platforms :instruments})
+  #{:providers :platforms :instruments :projects})
```
eereiter: I have to look further into this but I don't think this will work - Platforms and Projects sometimes have the same short name. So there may be a clash here where we will pick up the wrong value.  I have to look further into it.  Example Platform and Project short name Grace.

indexer-app/test/cmr/indexer/test/data/concepts/collection/project_keyword.clj
```diff
@@ -0,0 +1,52 @@
...
+    (f)))
+
+(use-fixtures :once (join-fixtures [redis-embedded-fixture/embedded-redis-server-fixture
+                                    redis-cache-fixture]))
+
+(deftest project-short-name->elastic-doc-test
+  (let [context create-context]
+    (testing "Project found in KMS"
+      (is (= {:bucket nil,
+              :bucket-lowercase nil,
```
eereiter: Why is bucket nil?  Why do you have bucket?

eereiter: I understand what bucket is: It is the alphabetical grouping by kms: example G - I.

search-app/src/cmr/search/services/json_parameters/conversion.clj
```diff
@@ -216,14 +216,17 @@
...
   [concept-type condition-name value]
-  (validate-nested-condition condition-name value)
-  (let [elastic-field-name (-> condition-name
-                               (inf/plural)
-                               (q2e/query-field->elastic-field concept-type))]
-    (nf/parse-nested-condition elastic-field-name
-                               value
-                               (case-sensitive-field? concept-type condition-name value)
-                               (:pattern value))))
+  (let [value (if (and (= condition-name :project) (string? value))
```
eereiter: Why are you changing the condition.  This code doesn't seem correct.

indexer-app/test/cmr/indexer/test/data/concepts/collection/project_keyword.clj
eereiter: But why are we indexing it?  The CMR doesn't care about the buckets - users don't want to search on it and we are not returning it.

eereiter: I know this is the testing code, but I am questioning the need for the full KMS structure.

eereiter (changes requested):
I need to review this a little more - but the code doesn't seem quite right.  I will add specifics in a little bit.

search-app/src/cmr/search/services/json_parameters/conversion.clj
eudoroolivares2016: This is so its reverse compatible with the existing API that can only search by shortname like this one `{
    "condition": {
        "or": [{"project": "ESI"}, {"project": "EVI"}]
    }
}` as an example above. single works of course I added a comment to clarify though

indexer-app/test/cmr/indexer/test/data/concepts/collection/project_keyword.clj
eudoroolivares2016: Yea I can take the bucket out of the index. I honestly don't think CMR even needs to know anything about the buckets at all from what I can tell but, it was just in the KMS record we pull down and gets passed along

common-app-lib/src/cmr/common_app/services/kms_lookup.clj
eudoroolivares2016: This was an excellent callout. I did test that condition and it does look like it works because this is adding a separete index to the cache even if they have the same name its actually getting indexed separately into elastic 
image

jmaeng72:
I'm going to request that you go over this PR with the CMR team along with a demo to fully understand the changes made here. Should be like a 10 min overview.

indexer-app/src/cmr/indexer/data/concepts/collection/project_keyword.clj
```diff
@@ -0,0 +1,31 @@
...
+          (repeat kf/FIELD_NOT_PRESENT)))
+
+(defn project-short-name->elastic-doc
+  "Converts an project short-name into an elastic document with the full nested hierarchy for
+  that short-name from the GCMD KMS keywords. If a field is not present in the KMS hierarchy, we
+  use a dummy value to indicate the field was not present."
+  [context short-name]
+  (let [full-project
+        (merge default-project-keyword-values
+               (kms-lookup/lookup-by-short-name context :projects short-name))
```
eereiter: I think we need to break up projects into its own cache where the short name is the key.  Using what we have now will only slow the entire system down because we will download from redis the entire projects list for every single call.  We need to break up the others too so make these calls faster.

indexer-app/src/cmr/indexer/data/concepts/collection/project_keyword.clj
```diff
@@ -0,0 +1,31 @@
...
+        {:keys [bucket short-name long-name uuid]
+         ;; Use the short-name from KMS if present, otherwise use the metadata short-name
+         :or {short-name short-name}} full-project]
+    {:bucket bucket
+     :bucket-lowercase (when bucket (string/lower-case bucket))
```
eereiter: The CMR doesn't make use of the bucket so :bucket and "bucket-lowercase should be deleted.

search-app/resources/schema/CollectionJSONQueryLanguage.json
```diff
@@ -177,8 +177,43 @@
...
+					"description": "Search against any of the subfields in the project hierarchy",
+					"oneOf": [{
+						"type": "string",
+						"$ref": "#/definitions/StringCondition"
+					}, {
+						"type": "object",
+						"additionalProperties": false,
+						"minProperties": 1,
+						"properties": {
+							"bucket": {
```
eereiter: Bucket should be deleted, because the CMR API doesn't have a use case to search by bucket.

indexer-app/src/cmr/indexer/data/concepts/collection.clj
```diff
@@ -281,10 +282,18 @@
...
+        projects-nested (map #(project/project-short-name->elastic-doc context %)
+                             (->> (map :ShortName (:Projects collection))
+                                  (map string/trim)))
+        project-short-names (map :short-name projects-nested)
+        project-long-names (->> projects-nested
+                                (concat (map util/map-keys->kebab-case (:Projects collection)))
+                                (keep :long-name)
+                                distinct
+                                (map string/trim)
+                                (remove string/blank?))
```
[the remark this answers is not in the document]
eereiter: Yeah, this is a breaking change.  The project-short-names function should still be defined as 

(->> (map :ShortName (:Projects collection))
                                 (map string/trim))

Otherwise if the value does not exist in KMS, the original UMM-C value will not be indexed and then can't be found.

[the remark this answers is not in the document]
eereiter: The point of all of these changes is to find which collections are impacted by a GCMD keyword change.  So we want to be able to find all collections where a specific UUID exists.  Then why do we care about all of the other project fields?  Its a lot of redundant information that doesn't need to exist in ES if we don't want to search for it and if it isn't stored to be given back to the user.  We can already search by projects short name.  Also project-uuid also already exists in the index.  So not sure that we need any of these changes.

indexer-app/test/cmr/indexer/test/data/concepts/collection/resolution.clj
eereiter: Yes, having the description after deftest isn't valid.  that is why the linter is flagging them.

search-app/src/cmr/search/services/json_parameters/conversion.clj
eereiter: Ah, I see because now project can be either a string or a nested condition, even though on line 69 it was changed from a string to a nested condition.

eereiter: project-uuid already exists at the main level. I don't think we need the nested condition. I think just changing the CollectionJSONQueryLanguage.json to include using project-uuid would be enough the changes of finding collections that match a GCMD keyword. I would even say add a new field called GCMD-UUID or something like that as an array of uuids for all of the keywords that you want to find the collections. That is really all you need to find them. Or we can also just put them into the keyword2 field. That should work too.

indexer-app/src/cmr/indexer/data/concepts/collection.clj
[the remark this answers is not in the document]
eudoroolivares2016: We can do it that way. The instruments and platforms are indexing both into this nested field so I'm not sure if there was a use-case for that previously I can if you want rewrite to just put only `uuid` in there and thus perhaps not need the nested-fields

indexer-app/src/cmr/indexer/data/concepts/collection/project_keyword.clj
eudoroolivares2016: This does add those projects to the `:kms-short-name-index` cache that already exists. The other fields in there are `platforms`, `providers` (which really maps to data-centers`, and `instruments` do you mean we'd make each of those its own `keywords-to-lookup-by-short-name`

common-app-lib/src/cmr/common_app/services/kms_lookup.clj
```diff
@@ -237,6 +264,21 @@
...
 
+(defn lookup-project-by-short-name
+  "Takes a kms-index and a short name and returns the UUID for
+  that short name. Returns nil if a keyword is not found. Comparison is made case insensitively."
+  [context short-name]
+  (try
+    (when-not (:ignore-kms-keywords context)
+      (let [project-cache (hash-cache/context->cache context kms-projects-cache-key)
+            [tm uuid] (util/time-execution (hash-cache/get-value project-cache kms-projects-cache-key (util/safe-lowercase short-name)))
+            _ (rl-util/log-redis-read-complete "lookup-project-by-short-name" kms-projects-cache-key tm)]
```
jceaser: it's the last statement of the let and it drops the return, you can just call that in the body

indexer-app/src/cmr/indexer/data/concepts/collection/project_keyword.clj
```diff
@@ -0,0 +1,12 @@
...
+  "Contains functions for converting project hierarchies into elastic documents"
+  (:require
+   [cmr.common-app.services.kms-lookup :as kms-lookup]))
+
+(defn project-short-name->elastic-doc
+  "Converts a project short-name into an elastic document with the uuid
+  for that short-name from the GCMD KMS keywords."
+  [context short-name]
+  (let [uuid (kms-lookup/lookup-project-by-short-name context short-name)]
+    (when uuid
```
jceaser, line 11: what is returned if no uuid? nil I think, is that expected?

jceaser (approved):
I had some questions, but nothing major

indexer-app/src/cmr/indexer/data/concepts/collection/project_keyword.clj
eudoroolivares2016: Yea it does return nil. I don't expect that is possible but, I tried locally deleting that `uuid` field from redis and reindexing just causes that not the be included since its missing. I can add a quick unit test into this module if you think that'd help

indexer-app/src/cmr/indexer/data/concepts/collection/project_keyword.clj
eudoroolivares2016: Done

search-app/src/cmr/search/services/json_parameters/conversion.clj
eudoroolivares2016: I did remove all this in favor of the other solution to using `keyword2`

common-app-lib/src/cmr/common_app/services/kms_lookup.clj
eudoroolivares2016: It is now its own cache see PR description for screenshot from redis insight

indexer-app/src/cmr/indexer/data/concepts/collection.clj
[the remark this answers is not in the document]
eudoroolivares2016: I did end up just removing the first way that we were doing this via using the JSON endpoint and thus we didn't need either the nested fields or any "new" indexes just appending the `uuids` to `keyword2`

indexer-app/test/cmr/indexer/test/data/concepts/keyword_util.clj
```diff
@@ -235,95 +235,6 @@
       :AverageFileSizeUnit nil
       :Media nil}]}})
 
-(def long-field-collection
```
eudoroolivares2016, line 238: this just wasn't being used

common-app-lib/src/cmr/common_app/services/kms_lookup.clj
```diff
@@ -36,6 +36,11 @@
   in the system hash cache map for fast lookups."
   :kms-short-name-index)
 
+(def kms-projects-cache-key
+  "The key used to store the data generated from KMS into a locations index cache
```
eereiter: locations?  should the word be projects?

common-app-lib/src/cmr/common_app/services/kms_lookup.clj
```diff
@@ -47,21 +52,30 @@
...
-  "The key used to store the data generated from KMS into a measurment index cache
+  "The key used to store the data generated from KMS into a measurement index cache
   in the system hash cache map for fast lookups."
   :kms-measurement-index)
 
 (def kms-cache-ttl
   "Time To Live value for KMS caches. nil means never expire."
   nil)
 
+;; TODO we will breakup this cache into composite ones and only index
```
eereiter, line 63: Isn't this TODO comment already done on the new line 72?  If so we need to delete this comment

indexer-app/test/cmr/indexer/test/data/concepts/collection/resolution.clj
```diff
@@ -6,7 +6,7 @@
     [cmr.indexer.data.concepts.collection.resolution :as res]))
 
 (deftest get-non-range-data-resolution-test
-  "Test the extraction of non range data resolutions."
+  (testing "Test the extraction of non range data resolutions.")
```
eereiter, line 9: Thank you!

indexer-app/test/cmr/indexer/test/data/concepts/keyword_util.clj
eereiter: Thanks for spotting this and removing it.

eereiter (approved):

zimzoom (approved):

common-app-lib/src/cmr/common_app/services/kms_lookup.clj
eudoroolivares2016: I think we'll do more of it in https://bugs.earthdata.nasa.gov/browse/CMR-11236 or I'll delete the comment then

PR #2415, opened 2026-04-24
CMR-10891: Add wildcard type fields support for search behind feature flag

zimzoom:
# Overview

### What is the objective?

Adds support for using Elasticsearch wildcard-type fields for granule wildcard searches, gated behind a feature flag so the behavior can be enabled only after the relevant indices are reindexed.

### What are the changes?

* Add feature flag defconfig `enable-wildcard-field-searches` -- when enabled, granule pattern searches now target dedicated wildcard fields for granule-ur and producer-granule-id, with separate case-sensitive and lowercase mappings. When the flag is disabled, wildcard searches continue using the existing standard fields, preserving prior behavior. Exact non-pattern searches are unchanged.

* Adds wildcard-field mapping helpers in elastic-utils-lib, granule-specific wildcard mappings in search-app, and unit coverage for enabled and disabled toggle behavior, including readable granule name handling.

search-app, elastic-utils-lib

# Required Checklist

- [x ] New and existing unit and int tests pass locally and remotely
- [x ] clj-kondo has been run locally and all errors in changed files are corrected

# Additional Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a configurable option (disabled by default) to use dedicated wildcard fields for pattern searches, improving wildcard-search behavior.
  * Granule searches can now target these wildcard-specific field variants when the option is enabled.

* **Tests**
  * Added unit tests verifying correct field selection for pattern vs exact granule searches and toggle behavior.

### DIFF
--- a/elastic-utils-lib/src/cmr/elastic_utils/search/es_query_to_elastic.clj
+++ b/elastic-utils-lib/src/cmr/elastic_utils/search/es_query_to_elastic.clj
@@ -57,8 +57,8 @@
   [query-str]
   (fix-double-escapes
    (string/replace query-str
-                query-string-reserved-characters-regex
-                "\\\\$1")))
+                   query-string-reserved-characters-regex
+                   "\\\\$1")))
 
 (defprotocol ConditionToElastic
   "Defines a function to map from a query to elastic search query"
@@ -145,6 +145,40 @@
   (get (field->lowercase-field-mappings concept-type) (keyword field)
        (str (name field) "-lowercase")))
 
+(defmulti field->wildcard-field-mappings
+  "Mapping of query model field names to Elasticsearch wildcard field names."
+  (fn [concept-type]
+    concept-type))
+
+(defmethod field->wildcard-field-mappings :default
+  [_]
+  ;; No field mappings specified by default
+  {})
+
+(defn field->wildcard-field
+  "Maps a field name to the name of the wildcard field to use within elastic.
+  If a mapping is not present it defaults to the regular elastic field."
+  [concept-type field]
+  (get (field->wildcard-field-mappings concept-type) (keyword field)
+       (query-field->elastic-field field concept-type)))
+
+(defmulti field->lowercase-wildcard-field-mappings
+  "Mapping of query model field names to Elasticsearch lowercase wildcard field names."
+  (fn [concept-type]
+    concept-type))
+
+(defmethod field->lowercase-wildcard-field-mappings :default
+  [_]
+  ;; No field mappings specified by default
+  {})
+
+(defn field->lowercase-wildcard-field
+  "Maps a field name to the name of the lowercase wildcard field to use within elastic.
+  If a mapping is not present it defaults to the lowercase field for the field."
+  [concept-type field]
+  (get (field->lowercase-wildcard-field-mappings concept-type) (keyword field)
+       (field->lowercase-field concept-type field)))
+
 (defn- range-condition->elastic
   "Convert a range condition to an elastic search condition. Execution
   should either by 'fielddata' or 'index'."
@@ -206,9 +240,15 @@
   cmr.common.services.search.query_model.StringCondition
   (condition->elastic
     [{:keys [field value case-sensitive? pattern?]} concept-type]
-    (let [field (if case-sensitive?
-                  (query-field->elastic-field field concept-type)
-                  (field->lowercase-field concept-type field))
+    (let [field (if (and pattern? (config/enable-wildcard-field-searches))
+                  ;; New behavior: use dedicated wildcard fields for pattern searches
+                  (if case-sensitive?
+                    (field->wildcard-field concept-type field)
+                    (field->lowercase-wildcard-field concept-type field))
+                  ;; Original behavior: use standard fields
+                  (if case-sensitive?
+                    (query-field->elastic-field field concept-type)
+                    (field->lowercase-field concept-type field)))
           value (if case-sensitive? value (string/lower-case value))]
       (if pattern?
         {:wildcard {field value}}
--- a/search-app/src/cmr/search/data/query_to_elastic.clj
+++ b/search-app/src/cmr/search/data/query_to_elastic.clj
@@ -306,6 +306,19 @@
       (merge default-mappings query-field->lowercase-granule-doc-values-fields-map)
       default-mappings)))
 
+(defmethod q2e/field->wildcard-field-mappings :granule
+  [_]
+  {:granule-ur :granule-ur-wildcard
+   ;; Both aliases below are needed because pre-existing granule queries use both names.
+   :producer-gran-id :producer-granule-id-wildcard
+   :producer-granule-id :producer-granule-id-wildcard})
+
+(defmethod q2e/field->lowercase-wildcard-field-mappings :granule
+  [_]
+  {:granule-ur "granule-ur-lowercase-wildcard"
+   :producer-gran-id "producer-granule-id-lowercase-wildcard"
+   :producer-granule-id "producer-granule-id-lowercase-wildcard"})
+
 (defn- keywords-in-query
   "Returns a list of keywords if the query contains a keyword condition or nil if not.
   Used to set sort and use function score for keyword queries."
@@ -396,7 +409,7 @@
                                       (string/replace #"^\"" "* ")
                                       (string/replace #"\"$" " *")
                                       (string/replace #"\\\"" "\""))
-                       :field :keyword-phrase)
+             :field :keyword-phrase)
       (if (and query-str
                (string/includes? trimmed-query-str "\\\""))
         (assoc condition :query-str (string/replace trimmed-query-str #"\\\"" "\""))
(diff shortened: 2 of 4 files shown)


elastic-utils-lib/src/cmr/elastic_utils/search/es_query_to_elastic.clj
```diff
@@ -145,6 +146,40 @@
...
        (str (name field) "-lowercase")))
 
+(defmulti field->wildcard-field-mappings
+  "Mapping of query model field names to Elasticsearch wildcard field names."
+  (fn [concept-type]
+    concept-type))
+
+(defmethod field->wildcard-field-mappings :default
+  [_]
+  ;; No wildcard field mappings specified by default
```
daniel-zamora: comment doesn't really add much

elastic-utils-lib/src/cmr/elastic_utils/search/es_query_to_elastic.clj
```diff
@@ -145,6 +146,40 @@
...
+       (query-field->elastic-field field concept-type)))
+
+(defmulti field->lowercase-wildcard-field-mappings
+  "Mapping of query model field names to Elasticsearch lowercase wildcard field names."
+  (fn [concept-type]
+    concept-type))
+
+(defmethod field->lowercase-wildcard-field-mappings :default
+  [_]
+  ;; No lowercase wildcard field mappings specified by default
```
daniel-zamora: comment doesn't really add much

elastic-utils-lib/src/cmr/elastic_utils/search/es_query_to_elastic.clj
```diff
@@ -206,9 +241,15 @@
   cmr.common.services.search.query_model.StringCondition
   (condition->elastic
     [{:keys [field value case-sensitive? pattern?]} concept-type]
-    (let [field (if case-sensitive?
-                  (query-field->elastic-field field concept-type)
-                  (field->lowercase-field concept-type field))
+    (let [field (if (and pattern? (config/enable-wildcard-field-searches))
+                  ;; New behavior: use dedicated wildcard fields for pattern searches
+                  (cond
```
daniel-zamora: cond feels out of place, it works but all the other stuff is using ifs

daniel-zamora: are you planning on adding more conds?

elastic-utils-lib/src/cmr/elastic_utils/search/es_query_to_elastic.clj
```diff
@@ -206,9 +241,15 @@
...
+                    case-sensitive? (field->wildcard-field concept-type field)
+                    :else (field->lowercase-wildcard-field concept-type field))
+                  ;; Original behavior: use standard fields
```
daniel-zamora, line 248: these comments don't add much

search-app/src/cmr/search/data/query_to_elastic.clj
```diff
@@ -306,6 +306,18 @@
       (merge default-mappings query-field->lowercase-granule-doc-values-fields-map)
       default-mappings)))
 
+(defmethod q2e/field->wildcard-field-mappings :granule
+  [_]
+  {:granule-ur "granule-ur-wildcard"
+   :producer-gran-id "producer-granule-id-wildcard"
```
daniel-zamora: can we get comments to explain the need for two seperate aliases for producer granule id

elastic-utils-lib/src/cmr/elastic_utils/search/es_query_to_elastic.clj
```diff
@@ -36,7 +36,8 @@
 (defn query-field->elastic-field
   "Returns the elastic field name for the equivalent query field name."
   [field concept-type]
-  (get (concept-type->field-mappings concept-type) (keyword field) field))
+  (keyword
```
daniel-zamora: why this needs to be a keyword would be good to add a comment for

elastic-utils-lib/src/cmr/elastic_utils/search/es_query_to_elastic.clj
```diff
@@ -145,6 +146,40 @@
   (get (field->lowercase-field-mappings concept-type) (keyword field)
```
daniel-zamora, line 145: this falls back to non keyword but query-field->elastic-field falls back to a keyword, its kina confusing why this is different

search-app/test/cmr/search/test/unit/data/query_to_elastic_converters/granule_wildcard_fields.clj
```diff
@@ -0,0 +1,86 @@
+(ns cmr.search.test.unit.data.query-to-elastic-converters.granule-wildcard-fields
+  "Tests wildcard field selection for granule string searches."
+  (:require
+   [clojure.test :refer :all]
+   [cmr.common.services.search.query-model :as qm]
+   [cmr.elastic-utils.config :as elastic-config]
+   [cmr.elastic-utils.search.es-params-converter :as params]
+   [cmr.elastic-utils.search.es-query-to-elastic :as q2e]
+   [cmr.search.data.query-to-elastic]
```
daniel-zamora, line 10: comment why we don't need an alias for these requires

elastic-utils-lib/src/cmr/elastic_utils/search/es_query_to_elastic.clj
zimzoom: This comment follows the style of the other default methods in this area of the code base

elastic-utils-lib/src/cmr/elastic_utils/search/es_query_to_elastic.clj
zimzoom: This comment follows the style of the other default methods in this area of the code base

elastic-utils-lib/src/cmr/elastic_utils/search/es_query_to_elastic.clj
zimzoom: .

elastic-utils-lib/src/cmr/elastic_utils/search/es_query_to_elastic.clj
zimzoom: These comments delineate a feature flag that will be removed soon

search-app/src/cmr/search/data/query_to_elastic.clj
zimzoom: Will add. Yeah sorry I meant to add that. Briefly, there was already two versions floating in the code with regard to the query model versus the mapping. So, I went with "granule" instead of "gran" because I preferred it.

elastic-utils-lib/src/cmr/elastic_utils/search/es_query_to_elastic.clj
zimzoom: This function is able to return either keyword or string based on whether a mapping is found. This was introducing difficulty after adding the feature flag so it seems appropriate to enforce a type here (keyword is used by the vast majority of the downstream consumers) and it's a good practice to do in a dynamically typed language to have more clarity. The "lowercase" versions of the helpers still return strings.

elastic-utils-lib/src/cmr/elastic_utils/search/es_query_to_elastic.clj
zimzoom: You're right, this was mistakenly left here instead of converted to if. Will fix

elastic-utils-lib/src/cmr/elastic_utils/search/es_query_to_elastic.clj
zimzoom: More detailed response:

An ambiguity already existed in `query-field->elastic-field` and introducing the feature flag exposed it. Without normalization, its return type depends on what comes in and what the mapping contains. Specifically, if a field in found in `concept-type->field-mappings`, the return type was whatever that mapping stored. If not found, the return type is whatever the caller passed as `field`.

Currently in main branch, that usually still ends up as keyword because both sides are usually keywords, but that's an implicit convention rather than an explicit contract. Wanting to make it explicit here because once I added the feature flag, the type behavior varied based on the state of the flag:

- wildcard branch:
     - `field->wildcard-field` / `field->lowercase-wildcard-field`
     -  returns strings for the newly mapped wildcard fields
- legacy branch:
     - case-sensitive path uses `query-field->elastic-field` which can return a keyword
     - case-insensitive path uses `field->lowercase-field` which returns a string

Once the toggle existed, the implicit behavior becomes visible because enabled and disabled had to coexist and be tested side by side. This was a pain. I added keyword normalization in query-field->elastic-field to make the legacy branch's contract explicit and stable, rather than letting its return type vary based on caller input or mapping presence while this feature introduces a second field-selection path.

elastic-utils-lib/src/cmr/elastic_utils/search/es_query_to_elastic.clj
zimzoom: This type ambiguity already existed, please see other comment "More detailed response"

search-app/test/cmr/search/test/unit/data/query_to_elastic_converters/granule_wildcard_fields.clj
zimzoom: Will add comment explaining this is for registering methods

elastic-utils-lib/src/cmr/elastic_utils/search/es_query_to_elastic.clj
daniel-zamora: if you really like the comments you may keep them

elastic-utils-lib/src/cmr/elastic_utils/search/es_query_to_elastic.clj
zimzoom: After writing all this I changed my own mind and moving this back to draft pending another update to this branch.

daniel-zamora (approved):

eudoroolivares2016 (approved):

PR #2417, opened 2026-04-28
CMR-11265: Update Jetty and Ring-Jetty-Adapter

jmaeng72:
# Overview

### What is the objective?

Update our jetty and ring-jetty-adapter dependencies

### What are the changes?

Updated the dependencies
Due to update of jetty, our unicode no longer default to utf-8 'under the hood' so had to explicitly set unicode to be utf-8

All

# Required Checklist

# Additional Checklist

## Summary by CodeRabbit

* **Chores**
  * Multiple modules now inherit managed dependency versions from a shared parent and enable parent-plugin support, removing many explicit version pins.
* **Bug Fixes**
  * API responses consistently include "; charset=utf-8" on JSON and XML Content-Type headers for UTF-8 encoding.
* **Tests**
  * Integration test updated to use explicit Unicode escapes for subscription native-id validation.


eereiter (approved):

ingest-app/project.clj
```diff
@@ -1,7 +1,7 @@
 (defproject nasa-cmr/cmr-ingest-app "0.1.0-SNAPSHOT"
   :description "Ingest is an external facing CMR service facilitating providers to create and  update their concepts in CMR. Internally it delegates concept persistence operations to metadata db and indexer micro services."
   :url "https://github.com/nasa/Common-Metadata-Repository/tree/master/ingest-app"
-  :dependencies [[camel-snake-kebab "0.4.2"]
+  :dependencies [[camel-snake-kebab "0.4.0"]
```
jceaser: wait, really, you had to go backwards?

zimzoom (approved):

ingest-app/src/cmr/ingest/api/core.clj
```diff
@@ -120,14 +120,14 @@
   ;; ring-json middleware will handle converting the body to json
   {:status (or (:status result)
                (ingest-status-code result))
-   :headers {"Content-Type" (mt/format->mime-type :json)}
+   :headers {"Content-Type" (str (mt/format->mime-type :json) "; charset=utf-8")}
```
jceaser, line 123: why is this not needed in more files, or are these the only tests that failed?

ingest-app/project.clj
jceaser: I think a comment is needed to prevent someone like me from moving this forward in the future not knowing why it was set backwards.

jceaser (changes requested):
A comment is needed for camel-snake-kebab. I have questions in other places, but these should not hold things back.

ingest-app/src/cmr/ingest/api/core.clj
jmaeng72: this was to pass tests, yes, although technically, all apps could be changed to this forcing charset

ingest-app/project.clj
jmaeng72: I meant to revert this back to the original version, doing that now

jceaser (approved):

ingest-app/src/cmr/ingest/api/core.clj
jmaeng72: After attempting to update the other apps with utf-8 forcing, I have decided against doing it. This lack of update will have CMR remain as it always was: relying on the system's unicode. I made this decision because to update the other apps to use utf-8 would require a huge refactoring of the sys int tests across many files and I did not think it is worth the effort.

PR #2436, opened 2026-05-19
CMR-11089 - Implement /bulk_index/between_date_time to reduce indexing batch sizes

jaortega527:
# Overview

### What is the objective?

Redesign bulk index with timeframe underneath so that large timeframes of data requested can be chunked and processed to improve efficiency.

### What are the changes?

Added a new /bulk_index/between_date_time endpoint that indexes provider/system concepts within an explicit revision-date range, chunking the work into smaller configurable batches. Updated /bulk_index/after_date_time to reuse that same bounded range logic by treating it as date_time -> now, with a max 168-hour (also default and configurable) window to prevent overly large requests. Added the service, dispatcher, message queue, Oracle SQL, config, and README wiring needed to support these changes. Added integration test coverage for the new endpoint, the safer /after_date_time behavior, and rejection of too large /after_date_time windows.

* bootstrap-app
* metadata-db-app

# Required Checklist

# Additional Checklist


bootstrap-app/src/cmr/bootstrap/data/bulk_index.clj
```diff
@@ -399,7 +519,46 @@
     {:message indexing-complete-message
      :max-revision-date (apply util/max-compare
                                (map :max-revision-date
-                                    (apply conj provider-response-map system-concept-response-map)))}))
+                                    (conj (vec provider-response-map)
+                                          system-concept-response-map)))}))
+
+(defn index-data-between-date-time
```
eereiter, line 489: Why is this a full function on its own?  why doesn't it just map through the provider-ids and call index-provider-data-between-date-time?

jaortega527: I won't pretent like I didn't just take index-data-later-than-date-time and create a new method for betwee-date-time, but looking at it now `index-provider-data-between-date-time` doesn't have the same return structure that supports aggregating status across providers. Do you see any issues with me updating `index-provider-data-between-date-time` to have the the same response structure?

bootstrap-app/src/cmr/bootstrap/api/bulk_index.clj
```diff
@@ -1,11 +1,83 @@
...
+
+(defn- validate-after-date-time-window
+  [start-date-time end-date-time]
+  (let [max-window-hours (bootstrap-config/bulk-index-after-date-time-max-window-hours)
+        max-end-date-time (time/plus start-date-time (time/hours max-window-hours))]
+    (when (time/after? end-date-time max-end-date-time)
+      (errors/throw-service-error
+       :invalid-data
+       (format (str "/bulk_index/after_date_time is limited to %d hours. "
+                    "Use /bulk_index/between_date_time with a smaller explicit range.")
```
zimzoom: Is it okay that introducing an hours limit is a breaking change in the API? I think since the limit is configurable, it is probably okay. You might want to confirm there are no callers that require more than 168 hours or that we have not required a longer time range in the recent past.

Also, I think the error message confusing the way it is written, mostly because it mentions two endpoints. If I hit after_date_time and then got this error message, I would feel like it's telling me to use another endpoint between_date_time instead and not understand why.

bootstrap-app/src/cmr/bootstrap/data/bulk_index.clj
```diff
@@ -348,6 +416,23 @@
     {:max-revision-date max-revision-date
      :num-indexed system-concept-count}))
 
+(defn- index-system-misc-concepts-between-date-times
+  "Index all system and miscellaneous concepts created within the given date-time range.
+  Returns a map of :max-revision-date and :num-indexed."
+  [system start-date-time end-date-time]
+  (let [system-concept-response-map (for [concept-type (concat system-concept-types misc-concept-types)]
```
zimzoom: I think this is indexing the misc concept types twice for the CMR/system provider? `system-concept-types` already includes `:variable`, `:service`, `:tool`, and
`:subscription`, and `misc-concept-types` contains those same four types. Because this code concatenates both lists, a `/bulk_index/between_date_time`
request that includes CMR, or omits provider_ids, will fetch and index those concept types twice for the same time window.

I'm DMing you a test I used to verify this.

I think you could fix by changing:

`(for [concept-type (concat system-concept-types misc-concept-types)]`

to:

`(for [concept-type (distinct (concat system-concept-types misc-concept-types))]`

zimzoom (changes requested):

zimzoom (approved):

eereiter (approved):

bootstrap-app/src/cmr/bootstrap/api/bulk_index.clj
```diff
@@ -1,11 +1,84 @@
...
+
+(defn- validate-after-date-time-window
+  [start-date-time end-date-time]
+  (let [max-window-hours (bootstrap-config/bulk-index-after-date-time-max-window-hours)
+        max-end-date-time (time/plus start-date-time (time/hours max-window-hours))]
+    (when (time/after? end-date-time max-end-date-time)
+      (errors/throw-service-error
+       :invalid-data
+       (format (str "The requested time window exceeds the /bulk_index/after_date_time limit of %d hours. "
+                    "Please use a smaller date_time value so the range to now is within %d hours.")
```
jceaser: does anyone else think these messages should be in messages.clj?

bootstrap-app/README.md
```diff
@@ -268,6 +272,26 @@ For a given list of providers (use provider `CMR` to index all system concepts):
...
+
+For a given list of providers (use provider `CMR` to index all system concepts):
+
+    curl -i \
+    	-X POST \
+    	-H "Content-Type: application/json" \
+    	-d '{"provider_ids": ["PROV1", "PROV2", "CMR"]}' \
+    	"http://localhost:3006/bulk_index/between_date_time?start_date_time=2015-02-02T10:00:00Z&end_date_time=2015-02-02T12:00:00Z"
+
+The `start_date_time` parameter is required. Callers can provide either `end_date_time` or `hours`.
```
jceaser, line 297: can you show an example with hours?

system-int-test/src/cmr/system_int_test/utils/bootstrap_util.clj
```diff
@@ -45,6 +45,37 @@
...
+                    :query-params {:synchronous true}
+                    :url (url/bulk-index-between-date-time-url start-date-time end-date-time)
+                    :body (json/generate-string {:provider_ids provider-ids})
+                    :content-type :json
+                    :accept :json
+                    :throw-exceptions false
+                    :connection-manager (s/conn-mgr)})
+         body (json/decode (:body response) true)]
+     (assoc body :status (:status response)))))
+
```
jceaser, line 78: I don't see the example on how to send a request with hours

bootstrap-app/src/cmr/bootstrap/data/bulk_index.clj
```diff
@@ -47,6 +51,21 @@
   "The list of miscellaneous concept types that are provider based but saved in a single system table"
   [:variable :service :tool :subscription])
 
+(def ^:private sql-timestamp-formatter
+  (time-format/formatter "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'"))
```
jceaser: there are functions in oracle-lib/src/cmr/oracle/connection.clj that deal with oracle dates, do any of these do what you need here?

oracle-timestamp->clj-time for one

bootstrap-app/src/cmr/bootstrap/data/bulk_index.clj
```diff
@@ -257,6 +276,55 @@
...
 
+(defn- find-concepts-between-date-times-sql
+  [provider concept-type start-date-time end-date-time]
+  (let [table (concept-tables/get-table-name provider concept-type)
+        provider-id (:provider-id provider)
+        provider-clause (when (or (= :access-group concept-type)
+                                  (and (some #{concept-type} misc-concept-types)
+                                       (not= system-concept-provider provider-id)))
+                          (format " and provider_id = '%s'" (sql-quote (:provider-id provider))))]
+    (format "select * from %s where revision_date >= %s and revision_date < %s%s"
```
jceaser: why are we doing sql statements in this package and not in the cmr.metadata-db.data package? Also, must we link to oracle here? We may want to drop oracle and it would be nice to only have to change metadata-db which should be the only package which knows what database we use.

jceaser: also, * (select all) costs a lot does it not? Do you need all columns?

bootstrap-app/src/cmr/bootstrap/data/bulk_index.clj
```diff
@@ -257,6 +276,55 @@
...
      :num-indexed num-indexed}))
 
+(defn- find-concepts-between-date-times-sql
+  [provider concept-type start-date-time end-date-time]
+  (let [table (concept-tables/get-table-name provider concept-type)
+        provider-id (:provider-id provider)
+        provider-clause (when (or (= :access-group concept-type)
+                                  (and (some #{concept-type} misc-concept-types)
+                                       (not= system-concept-provider provider-id)))
+                          (format " and provider_id = '%s'" (sql-quote (:provider-id provider))))]
```
jceaser: you already pulled out provider-id, and actually the caller does the same thing, why not just pass in provider id and not do this extraction 3 times

bootstrap-app/src/cmr/bootstrap/data/bulk_index.clj
```diff
@@ -257,6 +276,55 @@
...
+        provider-id (:provider-id provider)
+        params {:concept-type concept-type
+                :provider-id provider-id}
+        stmt (find-concepts-between-date-times-sql
+              provider concept-type start-date-time end-date-time)
+        _ (info (format (str "Fetching %s batches for provider %s with revision-date "
+                             "between %s and %s.")
+                        concept-type provider-id start-date-time end-date-time))
+        concept-batches (db/find-concepts-in-batches-with-stmt
+                         db provider params stmt (:db-batch-size system))
```
jceaser: don't you only need provider id in this function?

bootstrap-app/src/cmr/bootstrap/data/bulk_index.clj
```diff
@@ -257,6 +276,55 @@
...
+        num-of-concepts (apply + (map count concept-batches))
+        _ (info (format "Found %d %s(s) for provider %s."
+                        num-of-concepts (name concept-type) provider-id))
```
jceaser: are we doing granules or collections here? is the count high? will this cause any issues with too many scrolls?

jceaser (changes requested):

jceaser:
also, do you need all the info logs?

bootstrap-app/src/cmr/bootstrap/api/bulk_index.clj
jaortega527: Moved this message and the others i added to messages.clj

bootstrap-app/README.md
jaortega527: Updated the README.md to add an example of providing hours instead of an end date/time.

bootstrap-app/src/cmr/bootstrap/data/bulk_index.clj
jaortega527: Moved the SQL statements to metadata-db-app. Also updated the query to align with the style used for other queries and limit the columns selected to use `concept-type->columns` so only the columns needed to rebuild concepts are fetched instead of all columns.

bootstrap-app/src/cmr/bootstrap/data/bulk_index.clj
jaortega527: great feedback. Updated to pull out the provider-id once and pass it in as suggested

bootstrap-app/src/cmr/bootstrap/data/bulk_index.clj
jaortega527: addressed.

bootstrap-app/src/cmr/bootstrap/data/bulk_index.clj
jaortega527: I think both collections and granules can be indexed and I think that the count can be high depending on the provider and size of the date range. I don't know what kind of impact that can have on the scrolls.

jaortega527:
> also, do you need all the info logs?

fair comment. i went through and tried to trim the logging down.

system-int-test/src/cmr/system_int_test/utils/bootstrap_util.clj
jaortega527: added test for bulk_index/between-date-time with hours instead of an end date

bootstrap-app/src/cmr/bootstrap/api/messages.clj
```diff
@@ -73,10 +73,35 @@
     (:message result)
     (format "%s Processing data after %s for bulk indexing" msg/bulk-index-msg-general date-time)))
 
+(defn data-between-date-time
```
jceaser, line 76: public function, not documented. Please document all public functions you have added as the comment will show up in the editor when you hover over. AI can write it for you, but it tends to be to wordy

You can counter me by making function private if possible.

bootstrap-app/test/cmr/bootstrap/test/data/bulk_index_test.clj
```diff
@@ -0,0 +1,99 @@
...
+                  bulk-index/fetch-and-index-concepts-between-date-times
+                  (fn [_system provider concept-type start-date-time end-date-time]
+                    (swap! calls conj [(:provider-id provider)
+                                       concept-type
+                                       start-date-time
+                                       end-date-time])
+                    {:num-indexed 1 :max-revision-date start-date-time})
+                  bulk-index/index-system-misc-concepts-between-date-times
+                  (fn [_system start-date-time end-date-time]
+                    (swap! calls conj ["CMR" :system start-date-time end-date-time])
```
jceaser, line 57: wait, what is going on? are you trying to freeze time or do something else?

metadata-db-app/src/cmr/metadata_db/data/oracle/search.clj
```diff
@@ -330,6 +360,56 @@
...
+                      vec)]
+     (letfn [(find-batch
+               [start-index]
+               (j/with-db-transaction
+                 [conn db :isolation :read-committed :read-only? true]
+                 (let [clauses (-> (find-concepts-between-date-times-clauses
+                                    provider-id concept-type start-date-time end-date-time)
+                                   (conj `(>= :id ~start-index))
+                                   (conj `(< :id ~(+ start-index batch-size))))
+                       stmt (su/build (select columns
```
jceaser, line 388: how fast is this, could you select all concepts with no issues? I ask because you set an isolation mode and I want to know if you did this because you had issues in development.

jceaser (changes requested):
please comment on my questions.

bootstrap-app/test/cmr/bootstrap/test/data/bulk_index_test.clj
jaortega527: No. It isn't freezing time. I'm using the hard-coded start and end values, but the point of the test is to make sure that when provider IDs are empty or nil that the bulk index indexes all providers it can find plus the "CMR" system concepts. I won't pretend an LLM didn't write the test for me, but it was a test I asked it to write.

bootstrap-app/src/cmr/bootstrap/api/messages.clj
jaortega527: I can expand the docstring to 
```Index data with revision-date in the requested date-time range. The request must include start_date_time and may include either end_date_time or hours. If provider_ids is omitted, all providers and CMR system concepts are indexed.```

Attempted to provide better documentation without getting too wordy.

metadata-db-app/src/cmr/metadata_db/data/oracle/search.clj
jaortega527: Great question. The short answer is no, i didn't add the isolation mode because i ran into a dev issue. I used the read-only/read-committed pattern used in the existing `find-concepts-in-batches` function above (line 269). I didn't benchmark it so i can't speak to how fast it is. I saw that the /after-date-time endpoint used that batching approach so I followed that pattern.

bootstrap-app/src/cmr/bootstrap/data/bulk_index.clj
jaortega527: After moving the query code to metadata-db-app I also changed the code to match the query-builder style that is already in place and not using the string-query-builder style meant there was no need for the sql-timestamp-formatter helper function. This helper is no longer necessary and has been completely removed.

bootstrap-app/src/cmr/bootstrap/api/messages.clj
jaortega527: Updated

jceaser (approved):

PR #2437, opened 2026-05-19
CMR-11308: Changes to support better builds (AKA New build)

jceaser:
# Overview

A collection of changes made during the effort to restore CMR build. These changes are not mandatory to the build process, but were found to improve the situation.

1. kms cache testing was found to be missing some definitions which *sometimes* failed to pass
2. quartzite was published with a new name making it easy to find by maven
3. the builder image now sets the version of lein for more consistent builds
4. library exclusions were added to reduce warnings in the logs

### DIFF
--- a/system-int-test/test/cmr/system_int_test/ingest/granule/granule_validation_test.clj
+++ b/system-int-test/test/cmr/system_int_test/ingest/granule/granule_validation_test.clj
@@ -474,6 +474,7 @@
        (validate-granule-multipart granule-concept coll-concept))))
 
 (deftest validation-umm-g-granule-test
+  (ingest/enable-ingest-writes)
   (let [collection-attrs {:EntryTitle "correct"
                           :ShortName "S1"
                           :Version "V1"}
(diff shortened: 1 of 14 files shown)


system-int-test/test/cmr/system_int_test/ingest/granule/granule_validation_test.clj
```diff
@@ -23,7 +23,8 @@
     [cmr.umm.umm-spatial :as umm-s]))
 
 (use-fixtures :each (fn [f]
-                      (ingest/enable-ingest-writes)
+                      ;; TODO: test if this is needed
+                      ;;(ingest/enable-ingest-writes)
```
jceaser, line 27: we have been running without this, so I vote we leave it out

jmaeng72: that's fine, pls delete then

eereiter (approved):

jmaeng72 (changes requested):

jmaeng72 (approved):

PR #2439, opened 2026-05-27
CMR-11311 - Initializing additional caches to prevent CmrHashCache error

jaortega527:
# Overview

### What is the objective?

Ensure unit tests for umm-spec-lib, common-app-lib, indexer-app, and ingest-app have necessary KMS caches enabled to prevent a CmrHashCache error.

### What are the changes?

Created a create-kms-caches-for-testing method in common-app-lib that can be utilized in other modules where KMS caches are required. Additionally, a unit test was added to attempt to assist developers when new caches are added so they're also added for tests.

* umm-spec-lib
* common-app-lib
* indexer-app
* ingest-app

# Required Checklist

# Additional Checklist

### DIFF
--- a/common-app-lib/src/cmr/common_app/services/kms_lookup.clj
+++ b/common-app-lib/src/cmr/common_app/services/kms_lookup.clj
@@ -225,29 +225,6 @@
   []
   (create-kms-uuid-cache kms-temporal-keywords-cache-key))
 
-(defn create-all-kms-caches
-  "Creates all KMS caches required by create-kms-index and lookup functions.
-  Intended for centralized context setup to avoid missing-cache test failures when
-  new KMS cache keys are introduced."
-  []
-  {kms-short-name-cache-key (create-kms-short-name-cache)
-   kms-projects-cache-key (create-kms-project-uuid-cache)
-   kms-umm-c-cache-key (create-kms-umm-c-cache)
-   kms-location-cache-key (create-kms-location-cache)
-   kms-measurement-cache-key (create-kms-measurement-cache)
-   kms-processing-level-cache-key (create-kms-processing-level-uuid-cache)
-   kms-science-keywords-cache-key (create-kms-science-keywords-uuid-cache)
-   kms-platforms-cache-key (create-kms-platforms-uuid-cache)
-   kms-instruments-cache-key (create-kms-instruments-uuid-cache)
-   kms-providers-cache-key (create-kms-providers-uuid-cache)
-   kms-spatial-keywords-cache-key (create-kms-spatial-keywords-uuid-cache)
-   kms-concepts-cache-key (create-kms-concepts-uuid-cache)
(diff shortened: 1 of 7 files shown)


eereiter (approved):

common-app-lib/src/cmr/common_app/services/kms_lookup.clj
```diff
@@ -225,6 +225,29 @@
   []
   (create-kms-uuid-cache kms-temporal-keywords-cache-key))
 
+(defn create-all-kms-caches
```
eudoroolivares2016: Feels like if this is a test util we should find a different namespace for it

umm-spec-lib/src/cmr/umm_spec/test/location_keywords_helper.clj
```diff
@@ -22,13 +22,15 @@
                        {:category "CONTINENT", :type "AFRICA", :subregion-1 "CENTRAL AFRICA", :subregion-2 "ANGOLA", :uuid "9b0a194d-d617-4fed-9625-df176319892d"}
                        {:category "CONTINENT", :type "AFRICA", :subregion-1 "CENTRAL AFRICA", :subregion-2 "LAKE CHAD", :uuid "a1810ec4-2d03-4d98-b049-2cad380fb789"}]})
 
+(defn- create-kms-caches
+  "Creates KMS caches using the centralized helper."
+  []
+  (kms-lookup/create-all-kms-caches))
```
jceaser: is this function needed? can it just change all the other calls to the real function?

common-app-lib/src/cmr/common_app/services/kms_lookup.clj
jceaser: I think all calls to this are for test, so I don't have a comment on the name, but should it move to another package, something under a test folder? No production code uses this right?

jceaser: what about something like cmr.common-app.test.?

jceaser (changes requested):

umm-spec-lib/src/cmr/umm_spec/test/location_keywords_helper.clj
jaortega527: Good point. my first pass through, i had a fallback in case the `common-app-lib` jar that available at the time this test ran wasn't compatible, but decided against that and didn't clean up very well. i'll get this cleaned up

common-app-lib/src/cmr/common_app/services/kms_lookup.clj
jaortega527: nothing but tests are using that method. I am on board with that namespace.

jceaser (approved):

zimzoom (approved):

eudoroolivares2016 (approved):

PR #2441, opened 2026-06-01
CMR-1189: found another place that needed a change

jceaser:
# Overview

### What is the objective?

While testing it was found that an additional app needed to be change. This was probably missed while we had trouble compiling CMR in bamboo.

Since this change was missed here are the reasons for why I think all issues have been found:
* Locally calls to `lein dept :tree | grep  4.1.132.Final` in this project does not report any other missing libraries. This also affects all the apps that link to this library.
* `snyk test --all-projects --severity-threshold=high .` returns success

### DIFF
--- a/message-queue-lib/project.clj
+++ b/message-queue-lib/project.clj
@@ -8,7 +8,7 @@
 
 (def netty-version
   "The netty version to use."
-  "4.1.133.Final") ;; latest as of 2025-09-05
+  "4.1.134.Final") ;; latest as of 2026-05-05 for v3 - 2026-05-20 for v4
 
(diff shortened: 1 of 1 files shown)


eereiter (approved):

message-queue-lib/project.clj
```diff
@@ -8,7 +8,7 @@
 
 (def netty-version
   "The netty version to use."
-  "4.1.133.Final") ;; latest as of 2025-09-05
+  "4.1.134.Final") ;; latest as of 2026-05-05 for v3 - 2026-05-20 for v4
```
jceaser, line 11: I will clean up this comment after bamboo integration tests pass for version 134.

jmaeng72 (approved):

PR #2443, opened 2026-06-04
CMR-11252: fix collection granule agg cache

zimzoom:
# Overview

### What is the objective?

Fix internal error thrown when hitting `/indexer/caches/collection-granule-aggregation-cache"` in cloud environments. ("java.lang.RuntimeException: Invalid token: SIT-GRAN:")

Root cause:

- The indexer cache endpoint lists cache keys through RedisCache/get-keys.
- In cloud environments, shared Redis can contain unrelated raw keys such as "SIT-GRAN:".
- RedisCache/get-keys attempted to EDN-parse every scanned Redis key, so non-EDN keys caused the error.

### What are the changes?

- Hardened RedisCache/get-keys to skip Redis keys that cannot be EDN-deserialized.
- Added optional :key-match-pattern support to RedisCache so callers can scope Redis SCAN operations.
- Scoped collection-granule-aggregation-cache key listing to its own serialized Redis key.
- Moved ring/ring-jetty-adapter into normal redis-utils-lib dependencies so focused lein test commands have the required classpath.
- Added Redis cache regression tests covering non-EDN Redis keys and scoped key scans.

redis-utils-lib , indexer-app

# Required Checklist

# Additional Checklist

### DIFF
--- a/redis-utils-lib/project.clj
+++ b/redis-utils-lib/project.clj
@@ -23,7 +23,8 @@
                  ;; testcontainers needs a newer version of commons-compress, for now
                  ;; we will force it to use the latest version
                  [org.apache.commons/commons-compress]
-                 [org.testcontainers/testcontainers]]
+                 [org.testcontainers/testcontainers]
+                 [ring/ring-jetty-adapter "1.15.4"]]
   :plugins [[lein-exec "0.3.7"]
             [lein-parent "0.3.9"]
             [lein-shell "0.5.0"]]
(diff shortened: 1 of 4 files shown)


zimzoom:
I do not think we should be using EDN Redis keys. This change is to fix an immediate error.

redis-utils-lib/src/cmr/redis_utils/redis_cache.clj
```diff
@@ -20,6 +20,16 @@
...
 
+(def ^:private invalid-cache-key ::invalid-cache-key)
+
+(defn- safe-deserialize
+  "Deserializes v, marking keys that were not written through this cache."
+  [v]
+  (try
+    (deserialize v)
+    (catch RuntimeException _e
+      invalid-cache-key)))
```
zimzoom, line 31: Returning a sentinel instead of nil so that a valid EDN Redis key of nil would not be removed

redis-utils-lib/project.clj
```diff
@@ -23,7 +23,8 @@
                  ;; testcontainers needs a newer version of commons-compress, for now
                  ;; we will force it to use the latest version
                  [org.apache.commons/commons-compress]
-                 [org.testcontainers/testcontainers]]
+                 [org.testcontainers/testcontainers]
+                 [ring/ring-jetty-adapter "1.15.4"]]
```
zimzoom: i think we can leave out version number

eereiter (approved):

redis-utils-lib/src/cmr/redis_utils/redis_cache.clj
```diff
@@ -20,6 +20,16 @@
   [v]
   (when v (edn/read-string v)))
 
+(def ^:private invalid-cache-key ::invalid-cache-key)
```
jceaser, line 23: :: interesting

jceaser (approved):

PR #2458, opened 2026-07-01
CMR-11155: Index-set version control -- Part 2/2

jmaeng72:
# Overview

### What is the objective?

To allow for disaster recovery and keeping track of historical changes of index-set.

### What are the changes?

- Created new index-set table in database as new source of truth
- Created new /sync-with-database api to be able to recover index-sets from database if we lose the index in elastic
- Created new /index-sets/:id?revision_id=X to get different revision versions of the index-set for historical reference
- Index set is treated like a concept now, old revisions are deleted up to 10 versions and there is a new revision created for every create, update, delete. Delete index-set is tombstoned in the db

Indexer

# Required Checklist

# Additional Checklist


system-int-test/test/cmr/system_int_test/bootstrap/reshard_index_test.clj
```diff
@@ -110,41 +111,56 @@
...
                :reshard-index "1_small_collections_100_shards"
                :reshard-status "COMPLETE"}
               (bootstrap/get-reshard-status "1_small_collections" {:elastic-name gran-elastic-name :task-id task-id}))))
+
      (testing "finalizing the resharding"
        (is (= {:status 200
                :message "Resharding completed for index 1_small_collections"}
-              (bootstrap/finalize-reshard-index "1_small_collections" {:synchronous false :elastic-name gran-elastic-name}))))
+              (bootstrap/finalize-reshard-index "1_small_collections" {:synchronous false :elastic-name gran-elastic-name})))
+       (is (= (+ 3 initial-revision) (get-revision)) "Index-set revision should increment after starting, status-check, and finalizing reshard"))
```
jmaeng72, line 141: because some runs of this may be slow, it could take some time for reshard to finish, so I put 3 attempts here to prevent unnecessary failures on non-immediate completes

system-int-test/test/cmr/system_int_test/ingest/bulk_update/bulk_update_test.clj
```diff
@@ -944,7 +944,7 @@
             collection-response (ingest/bulk-update-task-status "PROV1" (:task-id response))
             collection-statuses (:collection-statuses collection-response)]
         (is (= "COMPLETE" (:task-status collection-response)))
-        (is (= "Collection with concept-id [C1200000009-PROV1] is deleted. Can not be updated."
```
jmaeng72, line 947: because we added a new index-set table which is treated like a concept, when these tests are run they were hardcoded to get the concept id as expected with the number of time the concept_seq_id would increment. Because index-set is also using the concept_seq_id to increment its id, the concept_seq_id was increased by one in all these tests.

metadata-db-app/src/cmr/metadata_db/api/routes.clj
```diff
@@ -31,7 +31,8 @@
    ;; XXX This is not a good pattern for large software systems; we need to
    ;;     find a different way to accomplish this goal ... possibly use protocols
    ;;     instead.
-   [cmr.metadata-db.data.oracle.concepts.generic-documents]))
+   [cmr.metadata-db.data.oracle.concepts.generic-documents]
```
jmaeng72, line 34: added to allow index set to be seen by metadata db during sys tests later

metadata-db-app/src/cmr/metadata_db/migrations/093_setup_index_sets_table.clj
```diff
@@ -0,0 +1,52 @@
+(ns cmr.metadata-db.migrations.093-setup-index-sets-table
+  (:require [config.mdb-migrate-helper :as h]))
+
+(def ^:private index-sets-column-sql
+  "id NUMBER,
+  concept_id VARCHAR(255) NOT NULL,
+  native_id VARCHAR(1030) NOT NULL,
```
jmaeng72, line 7: native_id is the id of the index-set

system-int-test/test/cmr/system_int_test/bootstrap/reshard_index_test.clj
```diff
@@ -110,41 +111,56 @@
...
          (is (= false (nil? (:task-id reshard-start-resp))))
 
-         ;; check status. Cannot finalize until status is checked.
-         (is (= {:status 200
-                 :original-index "1_small_collections_100_shards"
-                 :reshard-index "1_small_collections_50_shards"
-                 :reshard-status "COMPLETE"}
-                (bootstrap/get-reshard-status "1_small_collections_100_shards" {:elastic-name gran-elastic-name :task-id task-id})))))
+         (run! (fn [i]
+                 (Thread/sleep 2000) ;; wait for 2 secs
```
jmaeng72, line 155: put a wait here in case the reshard takes longer than expected

metadata-db-app/src/cmr/metadata_db/migrations/093_setup_index_sets_table.clj
```diff
@@ -0,0 +1,52 @@
...
+  metadata BLOB NOT NULL,
+  format VARCHAR(255) NOT NULL,
+  revision_id INTEGER DEFAULT 1 NOT NULL,
+  revision_date TIMESTAMP WITH TIME ZONE DEFAULT SYSTIMESTAMP NOT NULL,
+  deleted INTEGER DEFAULT 0 NOT NULL,
```
jmaeng72, line 12: deleted_time is missing... do we want this?

metadata-db-app/src/cmr/metadata_db/migrations/093_setup_index_sets_table.clj
```diff
@@ -0,0 +1,52 @@
...
+  user_id VARCHAR(30),
+  transaction_id INTEGER DEFAULT 0 NOT NULL")
```
jmaeng72, line 14: this may not be necessary col -- will double check if it's going into the queues and getting into race condition in ES

metadata-db-app/src/cmr/metadata_db/migrations/093_setup_index_sets_table.clj
```diff
@@ -0,0 +1,52 @@
+(ns cmr.metadata-db.migrations.093-setup-index-sets-table
+  (:require [config.mdb-migrate-helper :as h]))
+
+(def ^:private index-sets-column-sql
+  "id NUMBER,
+  concept_id VARCHAR(255) NOT NULL,
+  native_id VARCHAR(1030) NOT NULL,
+  metadata BLOB NOT NULL,
+  format VARCHAR(255) NOT NULL,
```
jmaeng72, line 9: if the format of the metadata is always the same, do we want this col? I can delete if we assume all metadata is going to be json

indexer-app/src/cmr/indexer/system.clj
```diff
@@ -80,6 +81,8 @@
    kl/kms-temporal-keywords-cache-key (kl/create-kms-temporal-keywords-uuid-cache)
    cgac/coll-gran-aggregate-cache-key (cgac/create-cache)
    hf/humanizer-cache-key (hf/create-cache-client)
+   augmenter/token-sid-cache-name (augmenter/create-token-sid-cache)
+   augmenter/token-user-id-cache-name (augmenter/create-token-user-id-cache)
```
eereiter, line 85: Why did these get added?

eereiter: Is this because when a request comes in the request Authorization token can be checked?

metadata-db-app/src/cmr/metadata_db/migrations/093_setup_index_sets_table.clj
eereiter: I don't think that we need the format column.

metadata-db-app/src/cmr/metadata_db/migrations/093_setup_index_sets_table.clj
eereiter: I don't think we need deleted_time, in reality it should never be deleted.

metadata-db-app/src/cmr/metadata_db/migrations/093_setup_index_sets_table.clj
eereiter: Yeah, I think you will need to keep this.

eereiter (approved):

indexer-app/src/cmr/indexer/services/index_set_service.clj
```diff
@@ -338,27 +343,33 @@
...
 (defn put-index-set
   "Upsert the given index-set to the index-sets index in ES."
-  [context index-set]
-  (let [split-index-set-map (split-index-set-by-cluster index-set)]
+  [context combined-index-set]
+  (let [;; Strip out revision-id and deleted if they were provided in the request body to avoid 422 errors from metadata-db.
+        clean-combined-index-set (update combined-index-set :index-set dissoc :revision-id :deleted)
+        split-index-set-map (split-index-set-by-cluster clean-combined-index-set)
+        ;; Save a single unified version to Oracle first to get the revision-id
+        revision-id (save-index-set-to-mdb context clean-combined-index-set)
```
zimzoom: Seems like this is saving a revision to Oracle before the validation 4 lines down (validate-requested-index-set), is that correct? Should we validate before saving?

indexer-app/src/cmr/indexer/services/index_set_service.clj
```diff
@@ -942,7 +987,7 @@
           (put-index-set context combined-index-set))))))
```
zimzoom, line 987: Seems like this will result in saving a duplicate revision to the DB during disaster recovery sync. It's reading the latest revision then calling `put-index-set` which turns around and writes another revision. This could push a genuine revision out of the 10 item limit, up to you if that matters

system-int-test/test/cmr/system_int_test/ingest/index_set_versioning_test.clj
```diff
@@ -0,0 +1,125 @@
...
+        (let [new-data (index/get-index-set-by-id id)]
+          (is (= 2 (get-in new-data [:index-set :revision-id])) "Oracle should have correctly incremented the revision"))))
+
+    (testing "Disaster Recovery: Restore lost ES cluster from Oracle"
+      (let [id 5555
+            is (data-index-set/sample-index-set id)]
+        (index/create-index-set is)
+        (is (some? (index/get-index-set-by-id id)))
+        ;; Simulate ES loss
+        (index/index-set-reset)
```
zimzoom: Suggestion to check the status of this reset call, e.g.: `(is (= 204 (:status (index/index-set-reset))))`

Because I don't understand how this test is working right now, since calling reset deletes from Oracle as well as removes from ES --> then the sync call below does not restore Oracle tombstones --> then test ends with asserting it is index-set 5555 specifically that was restored.

The suggested line could check for possibility that the reset failed to delete from Oracle

daniel-zamora (approved):

zimzoom (approved):

indexer-app/src/cmr/indexer/services/index_set_service.clj
jmaeng72: Yes definitely. I will update this. Good catch

indexer-app/src/cmr/indexer/services/index_set_service.clj
jmaeng72: Should be fine since the new revision will be the same as the old so we'll still have the right copy

system-int-test/test/cmr/system_int_test/ingest/index_set_versioning_test.clj
jmaeng72: I have added comments and the extra reset call check

indexer-app/src/cmr/indexer/system.clj
jmaeng72: I think this was a remanent of an old change I reverted, so I am going to delete this

metadata-db-app/src/cmr/metadata_db/migrations/093_setup_index_sets_table.clj
jmaeng72: since the indexer uses the concepts flow, format is required so I have decided to keep it

indexer-app/src/cmr/indexer/system.clj
jmaeng72: Actually, turns out we do need it to gain permissions to look at cache lists and check tokens

The following tests for these permissions under indexer:
cmr.system-int-test.admin.cache-api-test/cache-apis
cmr.system-int-test.admin.admin-permissions-test/ingest-management-permission-test

PR #2461, opened 2026-07-10
ACES-11360 Fix Int Test: ISO 19115 roundtrip strips trailing whitespace from Quality.QualityContentDetails.Strengths

jaortega527:
# Overview

### What is the objective?

This PR addresses where ISO 19115 roundtrips were silently removing trailing whitespace from `QualityContentDetails.Strengths` (and similar terminal quality detail fields) by aligning expected normalization to trim-based behavior.

### What are the changes?

- Updated test helper `expected-quality` in `umm-spec-lib/src/cmr/umm_spec/test/iso_shared.clj` to:
  - trim each `:QualityContentDetails` value
  - remove empty/nil detail values
  - remove `:QualityContentDetails` when empty after normalization

- Updated XML→UMM quality unit tests to match trim-normalized behavior and removed redundant edge-whitespace cases.

- Updated UMM→XML quality unit test expectation to match normalized serialized output behavior.

- Updated seeded generative roundtrip test in `umm-spec-lib/test/cmr/umm_spec/test/generate_and_parse.clj` from `1496683985472` to `1783446834701` for deterministic reproduction of the original issue.

* umm-spec-lib

# Required Checklist

# Additional Checklist

### DIFF
--- a/umm-spec-lib/src/cmr/umm_spec/umm_to_xml_mappings/iso_shared/quality.clj
+++ b/umm-spec-lib/src/cmr/umm_spec/umm_to_xml_mappings/iso_shared/quality.clj
@@ -9,14 +9,14 @@
   [c]
   (when-let [quality (:Quality c)]
     (let [{:keys [Summary QualityContentDetails]} quality
-          summary-part (str "Summary: " Summary)
+          summary-part (str "Summary: " Summary " ")
           ;; Map explicitly to maintain a reliable layout order
           details-order [:Strengths :Limitations :KnownIssues :Other]
           details-parts (for [k details-order
                               :let [v (get QualityContentDetails k)]
                               :when (seq v)]
-                          (str (name k) ": " v))
-          full-text (string/join " " (cons summary-part details-parts))]
+                          (str (name k) ": " v " "))
+          full-text (string/trim (apply str summary-part details-parts))]
       [:gmd:report
        [:gmd:DQ_QuantitativeAttributeAccuracy
         [:gmd:evaluationMethodDescription
--- a/umm-spec-lib/src/cmr/umm_spec/umm_to_xml_mappings/iso_shared/quality.clj
+++ b/umm-spec-lib/src/cmr/umm_spec/umm_to_xml_mappings/iso_shared/quality.clj
@@ -9,14 +9,14 @@
   [c]
   (when-let [quality (:Quality c)]
     (let [{:keys [Summary QualityContentDetails]} quality
-          summary-part (str "Summary: " Summary " ")
+          summary-part (str "Summary: " Summary)
           ;; Map explicitly to maintain a reliable layout order
           details-order [:Strengths :Limitations :KnownIssues :Other]
           details-parts (for [k details-order
                               :let [v (get QualityContentDetails k)]
                               :when (seq v)]
-                          (str (name k) ": " v " "))
-          full-text (string/trim (apply str summary-part details-parts))]
+                          (str (name k) ": " v))
+          full-text (string/join " " (cons summary-part details-parts))]
       [:gmd:report
        [:gmd:DQ_QuantitativeAttributeAccuracy
         [:gmd:evaluationMethodDescription
--- a/umm-spec-lib/src/cmr/umm_spec/xml_to_umm_mappings/iso_shared/quality.clj
+++ b/umm-spec-lib/src/cmr/umm_spec/xml_to_umm_mappings/iso_shared/quality.clj
@@ -12,39 +12,59 @@
 (def ^:private detail-keys
   [:Strengths :Limitations :KnownIssues :Other])
 
-(defn- preserve-last-detail-trailing-whitespace
-  "For ISO quality strings, preserves trailing whitespace on the last populated
-  quality detail field (if present) by restoring its raw value from the source
-  quality-string. This avoids data loss caused by generic description parsing trim."
+(defn- preserve-detail-edge-whitespace
+  "For ISO quality strings, preserves leading/trailing whitespace on populated
+   quality detail fields by restoring raw content from quality-string when
+   parsing trim caused data loss.
+
+   For non-terminal fields we remove one trailing separator space that is
+   introduced by the ISO writer between key/value pairs."
   [quality-string details]
-  (let [indexed-details (keep (fn [k]
-                                (let [detail-label (str (name k) ": ")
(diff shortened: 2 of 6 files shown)


umm-spec-lib/src/cmr/umm_spec/umm_to_xml_mappings/iso_shared/quality.clj
```diff
@@ -9,14 +9,14 @@
   [c]
   (when-let [quality (:Quality c)]
     (let [{:keys [Summary QualityContentDetails]} quality
-          summary-part (str "Summary: " Summary " ")
```
jaortega527, line 12: Removed the hardcoded trailing space

umm-spec-lib/src/cmr/umm_spec/umm_to_xml_mappings/iso_shared/quality.clj
```diff
@@ -9,14 +9,14 @@
...
+          summary-part (str "Summary: " Summary)
           ;; Map explicitly to maintain a reliable layout order
           details-order [:Strengths :Limitations :KnownIssues :Other]
           details-parts (for [k details-order
                               :let [v (get QualityContentDetails k)]
                               :when (seq v)]
-                          (str (name k) ": " v " "))
-          full-text (string/trim (apply str summary-part details-parts))]
+                          (str (name k) ": " v))
+          full-text (string/join " " (cons summary-part details-parts))]
```
jaortega527: Changed to use string/join shifting the control from the "append whitespace and then trim" to explicit separator joining.

umm-spec-lib/src/cmr/umm_spec/xml_to_umm_mappings/iso_shared/quality.clj
```diff
@@ -8,6 +9,63 @@
   "Returns the pattern that matches all the related fields in UMM-C Quality"
   (re-pattern "Summary:|Strengths:|Limitations:|KnownIssues:|Other:"))
 
+(def ^:private detail-keys
+  [:Strengths :Limitations :KnownIssues :Other])
+
+(defn- preserve-detail-edge-whitespace
```
jaortega527: `preserve-detail-edge-whitespace` compares each parsed detail value to the exact raw substring in the original ISO string. If the only mismatch is edge whitespace that trim removed, it restores that whitespace carefully:
- full raw for last field,
- raw minus one structural separator space for middle fields.

umm-spec-lib/test/cmr/umm_spec/test/generate_and_parse.clj
```diff
@@ -266,7 +266,7 @@
 ;; expected output with the result of the actual conversions. This test runs a record
 ;; through all of the supported formats.
 (deftest roundtrip-generated-collection-records-with-seed
-  (checking-with-seed "collection round tripping seed" 100 1496683985472
```
jaortega527, line 269: I am willing to remove this change. I changed the value to the seed value that reproduces the issue.

eereiter: Yeah, the original seed number also tested a specific issue from years ago. I think I am OK leaving it here as is or removing the checking-with-seed, but if we do remove this test, document on line 216 how to add a seed number because I won't remember how to do it when the time comes to use a different seed number to reproduce an error.  With this test and the one above, 200 tests are run using "random" values in the schema elements.  It just seems excessive to use 100 different tests, but on the other hand the tests "catch" these types of issues - although now we are coding to the test and not necessarily what the end user cares about.

umm-spec-lib/src/cmr/umm_spec/xml_to_umm_mappings/iso_shared/quality.clj
eereiter: We are parsing the same string twice.  Once on the old file line 17 or in this change on line 75 and again in this function.  

if you search for "convert-iso-description-string-to-map" in the xml_to_umm_mappings library there are several examples of how we dealt with the issue in the past.  

distributions_related_url is a good example:  
in file umm-spec-lib/src/cmr/umm_spec/xml_to_umm_mappings/iso_shared/distributions_related_url.clj on line 19 a regex pattern is laid out to parse a string, this same string exists in this file on new line 10.  Then on line 30 in the distributions_related_url file the string is parsed by the  "convert-iso-description-string-to-map" function.  

Then in umm-spec-lib/src/cmr/umm_spec/test/iso19115_expected_conversion.clj on line 395 we take the same starting generated UMM-C record and change what the expected outcome should be. That line calls expected-collection-related-urls which for the Description part of the object it trims the string.

I also noticed that the Quality for ISO expected only looks at the Summary and not the details: umm-spec-lib/src/cmr/umm_spec/test/iso19115_expected_conversion.clj online 422 it calls umm-spec-lib/src/cmr/umm_spec/test/iso_shared.clj/expected-quality. where it 
(when (:Summary quality)
    (assoc quality :Summary (string/trim (:Summary quality)))))

So I think what we should do is allow the previous parsing on old line 17 (new line 75) to be as is - not preserve space leading or trailing spaces (this may or may not affect search for values when users don't expect a space - But I think ES figures that out anyway) then in the test expected code, trim those strings. 

I can pair with you if you wish to walk you through what I wrote above.

eereiter: What I really want is to avoid writing custom code for every element in ISO where an element's string is encoded to contain multiple values which is what happened. Then a few years ago, we converged on the  convert-iso-description-string-to-map solution.

eereiter: if there is a round trip testing issue, we fix it in the test expected code if its possible - but it can be very frustrating

eereiter (changes requested):
We should "fix" this issue in the expected conversion test code.

umm-spec-lib/src/cmr/umm_spec/xml_to_umm_mappings/iso_shared/quality.clj
jaortega527: This makes sense. Instead of modifying the parser behavior to introduce another pass over the quality adding complexity (and potential flakiness) we'll update the expectations in the test.

eereiter (approved):

jceaser (approved):

PR #2462, opened 2026-07-14
CMR-11195: As a developer, I want incoming CMR search requests classified and rate limited by compute cost so that heavy requests cannot cascade into backend failures

daniel-zamora:
# Overview

### What is the objective?

Introduce a traffic-shaping proxy in front of CMR search that classifies requests into priority lanes (express/standard/heavy), applies distributed rate limiting via Redis semaphores, and caches responses to reduce backend load.

### What are the changes?

New search-proxy FastAPI service with three traffic lanes, each with configurable concurrency permits and response cache TTLs
Redis-backed distributed semaphores for per-lane in-flight request counting across instances
Response cache keyed on method, path, query, auth token, Accept header, cmr-search-after cursor, and POST body hash
Request classifier routing traffic to lanes based on spatial complexity, facets, temporal params, and other signals
Feature toggles for bypass, caching, load shedding, and classification via ECS environment variables
AWS infrastructure: ECS task definition, ECR repository, ALB listener rules (priority 83 bypass rule to prevent routing loop, priority 84 proxy rule)
Health endpoint exposing Redis, backend, and per-lane utilization status

List impacted applications

All CMR search traffic (collections and granules endpoints) in environments where enable_search_proxy = true
Internal ALB routing rules
ElastiCache Redis (new usage for semaphores and cache)

# Required Checklist

# Additional Checklist

### DIFF
--- a/search-proxy/src/proxy/app.py
+++ b/search-proxy/src/proxy/app.py
@@ -1,0 +1,355 @@
+import hashlib
+import logging
+import time
+import uuid
+from contextlib import asynccontextmanager
+from urllib.parse import parse_qs
+
+import httpx
+import redis.asyncio
+from fastapi import FastAPI, Request
+from fastapi.responses import JSONResponse, Response
+
+from proxy.cache import ResponseCache
+from proxy.classifier import classify_request
+from proxy.config import ProxySettings, load_lanes_config
+from proxy.lanes import LoadSheddingError, RequestLanes
+
+logger = logging.getLogger(__name__)
+
+# Hop-by-hop headers per RFC 2616 §13.5.1
+HOP_HEADERS = frozenset(
+    {
+        "connection",
+        "keep-alive",
+        "proxy-authenticate",
+        "proxy-authorization",
+        "te",
+        "trailers",
+        "transfer-encoding",
+        "upgrade",
+    }
+)
+
+
+@asynccontextmanager
+async def lifespan(app: FastAPI):
+    """Initialize shared resources on startup, clean up on shutdown."""
+    settings = ProxySettings()
+    lanes_config = load_lanes_config(settings.lanes_config)
+
+    app.state.settings = settings
+    app.state.lanes_config = lanes_config
+
+    # Connection-pooled httpx client for forwarding requests to the backend
+    app.state.backend = httpx.AsyncClient(
+        base_url=settings.backend_url,
+        timeout=settings.backend_timeout_seconds,
+        limits=httpx.Limits(
+            max_connections=settings.backend_max_connections,
+            max_keepalive_connections=settings.backend_max_keepalive,
+        ),
+    )
+    # Redis connection for distributed lane semaphores and response cache
+    app.state.redis = redis.asyncio.from_url(
+        settings.redis_url,
+        retry_on_timeout=True,
+        socket_connect_timeout=settings.redis_socket_connect_timeout,
+        socket_timeout=settings.redis_socket_timeout,
+        health_check_interval=settings.redis_health_check_interval,
+    )
+    app.state.lanes = RequestLanes(lanes_config, app.state.redis)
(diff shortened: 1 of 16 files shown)


zimzoom:
A README would be nice to have :)

search-proxy/src/proxy/lanes.py
```diff
@@ -0,0 +1,103 @@
...
+    def _lane_key(self, lane_name: str) -> str:
+        return f"lane:{lane_name}:active"
+
+    async def _try_acquire(self, lane_name: str, permits: int) -> bool:
+        """Atomically increment the lane counter and check against limit.
+
+        Uses INCR for atomicity — if the post-increment value exceeds
+        the permit limit, immediately DECR to roll back."""
+        key = self._lane_key(lane_name)
+        current = await self.redis.incr(key)
```
zimzoom, line 37: It seems that semaphore permits can leak, since there's no lease or TTL on the lane counters. If a request dies midflight for whatever reason before it gets to the decrement step, the counter is just permanently incremented.

search-proxy/src/proxy/app.py
```diff
@@ -0,0 +1,525 @@
...
+        dependencies["search"] = {"ok?": False, "problem": str(exc)}
+
+    # Lane utilization
+    lanes_config = request.app.state.lanes_config
+    redis_client = request.app.state.redis
+    for lane in lanes_config.lanes:
+        key = f"lane:{lane.name}:active"
+        active_raw = await redis_client.get(key)
+        active = int(active_raw) if active_raw else 0
+        lane_ok = active < lane.permits
```
zimzoom: As we discussed verbally, please double check that returning bad health checks when a lane is full will not cause ECS Agent/ALB to take action on the service

search-proxy/src/proxy/app.py
```diff
@@ -0,0 +1,525 @@
...
+            )
+
+    # Merge POST form body params into query params for classification
+    content_type = request.headers.get("content-type", "")
+    params = dict(request.query_params)
+
+    if request.method == "POST" and "application/x-www-form-urlencoded" in content_type:
+        body = await request.body()
+        try:
+            body_params = parse_qs(body.decode(), keep_blank_values=True)
```
zimzoom, line 314: Is this going to capture multipart form params? The classifier has a check for `_is_shapefile_request` that looks for both the multipart form Content-Type and also params shapefile or file. A request like that would have "shapefile"/"file" inside the multipart body but it seems this code here will only capture query params and URL-encoded POST body params

search-proxy/src/proxy/lanes.py
daniel-zamora: noted in the readme. TTL, doesn't work with a counter, you'd have to refactor to having individual cache key value pairs and count them manually if you wanted TTL to do anything.  The leak only really happens if the task dies mid flight or we run out of connections to redis.

search-proxy/src/proxy/app.py
daniel-zamora: not sure, I'll need to test.

zimzoom: Approving because I think classifier effectiveness is slightly out of scope for this ticket, but if your test shows this is an issue I think we should make a new ticket for it or fix now

zimzoom (approved):

search-proxy/src/proxy/classifier.py
```diff
@@ -0,0 +1,163 @@
...
+    "include_facets",
+    "online_only",
+    "cloud_cover",
+}
+
+HEAVY_PREFIXES = (
+    "temporal_facet[",
+    "cycle[",
+    "passes[",
+    "options[readable_granule_name][pattern]",
```
zimzoom, line 20: In order to capture all the problem wildcard searches we have identified, add these in addition to this existing line:

```
    "options[granule_ur][pattern]",
    "options[producer_granule_id][pattern]",
```

PR #2473, opened 2026-07-30
CMR-11357: update elasticsearch

zimzoom:


jceaser (approved):

dev-system/compose.yaml
```diff
@@ -11,7 +11,7 @@ services:
     build:
       context: .
       dockerfile: resources/elasticsearch/Dockerfile.elasticsearch
-    image: cmr-elasticsearch:8.19.14
+    image: cmr-elasticsearch:8.19.17
```
jmaeng72, line 14: our current elasticsearch cluster is 8.19.14, do we need to update our cluster as well?

es-spatial-plugin/project.clj
```diff
@@ -23,7 +23,8 @@
 (def es-deps-target-path
   "es-deps")
 
-(def elastic-version "8.19.14")
+(def elastic-version "8.19.17")
```
jmaeng72, line 26: any changes to es-spatial-plugin requires we create new AMI with these changes and update our elastic cluster to this new AMI across all ENVS. We need to make sure we do this.

We also need to test that our cluster will still work properly with this new es-spatial-plugin change. So pls test in WL before merging this

jmaeng72 (changes requested):

es-spatial-plugin/project.clj
zimzoom: Test executed as discussed

dev-system/compose.yaml
jmaeng72: I'm assuming that after the ES test in WL, we are saying this works fine

jmaeng72 (approved):

PR #2477, opened 2026-08-05
CMR-11386: Refactor search-proxy semaphore

daniel-zamora:
# Overview

### What is the objective?

Harden the search proxy's distributed lane semaphore against permit leaks from crashed tasks, fix several correctness bugs discovered during review of CMR-11195, and extend the request classifier to route two additional wildcard pattern params to the heavy lane.

### What are the changes?

Sorted-set semaphore (lanes.py) — replaced the previous INCR/DECR counter approach with a Redis sorted set where each in-flight request occupies one member (a UUID) scored by its expiry epoch (redis_now + permit_ttl). On each acquire, ZREMRANGEBYSCORE prunes entries whose score has passed before checking the count, so leaked permits from crashed tasks self-heal automatically after backend_timeout_seconds (300s default) without manual intervention. Redis server time is used for scores to avoid ECS task clock skew across instances.

Fail-open behavior was added to fix a bug where a Redis blip during acquire caused the exception to propagate up to the proxy handler, which only catches LoadSheddingError — so users were receiving dropped requests or 500 errors during any transient Redis unavailability. With fail-open, Redis exceptions during acquire are caught, logged, and the request is forwarded without a permit rather than returned as an error to the user.

Bug fixes (app.py):

- Hash truncation removed — auth token and POST body hashes were previously truncated to 16 characters ([:16]), reducing collision resistance and risking wrong cached responses being served to different users or bodies
- POST body double-read eliminated — await request.body() was called up to three times per request; body and body hash are now computed once up front and reused
- Health endpoint lane counts updated to use ZCOUNT key wall_now +inf (counts only non-expired sorted set entries) instead of GET key, which read the old INCR counter that no longer exists

Classifier (classifier.py) — added options[granule_ur][pattern] and options[producer_granule_id][pattern] to the heavy lane prefix list, consistent with the existing options[readable_granule_name][pattern] rule. Both trigger wildcard scans on large granule ID fields and carry the same performance profile.

README — updated concurrency, leaked permits, and configuration sections to reflect the sorted set implementation and corrected the Redis pool default (+50 → +100).

The search proxy (search-proxy/) only — specifically the Redis-backed lane semaphore, the request classifier, and the response cache key logic. No changes to CMR search itself or any other service.

# Required Checklist

# Additional Checklist

### DIFF
--- a/search-proxy/src/proxy/app.py
+++ b/search-proxy/src/proxy/app.py
@@ -91,12 +91,15 @@ async def lifespan(app: FastAPI):
         ),
     )
     # Redis connection for distributed lane semaphores and response cache.
-    # Pool size defaults to total lane permits + 50 so it always exceeds the
-    # maximum number of concurrent Redis operations (one per in-flight request).
+    # Pool size defaults to total lane permits + 100. Each acquire involves 3
+    # sequential Redis ops (TIME, ZREMRANGEBYSCORE, ZCARD, ZADD) plus a ZREM
+    # on release and a GET/SET for cache. Connections are released between ops
+    # so peak concurrent demand tracks concurrent requests, not ops per request.
+    # The +100 provides headroom for health checks and reconnect storms.
     # Override with CMR_PROXY_REDIS_MAX_CONNECTIONS if permits change without
     # a redeploy or if additional headroom is needed.
     redis_max_connections = settings.redis_max_connections or (
-        sum(lane.permits for lane in lanes_config.lanes) + 50
+        sum(lane.permits for lane in lanes_config.lanes) + 100
     )
     app.state.redis = redis.asyncio.from_url(
         settings.redis_url,
@@ -106,7 +109,11 @@ async def lifespan(app: FastAPI):
         health_check_interval=settings.redis_health_check_interval,
         max_connections=redis_max_connections,
     )
-    app.state.lanes = RequestLanes(lanes_config, app.state.redis)
+    app.state.lanes = RequestLanes(
+        lanes_config,
+        app.state.redis,
+        permit_ttl=int(settings.backend_timeout_seconds),
+    )
     app.state.cache = ResponseCache(app.state.redis, settings.max_cache_response_bytes)
     yield
     await app.state.backend.aclose()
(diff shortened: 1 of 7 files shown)


search-proxy/src/proxy/app.py
```diff
@@ -91,12 +91,15 @@ async def lifespan(app: FastAPI):
         ),
     )
     # Redis connection for distributed lane semaphores and response cache.
-    # Pool size defaults to total lane permits + 50 so it always exceeds the
-    # maximum number of concurrent Redis operations (one per in-flight request).
+    # Pool size defaults to total lane permits + 100. Each acquire involves 3
+    # sequential Redis ops (TIME, ZREMRANGEBYSCORE, ZCARD, ZADD) plus a ZREM
```
jmaeng72, line 95: you say 3 ops, but the list has 4

search-proxy/test/test_lanes.py
```diff
@@ -183,10 +183,35 @@ async def test_unknown_lane_falls_back_to_default(self, lanes):
...
+    async def test_no_release_attempted_on_fail_open(self, lanes):
+        """When Redis fails during acquire, no release is attempted (nothing was stored)."""
+        with patch.object(lanes.redis, "time", new=AsyncMock(side_effect=Exception("Redis down"))):
+            with patch.object(lanes.redis, "zrem", new=AsyncMock()) as mock_zrem:
+                async with lanes.acquire("heavy"):
+                    pass
+                mock_zrem.assert_not_called()
+
+    async def test_fail_open_does_not_propagate_exception(self, lanes):
+        """A Redis error during acquire must not surface to the caller as an exception."""
```
jmaeng72, line 202: Can you reword to match the format you used for the other tests "When..., Then"

search-proxy/test/test_lanes.py
```diff
@@ -216,3 +241,82 @@ async def test_does_not_raise_load_shedding_error(self, tight_lanes):
                     pass
             except LoadSheddingError:
                 pytest.fail("LoadSheddingError raised with load_shedding_enabled=False")
+
+
+class TestTTLExpiry:
+    async def test_expired_permit_is_pruned_on_next_acquire(self, redis_client):
+        """A leaked permit with an expired score is cleaned up on the next acquire."""
```
jmaeng72, line 248: Would like to request to reword tests to be in "When <scenario>, Then <expected result>" for easier reading and faster understanding by future devs

jmaeng72 (changes requested):

jmaeng72 (approved):

eereiter (approved):

PR #2483, opened 2026-08-13
CMR-11419: fix error handling for ES8 query complexity limit

zimzoom:
# Overview

### What is the objective?

Fix error handling for query complexity limit change in ES8.

### What are the changes?

The "payload too large" error handling has been expanded to cover both the ES7 and ES8 responses.

search-app

# Required Checklist

# Additional Checklist

### DIFF
--- a/elastic-utils-lib/src/cmr/elastic_utils/search/es_index.clj
+++ b/elastic-utils-lib/src/cmr/elastic_utils/search/es_index.clj
@@ -217,7 +217,7 @@
            :payload-too-large
            "The search is creating more buckets than allowed by CMR. Please narrow your search."))
 
-        (when (re-find #"maxClauseCount is set to 1024" body)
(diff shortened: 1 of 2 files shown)


jaortega527:
The only thing I could think of to improve the exception handling (and it would apply to the whole block) is to also check the status code, but the message checks are so specific I don't think it  would make any difference. The regex for both ES7 and ES8 is great. LGTM.

jaortega527 (approved):

zimzoom:
> The only thing I could think of to improve the exception handling (and it would apply to the whole block) is to also check the status code, but the message checks are so specific I don't think it would make any difference. The regex for both ES7 and ES8 is great. LGTM.

Yeah, I thought about adding this, but ultimately felt like it would be changing two things instead of one since the ES7 version did not check status code. Since another dev feels the same, I am going to go ahead and add it.

rushgeo (approved):

PR #2489, opened 2026-08-31
CMR-11195: As a developer, I want incoming CMR search requests classified and rate limited by compute cost so that heavy requests cannot cascade into backend failures

daniel-zamora:
# Overview

### What is the objective?

Introduce a traffic-shaping proxy in front of CMR search that classifies requests into priority lanes (express/standard/heavy), applies distributed rate limiting via Redis semaphores, and caches responses to reduce backend load.

### What are the changes?

New search-proxy FastAPI service with three traffic lanes, each with configurable concurrency permits and response cache TTLs
Redis-backed distributed semaphores for per-lane in-flight request counting across instances
Response cache keyed on method, path, query, auth token, Accept header, cmr-search-after cursor, and POST body hash
Request classifier routing traffic to lanes based on spatial complexity, facets, temporal params, and other signals
Feature toggles for bypass, caching, load shedding, and classification via ECS environment variables
AWS infrastructure: ECS task definition, ECR repository, ALB listener rules (priority 83 bypass rule to prevent routing loop, priority 84 proxy rule)
Health endpoint exposing Redis, backend, and per-lane utilization status

List impacted applications

All CMR search traffic (collections and granules endpoints) in environments where enable_search_proxy = true
Internal ALB routing rules
ElastiCache Redis (new usage for semaphores and cache)

# Required Checklist

# Additional Checklist

### DIFF
--- a/search-proxy/README.md
+++ b/search-proxy/README.md
@@ -89,7 +89,7 @@ Always returns HTTP 200. Used for ALB/ECS target group health checks so that Red
 
 ### `GET /health`
 
-Informational health check. Always returns HTTP 200 — dependencies report their status but do not affect the top-level `ok?`. The result is cached for 5 seconds (unhealthy results are not cached so recovery is visible immediately).
+Informational health check, not cached. Nothing automated polls it — ALB/ECS use `/health/shallow`. Currently always returns HTTP 200: dependencies report their status but do not affect the top-level `ok?`.
 
 ```json
 {
--- a/search-proxy/README.md
+++ b/search-proxy/README.md
@@ -45,7 +45,7 @@ All settings are environment variables with the `CMR_PROXY_` prefix.
 | `CMR_PROXY_BACKEND_TIMEOUT_SECONDS` | `300.0` | Backend request timeout |
 | `CMR_PROXY_BACKEND_MAX_CONNECTIONS` | `500` | httpx connection pool size |
 | `CMR_PROXY_BACKEND_MAX_KEEPALIVE` | `200` | httpx keepalive connection pool size |
-| `CMR_PROXY_REDIS_MAX_CONNECTIONS` | auto | Redis pool size; defaults to total lane permits + 50 |
+| `CMR_PROXY_REDIS_MAX_CONNECTIONS` | auto | Redis pool size; defaults to total lane permits + 100 |
 | `CMR_PROXY_REDIS_SOCKET_CONNECT_TIMEOUT` | `2.0` | Redis connection timeout in seconds |
 | `CMR_PROXY_REDIS_SOCKET_TIMEOUT` | `2.0` | Redis read/write timeout in seconds |
 | `CMR_PROXY_REDIS_HEALTH_CHECK_INTERVAL` | `30` | Seconds between Redis keepalive pings |
@@ -131,6 +131,6 @@ pytest
 
 ## Operational notes
 
-**Leaked permits**: If a task is killed mid-request or Redis is briefly unavailable during release, the permit entry is not removed. It will expire automatically once its TTL score passes (defaulting to the backend timeout, 300 seconds). If a lane shows sustained at-capacity pressure before entries age out, the sorted sets can be cleared directly from Redis: `lane:express:active`, `lane:standard:active`, `lane:heavy:active`.
+**Leaked permits**: A permit leaks when a task is killed before `_release` runs, or when Redis is briefly unavailable during release (the exception is swallowed so the ASGI handler can still return a response). Once a leaked entry's TTL score passes (defaulting to `backend_timeout_seconds`, 300 seconds), it stops affecting lane counts — the health endpoint's `ZCOUNT` filters on the current timestamp as a lower bound, and each acquire's `ZCARD` runs after `ZREMRANGEBYSCORE` prunes expired-score entries. Physical removal from Redis happens on the next acquire for that lane. Note: if Redis is unavailable during acquire, the fail-open path applies — no permit is stored and no release is attempted, so there is no leak in that case. To immediately reset a lane without waiting for TTL, delete its sorted set key from Redis: `lane:express:active`, `lane:standard:active`, `lane:heavy:active`.
 
 **Debugging**: Set `CMR_PROXY_LOG_LEVEL=DEBUG` to log backend response details including content encoding and actual byte counts. Remove when done — debug logging is verbose under load.
--- a/search-proxy/README.md
+++ b/search-proxy/README.md
@@ -36,8 +36,8 @@ All settings are environment variables with the `CMR_PROXY_` prefix.
 
 | Variable | Default | Description |
 |----------|---------|-------------|
-| `CMR_PROXY_BACKEND_URL` | required | CMR search base URL (no `/search` suffix) |
-| `CMR_PROXY_REDIS_URL` | required | Redis connection URL |
+| `CMR_PROXY_BACKEND_URL` | _none — required, startup fails if unset_ | CMR search base URL (no `/search` suffix) |
+| `CMR_PROXY_REDIS_URL` | _none — required, startup fails if unset_ | Redis connection URL |
 | `CMR_PROXY_LANES_CONFIG` | `lanes.json` | Path to lanes config file; used when `CMR_PROXY_LANES_JSON` is not set |
 | `CMR_PROXY_LANES_JSON` | — | Lanes config as a JSON string; takes precedence over `CMR_PROXY_LANES_CONFIG` when set. Intended for deployments that inject the value from Parameter Store as an environment variable |
 | `CMR_PROXY_LOG_LEVEL` | `INFO` | Log level (`DEBUG`, `INFO`, `WARNING`) |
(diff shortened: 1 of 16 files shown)


search-proxy/src/proxy/app.py
```diff
@@ -0,0 +1,570 @@
...
+def _extract_request_id(request: Request) -> str:
+    """Extract or generate a request ID."""
+    return (
+        request.headers.get("cmr-request-id")
+        or request.headers.get("x-request-id")
+        or str(uuid.uuid4())
+    )
+
+
+@app.api_route("/{path:path}", methods=["GET", "POST"])
```
jaortega527, line 286: Looking at the ALB listener rule it looks like all requests will go through the proxy if they match the`path_pattern` and `header` conditions. I see that this method is limited to `GET` and `POST` requests. Should it also allow `OPTIONS` requests? if a request comes to CMR from a browser and has the preflight request it will receive a 405 response from the proxy. Not blocking if we are sure that isn't a concern.

daniel-zamora: going to fix it in the alb rules

search-proxy/README.md
```diff
@@ -0,0 +1,137 @@
...
+{"errors": ["Service temporarily overloaded for heavy-tier queries"]}
+```
+
+## Configuration
+
+All settings are environment variables with the `CMR_PROXY_` prefix.
+
+| Variable | Default | Description |
+|----------|---------|-------------|
+| `CMR_PROXY_BACKEND_URL` | required | CMR search base URL (no `/search` suffix) |
```
jmaeng72: the default value of CMR_PROXY_BACKEND_URL is 'required'? That's what the column means right? "Default" = default value

search-proxy/README.md
```diff
@@ -0,0 +1,137 @@
...
+```
+HTTP 429 Too Many Requests
+Retry-After: 10
+
+{"errors": ["Service temporarily overloaded for heavy-tier queries"]}
+```
+
+## Configuration
+
+All settings are environment variables with the `CMR_PROXY_` prefix.
```
jmaeng72, line 35: are all these env vars required?

search-proxy/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Health endpoints
+
+### `GET /health/shallow`
+
+Always returns HTTP 200. Used for ALB/ECS target group health checks so that Redis or backend failures do not trigger task replacement.
+
+### `GET /health`
+
+Informational health check. Always returns HTTP 200 — dependencies report their status but do not affect the top-level `ok?`. The result is cached for 5 seconds (unhealthy results are not cached so recovery is visible immediately).
```
jmaeng72: It always returns 200? So what's the point of checking the health if it always returned 200?

search-proxy/README.md
```diff
@@ -0,0 +1,137 @@
...
+    "lane-standard": {"ok?": true, "active": 3, "permits": 150, "at_capacity": false},
+    "lane-heavy": {"ok?": true, "active": 0, "permits": 50, "at_capacity": false}
+  }
+}
+```
+
+When a lane is at capacity, `at_capacity` is `true` but `ok?` remains `true`. Use this endpoint to monitor lane utilization rather than to drive automated remediation.
+
+## Running locally
+
```
jmaeng72, line 110: which python version is required?

search-proxy/test/test_lanes.py
```diff
@@ -0,0 +1,322 @@
...
+            async with lanes.acquire("standard"):
+                with pytest.raises(LoadSheddingError) as exc_info:
+                    async with lanes.acquire("express"):
+                        pass
+                assert exc_info.value.lane_name == "express"
+
+
+class TestLoadShedding:
+    async def test_sheds_when_full(self, tight_lanes):
+        async with tight_lanes.acquire("heavy"):
```
jmaeng72, line 149: I think every test should have comment of "When... Then..." for easier reading. Although the naming of the python tests are clear enough in this case. So I would categorize this request as 'nice to have', if we can be consistent with tests having comments.

search-proxy/test/test_config.py
```diff
@@ -0,0 +1,388 @@
+import json
```
jmaeng72, line 1: Are these tests being auto-run in the CMR-Core bamboo build?

search-proxy/README.md
daniel-zamora: So that the ELB health check doesn't bounce the tasks because redis is down.  I can take another look and see if this is still needed.

search-proxy/test/test_config.py
daniel-zamora: Its not merged yet.

search-proxy/README.md
daniel-zamora: its in the dockerfile and the pyproject.toml

search-proxy/README.md
daniel-zamora: No.  They are just there for convenience.

search-proxy/README.md
daniel-zamora: Yea not this one.  I can format it differently but this one has to be set to CMR search's endpoint.

search-proxy/test/test_lanes.py
daniel-zamora: Yea, I think that is reasonable. Can we move it to a separate ticket?

search-proxy/src/proxy/app.py
```diff
@@ -0,0 +1,570 @@
...
+            "content_length_header": response.headers.get("content-length"),
+            "actual_content_bytes": len(response.content),
+        },
+    )
+    return response
+
+
+# Cached health check result with TTL-based expiration
+_health_cache: dict = {"result": None, "expires": 0.0}
+_HEALTH_CACHE_TTL = 5.0
```
jmaeng72: will this be a changing var in the future? If so, maybe we make this a more configurable value

search-proxy/src/proxy/lanes.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def _lane_key(self, lane_name: str) -> str:
+        return f"lane:{lane_name}:active"
+
+    async def _redis_now(self) -> float:
+        """Return current time from the Redis server to avoid ECS task clock skew."""
+        seconds, microseconds = await self.redis.time()
+        return seconds + microseconds / 1_000_000
+
+    async def _try_acquire(self, lane_name: str, permits: int, request_id: str, redis_now: float) -> bool:
```
jaortega527, line 49: I don't have a ton of redis experience, but as written, I don't think `_try_acquire` is atomic. If a lane has one permit, and 2 requests come in closely enough they could both call ZCARD both could receive a 0 and both call ZADD resulting in 2 active entries when only one should be allowed. it could look like this:

R1: ZCARD -> 0
R2: ZCARD -> 0
R1: since 0 is less than 1, ZADD R1
R2: since 0 is less than 1, ZADD R2

During a period of high load, this might result in a "recommended" lane limit but might not enforce that limit. 

I did some quick research and since the Redis instance is shared it is probably the best place to enforce an atomic ZCARD/ZADD call. This would require a Lua script that calls both the ZCARD and ZADD so those actions can be treated as a atomic and then the Lua script would be registered and called from lanes.py

Maybe this is tolerable since it should still bring improvement as written. I just want to confirm that we are okay with a potential soft limit.

search-proxy/src/proxy/app.py
daniel-zamora: Its no longer used as the ALB check target, so we don't even need to cache this, I think I might just take it out.   I can also have it return a non 200 for stuff not working downstream. I think when I was testing this stuff I was running into issues and never reverted this part back.  But I think that might be a separate ticket.

search-proxy/test/test_lanes.py
jmaeng72: yeah that's fine

search-proxy/src/proxy/lanes.py
daniel-zamora: Yes, I moved away from the atomic version of this proxy because it was orphaning permits and with this we have an effective TTL on all the permits. I don't think the race condition really has a huge impact.  If we run into a scenario where we are close to the limit it won't matter and at normal operation it won't matter.  Unless you disagree that some race condition around the edges can produce a scenario that we actually care about.

jaortega527: No, I think i'm just late to trying to understand. If you've already considered it then i don't see any reason to make any changes. I figured this was probably not a significant concern, but wanted to say something in case it hadn't already been considered.

jaortega527 (approved):

search-proxy/README.md
jmaeng72: what I mean is that the other variables have the actual default value in the 'Default' column, like CMR_PROXY_LANES_CONFIG is 'lanes.json'. But CMR_PROXY_BACKEND_URL is called 'required' which is not the same category. It should be the value of the default. And you can put that it is required in the description column

search-proxy/src/proxy/app.py
jmaeng72: I'm fine with separate ticket, just please remember to write it and put it in the sprints/backlog

search-proxy/README.md
jmaeng72: what was the conclusion to this?

search-proxy/README.md
daniel-zamora: Did you see how I updated it?  the backend url is configured by the deployment in any case.

search-proxy/README.md
daniel-zamora: yea we decided that we shouldn't have the search-proxy task fail and restart if its connection to redis or elastic is bad, or they have what would otherwise be a bad status.

search-proxy/src/proxy/app.py
daniel-zamora: I removed the caching.  Its no longer needed.

search-proxy/README.md
jmaeng72: oops was looking at wrong commit

jmaeng72 (approved):

PR #2490, opened 2026-08-31
CMR-11422: Add parameter to only allow validate keyword usage by provider (Enforcement list)

eudoroolivares2016:
# Overview

### What is the objective?

Adding a parameter store to be able to prevent providers in an enforced list e.g. the esdis providers +- to no longer be able to use the `validate-keyword: false` header which has allowed unsearchable metadata into the system

There is a temporary carveout for the KMS API itself until we resolve the cache delay issue either by removing the CMR cache entirely or having KMS issue a cache refresh to CMR `CMR-11524`. 

### What are the changes?

Added a new parameter store that is a comma sep list, pulled that into the conditional logic of the validate-keyword value, refactoring KMS lookup function. 

CMR ingest and metadata validation

### Testing
1. Create multiple providers if in SIT, set this parameter store in the ingest-app redeploy CMR
2. Ensure that the providers on that list are unable to use `validate-keyword` header. If they do it will always resolve to being true. Ensure that the KMS API if correcting a metadata keyword issue is still able to make sure of it. (SIT only testing possible) `https://cmr.sit.earthdata.nasa.gov/kms/metadata_correction`
3. Delete local keyword cache. Ensure that running `curl -i -X POST -H "Authorization: mock-echo-system-token" http://localhost:3006/caches/refresh/kms` regenerates all the caches as it does on main

# Required Checklist

# Additional Checklist

### DIFF
--- a/common-app-lib/src/cmr/common_app/services/kms_lookup.clj
+++ b/common-app-lib/src/cmr/common_app/services/kms_lookup.clj
@@ -443,8 +443,6 @@
 (defn- generate-lookup-by-related-urls-map
   "Create a map with the related url comparison map as keys to the UUID for that related url."
   [gcmd-keywords-map]
-  (def gkm gcmd-keywords-map)
-  (generate-lookup-by-comparison-map gcmd-keywords-map :related-urls))
 
 (defn- generate-lookup-by-measurement-name
   "Create a map with the measurement field values defined in UMM-Var map to the KMS keywords."
--- a/common-app-lib/src/cmr/common_app/services/kms_lookup.clj
+++ b/common-app-lib/src/cmr/common_app/services/kms_lookup.clj
@@ -443,6 +443,7 @@
 (defn- generate-lookup-by-related-urls-map
   "Create a map with the related url comparison map as keys to the UUID for that related url."
   [gcmd-keywords-map]
+  (generate-lookup-by-comparison-map gcmd-keywords-map :related-urls))
 
 (defn- generate-lookup-by-measurement-name
   "Create a map with the measurement field values defined in UMM-Var map to the KMS keywords."
--- a/common-app-lib/src/cmr/common_app/services/kms_lookup.clj
+++ b/common-app-lib/src/cmr/common_app/services/kms_lookup.clj
@@ -314,7 +314,6 @@
                  [keyword-scheme maps-by-short-name]))
              (select-keys gcmd-keywords-map keywords-to-lookup-by-short-name))))
 
-;; TODO remove this via any means necessary
 (def duplicate-keywords
   "Lookup table to account for any duplicate keywords. Will choose the preferred value.
   Common key is :uuid which is a field in the location-keyword map. "
@@ -446,12 +445,6 @@
   [gcmd-keywords-map]
   (def gkm gcmd-keywords-map)
   (generate-lookup-by-comparison-map gcmd-keywords-map :related-urls))
-
-(comment
-
-(tap> (generate-lookup-by-comparison-map gkm :related-urls))
-
-)
 
 (defn- generate-lookup-by-measurement-name
   "Create a map with the measurement field values defined in UMM-Var map to the KMS keywords."
--- a/common-app-lib/src/cmr/common_app/services/kms_lookup.clj
+++ b/common-app-lib/src/cmr/common_app/services/kms_lookup.clj
@@ -31,6 +31,10 @@
   (:import #_{:clj-kondo/ignore [:unused-import]}
            (clojure.lang ExceptionInfo)))
 
+;; ---------------------------------------------------------------------
+;; initializing kms cache keys
+;; ---------------------------------------------------------------------
+
 (def kms-short-name-cache-key
   "The key used to store the data generated from KMS into a short name index cache
   in the system hash cache map for fast lookups."
@@ -115,6 +119,11 @@
   "The key used to store the data generated from KMS into a temporal keywords index cache
   in the system hash cache map for fast lookups."
   :kms-temporal-keywords-index)
+
+
+;; ---------------------------------------------------------------------
+;; Initializing cache instances for kms keywords
+;; ---------------------------------------------------------------------
 
 (def kms-cache-ttl
   "Time To Live value for KMS caches. nil means never expire."
@@ -225,6 +234,10 @@
   []
   (create-kms-uuid-cache kms-temporal-keywords-cache-key))
 
+;; ---------------------------------------------------------------------
+;; Define rules for scheme lookups
+;; ---------------------------------------------------------------------
+
 (def kms-scheme->fields-for-umm-c-lookup
   "Maps the KMS keyword scheme to the list of fields that should be matched when
   comparing fields between KMS and UMM-C, UMM-G, UMM-S, UMM-T, or UMM-Var."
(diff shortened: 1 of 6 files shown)


common-app-lib/src/cmr/common_app/services/kms_lookup.clj
```diff
@@ -430,8 +444,15 @@
 (defn- generate-lookup-by-related-urls-map
   "Create a map with the related url comparison map as keys to the UUID for that related url."
   [gcmd-keywords-map]
+  (def gkm gcmd-keywords-map)
```
eereiter: Please remove debugging code.

common-app-lib/src/cmr/common_app/services/kms_lookup.clj
```diff
@@ -301,6 +314,7 @@
                  [keyword-scheme maps-by-short-name]))
              (select-keys gcmd-keywords-map keywords-to-lookup-by-short-name))))
 
+;; TODO remove this via any means necessary
```
eereiter: Cqn we write a CMR or ECSE ticket and put into this comment to remove duplicate-keywords

common-app-lib/src/cmr/common_app/services/kms_lookup.clj
```diff
@@ -430,8 +444,15 @@
...
   [gcmd-keywords-map]
+  (def gkm gcmd-keywords-map)
   (generate-lookup-by-comparison-map gcmd-keywords-map :related-urls))
 
+(comment
+
+(tap> (generate-lookup-by-comparison-map gkm :related-urls))
+
+)
+
```
eereiter, line 447: Please remove this comment block.

ingest-app/src/cmr/ingest/api/collections.clj
```diff
@@ -23,10 +23,16 @@
...
-  [headers]
-  (let [validate-keywords-value (if validate-keywords-default-true-enabled?
-                                  (if (= "false" (get headers VALIDATE_KEYWORDS_HEADER)) false true)
-                                  (= "true" (get headers VALIDATE_KEYWORDS_HEADER)))]
+  [headers provider-id]
+  (let [validate-keywords-value
+        ;; There is a temporary carveout for the KMS API itself until we resolve the cache delay issue either by removing the CMR cache entirely 
+        ;; or having KMS issue a cache refresh to CMR see CMR-11524.
+        (if (and (not (contains? headers SEND_KMS_METADATA_FIXER_HEADER))
+                 (contains? (set (ingest-config/keyword-enforced-providers)) provider-id))
```
eereiter: In order to read the parameter we have to redeploy CMR.  So here we are constantly converting to a set.  

Can we do the following:

;; 1. Define the cached set using delay
(def enforced-providers-cache
  (delay (set (ingest-config/keyword-enforced-providers))))

;; 2. Use it in your tool logic by dereferencing it with @
(defn provider-enforced? [provider-id]
  (contains? @enforced-providers-cache provider-id))

This way the set is only called once for the static keyword-enforced-providers list and cached in memory for the duration of the program lifecycle.

ingest-app/src/cmr/ingest/config.clj
```diff
@@ -158,3 +159,12 @@
...
   "This indicates whether or not ingest granule subscriptions are enabled."
   {:default true :type Boolean})
+
+(defconfig keyword-enforced-providers
+  "A list of providers for which keyword validation is enforced on ingest.
+   Should be an array of provider ids for which keywords must be present/valid.
+   Example  \"PROV1,PROV2\", would enforce keyword validation for PROV1 and PROV2.
+   If no providers should have keyword validation enforced, set to an empty array."
+  {:default []
+   :parser #(map (comp keyword string/trim) (string/split % #","))})
```
eereiter: So if the CMR_KEYWORD_ENFORCED_PROVIDERS parameter store does not exist an empty array [] is returned.  What if an empty string is defined in the parameter store ""  

When the input % is an empty string "", this snippet will evaluate to a list containing a single empty keyword: (:).

Step-by-Step Breakdown
(string/split "" #",")
Splitting an empty string does not return an empty list. It returns a vector containing the empty string itself: [""].
(string/trim "")
Trimming an empty string simply returns the empty string: "".
(keyword "")
Passing an empty string to keyword produces an empty keyword printed as : in Clojure.

Why This Is Dangerous
An empty keyword (:) is valid Clojure syntax, but it is rarely intended. 
It can cause silent bugs because:It evaluates to truthy in conditional checks.
It will not match a proper keyword like :provider-id.It consumes a tiny amount of memory in the keyword table that cannot be garbage collected.

;; Option 1: Remove blanks before processing (Recommended)
(->> (string/split % #",")
     (remove string/blank?)
     (map (comp keyword string/trim)))
;; Input: ""  => Output: ()

;; Option 2: Use keep to filter out blanks during the map step
(keep #(when-not (string/blank? %) 
         (keyword (string/trim %))) 
      (string/split % #","))
;; Input: ""  => Output: ()

eereiter (changes requested):
I have a couple of small changes

ingest-app/src/cmr/ingest/api/collections.clj
eudoroolivares2016: Interesting. So are we doing that for generics right now? I mean for the `generic-ingest-disabled-list` at least?

common-app-lib/src/cmr/common_app/services/kms_lookup.clj
eudoroolivares2016: Ah that was just a note to myself. I wanna talk about that a bit 2morrow. I had assumed we could just use something in the actual KMS keyword rather than hard-coding these

ingest-app/src/cmr/ingest/api/collections.clj
eereiter: I do not know, if we are not, we should. I just happened to notice the call in your code changes.

sylvesterkaczmarek (commented):
Provider enforcement is bypassed by the mere presence of `cmr-send-kms-metadata-fixer`; the new tests even accept the value `false`. Nothing in `get-validation-options` authenticates that request header, so an enforced provider can set it and regain the normal `validate-keywords=false` path. Gate this carveout on trusted KMS caller identity rather than a client-supplied header, and add a spoofing regression.

ingest-app/src/cmr/ingest/config.clj
```diff
@@ -158,3 +159,14 @@
...
+(defconfig keyword-enforced-providers
+  "A list of providers for which keyword validation is enforced on ingest.
+   Should be an array of provider ids for which keywords must be present/valid.
+   Example  \"PROV1,PROV2\", would enforce keyword validation for PROV1 and PROV2.
+   If no providers should have keyword validation enforced, set to an empty array."
+  {:default []
+   :parser #(->> (string/split % #",")
+              (remove string/blank?)
+              (map string/trim))
+  })
```
jceaser: while I like this syntax, it is not the clojure way to put closing brackets on a new line.

jceaser (changes requested):

ingest-app/src/cmr/ingest/config.clj
eudoroolivares2016: Ran it with [cljfmt](https://github.com/weavejester/cljfmt) it found some other stuff in there

eereiter (approved):

jceaser (approved):

zimzoom (approved):

